PR #5, opened 2016-12-17
Jest and Inquirer Integration [WIP]

ev1stensberg:
Commit allows further testing and integration with inquirer and jest
for our TDD approach. Regex could be changed to have specific test
names for our purposes.


ev1stensberg:
FWIW- requiring `rx` without having it as a dep should break, perhaps it's taking the local version from `inquirer`

lib/observable-questions.js
```diff
@@ -0,0 +1,12 @@
+const Rx = require('rx');
+const {inquirerInput} = require('./utils/inquirer-questions.js');
```
pksjce: Maybe name this as `inquirer-types` or `inquirer-question-types`?

lib/inquirer.js
```diff
@@ -0,0 +1,14 @@
+const inquirer = require('inquirer');
+const questions = require('./observable-questions.js');
+
+inquirer.prompt(questions).ui.process.subscribe(
+	(ans) => {
+		console.log(ans);
```
pksjce: How to test this though?

package.json
```diff
@@ -14,17 +14,32 @@
   "main": "./bin/cli.js",
   "scripts": {
     "lint": "eslint ./lib",
-    "install-commit-validator": "fit-commit-js install"
+    "install-commit-validator": "fit-commit-js install",
+    "test": "jest"
```
ev1stensberg: Could do later, when we put the entire readme up?

lib/inquirer.js
ev1stensberg: For yourself: Open in GH desktop: `npm link` could need a sudo flag. Type `webpack-cli --init`. Through jest: Simulate an inq instance if possible.

lib/observable-questions.js
ev1stensberg: Yeah, that's probably the best, I'll change that up

ev1stensberg:
FWIW - We can mock manually or use mocking patterns to do this automatically.

ev1stensberg:
Don't merge yet, [WIP]

package.json
pksjce: Also good to add a `test-watch` script?

lib/observable-questions.js
```diff
@@ -0,0 +1,12 @@
...
+const {inquirerInput} = require('./utils/inquirer-types.js');
+
+module.exports = Rx.Observable.create( (obs) => {
+	obs.next(inquirerInput('entryLogic',
+	'What is the name of the entry point in your application?'
+	));
+	obs.next(inquirerInput('outputLogic',
+	'What is the name of the output directory in your application?'
+	));
+	obs.onCompleted();
```
pksjce: Instead you could,
```
const questions = [
	inquirerInput('entryLogic','What is the name of the entry point in your application?'), 
	inquirerInput('outputLogic','What is the name of the output directory in your application?')];

module.exports = Rx.Observable.from(questions);
```

tests/setup-test.js
```diff
@@ -0,0 +1,27 @@
...
+	 it('Should recieve input and handle answers', function () {
+		 process.nextTick(function mockResponse() {
+			 stdin.send('index.js\n');
+		 });
+		 process.nextTick(function mockResponse() {
+			 stdin.send('output.js\n');
+		 });
+		 inquirer.prompt(questions).ui.process.subscribe(function (response) {
+			 expect(response.entryLogic).toBe('index.js')
+			 expect(response.outputLogic).toBe('output.js')
```
pksjce: I don't think you need to test if `inquirer` works. You only need to test what we do with the answers.

package.json
DanielaValero: I think we should also lint the tests, when we don't do it, the coding style of the tests will not be consistent with the style of the code under /lib

ev1stensberg: Doing what @pksjce said we need listed probably on wednesday. Got midterms. @DanielaValero Yes! Will do! Just have to get something testable up first, collecting information so I've got enough knowledge to do this 👍

DanielaValero:
Hi @ev1stensberg 

In spite of the work not yet being completed, therefore theoretically ready for a code review, I have done a quick check on the current state of your branch, and would like to share some ideas, perhaps you've got them already on your sight, but wanted to share them in case you don't.

* In order to fix the travis error, we need to add 'use strict' on top of the test files
* When we enable lintin on the test files, there will be some easy fixes that you will be able to apply with the help of the editorconfig and the beautifier, however, eslint is throwing no-undef on jasmine functions, such as `describe`, in order to fix them we would need to add to the list of environments in the eslint settings, `jasmine: true`
* I'd suggest that for the test files, we use the same name of the file under `./lib`, for which we are doing the testing. We could follow the same style as in webpack for the name, in there, the file has a sufix: `.test.js`, when we do this, for example a test for: `inquirer.js`, would be: `inquirer.test.js`. This way, would be more straight forward to find the test of the file we are at the moment looking at. 

Does it make sense?

package.json
DanielaValero: I'd also say that a watch for the test would be quite helpful when we want to code with TDD

Issue #6, opened 2016-12-20
Feature [--migrate]: Migrate from webpack 1 to 2 through webpack-cli


pksjce:
#### Migrations
- [x] loaders -> rules
- [x] pre/postLoaders -> rules with `enforce: 'pre/post'`
- [X] solve for `-loaders` suffix.
- [x] resolve.root -> resolve.modules
- [ ] remove `json-loader`
- [x] add `sourceMap: true` to `UglifyJsPlugin`
- [x] change `ExtractTextPlugin` syntax
- [x] remove `OccurendeOrderPlugin`
- [ ] add `LoaderOptionsPlugin` with `minimize:true` ,`context` and `debug`

#### NFR

- [x] Tests should pass
- [ ] Add yarn support
- [x] Add node 4 support

**Do you want to request a *feature* or report a *bug*?**
This is a feature request

**What is the current behavior?**
There is no automated tool to help migrate from webpack 1 to webpack 2.

**If the current behavior is a bug, please provide the steps to reproduce.**
NA

**What is the expected behavior?**
The expected behavior is to have webpack-cli help make this migration automatically.

**If this is a feature request, what is motivation or use case for changing the behavior?**
It is not clear from first glance as to what it takes to migrate breaking features for webpack2. Providing an automated option would be the best case scenario

**Please mention other relevant information such as the browser version, Node.js version, Operating System and programming language.**

Node.js version > 0.12.17

TheLarkInn:
Maybe we could create a schema diff map. (At least for future builds) v3, 4, etc. Then we could literally map values and pieces to v3 schema etc. 

Obvs a down the road feature but still relevant.

pksjce:
Yes, I'm really not sure what webpack1 schema was and so I'm pretty sure I will miss many cases while doing this.

DanielaValero:
I think we should add explicitly the task that we should validate the current webpack 1 config, and the generated webpack 2 config. This way we would avoid problems during the migration and ensure that the result is working fine

pksjce:
### Plan of action

```bash
webpack --migrate <transformation(optional)>
```

This would then work as follows.

```javascript
switch (arg1) {
  // other cases
  case 'migrate': 
    cliEngine.migrate(..restOfArgs);
    break;
}
```
### Migration logic

`cliEngine.js`

```javascript
const migrate = chosenTransforms => {
  const currentConfig = readCurrentWebpackConfig();
[...]

PR #26, opened 2016-12-27
Feat - Adds -loader suffix after loader manipulation

pksjce:
Handling `-loader` suffix from #6 

Yes, updated the test case accordingly

No

**Other information**


ev1stensberg:
❤️

PR #27, opened 2016-12-29
Integrate Addon Parser Structure, Improved Tests and Linting in bin and tests

ev1stensberg:
Initial Work on the addon parser. 

TODO: 

- [x] Webpack Addon integration

- [x] Linting
- [x] Tests

- [x] Refactor try-catch block in `bin/webpack` 
- [x] Parser


DanielaValero:
Hi @ev1stensberg is this PR ready for a review? the only commit it has at says WIP

ev1stensberg:
@DanielaValero Still stuff to do, I'll notify when it's ready.

ev1stensberg:
@DanielaValero  I think it's ready for merge now. Will have to wait for @pksjce for more changes. This serves as a good fundament for further work.

DanielaValero:
Cool, checking it now. There are some questions I've added so far, and code style related stuff also that we could improve to make it more maintainable, however we can do it as a second PR later on. 
We could for example have a PR to unify, add documentation and improve details in the code, that would make it more maintainable and easier for devs to digg in quickly

lib/parser/resolve-packages.js
```diff
@@ -0,0 +1,36 @@
...
+	});
+}
+function spawnChild(pkg) {
+	// spawn('npm', ['install', '--save', pkg], { stdio: 'inherit', customFds: [0, 1, 2] });
+	return spawn('npm', ['install', '--save', pkg]);
+}
+
+module.exports = function resolvePackages(option) {
+	option.packages.filter( (pkg) => {
+		processPromise(spawnChild(pkg)).then( (result) => {
```
DanielaValero: The linter is complaining about:

* 'result' is defined but never used
* 'packageModule' is assigned a value but never used
* 'err' is defined but never used

tests/__mocks__/parser/resolve.mock.js
```diff
@@ -0,0 +1,32 @@
...
+const path = require('path');
+
+function mockPromise(value) {
+	return (value || {}).then ? value : {
+		then: function (callback) {
+			return mockPromise(callback(value));
+		}
+	};
+}
+function spawnChild(pkg) {}
```
DanielaValero: Here the linter also complains about:
* 'pkg' is defined but never used 
* 'result' is defined but never used

tests/__mocks__/parser/resolve.mock.js
```diff
@@ -0,0 +1,32 @@
...
+
+function getLoc(option) {
```
ev1stensberg: location

bin/webpack.js
```diff
@@ -5,13 +5,30 @@
...
+		files.filter( (file) => {
+			delete require.cache[file];
+		});
+
+		// TODO: If path is wrong, leave it to higher powers until we publish on npm
+		// Could also install and load and install through github
+		return require(path.join(dirPath, 'webpack.js'))
+	});
+	*/
+}
```
ev1stensberg: In development it goes to the catch block immediately, in production this would try to resolve first. Breaks now, as we don't have `node_modules/webpack-cli`

lib/parser/resolve-packages.js
```diff
@@ -0,0 +1,36 @@
...
+const path = require('path');
+
+function processPromise(child) {
+	return new Promise(function(resolve, reject) { //eslint-disable-line
+		child.addListener('error', reject);
+		child.addListener('exit', resolve);
+	});
+}
+function spawnChild(pkg) {
+	// spawn('npm', ['install', '--save', pkg], { stdio: 'inherit', customFds: [0, 1, 2] });
```
ev1stensberg: Shows the output in development if there's error with a package and we want to debug

lib/parser/resolve-packages.js
ev1stensberg: We handle the error on ourselves and packageModule should be passed to `transform` later. Result is never used since we know it has succeeded if it doesn't make an error

lib/utils/npm-exists.js
```diff
@@ -0,0 +1,18 @@
+const got = require('got');
+
+const npmSource = {
+	taobao: 'https://npm.taobao.org',
+	npm: 'https://www.npmjs.org'
+};
+
+const constant = value => () => value;
```
ev1stensberg: Source code from `npm-exists`, would need refactor later.

ev1stensberg: Could leave this as good first contribution, make a issue and label it

tests/__mocks__/parser/resolve.mock.js
ev1stensberg: See previous comment.

bin/webpack.js
DanielaValero: Ok, perhaps then later we can run the block conditionally depending on NODE_ENV or so

lib/parser/resolve-packages.js
DanielaValero: We should not have lintin errors merged into master, In other parts we are doing `eslint-disable` for issues that are happening now, but as these ones, will be fixed later when we have more stuff in

lib/utils/npm-exists.js
DanielaValero: In that case, if we care using the source code of `npm-exists` then we should refer to it as a dependency and install it via npm, instead of copy paste the code into our project

tests/__mocks__/parser/resolve.mock.js
DanielaValero: In that case, we should call it `getLocation`, a good practice when naming variables is not to save bytes in words, the goal is to have names that are right away understood and that are meaningful. A good guide on these series of good practices to write clean code, was recently translated into JS from a really good book from Robert Martin's book [*Clean Code*](https://www.amazon.com/Clean-Code-Handbook-Software-Craftsmanship/dp/0132350882)

Here the link to it: https://github.com/ryanmcdermott/clean-code-javascript#variables

I understand that we want to do things quickly to setup the groundwork, however we should try to keep things as clean as possible, to avoid that the code grows unclean and later on is harder to clean.

lib/parser/resolve-packages.js
DanielaValero: Ok, perhaps then later we can run the block conditionally depending on NODE_ENV or so.

For now, would you add a comment on top of this section to explain why is commented, and how in development mode, we can use it?

pksjce:
@DanielaValero - I see that you've approved this. Let me merge it then

tests/__tests__/parser/validate.test.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
+require('babel-polyfill');
+const validateAddons = require('../../../lib/utils/validate-addons');
+
+describe('Addon Validation', () => {
+	it('Should notify if the package isn\'t prefixed', async () => {
+		try {
+			await validateAddons(['webpack']);
+		} catch (object) {
+			expect(object.toString()).toEqual('TypeError: Package isn\'t a valid name\n' +
```
okonet: Is there a reason not to use `toThrow` here?

```
expect(() => validateAddons(['webpack'])).toThrow(
			'Package isn\'t a valid name\n' +
			'It should be prefixed with \'webpack-addon\''
		);
```

works the same and is much more clear isn't it?

tests/__tests__/parser/validate.test.js
```diff
@@ -0,0 +1,25 @@
...
+	it('Should complain about only supplying \'webpack-addon\' as the name', async () => {
+		try {
+			await validateAddons(['webpack-addon']);
+		} catch (object) {
+			expect(object.toString()).toEqual('TypeError: \'webpack-addon\' is not a valid addon');
+		}
+	});
+	it('Should complain about unregistered packages', () => {
+		// error shows up in console as we're handling it instead of throwing an actual error.
+		expect(validateAddons(['webpack-addon-trinity'])).toBeDefined();
```
okonet: I'm wondering what this does. Anyways I'm having issues with Jest quitting on every run because of https://github.com/webpack/webpack-cli/pull/27/files#diff-3b9090788549586f699038f130f8a542R16 and the fact it isn't wrapped.

2017-02-03 at 17 41

ev1stensberg: Yeah, the tests should be improved, there's one guy doing a clean-up right now. If you look at the parser branch under tests, you can see what's left to be done. + More mocking. I was lazy when I first added the tests, as I knew this would change.

tests/__tests__/parser/validate.test.js
ev1stensberg: It's a thing I implemented to make the test run, ,see prev comment. This should be refactored, but was a nice way to get the error message that didn't pass using `toThrow`.

PR #40, opened 2017-01-26
Feature: Jscodeshift Transformations for --migrate

okonet:

feature

**Summary**

This is where the work on #6 is happening.

#### Migrations
- [x] loaders -> rules
- [x] pre/postLoaders -> rules with `enforce: 'pre/post'`
- [X] solve for `-loaders` suffix.
- [x] resolve.root -> resolve.modules
- [x] remove `json-loader`
- [x] add `sourceMap: true` to `UglifyJsPlugin`
- [x] change `ExtractTextPlugin` syntax
- [x] remove `OccurendeOrderPlugin`
- [x] Remove `webpack.optimize.DedupePlugin`
- [x] add `LoaderOptionsPlugin` with `minimize:true` ,~~`context`~~ and `debug`
- [x] change BannerPlugin (https://webpack.js.org/guides/migrating/#bannerplugin-breaking-change)

It should not

**Other information**

Issues:

* [x] If no value is set in query string it should convert it to `true`
* [x] Convert `String` to `Number`
* [ ] Support Arrays in query string
* [ ] Make sure special cases like https://github.com/webpack-contrib/imports-loader/issues/38 work
* [ ] Support identifiers for object keys. For now it always creates literals for keys. This is tricky. See https://mathiasbynens.be/notes/javascript-properties [Lets choose literals only for now]


ev1stensberg:
I'm assigning @pksjce for all related work on transformations. Good work! 🎉

okonet:
@ev1stensberg I'm a bit confused. Should I continue working on this?

ev1stensberg:
Oh yeah, I'm just making Pavithra head of reviewing and getting into the PR, evidently you should ask her for calibration on separation of work when she comes back

okonet:
During my work on this PR I run into some nasty issues with tabs and had to replace them with spaces for fixtures. I saw we have been using `eslint --fix` to re-format the result of the transformation in order to eliminate those inconsistencies. The problem with this is that:

1. It basically leaves our code untested since we compare our transformations + eslint transformation with the input
1. For some reason it completely strip the result of the `removeJsonLoaderTransform` and I spent a few hours debugging it :(
1. We should probably allow re-formating the code with eslint but _after_ all of the transforms are done (and not only in test runner)
1. This required some hard-code of for the jest test runner that is bundled with jscodeshift. This will make upgrading to the next version of jscodeshift very error prone if they decide to change the API internally.

I think it's much better to not rely on eslint formatting for our own tests. In order to be able to pass the `ast` down to transforms, I'd introduce a wrapper function that would take care of this. This would allow us to still use pre-bundled `defineTest` from `jscodeshift` and have modular transformations.

I’m also seeing some of the inconsistencies in the source code itself (as well as mixed usages of tabs and spaces). Should we switch to spaces to ensure we have fewer conflicts (I guess different editors and tools treat tabs differently) and enforce it with eslint? Now that we have auto-formatting in pre-commit hook with lint-staged, it will even fix these automatically. Thoughts?

okonet:
Looks like webpack repo itself is using tabs, so we should probably stick with it: https://github.com/webpack/webpack/blob/master/.editorconfig#L4

I'd still enforce it with eslint.

DanielaValero:
Hi @okonet 

About the tabs and spaces, I am with you on staying with tabs, so we are coherent with webpack. 

I am not the best expert in code transformations, but anyways will check the branch and the work done on this matter so far, so I can soon enough start also coding :)

okonet:
Another question is if we should check `devtool` option before applying this transform: https://webpack.js.org/guides/migrating/#uglifyjsplugin-sourcemap

pksjce:
@okonet - Thanks a lot for this! Was looking for a review to validate my approach to this. Looking forward to working with you. Will be adding my comments on this by tomorrow.

lib/transformations/defineTest.js
```diff
@@ -13,10 +13,6 @@
 
 const fs = require('fs');
 const path = require('path');
-const cli_engine = require('eslint').CLIEngine;
-const eslintrules = require(process.cwd() + '/.eslintrc.json');
-eslintrules.fix = true;
-const cli = new cli_engine(eslintrules);
```
pksjce: I did this to counteract [this issue](https://github.com/benjamn/recast/issues/242) with recast. But we are using a branch of recast right now, until this issue is fixed. So thanks for cleaning this up.

lib/transformations/index.js
```diff
@@ -1,7 +1,11 @@
 const loaderTransform = require('./loaders/loaders');
 const resolveTransform = require('./resolve/resolve');
+const removeJsonLoaderTransform = require('./removeJsonLoader/removeJsonLoader');
+const uglifyJsPluginTransform = require('./uglifyJsPluginTransform/uglifyJsPluginTransform');
```
pksjce: This file is `uglifyJsPlugin`

lib/transformations/loaders/loaders.js
```diff
@@ -3,114 +3,171 @@ const safeTraverse = require('../safeTraverse');
 module.exports = function(j, ast) {
 
 	const createArrayExpression = function(p) {
-		var objs = p.parent.node.value.value.split('!').map(val => j.objectExpression([j.property('init', 
-										j.identifier('loader'), 
-										j.literal(val)
-										)]));
-		var loaderArray = j.arrayExpression(objs);
+		let objs = p.parent.node.value.value.split('!')
```
pksjce: Note - Let's move such references(which are necessary) to some common meaningful methods.

lib/transformations/loaders/loaders.js
```diff
@@ -3,114 +3,171 @@ const safeTraverse = require('../safeTraverse');
...
 		p.parent.node.value = loaderArray;
 		return p;
 	};
 
 	const createLiteral = val => {
-		var literalVal = val;
-		if(val === 'true') literalVal = true;
-		if(val === 'false') literalVal = false;
+		let literalVal = val;
+		// We'll need String to native type conversions
```
pksjce: How to do this? Is there some generic way to convert?

lib/transformations/occurenceOrderPlugin/__testfixtures__/.editorconfig
```diff
@@ -0,0 +1,3 @@
+[*]
+indent_style = space
+indent_size = 4
```
pksjce: Does this need to be committed?

lib/transformations/occurenceOrderPlugin/__testfixtures__/occurenceOrderPlugin.output.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+    plugins: []
```
pksjce: Better to remove the plugins param if the array is empty?

pksjce:
Also, the `no-var` lint rule is breaking the build. Maybe turn it into a warning and slowly move up to error?

lib/transformations/occurenceOrderPlugin/__testfixtures__/.editorconfig
okonet: It does if you use editoconfig plugin in your IDE. Why not include it?

lib/transformations/occurenceOrderPlugin/__testfixtures__/occurenceOrderPlugin.output.js
okonet: I was thinking about it but still not sure about it. Do you think it won't be too obscure to just remove it from the config?

.eslintrc.json
```diff
@@ -29,6 +29,7 @@
             "always"
         ],
         "no-unused-vars": 1,
-        "no-console": 0
+        "no-console": 0,
+        "no-var": "error"
```
okonet: @pksjce I've added this rule to enable `eslint --fix` for it but I agree this should be in a separate PR at least. I'll revert the commit for now and rebase the branch.

lib/transformations/index.js
okonet: What do you mean? I'd add `Transform` suffix to all transforms just to make it more explicit. Thoughts?

lib/transformations/loaders/loaders.js
okonet: I'm not sure it will work in any context but I agree we should try reuse as much as possible. I'd wait till we see more repetitions before we start adding abstractions.

lib/transformations/loaders/loaders.js
okonet: Well, this is what it methods implements, basically. ↓

ev1stensberg:
@pksjce You've got conflicts

okonet:
I've fixed the confilcts 😎

okonet:
Some updates on this PR:

- Lots of changes in the folder structure. We could simplify even more since there are still lots of bootstrapping code in there but I wanted to stop somewhere :)
- Exporting `transform` function that takes source and list of transformations as arguments. This one is re-used in the `migrate` command

## Help needed:

- [ ] How to remove the right `debug: true`? Other plugins could use same key:value. https://github.com/webpack/webpack-cli/blob/2e4e011c6d0f8a210fad28c2db6d2f24a57a107a/lib/transformations/loaderOptionsPlugin/loaderOptionsPlugin.js#L9-L9
- [ ] It would be great to validate config after transformation. Is there a package for that?

pksjce:
For Validation you should use https://github.com/webpack/webpack/blob/master/lib/webpack.js#L58

ev1stensberg:
@okonet @pksjce In the parser branch, the validation logic is included. Should wait until that is merged perhaps? I'm aiming to get it ready before monday.

okonet:
I'm fine with merging this as soon as we get all transforms ready. We can continue in separate PRs after that.

lib/transformations/utils.js
```diff
@@ -0,0 +1,169 @@
...
+/*
+ * @function findPluginsRootNodes
+ *
+ * Finds the path to the `plugins: []` node
+ *
+ * @param j — jscodeshift API
+ * @param { Node } node - Node to start search from
+ * @returns Path
+ * */
+function findPluginsRootNodes(j, node) {
```
pksjce: Hey why can't this be like `findAnyRootNode(j, node, nodeName){}`

okonet: It can totally be if you need it to!

lib/transformations/utils.js
```diff
@@ -0,0 +1,169 @@
...
+ *
+ * @param j — jscodeshift API
+ * @param { string | number } key - Property key
+ * @param { string | number | boolean } value - Property value
+ * @returns Node
+ * */
+function createProperty(j, key, value) {
+	return j.property(
+		'init',
+		j.literal(key),
```
pksjce: This has to be `j.identifier(key)`. Will change this and also related snapshots.

lib/transformations/utils.test.js
```diff
@@ -0,0 +1,124 @@
...
+			const res = utils.createProperty(j, 'foo', 'bar');
+			expect(j(j.objectExpression([res])).toSource()).toMatchSnapshot();
+		});
+
+		it('should create properties for complex keys', () => {
+			const res = utils.createProperty(j, 'foo-bar', 'bar');
+			expect(j(j.objectExpression([res])).toSource()).toMatchSnapshot();
+		});
+
+		it('should create properties for non-literal keys', () => {
```
pksjce: Do we ever create properties with non-literal keys?

lib/transformations/utils.js
```diff
@@ -83,7 +83,7 @@ function findPluginsRootNodes(j, node) {
 function createProperty(j, key, value) {
 	return j.property(
 		'init',
-		j.literal(key),
+		j.identifier(key),
```
okonet: This won't work with anything more complex like `foo-bar` but I think it's okay in the scope of this project.

lib/transformations/__snapshots__/utils.test.js.snap
```diff
@@ -0,0 +1,52 @@
...
+`;
+
+exports[`utils createOrUpdatePluginByName should create a new plugin without arguments 1`] = `"{ plugins: [new Plugin()] }"`;
+
+exports[`utils createOrUpdatePluginByName should merge options objects 1`] = `
+"[new Plugin({
+  foo: true,
+  bar: \"baz\",
+  foo: false,
+  baz-long: true
```
okonet: Here: it's not a valid JS anymore

lib/transformations/utils.test.js
```diff
@@ -23,7 +23,7 @@ describe('utils', () => {
 			expect(j(j.objectExpression([res])).toSource()).toMatchSnapshot();
 		});
 
-		it('should create properties for non-literal keys', () => {
+		xit('should create properties for non-literal keys', () => {
```
okonet: Is this a typo?

lib/transformations/utils.js
pksjce: Yeah, so key in a Property can be Identifier or Literal, so for safety we'll keep it Literal?

pksjce:
Hey! @okonet - I have passed the tests and checked off `ExtractPlugin` changes. I will add more special cases in smaller PRs later? 
  - For "Support Arrays in query string" can you give an example?
  - For "Make sure special cases like webpack-contrib/imports-loader#38 work", shall we take it up in another PR?

This needs more work but ready to merge for now.

lib/transformations/utils.test.js
okonet: Don't think so but we should be aware of the issue.

lib/transformations/utils.js
okonet: Yeah but as far as I can see we don't use anything but identifiers in our code so I'd create a separate GHI, replace it to `j.identifier` and add a comment to the GHI for now. Since this style will break linting of lots of code I guess.

dentuzhik:
I know this is a stupid for this PR, but my searches resulted in pretty much nothing.
Is there any way to use this already, or this is still a WIP?

I tried to clone and run it on my configs which are constructed using [webpack-config](https://github.com/Fitbit/webpack-config), did not work out nicely :)

I have nothing agains manual upgrade, but having an automatic version is always better :)

ev1stensberg:
Yeah, you can use `npm link`, then `webpack-cli --migrate myconfig`

okonet:
@dentuzhik I'm not sure it will work with webpack-config. You can report issues with input and output but I'm not sure we'll support non-standard syntaxes in the beginning.

missing1984:
This is great, is there any way that i can provide a webpack1 config object instead of a file?

okonet:
@missing1984 what do you mean by that? stdin? Or what do you have in mind? Also, not at the moment and I'm not sure what's the use case.

missing1984:
@okonet In our use case, we break the webpack config into many small pieces and merge(webpack-merge) or layer them together to serve different deployment needs. Is the migration cli going to help?

dentuzhik:
@missing1984 I achieve the same thing via webpack-config package, I linked above and seems that in this case automatic migration will not be entirely possible.

okonet:
I think it will work if you operate on objects. The best way to know would be to try running it on your setup and let us know if something isn't working.

thom801:
@pksjce and @okonet you guys are amazing thank you for working so hard and creating this.

arthurgeron:
I'm a little lost, I've tried using `webpack-cli --migrate webpack.config.babel.js` but it says migrate is an unknown argument.

ev1stensberg:
Try without the dashes

PR #65, opened 2017-02-21
fix: Use real paths from argvs instead of dummy hard-coded file

okonet:
This PR is an ongoing work on #6 which enables to pass the path to webpack config as argument. It will overwrite it if confirmed by the user.

### DIFF
--- a/bin/webpack.js
+++ b/bin/webpack.js
@@ -162,9 +162,13 @@ var initInquirer = require('../lib/initialize');
 if(argv.init) {
 	initInquirer(argv._);
 } else if(argv.migrate) {
-	const dummyConfigLoc = require.resolve(path.join(process.cwd(), 'example/webpack.config.js'));
-	const outputConfigLoc = path.join(process.cwd(), 'example/neo-webpack.config.js');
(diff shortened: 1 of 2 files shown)


okonet:
Covers what?

bin/webpack.js
```diff
@@ -162,9 +162,13 @@ var initInquirer = require('../lib/initialize');
 if(argv.init) {
 	initInquirer(argv._);
 } else if(argv.migrate) {
-	const dummyConfigLoc = require.resolve(path.join(process.cwd(), 'example/webpack.config.js'));
-	const outputConfigLoc = path.join(process.cwd(), 'example/neo-webpack.config.js');
-	require('../lib/migrate.js')(dummyConfigLoc, outputConfigLoc);
+	const filePaths = argv._;
```
ev1stensberg: `._argv` is being used in `init` as well. Perhaps move the var outside the loop and pass the same variable to `init` too?

ev1stensberg:
lgtm ✅ 

-  optional fix left as comment

PR #67, opened 2017-02-22
fix: simplify validate-addons logic & npm-exists

ev1stensberg:
Simplifies the usage of validate-addons and replaces the manual mock of
nom-exists with the actual.


ev1stensberg:
Want to remove the mock implementation from so much here, but the entire parser is circular, we should try to use ´jest.mockfn` instead for some of this, so we can remove some manual mocks.

lib/utils/validate-addons.js
```diff
@@ -1,30 +1,28 @@
 const exists = require('./npm-exists');
 const resolvePackages = require('./resolve-packages');
 
+module.exports = (addons) => addons.map( pkg => checkExistence(pkg));
```
okonet: 1. Let's always use `function` declarations when defining functions
1. Let's name this function `validateAddons`
1. There is no need to create the closure in map / create a separate function
1. Please add JSDoc to the exported function since this is the API

```javascript
module.exports = function validateAddons(addons) {
  return addons.map(checkExistence);
}
```

ev1stensberg:
@okonet Changed. I like abstracting each check to an "pure" function in a sense that it looks less verbose. About the inline arrow functions, I agree, but I assumed the `checkExistence` function spoke for itself.

lib/utils/validate-addons.js
```diff
@@ -1,14 +1,27 @@
...
+* @function validateAddons
+*
+* Loops through an array and checks if a package is registered
+* on npm and throws an error if it is not from @checkExistence
+*
+* @param { Array } pkg - Array of packages to check existence of
+* @returns { <Function|Error> } resolvePackages - Returns an process to install the pkg
+*/
+
+module.exports = function validateAddons(addons) {
```
okonet: Do we need to export this function? It seems that it could be easily done in the code where you have to check.

Also, why not `addons.map(checkExistence)`?

lib/utils/validate-addons.js
```diff
@@ -1,14 +1,27 @@
...
+	return addons.map( pkg => checkExistence(pkg));
+};
 
 /*
 * @function checkExistence
```
okonet: I would export this function and do the `map` elsewhere. Should we rename it to `npmPackageExist`?

lib/utils/validate-addons.js
```diff
@@ -1,14 +1,27 @@
...
-module.exports = (addons) => addons.map( pkg => checkExistence(pkg));
+
+/*
+* @function validateAddons
+*
+* Loops through an array and checks if a package is registered
+* on npm and throws an error if it is not from @checkExistence
+*
+* @param { Array } pkg - Array of packages to check existence of
+* @returns { <Function|Error> } resolvePackages - Returns an process to install the pkg
```
okonet: It will return an `Array`, actually.

lib/utils/validate-addons.js
```diff
@@ -1,14 +1,27 @@
...
 
-module.exports = (addons) => addons.map( pkg => checkExistence(pkg));
+
+/*
+* @function validateAddons
+*
+* Loops through an array and checks if a package is registered
+* on npm and throws an error if it is not from @checkExistence
+*
+* @param { Array } pkg - Array of packages to check existence of
```
okonet: It takes an `Array<String>`

okonet:
I wouldn't commit coverage to the repo. The that it should work is to generate it during the check. Let's revert the .gitignore and merge. We need to work on the CI setup a bit more I think.

ev1stensberg:
@okonet Done. I'm leaving the merging to you :)

okonet:
I still can see the coverage file in the changes tab :/

ev1stensberg:
yeah, me2 :( Let me try deleting the folder.

ev1stensberg:
there we go 👍

okonet:
LGTM please squash and merge!

Issue #47, opened 2017-02-04
Integration testing : Obtain webpack 1 fixtures from the community to test 1->2 migrate command.


pksjce:
**Do you want to request a *feature* or report a *bug*?**
This is a requirement for integration testing.

**If this is a feature request, what is motivation or use case for changing the behavior?**
The idea is that once all the 1->2 migrate transformations are ready, we will do some integration testing on our side by asking the community for some webpack 1 fixture configs.

PR #75, opened 2017-02-27
feat: Add integration test runner

pksjce:
Adds scaffolding to chuck in test webpack configs.
Also add any fixes that break these config transformation.

Yep

Yep

**Summary**
Now all wild webpack v1 configs can be addes to the fixtures folder for integration testing.
If any snapshot does not work as expected, an issue can be raised for the relevant fix.

Required for #47 

Nope


lib/transformations/integration.test.js
```diff
@@ -0,0 +1,20 @@
+const fs = require('fs');
+const path = require('path');
+const transform = require('./').transform;
+const no_of_files = 1;
+
+const runTest = (currentConfigPath) => {
+
```
okonet: Can you please remove this empty line?

lib/transformations/integration.test.js
```diff
@@ -0,0 +1,20 @@
...
+	let currentConfig = fs.readFileSync(currentConfigPath, 'utf8');
```
okonet: Should be `const`?

lib/transformations/integration.test.js
```diff
@@ -0,0 +1,20 @@
+const fs = require('fs');
+const path = require('path');
+const transform = require('./').transform;
+const no_of_files = 1;
```
okonet: Please let's stick with `camelCased` convention for variable names.

lib/transformations/loaderOptionsPlugin/loaderOptionsPlugin.js
```diff
@@ -19,7 +20,7 @@ module.exports = function(j, ast) {
 
 	return ast
 		.find(j.ArrayExpression)
-		.filter(path => path.parent.value.key.name === 'plugins')
+		.filter(path => safeTraverse(path, ['parent', 'value', 'key', 'name']) === 'plugins')
```
okonet: 👍

lib/transformations/integration.test.js
```diff
@@ -0,0 +1,20 @@
...
+
+const runTest = (currentConfigPath) => {
+
+	let currentConfig = fs.readFileSync(currentConfigPath, 'utf8');
+	const outputConfig = transform(currentConfig);
+	expect(outputConfig).toMatchSnapshot();
+};
+
+describe('Run integration tests', () => {
+	for(let i =1; i<= no_of_files; i++){
```
okonet: Please fix formatting here. I'm wondering why lint-staged didn't fix it though.

lib/transformations/integration.test.js
```diff
@@ -0,0 +1,20 @@
...
+		test(`run test - webpack${i}.config.js`, () => {
```
okonet: Why use `it` everywhere. Let's switch for consistency.

lib/transformations/integration.test.js
okonet: I'm not sure how is this suppose to work... `no_of_files` is set to 1 so the loop will always be from 0 to 1.

pksjce: Not sure how to pick up the test files. I took a very naive approach. Should I use regex for this then?

pksjce:
Have separated the fix to it's own PR. Will close this one and create a new PR in some time.

Issue #76, opened 2017-02-27
--migrate :: TypeError: Cannot read property 'name' of undefined


arosenthal-r7:
**Do you want to request a *feature* or report a *bug*?**

bug

**What is the current behavior?**
```
NODE_ENV=development webpack-cli --migrate ./webpack.config.js ./webpack.config.js
/Users/arosenthal/dev/webpack-cli/lib/transformations/loaderOptionsPlugin/loaderOptionsPlugin.js:22
		.filter(path => path.parent.value.key.name === 'plugins')
		                                     ^

TypeError: Cannot read property 'name' of undefined
    at ast.find.filter.path (/Users/arosenthal/dev/webpack-cli/lib/transformations/loaderOptionsPlugin/loaderOptionsPlugin.js:22:40)
    at Array.filter (native)
    at Collection.filter (/Users/arosenthal/dev/webpack-cli/node_modules/jscodeshift/dist/Collection.js:65:46)
    at module.exports (/Users/arosenthal/dev/webpack-cli/lib/transformations/loaderOptionsPlugin/loaderOptionsPlugin.js:22:4)
    at transforms.forEach.f (/Users/arosenthal/dev/webpack-cli/lib/transformations/index.js:40:26)
    at Array.forEach (native)
    at transform (/Users/arosenthal/dev/webpack-cli/lib/transformations/index.js:40:13)
    at module.exports (/Users/arosenthal/dev/webpack-cli/lib/migrate.js:9:23)
    at Object.<anonymous> (/Users/arosenthal/dev/webpack-cli/bin/webpack.js:171:30)
    at Module._compile (module.js:570:32)
```

**If the current behavior is a bug, please provide the steps to reproduce.**

^^^ do that with this config:
```
const path = require('path');
const webpack = require('webpack');

module.exports = {
  devtool: '#source-map',

  entry: './src/index.js',

  eslint: {
    configFile: '.eslintrc',
    emitError: true,
    failOnError: true,
    failOnWarning: true,
    formatter: require('eslint-friendly-formatter')
  },

  module: {
    preLoaders: [
      {
        include: [
          path.resolve(__dirname, 'src')
        ],
        loader: 'eslint?',

        test: /\.js$/
      }
    ],

    loaders: [
      {
        loader: 'json',
        test: /\.json$/
      }, {
        include: [
          path.resolve(__dirname, 'src')
        ],
        loader: 'babel',
        test: /\.js$/
      }
    ]
  },

  output: {
[...]

Issue #69, opened 2017-02-22
Migrate: Plugins defined in variable results in TypeError


jouni-kantola:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**
I've declared plugins in a variable to be able to add extra plugins for production build (like `UglifyJsPlugin`), this results in:
```
node_modules/.bin/webpack-cli --migrate webpack.config.js
<path>/node_modules/webpack-cli/lib/transformations/loaderOptionsPlugin/loaderOptionsPlugin.js:21
		.filter(path => path.parent.value.key.name === 'plugins')
		                                     ^

TypeError: Cannot read property 'name' of undefined
```

**If the current behavior is a bug, please provide the steps to reproduce.**

Webpack config:
```
const webpack = require('webpack');

let plugins = [
  new webpack.DefinePlugin({
    alwaysTrue: true
  })
];

module.exports = {
  entry: {
    'app': './src/index.js'
  },
  output: {
    path: 'dist',
    filename: '[name].js'
  },
  module: {
    loaders: [
      {
        test: /\.js$/,
        exclude: /(node_modules)/,
        loader: 'babel-loader',
        query: {
          presets: ['env']
        }
      },
    ]
  },
  devtool: 'source-map',
  plugins: plugins
};
```

**What is the expected behavior?**
Be able to define a separate variable where plugins are specified.

**If this is a feature request, what is motivation or use case for changing the behavior?**
To have one config for different builds, where extra plugins are added for specific builds, then it's probably a common use case to define plugins in an array.

**Please mention other relevant information such as the browser version, Node.js version, Operating System and programming language.**
MacOS
node v7.4.0
webpack-cli: `https://github.com/webpack/webpack-cli.git#a46edbb10e443a0e256038212d4f44042c8e1dc2`

ev1stensberg:
Could you post a screenshot of stacktrace in console?

jouni-kantola:
@ev1stensberg, full output below:
```
> node_modules/.bin/webpack-cli --migrate webpack.config.js
/dev/webpack-output-by-build-type/node_modules/webpack-cli/lib/transformations/loaderOptionsPlugin/loaderOptionsPlugin.js:21
		.filter(path => path.parent.value.key.name === 'plugins')
		                                     ^

TypeError: Cannot read property 'name' of undefined
[...]

PR #77, opened 2017-02-28
fix: add safe traverse to loaderoptionsplugin

pksjce:
Bugfix

**Summary**
Solves #76, #69 

**Other information**


pksjce:
Good idea. Added one. The test case throws undefined error without the fix.

Issue #70, opened 2017-02-23
Migrate: Format output config to use prettier


TheLarkInn:
**Do you want to request a *feature* or report a *bug*?**
Feature

**What is the current behavior?**
Poorly formatted configs, or new ast nodes that are generated appear inline and hard to read after `--migrate`.

**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**
Be _prettier_ (put intended) than what is output now.

**If this is a feature request, what is motivation or use case for changing the behavior?**
I was able to add @jlongster's prettier and pass the string in to be formatted: 

Below is how I implemented this feature (at a very rough level). 

**transformations/index.js**

```javascript
  return pretty.format(ast.toSource(recastOptions), {singleQuote: true, tabWidth: 4});
```

### The results were, damn fine looking!

```javascript
var path = require('path');
var webpack = require('webpack');

// lets take this webpack v1 config and migrate it to v2 !!! <3333

module.exports = {
    debug: true,
    devtool: 'eval',
    entry: ['./src/index'],
    output: {
        path: path.join(__dirname, 'dist'),
        filename: 'index.js'
    },
    plugins: [
        new webpack.OccurenceOrderPlugin(),
        new webpack.optimize.UglifyJsPlugin({
            sourceMap: true
        }),
        new webpack.optimize.LoaderOptionsPlugin({
            debug: true,
            minimize: true
        })
    ],
    resolve: {
        modules: ['foo']
    },
    module: {
        rules: [
            {
                test: /\.js$/,
                use: 'babel-loader',
                include: path.join(__dirname, 'src')
            }
        ]
    }
};
```

Thoughts @pksjce @ev1stensberg @kenwheeler @okonet? Spoke with folks at prettier and they said if passing a direct string is a concern and overhead we could look into the future passing a direct AST.

TheLarkInn:
Could technically be some low hanging fruit for someone to tackle also!!! Given we all agree we should format the ast before its printed to string.

okonet:
I agree we should do that. My only concern is that it still not be a fit for many devs and keeping in mind prettier has almost zero configuration options we won't be able to please everyone. Which is okay for me :)

TheLarkInn:
Yes, that was one of my concerns as well. I would assume that the most suspect ones though would be spaces and tabs which I think we could in the future accept args and then pass to prettier as args/opts

pksjce:
[...]

PR #79, opened 2017-03-02
feat: Add prettier for formatting

pksjce:
Changing formatting to use prettier for  #70

Existing tests are modified

**Summary**
By formatting with prettier, we can use recast as is without worrying about https://github.com/benjamn/recast/issues/242.
Also due to this, we can add yarn support to our project!

Yes, formatting changes in all snapshots.

**Other information**


lib/transformations/index.js
```diff
@@ -38,7 +38,8 @@ function transform(source, transforms, options) {
 	}, options);
 	transforms = transforms || Object.keys(transformations).map(k => transformations[k]);
 	transforms.forEach(f => f(jscodeshift, ast));
-	return ast.toSource(recastOptions);
+	const astStr = ast.toSource(recastOptions);
```
okonet: I wouldn't call it `astStr` since. Let's call it `transformedSource`?

lib/transformations/index.js
```diff
@@ -38,7 +38,8 @@ function transform(source, transforms, options) {
...
+	return prettier.format(astStr, {singleQuote: true, tabWidth: 2});
```
okonet: Users should be able to pass prettier options. Also, some options like `singleQuote` and `tabWidth` might overlap recast options. We need to respect both or remove recast options completely.

lib/transformations/index.test.js
```diff
@@ -31,7 +31,8 @@ module.exports = {
 `;
 
 describe('transform', () => {
-	it('should not transform if no transformations defined', () => {
+	// WIll not be equal unless input is also from prettier
+	xit('should not transform if no transformations defined', () => {
```
okonet: Right, that's pretty much an issue to me. We only should apply `prettier` if we have transformed something.

bin/webpack.js
```diff
@@ -166,9 +166,11 @@ if(argv.init) {
 	if (!filePaths.length) {
 		throw new Error('Please specify a path to your webpack config');
 	}
+	const outputConfigName = filePaths[1] || `${filePaths[0]}v2.js`;
```
okonet: This will make git diff impossible and it will require an extra step from users to remove the old config and rename the new. IMO it defeats the purpose of using this tool. Why would someone like to keep the old config?

lib/transformations/index.js
pksjce: Lets drop recast options. Recast pretty print is not reliable. I'll add prettier options instead

lib/transformations/index.test.js
pksjce: Shall we applly prettier after showing diff to user? I'm not sure how to check if a transformation happened.. can AST's be compared?

bin/webpack.js
pksjce: Yes you are right. But can we keep this until testing is done?

lib/transformations/index.test.js
okonet: I think it can be done by checking the return value. Check out jscodeshift docs since I'm on mobile now

bin/webpack.js
TheLarkInn: @okonet I could see a benefit of being able to have the old one left around especially while still a [WIP]. There were cases on initial trials of migrate where I had wished I specified outputConfigName beacuse I had lost the original, etc.

lib/transformations/index.js
TheLarkInn: +1 Yes DAMP (Descriptive and Meaningful Phrases) ftw.

bin/webpack.js
TheLarkInn: At the least we should store it somewhere, but only as a temp trial for now, or if user specifies.

okonet: Okay but in this case I'd suggest renaming the original config to something with `webpack.config.orig.js` in order to be able to run `webpack` without specifying the path to the new one in case of the default name. Either way, I _guess_ for most people the next operation after transforming would be to try run the build and see if everything works. With a different name it's not possible.

pksjce:
Better to do this after #92

Issue #42, opened 2017-01-27
Scoping: A typical webpack-addon-[name] package


ev1stensberg:
#### 1. Discuss how we want to generate needed rules in a webpack-addon package 

</hr>

#### 2.  Discuss what to expected returned from an addons package for our transformer to run

</hr>

#### 3. Discuss on if we should allow `.babelrc` and such, to also be generated

</hr>

#### 4. Come up with an example of an webpack-addon

TheLarkInn:
I'm going to ping @developit and @addyosmani so they can help mold the user (addon author) and consumer (addon user) api and functionality. 

Below I'll list what I think feels like a great "user" surface api, and describe maybe a workflow.

## Hypothetical "Addon Author Dream Workflow"
@developit is the maintainer of Preact and using webpack makes it trivial for his users to transition from react. But he also has a few boilerplates for initial setups (beyond a react=>preact) migrate that he recommends to people and would love to make the process easier through a webpack addon. 

His imagines a user could simply do: `npm install webpack && webpack --init webpack-addon-preact` and the user would be prompted with any information they would like to know. (Maybe he could even `extend` the `webpack-addon-react` as well: another story).

### Addon Project Creation
Creates a new project locally and runs:
`npm install webpack-addon`, `$(npm bin)/webpack-addon --init`. This would create a best practice recommendation setup for a webpack addon. Including package.json configuration, the correct tags, and prompts @developit for the right information about his addon (name, email, etc similar to npm init). In addition he also has keywords, metadata and special fields that will let webpack users and our org track and list all packages on npm that have been published with this specific metadata. 

### Addon defaults, node api, and addon customization

The webpack-addon repo comes with a "base" set of inquirer questions, and config out of the box. The webpack-addon lib has convenience api's for:
  * Overriding the base config
  * Extending the base config
  * Composing the base config
  * Installing a npm package to a users project
  * Other custom steps <please add if you think of one>

Now all that @Developit has to do is specify the packages he'd like the user to have by default, prompt the user for additional "preact-friendly" libs based on questions, and maybe custom webpack config that he can map the AST transform to a inquirer question. 

He could even pull in a set of `webpack-addon.mixins` for "best-practice" styling, performance, "PWA" :hot:,  etc. configs that could be composed together as well.

### Addon deployment

From there `webpack-addon-preact` can be published simply by running `webpack-addon publish [minor|major|patch]`. This will publish to npm given that his user is already logged in and he has permissions.

[...]

PR #81, opened 2017-03-04
[WIP] - Using webpack-chain for our transforms

pksjce:
This PR integrates webpack chain to create configuration presets from scratch for init creation of webpack config.

WIP - Can add snapshots

WIP

**Summary**
`webpack-chain` seemed like the perfect config builder for us and so I tested it here. It covers our cases pretty neatly and @eliperelman was super fast in fixing any issue I faced.

**Other information**

### DIFF
--- a/lib/init-transforms/entry.js
+++ b/lib/init-transforms/entry.js
@@ -1,5 +1,5 @@
 const run = (config, answer) => {
-	config.entry(answer).end();
+	config.entry('index').add(answer).end();
 	return config;
 };
 
--- a/lib/init-transforms/entry.js
+++ b/lib/init-transforms/entry.js
@@ -1,0 +1,6 @@
+const run = (config, answer) => {
+	config.entry(answer).end();
+	return config;
+};
+
+module.exports = run;
--- a/lib/init-transforms/outputFileName.js
+++ b/lib/init-transforms/outputFileName.js
@@ -1,0 +1,3 @@
+module.exports = (config, answer) => {
+	return config.output.filename(answer).end();
+};
--- a/lib/init-transforms/outputPath.js
+++ b/lib/init-transforms/outputPath.js
@@ -1,0 +1,3 @@
+module.exports = (config, answer) => {
+	return config.output.path(answer).end();
+};
--- a/lib/inquirer-prompt.js
+++ b/lib/inquirer-prompt.js
@@ -15,38 +15,15 @@ const chalk = require('chalk');
 
 module.exports = function prompt(questions, config) {
 	inquirer.prompt(questions).ui.process
-	.reduce(function (newOpts, ans) {
-		return attachAnswers(newOpts, ans);
-	}, Object.assign(config, require('./utils/initial-config')))
-	.flatMap(newOpts => {
-		const parser = require('./parser/index');
-		return parser(null, checkEmptyAnswers(newOpts));
-	})
-	.subscribeOnError( (err) => process.stdout.write(err.toString()));
-};
-
-/*
-* @function attachAnswers
-*
-* Adds the answers from the inquirer instance if the type of the question is
-* equal to the configuration property
-*
-* @param { Object } config - initial questions provided from --init
-* @param { Object } answers - The answers passed from the inquirer instance
-* @returns { Object } newConfig - An new object with the answers added to it
-*/
-
-function attachAnswers(config, answers) {
-	let newConfig;
-	Object.keys(config).forEach( (configName) => {
-		if(configName == answers.name) {
-			newConfig = Object.assign(config, {
-				[configName]: answers.answer
-			});
-		}
+	.subscribe(ans => {
+		config = require(`./init-transforms/${ans.name}`)(config, ans.answer);
+	}, err => process.stdout.write(err.toString()),
+	() => {
+		console.log(chalk.white('We managed to generate a webpack config for you.'));
+		// TODO pass through prettier
+		console.log(chalk.green(JSON.stringify(config.toConfig())));
 	});
-	return newConfig;
-}
+};
 
 /*
 * @function checkEmptyAnswers
--- a/lib/inquirer-prompt.js
+++ b/lib/inquirer-prompt.js
@@ -17,11 +17,12 @@ module.exports = function prompt(questions, config) {
 	inquirer.prompt(questions).ui.process
 	.subscribe(ans => {
 		config = require(`./init-transforms/${ans.name}`)(config, ans.answer);
+
 	}, err => process.stdout.write(err.toString()),
 	() => {
 		console.log(chalk.white('We managed to generate a webpack config for you.'));
 		// TODO pass through prettier
-		console.log(chalk.green(JSON.stringify(config.toConfig())));
+		console.log(chalk.green(config.toConfig()));
 	});
 };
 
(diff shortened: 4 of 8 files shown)


ev1stensberg:
Sweet 👍

ev1stensberg:
For `--init` my concern is that we need to resolve the input from inquirer somewhere before passing it to `webpack-chain` and later insert the given obj props such as entry and output. How are you doing this in `--migrate`? Transforming first and then sending to `webpack-chain`? Should we do the same with `--init` ?

lib/init-transforms/entry.js
```diff
@@ -0,0 +1,6 @@
+const run = (config, answer) => {
+	config.entry('index').add(answer).end();
```
eliperelman: Unless you are going to be doing more chaining here, you can leave off `.end()`, since you are just returning `config` anyway.

ev1stensberg:
@pksjce If you've done the validation part, I'll close my PR. I see this is about init :)

lib/init-transforms/entry.js
pksjce: Sometimes it returned an object which was one level up starting with a `parentPath`
. When I added `.end()` it worked consistently everywhere.

pksjce:
Can't use `webpack-chain` to transform existing config. So we can use it for `init` only. 
Will try to add validations in a couple of days.

ev1stensberg:
@pksjce Not what I meant. The way we build up a webpack config is that we run parser as many times as it needs before exiting ( at least that is the goal ). If we stop after one package, we can't provide multiple packages to the end user. I later want parser to run in parallel to shorten down the execution time.

ev1stensberg:
Forgot to mention that we later need to add a step trying to merge some `validate-addons` packges to cut down the `x` times the parser runs, but that's another thread. If something was unclear in what I've written, I'm happy to rephrase :)

pksjce:
I just sent this PR as a POC of what could be possible. You are more familiar with this codebase. Let's both iterate on this?

pksjce:
actually webpack-chain has an issue where he working on immutable chains. Hopefully API won't change when the transforms are immutable.

ev1stensberg:
@pksjce Sure :) I'll put up a wiki page on the `--init` feature now, what we're suppose to do and so on, and let's take it from there. I think this approach is the best one for building a config, as most of us thinks :)

lib/init-transforms/entry.js
eliperelman: If you are going to return the chain, that's different. Although there isn't much point in returning any config since its all mutable.

lib/inquirer-prompt.js
```diff
@@ -15,38 +15,16 @@ const chalk = require('chalk');
...
-*
-* Adds the answers from the inquirer instance if the type of the question is
-* equal to the configuration property
-*
-* @param { Object } config - initial questions provided from --init
-* @param { Object } answers - The answers passed from the inquirer instance
-* @returns { Object } newConfig - An new object with the answers added to it
-*/
+	.subscribe(ans => {
+		config = require(`./init-transforms/${ans.name}`)(config, ans.answer);
```
ev1stensberg: Mutated here @eliperelman

eliperelman: If `config` here is an instance of webpack-chain, then returning it here won't do anything since all interactions with the API all build up an internal configuration object. Which means:

```js
const cached = config;
config = require(`./init-transforms/${ans.name}`)(config, ans.answer);

assert(config === cached);
```

Returning it won't change what you've passed in, unless you return a different webpack-chain instance, which isn't happening here. It could technically just be:

```js
require(`./init-transforms/${ans.name}`)(config, ans.answer);
```

ev1stensberg: >> Returning it won't change what you've passed in, unless you return a different webpack-chain instance, which isn't happening here. It could technically just be:

That's what I was thinking about. This module should/is recursive

lib/utils/initial-questions.js
```diff
@@ -25,7 +25,10 @@ function inquirerInput(name, message) {
 
 const questions = [
 	inquirerInput('entry','What is the name of the entry point in your application?'),
-	inquirerInput('output','What is the name of the output directory in your application?')
+	inquirerInput('outputPath','What is the name of the output directory in your application?'),
```
okonet: Missing whitespace. @pksjce I'm wondering if you got the lint-staged as a pre-commit hook in place since it should fix such issues automatically. Or are we missing an eslint rule for that?

lib/inquirer-prompt.js
```diff
@@ -15,38 +15,16 @@ const chalk = require('chalk');
...
-	let newConfig;
-	Object.keys(config).forEach( (configName) => {
-		if(configName == answers.name) {
-			newConfig = Object.assign(config, {
-				[configName]: answers.answer
-			});
-		}
+	}, err => process.stdout.write(err.toString()),
+	() => {
+		console.log(chalk.white('We managed to generate a webpack config for you.'));
```
okonet: I'd just write:

`✓ Webpack config generated! Please review it` and I'd use `chalk.green` for it.

lib/inquirer-prompt.js
```diff
@@ -15,38 +15,16 @@ const chalk = require('chalk');
...
+		// TODO pass through prettier
+		console.log(chalk.green(config.toConfig()));
```
okonet: And this one I'd put in a less distracting color. gray should work just as fine I think.

lib/init-transforms/outputPath.js
```diff
@@ -0,0 +1,3 @@
+module.exports = (config, answer) => {
```
okonet: I'd prefer `function` syntax everywhere besides inlined closures. In this case I think `function` is just more explicit.

lib/init-transforms/outputFileName.js
```diff
@@ -0,0 +1,3 @@
+module.exports = (config, answer) => {
```
okonet: Same as ↓: let's use `function` keyword here.

lib/init-transforms/entry.js
```diff
@@ -0,0 +1,6 @@
+const run = (config, answer) => {
```
okonet: Why not `function run(config, answer) {}`? Makes it more explicit IMO

lib/inquirer-prompt.js
pksjce: @eliperelman - Got it :)

lib/init-transforms/outputPath.js
pksjce: Hmmm.. Sure, will do.

pksjce:
@ev1stensberg - How do you want to take this forward?
1. We can make a `webpack-default-init` addon which can be run when no package name is specified. I could start writing that. And then as suggested in #42, we can write specific transforms.
But this default package itself can be a composition of smaller reusable packages.

2. Not sure how to make this immutable. I see in #42 that you guys agreed on precedence order. Why dont we stick to that? last one wins?

3. I think I will close this PR and open one with solely the addons.

ev1stensberg:
@pksjce I'm thinking about integrating the bare minimum inside our source code, so we don't have to install a package to get the default behaviour for init. This way we save time, by not installing. Somewhat more verbose code, but I think it is well worth it. (1)

Use obj.assign? (2)

Also, let's focus on the transforms for now, while trying to find out where we can use yeoman + webpack-chain.

pksjce:
@ev1stensberg is planning on better integration in a new PR

PR #82, opened 2017-03-04
Add webpack validation to init and initial parser transformations

ev1stensberg:

Let's us validate an webpack config after creating one and also introduces the first transformation rules.

Yes

No

**Summary**

Adds validation and initial transformation rules to get a "first usable config" with only basic rules. 

Yes

**Other information**

We should get @eliperelman involved, we may not need transformation rules for the parser. PR should be merged and then we could refactor later, which is the next step in the process ( see if we can integrate webpack-chain )


ev1stensberg:
Moving this to #81

Issue #84, opened 2017-03-09
Feature - using stats presets from CLI


bigfanjs:
**Do you want to request a *feature* or report a *bug*?**
*feature* Webpack CLI

**What is the current behavior?**
You have to create a config file to use stats presets like errors-only

**What is the expected behavior?**
Webpack CLI should display information using a preset just like what happens in configFiles

> webpack entry.js bundle.js --stats errors-only
       [0] ./entry1.js 0 bytes {0} [built]

**Please mention other relevant information such as the browser version, Node.js version, Operating System and programming language.**
OS: Windows 10
Node.js(v7.6.0)
Lang: JS

ev1stensberg:
Could you elaborate more on expected behaviour and replace current behaviour with the actual behaviour?

AdelJS:
I think the idea is pretty obvious; The stats shortcuts(errors-only, none, minimal, normal, verbose) that you've probably been using in your config files (see below) simply have to be available in the command line as well.

````js
{
  stats: "errors-only"
}
````
Which outputs this:
> [0] ./entry1.js 0 bytes {0} [built]

In case of this *errors-only* preset, if you want to achieve it through the cli, even if you could hide modules(using --hide-modules command) there still would be a hash, version and time in the output. And you would end up creating a webpack.config.js for such a straightforward configuration.

The actual behavior has to be something like this:
> webpack entry.js bundle.js --stats errors-only
[0] ./entry1.js 0 bytes {0} [built]

It's a matter of a couple lines of code.

Thank you!

AdelJS:
I think I'll clone Webpack and add this feature myself!

ev1stensberg:
We'll see what we can do :) This is not prio nr.1, so if you wanna implement this yourself, feel free! We'll be here if you want some help on where to put it.

ev1stensberg:
FWIW- You should send the PR to webpack directly, as this is under `bin`. Planning to merge that into our bin folder later.

AdelJS:
Oh thank you, I appreciate it. OK, I've never actually contributed to an open source project before but I would love to, and especially for Webpack. I am very good at JavaScript but I am sucks at open source contribution.

Would you please help me with the commit format and testing my work.

ev1stensberg:
No problems! I recommend you to read a bit about stats to start of with. You can find places where webpack use stats object by globally searching the option in the webpack repo from your text editor.

[...]

PR #85, opened 2017-03-11
Feature: adds a stats option

bigfanjs:
display information using presets as shortcuts

Feature: adds a stats option

display information using presets as shortcuts

*feature*

Actually Not!

**Summary**
One cool feature in webpack config files is that you can display stats information about your bundling process using presets as shortcuts. I would love this feature to be avaliable on the CLI as well. #84

### DIFF
--- a/bin/process-options.js
+++ b/bin/process-options.js
@@ -21,6 +21,10 @@ module.exports = function processOptions(yargs, argv) {
 
 	var firstOptions = Array.isArray(options) ? (options[0] || {}) : options;
 
+	ifArg('stats', function (value) {
+		options.stats = value;
+	});
+
 	if(typeof options.stats === 'boolean' || typeof options.stats === 'string') {
 		var statsPresetToOptions = require('webpack/lib/Stats.js').presetToOptions;
 		options.stats = statsPresetToOptions(options.stats);
@@ -50,9 +54,11 @@ module.exports = function processOptions(yargs, argv) {
 		outputOptions.assetsSort = value;
 	});
 
-	ifArg('display-exclude', function(value) {
-		outputOptions.exclude = value;
-	});
+	if (!argv.stats) {
+		ifArg('display-exclude', function(value) {
+			outputOptions.exclude = value;
+		});
+	}
 
 	if(!outputOptions.json) {
 		if(typeof outputOptions.cached === 'undefined')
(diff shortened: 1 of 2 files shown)


bin/process-options.js
```diff
@@ -50,54 +54,60 @@ module.exports = function processOptions(yargs, argv) {
...
 		if(typeof outputOptions.cached === 'undefined')
 			outputOptions.cached = false;
 		if(typeof outputOptions.cachedAssets === 'undefined')
 			outputOptions.cachedAssets = false;
 
-		ifArg('display-chunks', function(bool) {
-			outputOptions.modules = !bool;
-			outputOptions.chunks = bool;
-		});
+		if (!argv.stats) {
```
ev1stensberg: Display chunks is used without stats, would be removal of an argv here too

bin/webpack.js
```diff
@@ -140,12 +139,30 @@ yargs.options({
 		type: 'boolean',
 		group: DISPLAY_GROUP,
 		describe: 'Show more details'
+	},
+	'stats': {
+		type: 'string',
+		describe: 'Display information using a preset(Accepted presets: none, errors-only, minimal, normal, verbose)',
```
ev1stensberg: We should try to make this better looking in the terminal. Shorten the text, and make the options be inline. Should also split this string to two lines

AdelJS:
OK, this is how we use the options:

say we have an entry point(entry.js)
which contain:
````js
console.log('I am the entry point');
````

This one only output if there are errors. And since we don't have any in our entry point here is what happens:
````bash
> webpack entry.js bundle.js --stats errors-only
[0] ./entry.js 0 bytes {0} [built]
````

and now lets add some code which we did not set up a loader for:
````js
const el = <div></div>;
console.log( el );
````
and by running the command again:
````bash
> webpack entry.js bundle.js --stats errors-only
ERROR in ./entry.js
Module parse failed: C:\path\to\project\entry.js Unexpected token (1:11)
You may need an appropriate loader to handle this file type.
| const el = <div></div>;
|
| console.log( el );
````

The second preset is **none** which does almost the same except it does not output any stack trace even if there is an error:
````bash
> webpack entry.js bundle.js --stats errors-only
[0] ./entry.js 232 bytes {0} [built] [failed] [1 error]
````

Unlike the **errors-only** preset and in the case of a correct code the **none** preset output, the modules with no colors.

Just like **errors-only** the **minimal** preset outputs when an error happens plus the chunks:

````js
> webpack entry.js bundle.js --stats minimal
chunk    {0} bundle.js (main) 49 bytes [entry] [rendered]
   [0] ./entry.js 49 bytes {0} [built]
````

As the docs say, the **normal** preset makes a standard output which includes hash, time and chunk:
````bash
> webpack entry.js bundle.js --stats normal
Hash: d91d1b54ca54c9ff8deb
Time: 88ms
chunk    {0} bundle.js (main) 49 bytes [entry] [rendered]
   [0] ./entry.js 49 bytes {0} [built]
````

And finally the **verbose** preset outputs everything:
````bash
> webpack entry.js bundle.js --stats verbose
Hash: d91d1b54ca54c9ff8deb
Version: webpack 2.2.1
Time: 121ms
     Asset     Size  Chunks             Chunk Names
bundle.js  2.68 kB       0  [emitted]  main
Entrypoint main = bundle.js
chunk    {0} bundle.js (main) 49 bytes [entry] [rendered]
    [0] ./entry.js 49 bytes {0} [depth 0] [built]
   [0] ./entry.js 49 bytes {0} [depth 0] [built]
````


Setting both stats argument and any other argument matching *hide-modules* or *verbose* or those starting with *display-* I throw an error says you have to pick only one:

````bash
> webpack entry.js bundle.js --stats verbose --verbose
C:\Users\Adel\my-clones\webpack-cli\bin\webpack.js:161
                        throw new Error(`Please omit ${ key } if you are using stats`);
                        ^

Error: Please omit verbose if you are using stats
    at Object.<anonymous> (C:\Users\Adel\my-clones\webpack-cli\bin\webpack.js:161:10)
    at Module._compile (module.js:571:32)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
    at Module.runMain (module.js:605:10)
    at run (bootstrap_node.js:422:7)
    at startup (bootstrap_node.js:143:9)
    at bootstrap_node.js:537:3
````

AdelJS:
While reviewing my changes I've come across a weird behavior in the bin/process-options.js file:

The Yargs library if given an option with a Boolean type, it defaults to false if not set right?

In the bin/process-options.js there is an .ifArg() method which makes sure an argument is not undefined before executing the callback function. Which means if the type of an argument is Boolean it would execute the callback.

First, I don't know why instead of just explicitly grabbing it from the argv array, they are passing those boolean arguments through this method checking whether their type are undefined or not before getting the value and finally use it. That is actually not important and especially if the performance doesn't effect.

What is actually important is that when given a boolean option these ifArg() methods which work quite well and with no side-effects in the regular control of displaying information, seem like they kinda override the output options of stats preset as long as there is nothing to stop them from working.

For example:
forget about my changes and suppose you have a config file:

````js
module.exports = {
   entry: './entry.js',
   output: {
       filename: 'bundle.js'
   },
   stats: 'verbose'
};
````
This gonna output:
````bash
> webpack
Hash: 914c60bcad1406c9ac9b
Version: webpack 2.2.1
Time: 54ms
     Asset     Size  Chunks             Chunk Names
bundle2.js  2.51 kB       0  [emitted]  main
   [0] ./entry2.js 0 bytes {0} [built]
````

But the presetToOtions() static method of the Stats class in the webpack/lib/stats.js file says that *verbose* should output almost everything.

But when taking my changes into account it seems like it works because I prevent those ifArg() methods from working when stats option is defined. I don't know if that is a mistake or just they meant it to be like that!

ev1stensberg:
@AdelJS It is the correct behaviour. Generally speaking, verbose is suppose to print everything. If an object is not defined, it defaults to false, that's regular JS. I think that the reason for `ifArg` is to let the function execute a callback and that we can get a variable passed to assign our `argv` to.

pksjce:
If I'm understanding correctly, what you are saying is that the `stats` option does not work as documented? So this change makes it correct along with a warning not to use certain options if you are already using `stats`. Correct? 
Can you check if this is fixed on `webpack` repository somehow? There were a bunch of PRs related to `stats` [here](https://github.com/webpack/webpack/pulls?utf8=%E2%9C%93&q=is%3Apr%20is%3Aclosed%20stats%20)

bigfanjs:
Thank you @pksjce! ^^. So Sorry! No longer interested in this PR!

PR #86, opened 2017-03-15
Validation and parser transformations

ev1stensberg:
This PR is the initial work on the transformations. @pksjce you can add your work at the respective folders or tweak it as you want to. I suggest that @okonet and @pksjce is starting on the transformations, while me and @eliperelman integrate webpack-chain, yeoman and get our addon system viable. Let's make a final push for first release, it's not far away if we put some work into it the next couple of weeks.

### DIFF
--- a/lib/inquirer-prompt.js
+++ b/lib/inquirer-prompt.js
@@ -18,11 +18,13 @@ module.exports = function prompt(questions, config) {
 	.reduce(function (newOpts, ans) {
 		return attachAnswers(newOpts, ans);
 	}, Object.assign(config, require('./utils/initial-config')))
-	.flatMap(newOpts => {
+	.map(function(newOpts) {
 		const parser = require('./parser/index');
 		return parser(null, checkEmptyAnswers(newOpts));
 	})
-	.subscribeOnError( (err) => process.stdout.write(err.toString()));
+	.subscribeOnError((err) => {
+		process.stdout.write(err.toString());
+	});
 };
 
 /*
--- a/lib/inquirer-prompt.js
+++ b/lib/inquirer-prompt.js
@@ -23,7 +23,8 @@ module.exports = function prompt(questions, config) {
 		return parser(null, checkEmptyAnswers(newOpts));
 	})
 	.subscribeOnError((err) => {
-		process.stdout.write(err.toString());
+		console.log(err);
+	//	process.stdout.write(err.toString());
 	});
 };
 
--- a/lib/inquirer-prompt.js
+++ b/lib/inquirer-prompt.js
@@ -18,7 +18,7 @@ module.exports = function prompt(questions, config) {
 	.reduce(function (newOpts, ans) {
 		return attachAnswers(newOpts, ans);
 	}, Object.assign(config, require('./utils/initial-config')))
-	.map(function(newOpts) {
+	.map( (newOpts) => {
 		const parser = require('./parser/index');
 		return parser(null, checkEmptyAnswers(newOpts));
 	})
--- a/lib/parser/init-transform.js
+++ b/lib/parser/init-transform.js
@@ -1,0 +1,11 @@
+const jscodeshift = require('jscodeshift');
+const fs = require('fs');
+const path = require('path');
+
+module.exports = function initTransform(opts) {
+
+	const templateFP = path.join(__dirname, 'utils', 'template.js');
+	const template = fs.readFileSync(templateFP, 'utf8');
+	const ast = jscodeshift(template);
+
+};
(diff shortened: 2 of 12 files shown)


lib/parser/init-transform.js
```diff
@@ -0,0 +1,18 @@
+const fs = require('fs');
+const path = require('path');
+const transform = require('./transformations').transform;
+
+module.exports = function initTransform(opts) {
+	global.options = opts;
+	const templateFP = path.join(__dirname, 'utils', 'template.js');
+	const template = fs.readFileSync(templateFP, 'utf8');
+	const ast = transform(template);
```
okonet: This is confusing since you'll get the new source and not the AST

lib/parser/init-transform.js
```diff
@@ -0,0 +1,18 @@
...
+	try {
+		const logicPath = process.cwd() + '/webpack.config.js';
+		fs.writeFileSync(logicPath, ast, 'utf8');
```
okonet: ... and this made me think for a couple of seconds by looking at this. Let's rename `ast` to something like `new Source`

lib/parser/init-transform.js
```diff
@@ -0,0 +1,18 @@
+const fs = require('fs');
+const path = require('path');
+const transform = require('./transformations').transform;
+
+module.exports = function initTransform(opts) {
+	global.options = opts;
+	const templateFP = path.join(__dirname, 'utils', 'template.js');
+	const template = fs.readFileSync(templateFP, 'utf8');
```
okonet: It looks like you never remove the tmp file. You might want to take a look at https://www.npmjs.com/package/tmp to automate this.

lib/transformations/__snapshots__/utils.test.js.snap
```diff
@@ -1,24 +1,8 @@
...
-exports[`utils createLiteral should create basic literal 1`] = `
-Object {
-  "comments": null,
-  "loc": null,
-  "regex": null,
-  "type": "Literal",
-  "value": "strintLiteral",
-}
-`;
+exports[`utils createLiteral should create basic literal 1`] = `"\\"stringLiteral\\""`;
```
okonet: Hmmm, seems like this is from another commit and doesn't relate to this PR. Do you mind rebasing it?

lib/parser/init-transform.js
ev1stensberg: I'm aware of this :) Thanks for the heads-up

lib/parser/init-transform.js
ev1stensberg: Is the initial work, I want @pksjce to work on this, I'm not fluent enough in writing AST's yet.

lib/parser/init-transform.js
ev1stensberg: Sure

lib/transformations/__snapshots__/utils.test.js.snap
ev1stensberg: Sure!

lib/inquirer-prompt.js
```diff
@@ -18,11 +18,14 @@ module.exports = function prompt(questions, config) {
 	.reduce(function (newOpts, ans) {
 		return attachAnswers(newOpts, ans);
 	}, Object.assign(config, require('./utils/initial-config')))
-	.flatMap(newOpts => {
+	.map(function(newOpts) {
```
ev1stensberg: since we used function at reduce, I figured out we should stick with the convention

okonet: Hmm, I thought the convention is to use `function` only for declarations, not for closures :/ Can you point me to the other usage?

lib/parser/init-transform.js
okonet: 👍

lib/inquirer-prompt.js
ev1stensberg: I'll change it if so, didn't know that was the rule :)

ev1stensberg:
Okay, done with what I think needs to be addressed first @okonet.  I think that we further need to discuss the transformations and how to use that effectively, with some of the current rules if doable. Let's bring it up tomorrow at the meeting :)

ev1stensberg:
@okonet readfile didn't make validations work. I guess we can try change this a bit or try to validate after writing?

ev1stensberg:
Closing this PR for now, maybe reopening later. We should first:

1. Add yeoman support
2. Hook the generators up with our internals for validation and attachments of objects to webpack-chain
3. Validate the config

PR #105, opened 2017-04-07
Webpack-CLI version 1

ev1stensberg:
# Webpack CLI version 1 

- AST's

- Prettier

- Windows support 

- Documentation

- Tests

- Generator with adapter for us later to use for path validation before sending to AST's


lib/creator/transformations/index.js
```diff
@@ -11,9 +17,30 @@
 * @returns { <Void> } TODO
 */
 
-module.exports = function runTransform(transformObject) {
+const transformsObject = {
+	entryTransform,
+	outputTransform
+};
```
pksjce: This should be passed to `runTransform` somehow.

ev1stensberg: Doing through `p-lazy` I suppose ;)

ev1stensberg: Don't review yet, will notify when there's something better to review, this isn't nearly done at all

lib/creator/transformations/output/output.js
```diff
@@ -0,0 +1,24 @@
+const webpackOutputTypes = require('./output-types');
+const utils = require('../../../transformations/utils');
+
+module.exports = function(j, ast, webpackProperties) {
+	if(!webpackProperties['output']) {
+		return ast;
+	} else if(webpackProperties['output'].length) {
+		throw new Error('Supplying output with only no options is not supported.');
```
ev1stensberg: Remind me to change this l8tr :)

pksjce:
Awesome work @ev1stensberg
^following

ev1stensberg:
@pksjce I'll finish the todo and implement safeTraverse. Removed `utils.createProperty`, cause they result in string values and different nodes than I want.

ev1stensberg:
Won't remove the code from the default generator yet, as I've got to build a default scaffold once this is done. Codacy is irrelevant for now, as some of the code used as utils, will be moved later.

lib/creator/index.js
```diff
@@ -15,18 +16,26 @@ const runTransform = require('./transformations/index');
 */
 
 function creator(options) {
-	const env = yeoman.createEnv(null, null, new WebpackAdapter());
-	const generatorName = options ? replaceGeneratorName(path.basename(options)) : 'webpack-default-generator';
+	let env = yeoman.createEnv('webpack', null, new WebpackAdapter());
+	const generatorName = options ? replaceGeneratorName(path.basename(options[0])) : 'webpack-default-generator';
```
okonet: This will be a runtime error if options isn't Array

lib/creator/transformations/context/context.js
```diff
@@ -0,0 +1,19 @@
+module.exports = function(j, ast, webpackProperties) {
+	function createContextProperty(p) {
+		if(p.parent.value.type === 'AssignmentExpression') {
+			if(webpackProperties['context'].__paths) {
+				let contextProp = webpackProperties['context'].__paths[0].value.program.body[0].expression;
+				return p.value.properties.push(j.property('init', j.identifier('context'), contextProp));
+			} else {
```
okonet: No need for this `else` I belive

lib/creator/transformations/context/context.js
```diff
@@ -0,0 +1,19 @@
...
+				return p.value.properties.push(
+					j.property('init', j.identifier('context'), j.literal(webpackProperties['context']))
+				);
+			}
+		}
+	}
+	if(webpackProperties['context'] && webpackProperties['context'].length) {
+		return ast.find(j.ObjectExpression).filter(p => createContextProperty(p));
+	} else {
```
okonet: Same here?

lib/creator/transformations/module/module.js
```diff
@@ -0,0 +1,164 @@
...
+														Object.keys(subProps[key][optionProperty]).forEach( subOptionProp => {
+															if(typeof(subProps[key][optionProperty][subOptionProp]) === 'string') {
+																loaderProperty.properties.push(
+																	j.property('init',
+																	j.identifier(subOptionProp),
+																	j.literal(subProps[key][optionProperty][subOptionProp])
+																)
+															);
+															} else {
+																let subSubProps = j.objectExpression([]);
```
okonet: Sorry but this is not possible to read. Line lengths are way too over the eslint limit and it should be split into smaller functions I think.

lib/creator/transformations/node/node.js
```diff
@@ -0,0 +1,21 @@
...
+				Object.keys(webpackProperties.node).forEach( (webpackProp) => {
+					prop.value.properties.push(
+						j.property('init', j.identifier(webpackProp), j.literal(webpackProperties.node[webpackProp]))
+					);
+				});
+			});
+		}
+	}
+	if(webpackProperties['node'] && typeof(webpackProperties['node']) === 'object') {
+		return ast.find(j.ObjectExpression).filter(p => createNodeProperty(p));
```
okonet: No need for the function. Pass the `createNodeProperty` to the filter.

lib/creator/yeoman/webpack-generator.js
```diff
@@ -1,22 +1,141 @@
...
 	prompting() {
-		return this.prompt([Input('entry', 'What is the name of the entry point in your application?'),
-			Input('output', 'What is the name of the output directory in your application?')]).then( (answer) => {
-				this.configuration.webpackOptions = answer;
-			});
-	}
-	config() {}
-	childDependencies() {
-		this.configuration.childDependencies = ['webpack-addons-preact'];
+		this.configuration.webpackOptions.entry = {
```
pksjce: Last I saw, there were inquirer questions here.

lib/creator/index.js
```diff
@@ -15,18 +16,30 @@ const runTransform = require('./transformations/index');
...
 		env.registerStub(defaultGenerator, 'webpack-default-generator');
 	}
 
 	env.run(generatorName)
 	.on('end', () => {
-		return runTransform(env.getArgument('configuration'));
+		if(generatorName !== 'webpack-default-generator') {
+			//HACK / FIXME
+			env = env.options.env;
+			return runTransform(env.configuration);
```
pksjce: Send `env.configuration.webpackOptions`

lib/creator/yeoman/webpack-generator.js
ev1stensberg: Removed for now, testing from addon perspective, will add a full scaffold for this later.

lib/creator/index.js
ev1stensberg: Levels of abstraction in the transform relying on props on `configuration` too, not only webpackOptions.

lib/creator/index.js
```diff
@@ -15,18 +16,30 @@ const runTransform = require('./transformations/index');
...
 	} else {
 		env.registerStub(defaultGenerator, 'webpack-default-generator');
 	}
 
 	env.run(generatorName)
 	.on('end', () => {
-		return runTransform(env.getArgument('configuration'));
+		if(generatorName !== 'webpack-default-generator') {
+			//HACK / FIXME
+			env = env.options.env;
```
ev1stensberg: https://github.com/yeoman/generator/pull/1011, ongoing discussion on how to extract the config file and update it from yeomans-side.

ev1stensberg:
Don't merge just yet if everything is ok. Some work left.

lib/creator/index.js
ev1stensberg: But it is :) getting caught in `resolve-packages`

ev1stensberg:
@okonet @pksjce Anything you want changed before merge? Tried my best at the styling part @okonet

lib/creator/transformations/devtool/devtool.js
```diff
@@ -5,8 +5,11 @@ module.exports = function(j, ast, yeomanConfig) {
 	const webpackProperties = yeomanConfig.webpackOptions;
 
 	function createDevToolProperty(p) {
-		const propValue = j.literal(webpackProperties['devtool']);
-		transformUtils.checkIfExistsAndAddValue(j, p, 'devtools', propValue);
```
pksjce: Why would you remove this. Its breaking the snapshots. Please revert

lib/creator/transformations/devtool/__snapshots__/devtool.test.js.snap
```diff
@@ -7,18 +7,33 @@ exports[`devtool transforms correctly using "devtools-0" data 1`] = `
...
-	entry: 'index.js',
-	output: {
+    entry: 'index.js',
+    output: {
 		filename: 'bundle.js'
 	},
-	devtools: 'source-map'
+
+    devtools: 'eval',
+    devtool: 'cheap-module-source-map'
```
pksjce: This is wrong

lib/creator/index.js
ev1stensberg: Maybe we could find a way to destruct & send each option to respective transforms?

lib/creator/transformations/devtool/devtool.js
ev1stensberg: Oh right. Totally forgot. It's literal value. We need to change this based on if it's an actual literal or if it should be treated as an identifier

ev1stensberg:
@pksjce Let's make some utility around creating transforms for nested props on the future transform, such as arrays.

ev1stensberg:
So I've done some changes. I'm trying to get this done by monday and there's still things left. What I feel is essential right now is cutting down the boilerplate code and reduce the amount of specific code. We are basically doing the same transform strategies across all the properties, so I'll try to make some general functions like the ones I've started with.

ev1stensberg:
@pksjce We should try making end-to-end test instead of integration tests for migrate and init, ( the whole process ) , as this is breaking the tests due async work taking to long in node.v4

lib/creator/transformations/context/__testfixtures__/context-2.input.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+	entry: 'index.js',
+	output: {
+		filename: 'bundle.js'
+	}
+}
```
ev1stensberg: not gonna work. I was planning to integrate it for every transform in v2 for `webpack-cli --add -someRules`

bin/convert-argv.js
```diff
@@ -122,7 +122,7 @@ module.exports = function(yargs, argv, convertOptions) {
 	function processConfiguredOptions(options) {
 		if(options === null || typeof options !== 'object') {
 			console.error('Config did not export an object or a function returning an object.');
-			process.exit(-1);
+			process.exit(-1); // eslint-disable-line
```
okonet: I'm not sure it's good to do this. AFAIK you should use `process.exitCode` instead.

bin/webpack.js
```diff
@@ -1,45 +1,26 @@
...
-		});
-
-		// TODO: If path is wrong, leave it to higher powers until we publish on npm
-		// Could also install and load and install through github
-		return require(path.join(dirPath, 'webpack.js'))
-	});
-	*/
-}
+	} //eslint-disable-next-line
+} catch(e) {}
```
okonet: No eror handling?

bin/webpack.js
```diff
@@ -1,45 +1,26 @@
...
-	.usage('webpack-cli ' + require('../package.json').version + '\n' +
-		'Usage: https://webpack.github.io/docs/cli.html\n' +
+	.usage('webpack ' + require('../package.json').version + '\n' +
+		'Usage: https://webpack.js.org/api/cli/\n' +
 		'Usage without config file: webpack <entry> [<entry>] <output>\n' +
 		'Usage with config file: webpack');
 
 require('./config-yargs')(yargs);
 
+
```
okonet: Please remove this extra line

bin/webpack.js
ev1stensberg: I did a try on this earlier, figure its best to push a commit for this after we've got it linked with the webpack repo to make sure the catch case is working ( see how they do it in above block )

bin/webpack.js
ev1stensberg: Extends options provided from the old cli

bin/convert-argv.js
ev1stensberg: Got any place I could read up on why? ( Also taken from the old cli )

lib/creator/index.js
addyosmani: > Going to close this, as we're going to transition into using yeoman generators explicitly pretty soon. For now, running against a local fork is fine, but I'm a bit nervous about how the this.composeWith will turn out instead of env.configuration, when we run the webpack-cli generator in near future.

Based on this comment from Even, I think transitioning to explicit generator usage was the right call here. Let's file bugs upstream if there's more we can do to make this type of transition easier.

lib/creator/yeoman/utils/entry.js
```diff
@@ -0,0 +1,50 @@
...
+		let entryIdentifiers;
+		let entryProperties;
+		if(answer['entryType'].indexOf('Array') >= 0) {
+			self.prompt([
+				Input('arrayNumber', 'write your entry points seperated by comma')
+			]).then( (arrayAnswer) => {
+				webpackEntryPoint = arrayAnswer['arrayNumber'].split(',');
+				resolve(webpackEntryPoint.map( (prop) => {
+					let newProp = `'${prop}'`;
+					return newProp;
```
addyosmani: Is there a reason not to directly return `'${prop}'` here? (i.e drop line 15)

lib/creator/yeoman/webpack-generator.js
```diff
@@ -1,138 +1,92 @@
...
+			entryQuestions(self, entryTypeAnswer).then(entryOptions => {
+				this.configuration.config.webpackOptions.entry = entryOptions;
+			}).then( () => {
+
+				this.prompt([
+					CheckList('outputType', 'What kind of output properties do you want?', [
+						'chunkFilename',
+						'filename',
+						'path'
+					])
```
addyosmani: I wonder if the average user or beginner is going to know what each of these mean. I'd almost suggest not prompting, scaffold by default and include some boilerplate comments in case they need to customize or remove. Interested in @pksjce's thoughts here in case I'm off base :)

lib/creator/yeoman/webpack-generator.js
```diff
@@ -1,138 +1,92 @@
...
+				]).then( (outputTypeAnswer) => {
+					// Ask different questions to output
+					OutputQuestions(self, outputTypeAnswer['outputType']).then(outputOptions => {
+						this.configuration.config.webpackOptions.output = outputOptions;
+					}).then( () => {
+						// Ask if the user wants to use extractPlugin
+						this.prompt([
+							Input('extractPlugin', 'If you got a css file, what\'s it named?')
```
addyosmani: "What file should your CSS be saved to? Hit enter to skip" or similar?

lib/creator/yeoman/webpack-generator.js
```diff
@@ -1,138 +1,92 @@
...
-								hello: "'world'"
+		let done = this.async();
+		let self = this;
+
+		this.configuration.config.webpackOptions.module = getDefaultLoaders();
+		this.configuration.config.webpackOptions.plugins = getDefaultPlugins();
+		this.prompt([
+			RawList('entryType', 'What kind of entry point do you want?', ['Array', 'Function', 'String', 'Object'])
+		]).then( (entryTypeAnswer) => {
+			// Ask different questions for entry points
```
addyosmani: I personally think this question is a little low level for the target audience. I'd shoot for a few levels higher up - like, "Which JavaScript file is your main entry point?" then pick the right choice for them. Or you could ask another question like...

"Would you like to define one or more main JavaScript bundles?"

If one, let's just go for one entry point with the next question being "Which file would you like to be your main entry point?"

if it's more than one, could iterate across the number of bundles and let them specify the entry for each one.

lib/creator/yeoman/webpack-generator.js
```diff
@@ -1,138 +1,92 @@
...
+								this.configuration.config.webpackOptions.plugins.push(
+									'new ExtractTextPlugin(\'styles.css\')'
+								);
+								this.configuration.config.webpackOptions.module.rules.push({
+									test: new RegExp(/\.css$/),
+									use: `ExtractTextPlugin.extract({
+										fallback: 'style-loader',
+										use: 'css-loader'
+									})`
+								});
```
addyosmani: Should we also ask any questions about their HTML? I see a lot of advanced projects using the HTML Webpack plugin for wiring up scripts. Perhaps it's worth also including here or is that overkill?

lib/creator/yeoman/webpack-generator.js
addyosmani: I also wonder if we might be able to (by default) include some notion of the Webpack performance budgets to get them in that frame of mind earlier on.

lib/creator/yeoman/webpack-generator.js
ev1stensberg: Could be overkill as well as another callback that needs to be added. I'm really favoring rxJS w/ inquirer for prompting, cause the amount of callbacks are unbelievable with regular syntax. I'm open to the suggestion, this would inject the webpack bundle for them, right? If so, definitely should be added.

lib/creator/yeoman/utils/entry.js
ev1stensberg: Was logging output in dev, will remove.

lib/creator/yeoman/utils/output.js
```diff
@@ -0,0 +1,16 @@
+const Input = require('webpack-addons').Input;
+
+module.exports = (self, answer) => {
+	return new Promise( (resolve) => {
```
sindresorhus: Same as earlier, don't wrap it in a Promise constructor.

lib/creator/yeoman/utils/entry.js
```diff
@@ -0,0 +1,50 @@
+const Input = require('webpack-addons').Input;
+
+module.exports = (self, answer) => {
+	return new Promise( (resolve) => {
```
ev1stensberg: http://bluebirdjs.com/docs/anti-patterns.html#the-explicit-construction-anti-pattern I'm sure you're refering to this, but I'm not sure how to make this async without using the antipattern. If you could elaborate that would be great. Using a promise here to avoid breaking inquirer. If you remove the promise the prompts get all invoked at once

lib/creator/yeoman/utils/output.js
ev1stensberg: Deleted in newest commit from @addyosmani ´s comment on UX.

bin/webpack.js
```diff
@@ -1,45 +1,25 @@
...
-
-process.title = 'webpack';
 
+// Local version replace global one
 try {
-	var localWebpack = require.resolve(path.join(process.cwd(), 'node_modules', 'webpack-cli', 'bin', 'webpack.js'));
-	if(localWebpack && path.relative(localWebpack, __filename) !== '') {
+	var localWebpack = require.resolve(path.join(process.cwd(), 'node_modules',
+	'webpack', 'node_modules', 'webpack-cli', 'bin', 'webpack.js'));
+	if(__filename !== localWebpack) {
```
ev1stensberg: Added in new commit, thanks for the heads-up! A bit hard to test this in dev, but used `resolveCwd.silent`, thanks for an excellent api!

lib/creator/yeoman/webpack-generator.js
ev1stensberg: Check the latest commit and tell me what you think!

lib/creator/yeoman/webpack-generator.js
ev1stensberg: Removed this and just asked for path

lib/creator/yeoman/webpack-generator.js
ev1stensberg: Changed in latest commit

lib/creator/yeoman/utils/entry.js
sindresorhus: It's already async, so not sure what you're talking about, but here's how you could do it (Note how the promise is always returned):

```js
const InputValidate = require('webpack-addons').InputValidate;
const validate = require('./validate');

module.exports = (self, answer) => {
	let webpackEntryPoint;

	let entryIdentifiers;
	let entryProperties;

	let ret;

	if(answer['entryType'] == 'Multiple') {
		ret = self.prompt([
			InputValidate(
				'multipleEntries',
				'Write the name you want for your modules seperated by comma',
				validate
			)
		]).then( (multipleEntriesAnswer) => {
			webpackEntryPoint = {};
			entryIdentifiers = multipleEntriesAnswer['multipleEntries'].split(',');
			return self.prompt([
				InputValidate(
					'objectProperties',
					'Write the location of those modules seperated by comma',
					validate
				)
			]).then( (objectPropAnswer) => {
				entryProperties = objectPropAnswer['objectProperties'].split(',');
				for(let k = 0; k < entryIdentifiers.length; k++) {
					if(entryProperties[k].charAt(0) == '(' ||  entryProperties[k].charAt(0) == '[') {
						webpackEntryPoint[entryIdentifiers[k]] = entryProperties[k];
					} else {
						webpackEntryPoint[entryIdentifiers[k]] = `'${entryProperties[k]}.js'`;
					}
				}
				return webpackEntryPoint;
			});
		});
	}
	else {
		ret = self.prompt([
			InputValidate(
				'singularEntry',
				'Write the location of module you would like to use',
				validate
			)
		]).then( (singularAnswer) => `'${singularAnswer['singularEntry']}.js'`);
	}

	return ret;
};
```

ev1stensberg: You're assigning the promise to a value instead of directly calling it 💯  Thanks! Just tested it and it has the behaviour I wanted. When I did it without referencing an object the callbacks for multiple entries were all outputting to the terminal. Thanks for reviewing! Pushing a commit shortly. Do you mind doing  a quick overview after that?

lib/creator/yeoman/utils/entry.js
```diff
@@ -6,44 +6,32 @@ module.exports = (self, answer) => {
 
 		let entryIdentifiers;
 		let entryProperties;
-		if(answer['entryType'].indexOf('Array') >= 0) {
+		if(answer['entryType'] == 'Multiple') {
```
okonet: `==` -> `===`

lib/creator/yeoman/utils/entry.js
```diff
@@ -6,44 +6,32 @@ module.exports = (self, answer) => {
...
+						}
 					}
 					resolve(webpackEntryPoint);
 				});
 			});
 		}
 		else {
 			self.prompt([
-				Input('singularEntry', 'What\'s your entry point?')
+				Input('singularEntry', 'Write the location of module you would like to use')
```
okonet: I'm not sure `write` is a good verb here. I'd say `type`?

lib/creator/yeoman/webpack-generator.js
```diff
@@ -24,65 +22,68 @@ module.exports = class WebpackGenerator extends Generator {
 	prompting() {
 		let done = this.async();
 		let self = this;
-
+		let oneOrMoreEntries;
 		this.configuration.config.webpackOptions.module = getDefaultLoaders();
 		this.configuration.config.webpackOptions.plugins = getDefaultPlugins();
 		this.prompt([
-			RawList('entryType', 'What kind of entry point do you want?', ['Array', 'Function', 'String', 'Object'])
+			RawList('entryType', 'Would you like to bundle a single or multiple javascript modules?',
```
okonet: Please put each argument on a new line

pksjce:
Hey,
Lot's of work done on this! Great to have more eyes on this code!

@ev1stensberg - I checked out the latest and ran `webpack-cli --init` to see usage. I have a few small concerns. 
Here is the screenshot.
screen shot 2017-05-05 at 10 54 56 pm

1. Asking to add "Remember to add string quotes to your bundle entries and file extensions!" is a bit much. Since it's only an init tool, we need not support too many complex options. We can just assume that what the user has entered is a string. Don't you think so?
2. " Would you like to bundle a single or multiple javascript modules?" could be simplfied to "Will you be creating multiple bundles?" with "Yes" or "No" answer.
3.  "Type the location of module you would like to use" . This is a bit vague and we need to use simpler terminology.
4. Once all the prompts are answered, before the `npm install` runs, lets show a message saying "Created config, installing dependencies..." or something like that?

**But**, I would suggest an alternate set of questions to ask. The script would be something like
```bash
Welcome to webpack! 
We will ask you a few questions to create your configuration file.

1. Enter the name of your entry file - index.js
2. Enter the path for the output file - dist/bundle.js
3. Do you use ES6 in your project? Y/n
4. Do you use one of the below CSS solutions?
a. SASS
b. LESS
c. PostCss
5. Do you want to bundle the CSS to a separate file? Y/n

Creating "webpack.config.js"

module.exports = {
...
}

Do you want to save this file and install dependencies? Y/n
Installing dependencies....
..
..

Congratulations! Your configuration file has been created at `webpack,config.js`

```
I feel we have the ast support to convert all the above questions to a configuration. This should suffice to be the default `init` functionality.

Plugin packages could extend these questions to include framework/library specific questions like preact/react etc.

ev1stensberg:
@pksjce Hi! Great to have you reviewing! 

Here's my replies to your concerns:

1. I feel like we should warn the user, or at least mention it in the docs somewhere. I don't really want to change the entries section, unless it's **really needed**. The reason is that we are with this logic allowing: 

- CommonChunkPlugin on multiple entries by default
- Tooltip for CommonsChunk as well, to help users get a notion of perf, continued from a discussion and mention of this with @addyosmani 
- Super dynamic scaffolding for entries, people can add all sorts of types, and reminding people to add quotes is a small disadvantage. 

We can, however, try to implement the string addition and remove this burden from the users. Not guaranteeing it will work 100%, but I'll try implementing it along with file-extensions for single entries. I was thinking about consistency when adding the `file end` note.

2. I completely agree, will proceed to replace 
3. Yes, same here
4. Agree on this point too, this is consistent with `migrate` and allows people to reconsider.

For the CSS part, will refactor. Do you have a chance to review the ast part too?

A question for me is: What is missing after the fixes mentioned to merge and push `v1`?

ev1stensberg:
A small note to add is that `migrate` uses a task runner for a single `webpack.config`. I'm a bit worried about this, as we support multiple package scaffolds in `init`, which will make the terminal polluted. Perhaps we should ask if the user wants to write the file instead? Can try to integrate the task-runner, but again, no promises. Fixed what you mentioned, except the CSS part, where we need to discuss. 

Typing entry now lets you add things such as `Array`, `Function`, `ArrowFunction`, `String` and properties on strings that aren't litteral, such as `path.join` or `process.cwd()`.

addyosmani:
Fwiw, the proposal by @pksjce in "But, I would suggest an alternate set of questions to ask. The script would be something like" sounds like it would tweak the prompts to be a little more beginner friendly. Perhaps 'Do you use ES2015 in your project?' instead of ES6 :)

Silly question: why wouldn't the user want to create their Webpack config in the last step of https://github.com/webpack/webpack-cli/pull/105/commits/0ba9d7d1130e2b4fa2d1e7e4b1d87c830a83d965? It seems extraneous to me. At least with Yeoman generators, the UX we typically use is asking questions and on the final question then kicking of writing to disk.

Reading through @ev1stensberg's work in https://github.com/webpack/webpack-cli/pull/105/commits/c58f79331ab88bfddb78bb9e4a43da316c252eb4, I do like the idea of minimal boilerplate for encouraging the use of common vendor chunks and ExtractTextPlugin. 

Another silly q: To what extent if any should we be switching over to babel-minify if the user opts for ES2015 support with Babel? Are we encouraging ES2015 -> ES5 -> Uglify that code?

ev1stensberg:
@pksjce @addyosmani Corrected the generator as of the comments. The last thing left is the CSS/SASS/LESS prompts, do we want those? I'm not sure, but open for input. Removed the confirm, even though I side with @pksjce on the confirm part, as this is consistent with `migrate`.

ev1stensberg:
Also, heard some things about `babel-minify` resulting in a bigger bundle, or something in that alley. Pinging @kdzwinel for his input.

kdzwinel:
Yeah, we did some tests couple of weeks back - babili wasn't terribly bigger (~10%), but it was way slower than babel+uglify.

```
Babel + uglify:
23.87s - 1.06 MB

Babili:
49.24s - 1.18 MB
```

This may have changed in the meantime though (babili is a beta project after all).

addyosmani:
cc @hzoo as an fyi on the babili part of the discussion

👍 for the current status of this work landing @ev1stensberg 

> Also, I think we should enable extractTextPlugin by default and skip the bundle CSS in one file, this is better for perf, yes?

👍

anyulled:
migrate.md file is empty. Is it intentional?

ev1stensberg:
@anyulled Yep, left it for @okonet or @pksjce , but this could be a good first contribution!

PR #110, opened 2017-04-14
Make use of utils for the ast transformations

pksjce:
Would be great if you were able to replicate in the other transformations as well. Will try to cover a couple here.

### DIFF
--- a/lib/creator/transformations/context/context.js
+++ b/lib/creator/transformations/context/context.js
@@ -10,7 +10,7 @@ module.exports = function(j, ast, yeomanConfig) {
 
 	if(webpackProperties['context']) {
 		return ast.find(j.ObjectExpression)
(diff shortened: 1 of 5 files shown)


pksjce:
@ev1stensberg - I hope this helps!

ev1stensberg:
Would be great, but we're creating a property first on the module.exports, needs to be done before the `filter`. Which again means we're creating dupe nodes ( when I tried using `filter` and `forEach`). I used filter as a safe-mechanism for creating the value once.

ev1stensberg:
I like the approach you've made, so it's worth implementing, nonetheless.

ev1stensberg:
Run `webpack-cli --init webpack-addons-pwa` to see if your changes is viable. Needs a rebase first though

PR #115, opened 2017-04-16
Tests for entry.js

pksjce:


pksjce:
Yep, it's here.
https://github.com/webpack/webpack/blob/master/schemas/webpackOptionsSchema.json#L67

ev1stensberg:
As mentioned at twitter, I changed some of the code. We need to use identifiers for every value, as we don't know what  the user wants to do. Provides flexibility, as we can now ast functions. 

https://github.com/ev1stensberg/webpack-addons-demo#important-with-string-values-you-need-to-wrap-your-strings-once-again-this-is-because-we-can-declare-some-other-functionality-using-only---while--mystring--resolves-to-a-string

PR #121, opened 2017-05-04
Better CLI

okonet:
feature / refactoring

No, we need to introduce e2e tests for that

Yes

**Summary**

Refactors the CLI code so we use commands for `migrate` and `init`. 
Made use of yargs to simplify the source code. 
Consolidated all webpack options into one file.

Yes, the commands should be now called with `webpack-cli migrate` instead of `webpack-cli --migrate`

**Other information**


ev1stensberg:
Ready for review @okonet ?

okonet:
Almost there.

okonet:
So after spending a good amount of time on this, I reached a dead end it seems. I can't think of a good way of doing this change as a non-breaking change to the existing webpack CLI.

The problem is that current CLI syntax looks like this `webpack <entry> [<entry>] <output>`. 

Making `migrate` and `init` tasks act as commands means, the syntax to use them will be `webpack init` which might be ambiguous to the current webpack CLI.

Another issue I tried to solve is to not display all options by default when running `webpack --help`. This also seems impossible with a non-breaking change.

A breaking change means an introduction of an additional command for webpack default behavior. For example `compile` or `start`.

There is probably a solution for this that involves parsing of arguments but I'm not sure yet how reliable this can be implemented. Any feedback is appreciated!

@sokra I think you should be involved in this discussion since implications can be pretty big.

TheLarkInn:
Any good ideas @sokra

okonet:
We discussed on Slack but I'm a bit drawn by work now so I'll continue on this in a week or so.

bin/webpack.js
```diff
@@ -19,360 +19,22 @@ if (localCLI && path.relative(localCLI, __filename) !== '') {
...
-}
+yargs
+	.usage(`webpack-cli ${require('../package.json').version}
+
+Usage: $0 <cmd> [args]
+
+Usage without config file: webpack <entry> [<entry>] <output>
+Usage with config file: webpack
+`)
+	.epilogue('for more information see https://webpack.github.io/docs/cli.html')
```
pksjce: This should point to new docs?

DanielaValero:
@okonet anything here for me to help with?

okonet:
More free time would be great :( Seriously though, having tests would help dramatically before I refactor even further. Not sure how much of this is covered.

ev1stensberg:
@okonet we should get that react test suite up and running first, yea?

okonet:
What react test suit? You mean Jest snapshot tests? yes, that would be great to have. We need 100% coverage if we don't want break current implementation.

ev1stensberg:
@okonet You mentioned on Slack that instead of Soren, you could use this https://github.com/vadimdemedes/ink for testing

okonet:
@ev1stensberg it's not for testing but for the CLI itself and I'd like to look into it.

ev1stensberg:
@okonet gotcha! We would need to come up with a way to test E2E and produce results, both for the new features but also the "old" CLI commands that webpack have.

ev1stensberg:
May be better to redo this in another fresh PR after its been merged with webpack, closing @okonet , reopen if you want to continue to work on this branch!

Issue #135, opened 2017-05-12
webpack-cli init doesn't install webpack, uglifyjs-webpack-plugin, babel-preset-es2015


nilliams:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
```
$ webpack-cli init

Insecure about some of the questions?

https://github.com/webpack/webpack-cli/blob/master/INIT.md

? Will you be creating multiple bundles? No
? Which module will be the first to enter the application? index.js
? Which folder will your generated bundles be in? [default: dist]:
? Are you going to use this in production? No
? Will you be using ES2015? Yes
? Will you use one of the below CSS solutions? No
? If you want to bundle your CSS files, what will you name the bundle? (press enter to skip)
? Name your 'webpack.[name].js?' [default: 'config']:
...
├─┬ babel-core@6.24.1
│ ├─┬ babel-code-frame@6.22.0
│ │ ├─┬ chalk@1.1.3
...
└── UNMET PEER DEPENDENCY webpack@2
```

Then:
```
$ webpack-cli
module.js:471
    throw err;
    ^

Error: Cannot find module 'webpack'
```

If install webpack (`npm i -S webpack`), next error is:
```
$ webpack-cli
module.js:471
    throw err;
    ^

Error: Cannot find module 'uglifyjs-webpack-plugin'
```

If install that, we get (after fixing `"./dist" is not an absolute path` which is I see is fixed now):

```
$ webpack-cli
Hash: 5492c18c1852ed5bd45b
Version: webpack 2.5.1
Time: 3144ms
         Asset     Size  Chunks             Chunk Names
main.bundle.js  1.99 kB       0  [emitted]  main
   [0] ./index.js 1.49 kB {0} [built] [failed] [1 error]

ERROR in ./index.js
Module build failed: Error: Couldn't find preset "es2015" relative to directory
```

**What is the expected behavior?**

`webpack-cli init` to install all needed deps

**Version**
```
$ webpack-cli --version
1.3.1
```

ev1stensberg:
Hi! The `dist` output is fixed in #125 . It should install your dependencies, what's your repository structure, and where do you run the command from?

nilliams:
Hmm, it does seem odd no-one else has hit this!

Thought it might be an env thing, so now tried in 2 envs, (empty dir each time):

- 2 x in MacOS
- 2 x in Ubuntu (via Vagrant)

... and get the same.

[...]

PR #138, opened 2017-05-13
fix: Install correct (`es2015`) babel preset to match generated config

nilliams:
Bugfix

No (none in repo)

N/A

**Summary**
Raised in #135, but I didn't address this outstanding problem in the previous PR.

The config generated uses `presets: ['es2015']`, see: https://github.com/webpack/webpack-cli/blob/master/lib/creator/yeoman/utils/module.js#L8

... but are previously installing `babel-preset-env` instead, which results in: `Error: Couldn't find preset "es2015"` when user answers `'y'` to 'Will you be using ES2015?', then runs `webpack`.

No


nilliams:
Thanks! Sorry I did mention that this was outstanding in the last comment in that issue. I wasn't sure what the intended babel preset was at the time, but it became clear this is the only fix that made sense after a bit of reading into babel-preset-env, and sleep 😀

PR #160, opened 2017-05-25
feat: replace yarn backup w/ normal yarn

ev1stensberg:
CC-ing @pksjce


ev1stensberg:
Why did we disable this in the first place? I know there was a yarn issue with recast but has that been resolved?

DanielaValero:
Actually this is a question that I wanted to ask you, are we using npm or yarn? After installing again the project, I got the next unstaged files, which confuses me a bit:

```
	package-lock.json
	yarn.lock
```

Maybe we should consider using only one?

ev1stensberg:
Why is that better over both? Users use both

DanielaValero:
Hi @ev1stensberg 

Sorry for my late reply. 

As for this project, one could argue, to have both installed, and maintain both. Which would be ok, not a biggie. However, as maintainers we need to put some (small but still present) extra effort on keeping them both up to date.

I've been there at the beginning of the project, and was probably me the one who added yarn to the project at the first place. But, after having got a long break on the cli, and having come back, the first thing I did was to install all the dependencies. And to be honest, I really did not get what should I use to do so, yarn or npm?. I read the contributing.md (that I wrote) and still was not clear.

After having got a new fresh look into this, I'd recommend that we just make this easier for us, and keep only one. Yes, it is true that users use yarn, or npm, but that is not really a strong argument to keep both in the project. If both do the same, what is the benefit for us of having two libraries doing the same thing, which in this case, increase maintenance effort and/or confusion at the entry of new devs/coming back devs?

ematipico:
When I started my devs on the cli I was confused as well and when I saw the .yarn file (even though it was a backup file) I guessed I had to use yarn at first place.

Although I was confused. :smile: 
I guess it would be better to use just one and keep it up to date

okonet:
Yeah, it's quite confusing. Since webpack main repo doesn't use yarn I'd say let's stick with npm for the consistency sake.

This is what I would do:

1. Increase `node` and `npm` version to `>= 5.x` in `package.json`
1. Add `package-lock.json`
1. Remove `yarn.lock`

Thoughts?

ev1stensberg:
Users are complaining that we aren't supporting yarn though?

DanielaValero:
@okonet I added already .2 of your list in #173, I could add in there also the .1 and .3, and update the contributing.d

@ev1stensberg do you know the reasons why the users complain about it? and from which medium are they complaining? with more info about it we could make a more informed decision.

okonet:
@ev1stensberg what users? This is only relevant for contributors.

ematipico:
Actually, webpack uses yarn instead of npm

PR #164, opened 2017-06-12
chore(tests): Use node env for Jest to avoid JSDOM overhead

okonet:
build related change

Not relevant

Not relevant

**Summary**

Jest can be configured to run in Node env only without emulating browser with JSDOM: https://facebook.github.io/jest/docs/en/configuration.html#testenvironment-string

This improves tests running speed significantly:

### Before

```
Test Suites: 31 passed, 31 total
Tests:       1 skipped, 126 passed, 127 total
Snapshots:   110 passed, 110 total
Time:        24.577s
```

### After

```
Test Suites: 31 passed, 31 total
Tests:       1 skipped, 126 passed, 127 total
Snapshots:   110 passed, 110 total
Time:        15.168s, estimated 19s
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,5 +73,8 @@
       "git add"
     ]
   },
-  "pre-commit": "lint:staged"
+  "pre-commit": "lint:staged",
+  "jest": {
+    "testEnvironment": "node"
+  }
 }


cpojer:
I don't think this is correct. switching from jsdom to node will improve Jest's startup time by about 400ms but it should be a one time cost. Based on your output, it seems like the second run was using a cache but the first one wasn't (as indicated by "estimated xx seconds").

Can you try:
* Run Jest once with either environment.
* Run Jest again, this time it will use the cache.
* Switch the environment, it will keep using the cache.

And then compare timings for #2 and #3?

okonet:
@cpojer thanks for looking at it. Here are my timings:


Second run with `node` env:
`Time:        17.528s`

Switching back to defaults:

```
Time:        22.979s
```

Second run defaults:

```
Time:        17.542s
```

So it seems you're right about the caching.

cpojer:
Still worth doing, as you are running this code in a node environment.

PR #170, opened 2017-07-04
fix: Give examples that are valid (and clearer).

nilliams:
Bugfix

N/A

N/A

**Summary**
When using `webpack-cli init` and answering the questions for 'multiple' entry files, the examples given are misleading.
 
The main issue is that following the format of one example will lead to a `.js.js` extension, so the resulting config will not be valid.

See [comments here](https://github.com/webpack/webpack-cli/commit/37c40f046bec586819263361e56fc90dc3309e08) for more details.

No

### DIFF
--- a/lib/creator/yeoman/utils/entry.js
+++ b/lib/creator/yeoman/utils/entry.js
@@ -8,7 +8,7 @@ module.exports = (self, answer) => {
 		result = self.prompt([
 			InputValidate(
 				'multipleEntries',
-				'Type the name you want for your modules (entry files), separated by comma [example: \'app\']',
+				'Type the names you want for your modules (entry files), separated by comma [example: \'app,vendor\']',
 				validate
 			)
 		]).then( (multipleEntriesAnswer) => {
@@ -41,7 +41,7 @@ module.exports = (self, answer) => {
 			return forEachPromise(entryIdentifiers, (entryProp) => self.prompt([
 				InputValidate(
 					`${entryProp}`,
-					`What is the location of '${entryProp}'? [example: '../dist/app.js']`,
+					`What is the location of '${entryProp}'? [example: './${entryProp}']`,
 					validate
 				)
 			])).then(propAns => {
--- a/lib/creator/yeoman/utils/entry.js
+++ b/lib/creator/yeoman/utils/entry.js
@@ -41,7 +41,7 @@ module.exports = (self, answer) => {
 			return forEachPromise(entryIdentifiers, (entryProp) => self.prompt([
 				InputValidate(
 					`${entryProp}`,
-					`What is the location of '${entryProp}'? [example: './app']`,
+					`What is the location of '${entryProp}'? [example: './src/${entryProp}']`,
 					validate
 				)
 			])).then(propAns => {
@@ -65,7 +65,7 @@ module.exports = (self, answer) => {
 		result = self.prompt([
 			InputValidate(
 				'singularEntry',
-				'Which module will be the first to enter the application? [example: \'./index\']',
+				'Which module will be the first to enter the application? [example: \'./src/index\']',
 				validate
 			)
 		]).then( (singularAnswer) => `'${singularAnswer['singularEntry']}'`);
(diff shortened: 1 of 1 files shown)


lib/creator/yeoman/utils/entry.js
```diff
@@ -8,7 +8,7 @@ module.exports = (self, answer) => {
 		result = self.prompt([
 			InputValidate(
 				'multipleEntries',
-				'Type the name you want for your modules (entry files), separated by comma [example: \'app\']',
+				'Type the names you want for your modules (entry files), separated by comma [example: \'app,vendor\']',
```
ev1stensberg: singular? @TheLarkInn

lib/creator/yeoman/utils/entry.js
```diff
@@ -41,7 +41,7 @@ module.exports = (self, answer) => {
 			return forEachPromise(entryIdentifiers, (entryProp) => self.prompt([
 				InputValidate(
 					`${entryProp}`,
-					`What is the location of '${entryProp}'? [example: '../dist/app.js']`,
+					`What is the location of '${entryProp}'? [example: './${entryProp}']`,
```
ev1stensberg: Don't use entryprop. Change app directory instead.

ev1stensberg:
Awesome! Left some comments!

nilliams:
Sure, done!

lib/creator/yeoman/utils/entry.js
ev1stensberg: This wasn't really needed

lib/creator/yeoman/utils/entry.js
```diff
@@ -41,7 +41,7 @@ module.exports = (self, answer) => {
 			return forEachPromise(entryIdentifiers, (entryProp) => self.prompt([
 				InputValidate(
 					`${entryProp}`,
-					`What is the location of '${entryProp}'? [example: '../dist/app.js']`,
+					`What is the location of '${entryProp}'? [example: './app']`,
```
ev1stensberg: This will produce a broken build `./app` isn't a valid FP

lib/creator/yeoman/utils/entry.js
nilliams: Surely the example should match the question. The question clearly wants **multiple** entry files, **comma-separated**, so the example should match.

If I have misinterpreted this I suggest the question needs re-wording.

Not sure if we're having a language issue here, perhaps input from @TheLarkInn would help.

lib/creator/yeoman/utils/entry.js
nilliams: FP = file path?

`./app` gets converted to './app.js'. Please run the generator yourself to see the output.

ev1stensberg: Wasn't the entire point of this PR to change the example from `./dist` to something else? If so, I think it would be better with `./build/app`!

nilliams: It needs an 'entry file' location, entry files are **input** files, not **output**, so `build` makes no more sense than `dist`.

See the `app` and `vendors` examples here: https://webpack.js.org/concepts/entry-points/#separate-app-and-vendor-entries

lib/creator/yeoman/utils/entry.js
ev1stensberg: I guess we can have it like that then.. :)

lib/creator/yeoman/utils/entry.js
ev1stensberg: Change this to `[`example: ../src(or build)/{entryProp}.js`]`

nilliams: Done!

nilliams: webpack-cli-examples

ev1stensberg: Sweet!

DanielaValero:
lgtm. Will merge. @nilliams thanks for the PR!

Issue #179, opened 2017-08-27
Feature: Generate Custom Loaders and Plugins(?)


TheLarkInn:
**Do you want to request a *feature* or report a *bug*?**
Feature!!! <3

**What is the current behavior?**
Currently it takes a bit of setup to have a "webpack-defaults"/"webpack-contrib" style loader or plugin with some bare tests. 

**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**
I think it would be perfect for users to have an approachable way to create custom plugins and loaders via the cli. 

webpack-cli --generate plugin `my-webpack-plugin` (following naming conventions OOB)

webpack-cli- --generate loader `my-loader` (following naming conventions OOB)

This could generate a 'bootstrapped' repo that contains example functional and unit tests for plugins and loaders. 

**If this is a feature request, what is motivation or use case for changing the behavior?**
Motivation for this feature comes from me creating a "best practice" [loader-boilerplate repo](https://github.com/thelarkinn/example-webpack-loader) for one of my https://webpack.academy courses. And I thought to myself: 

> Holy crap it would be nice if I just had to gen this from a command and have some EZPZ scaffold.

So yeah! Feel free and leverage my repo as an example for this.

PR #183, opened 2017-09-05
Feature/generate loader

ianjsikes:

Feature

A few

**Summary**

See #179 for feature request.
Adds two new commands, `webpack-cli generate-loader` and `webpack-cli generate-plugin`. These commands generate `webpack-defaults`-compliant starter projects for Webpack loaders and plugins, including basic tests.

No

**Other information**
The starter projects are based off of @TheLarkInn's [example loader][ex] repo.

[ex]: https://github.com/TheLarkInn/example-webpack-loader


ev1stensberg:
Heyo! Thanks for submitting a PR, will review later this week! Awesome work! CC-ing @okonet

TheLarkInn:
@ianjsikes this looks incredible. Thank you so much for this work. This will be incredibly impactful on the webpack community. 🙇🙇🙇

okonet:
That's great! @ianjsikes do you think it would make sense to add tests for the generator?

ianjsikes:
@okonet definitely. I'll get started on that.

ianjsikes:
So in attempting to add tests for the generator, I discovered that anything extending `yeoman-generator` will not work on node v4 or v5 (including the existing `webpack-cli init` generator) since `yeoman-generator` is written and distributed in ES6.

I can't figure out a way to write generator tests that will pass on v4 or v5 since `yeoman-generator` doesn't seem to work on those versions.

okonet:
Does this mean that the `webpack init` will also not work on Node v4? This is a blocker then and we need to transpile the code in that case.

bin/webpack.js
```diff
@@ -167,6 +167,10 @@ if(argv._.includes('init')) {
 	const inputConfigPath = path.resolve(process.cwd(), filePaths[0]);
 
 	return require('../lib/migrate.js')(inputConfigPath, inputConfigPath);
+} else if(argv._.includes('generate-loader')) {
```
ev1stensberg: If we were to support v4 this would be a thing that needs change. My self included, have used `.includes`, but I think that's v4<

lib/generate-loader/loader-generator.js
```diff
@@ -0,0 +1,134 @@
...
+				this.props.name +
+				'\nI will create this folder for you.'
+			);
+			mkdirp(this.props.name);
+			this.destinationRoot(this.destinationPath(this.props.name));
+		}
+	}
+
+	writing() {
+		/**
```
ev1stensberg: Single line comment

lib/generate-loader/loader-generator.js
```diff
@@ -0,0 +1,134 @@
...
+			this.destinationPath('src/index.js'),
+			{ name: this.props.name }
+		);
+
+		this.fs.copy(
+			path.join(__dirname, 'templates', 'src', 'cjs.js.tpl'),
+			this.destinationPath('src/cjs.js')
+		);
+
+		/**
```
ev1stensberg: Single line comment

lib/generate-loader/loader-generator.js
```diff
@@ -0,0 +1,134 @@
...
+			this.destinationPath('test/functional.test.js')
+		);
+
+		this.fs.copy(
+			path.join(__dirname, 'templates', 'test', 'fixtures', 'simple-file.js.tpl'),
+			this.destinationPath('test/fixtures/simple-file.js')
+		);
+
+		/**
+		 * Examples
```
ev1stensberg: Single line comment

lib/generate-loader/templates/test/unit.test.js.tpl
```diff
@@ -0,0 +1,32 @@
+import fs from 'fs';
+import { runLoaders } from 'loader-runner';
+import Promise from 'bluebird';
```
ev1stensberg: This should be at 2nd line

ianjsikes:
I made the changes @ev1stensberg requested and opened #184 for fixing the `yeoman-generator` issue.

lib/generate-loader/loader-generator.js
```diff
@@ -0,0 +1,103 @@
...
+
+	default() {
+		if (path.basename(this.destinationPath()) !== this.props.name) {
+			this.log(
+				'Your loader must be inside a folder named ' +
+				this.props.name +
+				'\nI will create this folder for you.'
+			);
+			mkdirp(this.props.name);
+			this.destinationRoot(this.destinationPath(this.props.name));
```
ev1stensberg: Store `destinationPath(this.props.name)` in a variable for new contributors to easier know what it is

lib/generate-loader/loader-generator.js
```diff
@@ -0,0 +1,103 @@
...
+		}
+	}
+
+	writing() {
+		const copy = (filePath) => {
```
ev1stensberg: Leave a comment/ description of what this is doing

lib/generate-plugin/plugin-generator.js
```diff
@@ -0,0 +1,89 @@
...
+			},
+		];
+
+		return this.prompt(prompts).then(props => {
+			this.props = props;
+		});
+	}
+
+	default() {
+		if (path.basename(this.destinationPath()) !== this.props.name) {
```
ev1stensberg: Use local var instead

lib/generate-plugin/plugin-generator.js
```diff
@@ -0,0 +1,89 @@
...
+				this.props.name +
+				'\nI will create this folder for you.'
+			);
+			mkdirp(this.props.name);
+			this.destinationRoot(this.destinationPath(this.props.name));
+		}
+	}
+
+	writing() {
+		const copy = (filePath) => {
```
ev1stensberg: Maybe abstract this into a util file so you'd avoid re-declaring the same function twice, if it contains the same logic

lib/generate-loader/index.js
```diff
@@ -0,0 +1,16 @@
+const yeoman = require('yeoman-environment');
+const LoaderGenerator = require('./loader-generator').LoaderGenerator;
+
+/**
```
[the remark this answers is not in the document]
ianjsikes: Without the double asterisks (`/**`), it isn't recognized as a [JSDoc comment](http://usejsdoc.org/about-getting-started.html#adding-documentation-comments-to-your-code). I could still remove it but you won't get any code hints from your editor.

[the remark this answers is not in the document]
okonet: No, don't remove it. We should follow JSDoc everywhere. ESLint should be setup to enforce that.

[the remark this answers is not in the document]
ev1stensberg: Weird, we've used it like the one I've proposed, may need to change everything then?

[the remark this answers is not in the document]
okonet: Yeah, we should change it everywhere then.

ianjsikes:
@okonet @ev1stensberg is there anything else that needs to be done on this PR?

ev1stensberg:
I'll have a final look and merge if it looks fine @ianjsikes !

ev1stensberg:
Remove the `//eslint-disable`'s and you're set. I don't believe we'd need to worry about that when initializing a project, because we do not know their eslint config settings

ev1stensberg:
LGTM, let's wait for tests and then merge if everything checks out

ev1stensberg:
Thanks @ianjsikes ❤️

PR #186, opened 2017-09-21
Add unit tests for CLI and add flow compilation

ev1stensberg:
This PR enables us to test the webpack CLI as it stands in core, which in turns allows us to merge into the core repository


ev1stensberg:
Once this gets a ok to merge, we'd need to remove the local import, but it's to emphasize that the CLI tests runs well locally if you clone and run the branch.

ev1stensberg:
Could anyone review this?

package.json
```diff
@@ -73,9 +77,13 @@
     "eslint-plugin-flowtype": "^2.35.1",
     "eslint-plugin-node": "^5.1.0",
     "flow-bin": "^0.49.1",
+    "flow-remove-types": "^1.2.1",
     "husky": "^0.14.3",
     "jest": "^20.0.4",
     "jest-cli": "^20.0.4",
-    "lint-staged": "^4.1.3"
+    "lint-staged": "^4.1.3",
+    "mocha": "^3.5.3",
```
okonet: We're using Jest for tests. Why is this needed?

package.json
```diff
@@ -73,9 +77,13 @@
...
+    "prettier-eslint-cli": "^4.3.2",
+    "should": "^13.1.0"
```
okonet: Same, jest already has everything we need.

ev1stensberg:
Yep, sure. Mocha was in old webpack, was easier to port those

ev1stensberg:
@okonet done!

bin/convert-argv.js
```diff
@@ -233,21 +242,22 @@ module.exports = function(yargs, argv, convertOptions) {
 		}
 
 		function mapArgToBoolean(name, optionName) {
+			if (options[optionName || name]) {
+				options[name] = true;
+			}
+			//eslint-disable-next-line
+			if (name && options[name] == true) {
```
ematipico: Shouldn't be `===`?

ev1stensberg: Implemented by webpack/core. You should just look at the test for now

ev1stensberg: Also, let's keep it non-strict equal, cause it could be stringified

test/BinTestCases.test.js
```diff
@@ -0,0 +1,145 @@
+/* globals describe it before */
+"use strict";
+
+const path = require("path");
+const fs = require("fs");
+const spawn = require("child_process").spawn;
+
+function loadOptsFile(optsPath) {
+	// Options file parser from Mocha
```
ematipico: Comment can be removed as we decided to stick with Jest

ev1stensberg: Nice catch!

ev1stensberg: Actually, this should be kept, if you see next line

bin/webpack.js
```diff
@@ -1,36 +1,45 @@
...
 */
-"use strict";
-
 var path = require("path");
 var resolveCwd = require("resolve-cwd");
-
+// Local version replace global one
 var localCLI = resolveCwd.silent("webpack-cli/bin/webpack");
 
+if (process.argv.slice(2).includes("init")) {
```
ematipico: `.includes` is not supported by Node 5 and 4. We need to change it. Also below
http://node.green/#ES2015-built-ins-well-known-symbols-Symbol-match--String-prototype-includes

ematipico:
Some tests are failing on Node 5 and 4

ev1stensberg:
Yep, cause we use `include`, changed in the `feature` PR

ematipico:
Great, so it's under control :)

ev1stensberg:
Any comments?

ematipico:
It looks fine to me, there are just some conflicts but overall looks good

ev1stensberg:
V4 failing cause there's too many async operations going on, LGTM for now I guess. @ematipico your opinion?

ematipico:
Fine by me. We also added the removing of the types before publishing, which is awesome!

Issue #22, opened 2016-12-24
Feature: Support --quiet on the command-line


bebraw:
Currently, `webpack --quiet` ignores the files passed, while still outputting to the console:

```
Hash: 6dbc5ab44333d067d673
Version: webpack 1.4.0-beta6
Time: 11ms
```

It would be nice to have the `--quiet` option work as expected, i.e. silence all non-error messages.

**Moved from https://github.com/webpack/webpack/issues/490**

DanielaValero:
There was a PR some time ago in webpack, see webpack/webpack#517

georgiosd:
The `stats` property doesn't seem to get picked up when returning multiple configuration objects.

DanielaValero:
@ev1stensberg, @pksjce what you guys think, should this be part of our roadmap to v1 ?

pksjce:
Yes, the `--quiet` option can be implemented. In the compiler callback, we just need to suppress all ouput.
But fixing `stats` bugs might be harder. Might require changes in webpack core.

DanielaValero:
is somehow the `stats` bug related to the `--quiet` option? The relationship I see is that the `--quiet` silence the `stats`. If this is the only relationship, then I'd say that the `stats` bug should be raised as a separate github issue, so we don't mix two different things

bebraw:
Btw, if you are open to changing naming, consider `--silent`/`-s` to follow npm convention.

ev1stensberg:
We've got `-s --save`, I think. That would need to be changed, dunno if people would like that. I'm open to it though

bebraw:
Ah, yeah. Namespace collisions are tough.

ev1stensberg:
@georgiosd Could you come up with a test repro on the stats and comment it at #84 ?

ematipico:
Feature implemented: `--silent`

PR #200, opened 2017-12-12
Improvement: add an option to watch messaging. Add .idea to .gitignore

EugeneHlushko:

Adds `infoVerbosity` option which is set to `info` by default. In order for user to be able to control the amount of extra messaging output. 

Provides clear indication of when compilation had started and when it had finished if such verbosity is requested

Fixes webpack/webpack#6051 and webpack/webpack#5895

Yes
Also added PR for json schema validation of the option: webpack/webpack#6332
UPD: we wont add these to webpkac/webpack's schema, instead we can create our own in webpack-cli

*Not yet had been approved*

**Summary**

See webpack/webpack#6051 and webpack/webpack#5895

no

**Other information**

### DIFF
--- a/bin/webpack.js
+++ b/bin/webpack.js
@@ -188,10 +188,11 @@
 			group: DISPLAY_GROUP,
 			describe: "Show more details"
 		},
-		"display-watch-messages": {
-			type: "boolean",
+		"info-verbosity": {
+			type: "string",
+			default: "info",
 			group: DISPLAY_GROUP,
-			describe: "Display watch lifecycle messages e.g. Started watching files..."
+			describe: "Controls the output of lifecycle messaging e.g. Started watching files... (verbose, info, none)"
 		}
 	});
 
@@ -404,8 +405,8 @@
 				}
 			});
 
-			ifArg("display-watch-messages", function(bool) {
-				if (bool) outputOptions.displayWatchMessages = true;
+			ifArg("info-verbosity", function(value) {
+				outputOptions.infoVerbosity = value;
 			});
 
 			var webpack = require("webpack/lib/webpack.js");
@@ -474,7 +475,7 @@
 					process.stdin.resume();
 				}
 				compiler.watch(watchOptions, compilerCallback);
-				if (outputOptions.displayWatchMessages)
+				if (outputOptions.infoVerbosity !== "none")
 					console.log("\nWebpack is watching the files…\n");
 			} else compiler.run(compilerCallback);
 		}
--- a/bin/webpack.js
+++ b/bin/webpack.js
@@ -438,6 +438,12 @@
 				);
 			}
 
+			if (outputOptions.infoVerbosity === "verbose") {
+				compiler.hooks.beforeCompile.tap("WebpackInfo", (compilation) => {
+					console.log("\nCompilation starting…\n");
+				});
+			}
+
 			function compilerCallback(err, stats) {
 				if (!options.watch || err) {
 					// Do not keep cache anymore
@@ -477,6 +483,12 @@
 				compiler.watch(watchOptions, compilerCallback);
 				if (outputOptions.infoVerbosity !== "none")
 					console.log("\nWebpack is watching the files…\n");
+
+				if (outputOptions.infoVerbosity === "verbose") {
+					compiler.hooks.afterCompile.tap("WebpackInfo", (compilation) => {
+						console.log("\nCompilation finished\n");
+					});
+				}
 			} else compiler.run(compilerCallback);
 		}
 
--- a/bin/webpack.js
+++ b/bin/webpack.js
@@ -406,7 +406,7 @@
 			});
 
 			ifArg("info-verbosity", function(value) {
-				if (typeof value !== "string" || !["none", "info", "verbose"].includes(value))
+				if (!["none", "info", "verbose"].includes(value))
 					throw new Error("Invalid configuration object. \n configuration['info-verbosity'] should be one of these:\n \"none\" | \"info\" | \"verbose\"");
 				outputOptions.infoVerbosity = value;
 			});
--- a/bin/webpack.js
+++ b/bin/webpack.js
@@ -406,6 +406,8 @@
 			});
 
 			ifArg("info-verbosity", function(value) {
+				if (typeof value !== "string" || !["none", "info", "verbose"].includes(value))
+					throw new Error("Invalid configuration object. \n configuration['info-verbosity'] should be one of these:\n \"none\" | \"info\" | \"verbose\"");
 				outputOptions.infoVerbosity = value;
 			});
 
--- a/bin/webpack.js
+++ b/bin/webpack.js
@@ -444,6 +444,9 @@
 				compiler.hooks.beforeCompile.tap("WebpackInfo", (compilation) => {
 					console.log("\nCompilation starting…\n");
 				});
+				compiler.hooks.afterCompile.tap("WebpackInfo", (compilation) => {
+					console.log("\nCompilation finished\n");
+				});
 			}
 
 			function compilerCallback(err, stats) {
@@ -485,12 +488,6 @@
 				compiler.watch(watchOptions, compilerCallback);
 				if (outputOptions.infoVerbosity !== "none")
 					console.log("\nWebpack is watching the files…\n");
-
-				if (outputOptions.infoVerbosity === "verbose") {
-					compiler.hooks.afterCompile.tap("WebpackInfo", (compilation) => {
-						console.log("\nCompilation finished\n");
-					});
-				}
 			} else compiler.run(compilerCallback);
 		}
 
--- a/bin/webpack.js
+++ b/bin/webpack.js
@@ -187,6 +187,11 @@
 			type: "boolean",
 			group: DISPLAY_GROUP,
 			describe: "Show more details"
+		},
+		"display-watch-messages": {
+			type: "boolean",
+			group: DISPLAY_GROUP,
+			describe: "Display watch lifecycle messages e.g. Started watching files..."
 		}
 	});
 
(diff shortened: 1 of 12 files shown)


EugeneHlushko:
Hi @sokra, here's is migrated pr to webpack-cli. Couple of tests are now failing due to message being off by default. Should i turn it on by default or should i change the tests to not expect the messages when the param is not there?

ev1stensberg:
@EugeneHlushko update the tests, would be convenient if you add a way to test this locally, not just via tests too / screenshot 👍

EugeneHlushko:
Hi @ev1stensberg 

i've updated the tests with config to expect the string and default ones to not expect it. Do you think that is enough or should i just add this option to all old tests and create a separate test case to see that string?

EugeneHlushko:
Any update on this one @ev1stensberg ?

ev1stensberg:
Is there an issue on this? If not , you'd have to open one and @sokra must approve the feature

sokra:
See also https://github.com/webpack/webpack/issues/766

I guess it should be on by default.

See also https://github.com/webpack/webpack/issues/5895

I guess it would a a good idea to combine these features into something more general like `info-verbosity`...

EugeneHlushko:
Got your point @sokra 
`info-verbosity` sounds more like a level setting rather than a bool. In that case i would make it: `verbose` and `silent` in this PR, and in future we could always add third option and so on.

And by default it would be set to `verbose` per @ev1stensberg 's request.

If you are ok with that the plan for me would be adding that option within the scope of this PR, adding additional test case for `silent` mode. And the next PR would then cover more messages like `Build started...` and `Build finished` in the watch mode.

Please comment or approve the course.

P.S. Updating the description with permalinks to issues

ev1stensberg:
@EugeneHlushko @sokra  I actually think that we should try to align this with #22 , as this is the same issue?

EugeneHlushko:
Not exactly the same but similar. If you read the issues, its more about being able to not see extra text except for the compilation output, and another one wants to see clear messages on `start` and `end` of the compilation process. But we indeed can also respect this flag for `hash` and `version` etc. Can include it into the scope of current PR, what you think?

ev1stensberg:
I think you should consult with @sokra, it would be nice if we'd combine these features into one alias

EugeneHlushko:
Ok lets see what @sokra can say on this thread after what i've wrote yesterday: https://github.com/webpack/webpack-cli/pull/200#issuecomment-351736138

EugeneHlushko:
Hi @sokra , @ev1stensberg 
As this isnt getting additional comments i've implemented the last suggested solution and added a new test case. Please review

EugeneHlushko:
Any updates on this PR? the two issues opened are not getting closed if this is freezed. @sokra @ev1stensberg

sokra:
> info-verbosity sounds more like a level setting rather than a bool.

yep, in my opinion it should be a level settings.

I would default it to "info".

In "verbose" you could add the messages for start, end invalid...

EugeneHlushko:
Hi @ev1stensberg and @sokra 
I did update the PR accordingly

EugeneHlushko:
Should i rebase or is this still under discussion @sokra ?

bin/webpack.js
```diff
@@ -378,6 +384,10 @@ yargs.parse(process.argv.slice(2), (err, argv, output) => {
 			}
 		});
 
+		ifArg("info-verbosity", function(value) {
+			outputOptions.infoVerbosity = value;
```
sokra: You don't have to put this into `outputOptions`. A local variable would be fine.

ev1stensberg:
Could you rebase?  @EugeneHlushko

EugeneHlushko:
Yup, on it. I will leave it in outputOptions as i thought to use it in the plugins too, but if you insist i can change to local var

EugeneHlushko:
done @ev1stensberg

ev1stensberg:
Hi @EugeneHlushko . Thanks! Will review tomorrow!

EugeneHlushko:
@ev1stensberg that one stretched out, any updates?

ev1stensberg:
Hey, sorry, forgot, hectic these days. I'll review tomorrow @EugeneHlushko

bin/webpack.js
```diff
@@ -183,6 +183,12 @@
 			type: "boolean",
 			group: DISPLAY_GROUP,
 			describe: "Show more details"
+		},
+		"info-verbosity": {
+			type: "string",
+			default: "info",
+			group: DISPLAY_GROUP,
+			describe: "Controls the output of lifecycle messaging e.g. Started watching files... (verbose, info, none)"
```
ev1stensberg: What's the difference between `info` and `verbose` ? Seems to do the same thing.

bin/webpack.js
```diff
@@ -382,6 +388,10 @@
 				}
 			});
 
+			ifArg("info-verbosity", function(value) {
+				outputOptions.infoVerbosity = value;
```
ev1stensberg: If `info-verbosity` is not a string, i.e boolean, it will still output. Would be nice with an error that will match against `webpack-schema`. https://github.com/webpack/webpack/blob/master/schemas/webpackOptionsSchema.json

You should send a PR to add `info-verbosity` and link it to this.

bin/webpack.js
```diff
@@ -448,7 +458,8 @@
 					process.stdin.resume();
 				}
 				compiler.watch(watchOptions, compilerCallback);
-				console.log("\nWebpack is watching the files…\n");
+				if (outputOptions.infoVerbosity !== "none")
```
ev1stensberg: See above

bin/webpack.js
EugeneHlushko: I will modify within this scope (the schema), compeltely forgot about it

bin/webpack.js
EugeneHlushko: Well within this PR yes, but the next one should add two more messages they will be verbose only (see linked issues)

EugeneHlushko:
Thank you @ev1stensberg 👍 
I have answered the level purpose question in the code comment and schema will be updated later on today, i will ping you when its done.

bin/webpack.js
ev1stensberg: Could you make it in this PR as well? Ask any of the maintainers if you got any questions!

EugeneHlushko: Probably it will stretch out for another month in that case, any good reason for that?

ev1stensberg: Leaving code unfinished is a bad habit 💃

EugeneHlushko: But if you promise to respond quickly im on it! :)

ev1stensberg: Pinky Promise 😆

bin/webpack.js
EugeneHlushko: schema added and pr's linked in description section

ev1stensberg: Cool beans 🐻

EugeneHlushko:
@ev1stensberg feature for webpack/webpack#5895 and test coverage added

EugeneHlushko:
I've also added temporary validation, we can deprecate it as soon as jsonschema is implemented.

EugeneHlushko:
But i did add that, or do you mean anything else?
Here:
https://github.com/webpack/webpack-cli/pull/200/files#diff-877a99a662829398e8136ee938cf1c93R443
and here:
https://github.com/webpack/webpack-cli/pull/200/files#diff-877a99a662829398e8136ee938cf1c93R489

EugeneHlushko:
Hi @ev1stensberg 
Forgot to mention you in the last comment.
Verbose is already added in previous commits:
https://github.com/webpack/webpack-cli/pull/200/files#diff-877a99a662829398e8136ee938cf1c93R443

ev1stensberg:
Let me take another look later tonight

EugeneHlushko:
Sure thing mate, looking forward to feedback if any @ev1stensberg

bin/webpack.js
```diff
@@ -399,6 +405,12 @@
 				}
 			});
 
+			ifArg("info-verbosity", function(value) {
+				if (typeof value !== "string" || !["none", "info", "verbose"].includes(value))
+					throw new Error("Invalid configuration object. \n configuration['info-verbosity'] should be one of these:\n \"none\" | \"info\" | \"verbose\"");
```
ev1stensberg: I think we've got a error helper for this

EugeneHlushko: I maybe misunderstood, but couldnt find the usage in codebase of error helper, could you please point me into the right direction here? All i've found is `bin/errorHelpers.js` which basically works with existing error but doesnt throw one for me

ev1stensberg: Hm, okay, so if you look in the function inside the bin folder that I believe is called convert-argv, there's a processing of the options, and if it's not matching the schema, it throws. At least that is what I think it's doing under the hood. I'm not on a computer right now, so let me get back to you in a bit

EugeneHlushko: yeah but seems it doesnt support enum which could substitue `![].includes` call as jsonschema does?

ev1stensberg: It should only be a string

bin/webpack.js
```diff
@@ -399,6 +405,12 @@
 				}
 			});
 
+			ifArg("info-verbosity", function(value) {
+				if (!["none", "info", "verbose"].includes(value))
```
EugeneHlushko: @ev1stensberg can we keep this check until schema lands so there is no misuse?

ev1stensberg: Could you create an issue for it? Mergeable after

ev1stensberg:
Do you mind adding `webpack-dev-server` as the same dependency as `webpack`? Seems to get linting errors

bin/webpack.js
EugeneHlushko: done #249 @ev1stensberg

EugeneHlushko:
@ev1stensberg added `webpack-dev-server` to `allowModules` section

PR #217, opened 2017-12-16
feature(defaults): remove entry and output.path validation

TheLarkInn:

Feature
N/A(?)
Does it need to be documented? (We can discuss).

**Summary**
This PR is a dependent PR for webpack/webpack#6140. The aforementioned PR removes validation tests in favor of sensible defaults. 

Turns out removing them exposes outdated entry and output.path validations that lived in the cli at one time. This PR removes those validations. 

**Other information**
Also bumping optional dependency version for webpack to alpha.1


package.json
```diff
@@ -86,7 +86,7 @@
   },
   "optionalDependencies": {
     "schema-utils": "^0.4.2",
-    "webpack": "^4.0.0-alpha.0",
+    "webpack": "^4.0.0-alpha.1"
```
azz: Missing comma :)

PR #227, opened 2017-12-22
feat(0CJS): add remaining compat for new defaults to output path

TheLarkInn:

N/A (?)

Will create issue and add to backlog from webpack/webpack

**Summary**
This disables `webpack in out` for entry and output properties now that `webpack` alone will generate the defaults. 

* Instead if you would like to pass an array of entries, and output, you must specify `-o`: `webpack entry1.js entry2.js -o "[name].js"`

* If a relpath is specified, the defaulted output.path property will be overridden with the specified relpath, and filename will be name of the file from relpath. If no `./` infront of value, then output.path will yield to webpack's default to `path.join(process.cwd(), "dist")` .

* mode will also be defaulted so mode checks are not needed either. 

**Yes**, this change forces you to have to change cli procedure for users who use `webpack in out`, as this will now (with output defaulted), convert to `entry: [arg1, arg2, etc.]`.

**Other information**

### DIFF
--- a/bin/convert-argv.js
+++ b/bin/convert-argv.js
@@ -7,11 +7,10 @@ module.exports = function(yargs, argv, convertOptions) {
 	var options = [];
 	// Shortcuts
 	if (argv.o) {
-		var pathMeta = argv.o.split(path.sep);
+		argv["output-filename"] = path.basename(argv.o);
 
-		argv["output-filename"] = pathMeta.pop();
-		if (pathMeta[0] === ".") {
-			argv["output-path"] = path.resolve(process.cwd(), pathMeta.join(path.sep));
+		if (!path.isAbsolute(argv.o)) {
+			argv["output-path"] = path.resolve(process.cwd(), path.dirname(argv.o));
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


bin/convert-argv.js
```diff
@@ -6,6 +6,15 @@ var prepareOptions = require("./prepareOptions");
 module.exports = function(yargs, argv, convertOptions) {
 	var options = [];
 	// Shortcuts
+	if (argv.o) {
+		var pathMeta = argv.o.split(path.sep);
+
+		argv["output-filename"] = pathMeta.pop();
```
michael-ciniawsky:
```js
path.basename(argv.o)
``` 

?

bin/convert-argv.js
```diff
@@ -6,6 +6,15 @@ var prepareOptions = require("./prepareOptions");
...
+		if (pathMeta[0] === ".") {
+			argv["output-path"] = path.resolve(process.cwd(), pathMeta.join(path.sep));
```
michael-ciniawsky:
```js
path.resolve(process.cwd(), path.dirname(argv.o))
```

?

bin/convert-argv.js
```diff
@@ -6,6 +6,15 @@ var prepareOptions = require("./prepareOptions");
 module.exports = function(yargs, argv, convertOptions) {
 	var options = [];
 	// Shortcuts
+	if (argv.o) {
+		var pathMeta = argv.o.split(path.sep);
+
+		argv["output-filename"] = pathMeta.pop();
+		if (pathMeta[0] === ".") {
```
michael-ciniawsky:
```js
if (!path.isAbsolute(argv.o))
```

?

sokra:
Why is there no test for this?

Why does the CLI capture coverage for `bin/*.js`?

PR #239, opened 2018-01-06
rename bot to cli bot

ev1stensberg:


ev1stensberg:
@ematipico

ev1stensberg:
hm, okay... So what's the purpose of the name tag then, if its bound to the organization name?

ematipico:
Because you can publish on AWS multiple bots with different names. @sokra suggested to use the bot that we already have since it is already up and running for the current organisation.

ev1stensberg:
Gotcha, okay let's close this, its picky-mcnittywork

sokra:
The propose of the name tag is to avoid two bots acting on the same repo. The deployed bot only operates on repos with the current name in the configuration.

Two bot operating on the same repo can have bad side-effects like bot talking with each other in a infinite loop...

ev1stensberg:
That would be really cool to watch happen @sokra 😆

PR #238, opened 2018-01-06
Fixing failing tests

ematipico:
**What kind of change does this PR introduce?**

Bugfix existing tests.

**Did you add tests for your changes?**
Yes

**If relevant, did you update the documentation?**
No

**Summary**
Fixing some failing tests, probably due to the new version of webpack

**Does this PR introduce a breaking change?**

No

**Other information**

PR #229, opened 2017-12-30
reduce coupling between webpack and webpack-cli

sokra:
**What kind of change does this PR introduce?**
refactoring

**Did you add tests for your changes?**
existing tests

**If relevant, did you update the documentation?**
N/A

**Summary**
* reduce coupling between webpack and webpack-cli
* `require("webpack/lib/DefinePlugin")` -> `require("webpack").DefinePlugin`
* move two files from webpack to webpack-cli. They are only used by webpack-cli.

**Does this PR introduce a breaking change?**

yes, since `webpack.Stats` is not exposed by webpack versions <= 4.0.0-alpha.2, this change will only work with higher versions. It was exposed by https://github.com/webpack/webpack/commit/289ba67ed68483a3e41204091148eb55ceb49e01

**Other information**
The goal is to reduce the coupling between webpack and webpack-cli.
webpack-cli should only use the public webpack API.

PR #240, opened 2018-01-08
[feature] configuration validation

jbottigliero:
- uses JSON schema to validate configuration types.
  - validated via `webpack/lib/validateSchema`
- adds tests for various configuration types
  - adds negative tests for an invalid type and array.

Feature

Yes – I'll probably add more based on actual desired implementation.

N/A – for now.

**Summary**

This is an port of webpack/webpack#6255 which is intended to resolve webpack/webpack#4977.

> No, it's not expected to... However, it may cause previously invalid configuration objects to throw errors not seen before.

**Other information**

Some outstanding questions (from webpack/webpack#6255):

> - Is using JSON schema validation too much for this?
>   - It does seem like it might be better to combine this validation with the configuration options validation. This would be a much larger change and is worthy of a discussion.
> - Should it be using the `WebpackOptionsValidationError` class? The updated output still leaves a bit to be desired. This could be accomplished using the same class or it's own.

With the code now being in separate repositories/modules, I could see how the proposed solution is less than ideal. I'm up for discussing how to move forward!


jbottigliero:
I _believe_ the failing tests here are addressed in #238

ematipico:
Yeah exactly, these are tests due to the new version of `webpack` and how it outputs the entry points.

ev1stensberg:
Let's wait for the fix on that to land and then do a review/rebase

bin/convert-argv.js
```diff
@@ -3,6 +3,10 @@ var fs = require("fs");
 fs.existsSync = fs.existsSync || path.existsSync;
 var interpret = require("interpret");
 var prepareOptions = require("./prepareOptions");
+var webpackConfigurationSchema = require("../schemas/webpackConfigurationSchema.json");
+var validateSchema = require("webpack/lib/validateSchema");
+var WebpackOptionsValidationError = require("webpack/lib/WebpackOptionsValidationError");
```
ooflorent: Please use `webpack.WebpackOptionsValidationError` instead

bin/convert-argv.js
```diff
@@ -3,6 +3,10 @@ var fs = require("fs");
 fs.existsSync = fs.existsSync || path.existsSync;
 var interpret = require("interpret");
 var prepareOptions = require("./prepareOptions");
+var webpackConfigurationSchema = require("../schemas/webpackConfigurationSchema.json");
+var validateSchema = require("webpack/lib/validateSchema");
```
ooflorent: Please use `webpack.validateSchema ` instead

ev1stensberg:
@jbottigliero mind rebasing against the master branch?

jbottigliero:
@ev1stensberg – rebased and updated. I wasn't able to reproduce the test failure in `node@6` locally, should I just re-trigger the build? Any insight here would be helpful!

@ooflorent – for the usage of `webpack.*`... I was basing my implementation/usage of webpack lib files off of `lib/commands/migration.js`. I'm not super familiar with the build process and this package's association between `webpack` core, yet. Is it safe to require webpack core here? I'll dig through the code myself, but any advice in direction would be helpful!

ev1stensberg:
@jbottigliero sure, happens sometimes, let's do another try

ev1stensberg:
Re-started your build, let's cross fingers 🤞

bin/convert-argv.js
TheLarkInn: Can you switch this over @jbottigliero?

jbottigliero:
> @ooflorent – for the usage of `webpack.*`... I was basing my implementation/usage of webpack lib files off of `lib/commands/migration.js`. I'm not super familiar with the build process and this package's association between `webpack` core, yet. Is it safe to require webpack core here? I'll dig through the code myself, but any advice in direction would be helpful!

Updated based on #229.

cc. @TheLarkInn

jbottigliero:
Thanks for taking a look @ev1stensberg and @fokusferit.

After taking another look at the code in `bin/convert-argv.js`, maybe the validation could be more strict? 

When `processConfiguredOptions` [is called](https://github.com/webpack/webpack-cli/blob/master/bin/convert-argv.js#L141), it looks like the configuration can only ever be a `Promise`, object or array of objects.

When you're using a `function` for a configuration, it's [_actually_ handled by `prepareOptions`](https://github.com/webpack/webpack-cli/blob/master/bin/convert-argv.js#L130).

Seems like I could/should also add a case for the `Promise` configuration.

fokusferit:
Hey @jbottigliero what do you think regarding about adding a test case for `Promise`configuration? Just wanted to follow up :)

ev1stensberg:
If we can strip the whole function and add a Promise clause instead that should be preferable

jbottigliero:
@fokusferit, @ev1stensberg, just took a look at this...

Unless I'm mistaking there are few things that need to happen in order to make this work (at least to validate the `Promise` instance).

It appears the `ajv-keywords` package [doesn't actually support `"instanceof": "Promise"`](https://github.com/epoberezkin/ajv-keywords/blob/master/keywords/instanceof.js#L3). I plan to open a ticket there to discuss whether or not it could/should be added – I'll follow-up on this.

An alternative would be to add a custom validator similar to[`webpack`'s addition of `absolutePath`
](https://github.com/webpack/webpack/blob/master/schemas/ajv.absolutePath.js)... because we're sharing the validator, I'll need to open a PR against `webpack` core.

I think ideally we get `Promise` added to the `ajv-keywords` package.

For now, I've added two new tests and resolved merge conflicts.

jbottigliero:
(tests are passing now because it's still just `"instanceof": "Function"`)

jbottigliero:
Support for `"instanceof": "Promise"` was just merged – epoberezkin/ajv-keywords#51 (closed epoberezkin/ajv-keywords#51).

Once released, I'll open a PR bumping `webpack` core's dependency version and update here.

ev1stensberg:
Maybe we can keep it like this an open an issue up for discussion? Instead of rushing it

jbottigliero:
@ev1stensberg works for me. I _did_ just open webpack/webpack#6430...

The dependency version bump in core _does_ feel pretty heavy-handed based on the original intent of this PR.

I'll be watching for feedback, thanks again.

ev1stensberg:
Can you make an issue at webpack-cli with what we've just discussed and link to the  6430 and this pr? Thanks!

jbottigliero:
The [console.error](https://github.com/webpack/webpack-cli/pull/240/files#diff-daabbe74e88640829a3ad9e9899efd8cR163) will log:

```
Invalid configuration object. Webpack has been initialised using a configuration object that does not match the API schema.
 - configuration should be one of these:
   object | [object] | function
   Details:
    * configuration should be an object.
      -> A webpack configuration object.
    * configuration should be an array:
      [object]
      -> An array of webpack configuration objects.
    * configuration should be an instance of function
      -> A promise that resolves with a configuration object, or an array of configuration objects.
```

I can certainly add something along the lines of `Received: ${typeof options} (${options})` to this.

ev1stensberg:
Previous looked fine, but some extra indication would be nice, like the schema, but:

```

Webpack has been initialised using a configuration object that does not match the API schema. Recieved: $type 
...
```

ev1stensberg:
But that might be a schema thing, so LGTM :shipit:

bin/convert-argv.js
```diff
@@ -148,9 +152,17 @@ module.exports = function(...args) {
...
+			options
+		);
+		if (webpackConfigurationValidationErrors.length) {
+			var error = new WebpackOptionsValidationError(
+				webpackConfigurationValidationErrors
+			);
 			console.error(
-				"Config did not export an object or a function returning an object."
+				error.message,
+				`\nReceived: ${typeof options} : ${JSON.stringify(options, null, 2)}`
```
ev1stensberg: Check the json stringify callback to b correct @fokusferit @ematipico

ematipico: I would add a newline `\n` straight after `:`. But the rest looks good to me

ematipico:
I really like the correct indentation of the configuration error, it makes more cleaner and easier to read!

ev1stensberg:
Thanks ☝️💙💙

PR #241, opened 2018-01-11
[feature] --silent

jbottigliero:
- Adds `--silent` option for suppressing output to stdout [#22]

feature

Yes - the test is a duplicate of `test/binCases/stats/multi-config`, with some modifications to try and force _more_ output.

webpack/webpack.js.org#1781

**Summary**
Related Issue: #22 

This is possibly a naïve implementation, but seems to get the job done now that `webpack-cli` is it's own module.

Currently this change will **not** suppress output for the "non-compilation" commands. I can update this PR if it _should_, however most seem like they would not benefit from a `--silent` flag.

As far as the name for the actual flag, I can change that as well...  @ev1stensberg [previously mentioned](https://github.com/webpack/webpack-cli/issues/22#issuecomment-277516967) `-s/--save` was taken at one point, but it looks open now?

No


ev1stensberg:
Thanks! Could you remove the quiet flag and its code while at it? Cc @bebraw

jbottigliero:
@ev1stensberg – I'm not so sure it was ever actually implemented. I'm not seeing anything in `webpack@[next|master|webpack-1|webpack-2]`, or here... and based on the conversation in webpack/webpack#490 it seems like it could have just been abandoned before adding a ticket here (#22)

ev1stensberg:
Remember adding the feature, maybe it has been discarded. I'll check and get back to you during the day

PR #242, opened 2018-01-12
fix(json): json parameter now enables modules

axe312ger:
bugfix

Nope, wanted to extend tests for `--json` but couldn't find any. Should I do so? (tests are actually failing locally for me, even without my change)

nope, not needed IMHO

**Summary**

Fixes https://github.com/webpack/webpack/issues/6307

nope

**Other information**

### DIFF
--- a/bin/webpack.js
+++ b/bin/webpack.js
@@ -281,7 +281,10 @@
 			});
 
 			ifArg("json", function(bool) {
-				if (bool) outputOptions.json = bool;
+				if (bool) {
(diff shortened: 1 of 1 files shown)


ev1stensberg:
@axe312ger could you add a test to this?  Also, print the failing tests here as a screenshot :)

axe312ger:
Failed in node 6 only via a timeout. Maybe just retrigger this one build?

ematipico:
I have it a rebuild, let's see how it turns

Issue #124, opened 2017-05-08
Error message when no options are passed


montogeek:
Hi!

Congrats for the release!

I am getting a duplicate message when no options are passed to the CLI, I know it needs at least one option.

It would be nice to have `-h` printed, or a little more user friendly error printed.

![image](https://cloud.githubusercontent.com/assets/1002461/25823097/914a1952-3400-11e7-9384-7bded95981a3.png)

ev1stensberg:
I can see why this is important! @okonet is doing some stuff to revamp the CLI #121 , and we're planning to make the old CLI better visually too :)

ev1stensberg:
@kingdaro This one might be nice too, if no args present, add `-h` to the argv array

kingdaro:
Sure, I can pick this one up another time.

sumit-gupta91:
Hi, I would like to pick this bug up if no one is working on it.

kingdaro:
@sumit-gupta91 You can go ahead. I have a solution but I'm not all that happy with it.

PR #252, opened 2018-01-26
Show help on no command

kingdaro:
Shows the full help content when no arguments are given to the CLI, and when the webpack build fails. 

Yes.

**Summary**
See https://github.com/webpack/webpack-cli/issues/124

Before, when running the webpack build without a config, the CLI runs webpack with the default config, and fails with an unhelpful, possibly confusing error message. The original issue suggests to show help to solve this issue.

The reasons behind these specific conditions is to allow the default configuration to run as-is without arguments.

No

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # IDE folders
 .idea
+.vscode
 
 # Mac OS X Finder
 .DS_Store
--- a/bin/webpack.js
+++ b/bin/webpack.js
@@ -192,7 +192,8 @@
 			type: "string",
 			default: "info",
 			group: DISPLAY_GROUP,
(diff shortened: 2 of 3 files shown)


ematipico:
Hi @kingdaro, just noticed now that there's also another PR that tries to fix the same issue. I am confused. But I looked at the changes and it looked different from #254 ! Could you provide a small background inside the description of the PR? Please use the template provided :)

kingdaro:
Sure, updating it in just a moment.

kingdaro:
Right now in this PR it shows the _full_ help text, but I like the idea from the other PR that instead says "run again with --help for more options" for brevity. That's probably the better approach.

PR #263, opened 2018-02-09
Refactor bin directory

EugeneHlushko:

Refactor bin directory to be aligned with code style according to node version support.

it is not necessary

**Summary**

no
**Other information**

### DIFF
--- a/bin/config-yargs.js
+++ b/bin/config-yargs.js
@@ -1,11 +1,11 @@
 var CONFIG_GROUP = "Config options:";
 var BASIC_GROUP = "Basic options:";
-var MODULE_GROUP = "Module options:";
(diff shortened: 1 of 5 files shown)


bin/config-yargs.js
```diff
@@ -1,11 +1,11 @@
 var CONFIG_GROUP = "Config options:";
```
ematipico: I guess also these can be `const`s?

EugeneHlushko: wow how did they get through 👍

EugeneHlushko:
The failure doesnt look valid :/

ematipico:
Kicked a new build. Probably it's a random failure

EugeneHlushko:
rebased onto master just in case

PR #264, opened 2018-02-11
Issue 249 fixed and other enums refactored

### LINKED ISSUE #249: Add info-verbosity into json schema

Feature

info-verbosity enum range is set within the code

info-verbosity enum is set in the schema, so there is no validation for it outside of validateOptions
We are adding this option before json schema landed, but PR is there and should go in soon.

**Please mention other relevant information such as the browser version, Node.js version, Operating System and programming language.**

Work can be started when #240 lands into master

EugeneHlushko:

Validation in proper place

yes

**Summary**
FIxes #249 and small refactor of similar enum string options

**Other information**

### DIFF
--- a/test/binCases/errors/info-verbosity/stdin.js
+++ b/test/binCases/errors/info-verbosity/stdin.js
@@ -1,7 +1,6 @@
 "use strict";
 
 module.exports = function testAssertions(code, stdout, stderr) {
-	expect(stderr[3]).toContain("Error: Invalid configuration object.");
-	expect(stderr[4]).toContain("configuration['info-verbosity'] should be one of these:");
-	expect(stderr[5]).toContain("\"none\" | \"info\" | \"verbose\"");
+	expect(stderr[stderr.length - 3]).toContain("Invalid values:");
+	expect(stderr[stderr.length - 2]).toContain("  Argument: info-verbosity, Given: \"false\", Choices: \"none\", \"info\", \"verbose\"");
 };
--- a/bin/config-yargs.js
+++ b/bin/config-yargs.js
@@ -88,7 +88,8 @@ module.exports = function(yargs) {
 			},
 			mode: {
 				type: "string",
-				describe: "Mode to use (production or development)",
+				choices: ["development", "production"],
+				describe: "Mode to use",
 				group: CONFIG_GROUP,
 				requiresArg: true
 			},
--- a/bin/webpack.js
+++ b/bin/webpack.js
@@ -179,9 +179,9 @@
 		},
 		display: {
 			type: "string",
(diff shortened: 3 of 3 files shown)


EugeneHlushko:
Hi @ev1stensberg
When we run with `--display` what is the expected behavior? because option has nothing for empty string or am i wrong, see second commit

ev1stensberg:
Maybe this is apart of the 0CJS strategy, don't really know. Anyways, I'm good with this, the schema validation will pick up on the missing things I assume.

test/binCases/errors/info-verbosity/stdin.js
```diff
@@ -1,7 +1,6 @@
 "use strict";
 
 module.exports = function testAssertions(code, stdout, stderr) {
-	expect(stderr[3]).toContain("Error: Invalid configuration object.");
-	expect(stderr[4]).toContain("configuration['info-verbosity'] should be one of these:");
-	expect(stderr[5]).toContain("\"none\" | \"info\" | \"verbose\"");
+	expect(stderr[stderr.length - 3]).toContain("Invalid values:");
+	expect(stderr[stderr.length - 2]).toContain("  Argument: info-verbosity, Given: \"false\", Choices: \"none\", \"info\", \"verbose\"");
```
ev1stensberg: Maybe this could be done without slicing at the end of the array? If you already know the size of it?

EugeneHlushko: Total length will vary because adding more options into `yargs.options` will change the total length as invalid options do print out the whole usage info, but i know that -3 and -2 will be error specific values that i want to check

EugeneHlushko:
Ok in that case it is good (re `""`)

test/binCases/errors/info-verbosity/stdin.js
ev1stensberg: gotcha, but the tests are static, yes? So the size will always be constant

ev1stensberg: Unless there's 9 different tests for 1 case here ?

EugeneHlushko: yeah sorry, so any update to their description or removal\addition will change lines amount

ev1stensberg: right. Let's keep it for now, could change it later if needed

EugeneHlushko: true, thank you!

EugeneHlushko:
Not clear if there is an action point left here @ev1stensberg , can you add review pls

EugeneHlushko:
haha thank you ❤️

PR #272, opened 2018-02-20
misc(refactor): reduce code duplication use process.exitCode instead of process.exit

EugeneHlushko:

* small code duplication reduce
* use `process.exitCode = 1` vs `process.exit(1)` 

https://nodejs.org/dist/latest-v9.x/docs/api/process.html#process_process_exit_code

>Rather than calling process.exit() directly, the code should set the process.exitCode and allow the process to exit naturally by avoiding scheduling any additional work for the event loop:

no

not relevant

**Summary**
Progressing in the `bin` directory

no

**Other information**
none

### DIFF
--- a/bin/convert-argv.js
+++ b/bin/convert-argv.js
@@ -237,7 +237,7 @@ module.exports = function(...args) {
 		function ifArg(name, fn, init, finalize) {
 			const isArray = Array.isArray(argv[name]);
 			const isSet = typeof argv[name] !== "undefined" && argv[name] !== null;
-			if (!isArray || !isSet) return;
+			if (!isArray && !isSet) return;
 
 			init && init();
 			if (isArray) argv[name].forEach(fn);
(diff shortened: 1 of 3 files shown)


ev1stensberg:
Mind checking out the failing test cases? Might be cause exitCode is async, so the tests get polluted

bin/convert-argv.js
```diff
@@ -235,23 +235,14 @@ module.exports = function(...args) {
...
-					init();
-				}
-				fn(argv[name], -1);
-				if (finalize) {
-					finalize();
-				}
-			}
+			const isArray = Array.isArray(argv[name]);
+			const isSet = typeof argv[name] !== "undefined" && argv[name] !== null;
+			if (!isArray || !isSet) return;
```
EugeneHlushko: Hmm there was wrong operator in a con here, should be `if (!isArray && !isSet) return;`

Issue #271, opened 2018-02-20
Empty Module.exports created with webpack-cli init


wangtoday:
Hi Guys, 

I was trying to use webpack-cli to generate the webpack.config.js file.  I follwed the stpes blow:

1.  In a empty folder,  run **npm init** first.
2.  run **webpack-cli init**
3.  fill with the below answers.

>  ? Will your application have multiple bundles? **No**
? Which module will be the first to enter the application? [example: './src/index'] **index.js**
? Which folder will your generated bundles be in? [default: dist]: **dist**
? Are you going to use this in production? **Yes**
? Will you be using ES2015? **Yes**
? Will you use one of the below CSS solutions? **LESS**
? If you want to bundle your CSS files, what will you name the bundle? (press enter to skip) **stylesheet.min.css**
? Name your 'webpack.[name].js?' [default: 'prod']: **config**

However, once it done, I got a webpack.config.js file without any description in the Module.export ,
```

const webpack = require('webpack');
const path = require('path');

/*
 * We've enabled UglifyJSPlugin for you! This minifies your app
 * in order to load faster and run less javascript.
 *
 * https://github.com/webpack-contrib/uglifyjs-webpack-plugin
 *
 */

const UglifyJSPlugin = require('uglifyjs-webpack-plugin');

/*
 * We've enabled ExtractTextPlugin for you. This allows your app to
 * use css modules that will be moved into a separate CSS file instead of inside
 * one of your module entries!
 *
 * https://github.com/webpack-contrib/extract-text-webpack-plugin
 *
 */

const ExtractTextPlugin = require('extract-text-webpack-plugin');

module.exports = {};
```

I thought there should be some basic configuration setting inside the module.exports, but it just return a empty obj.  I followed by [Getting started with the NEW webpack cli](https://www.youtube.com/watch?v=mE9IvFzXDFc&list=PLw5h0DiJ-9PDnVL1s-Y3HLWK8FYaCK8-e). Thought somethins is not correct with my output of webpackconfig.js, but can not figure where is wrong. 

Sorry, I am new to webpack.

ev1stensberg:
Thanks for submitting an issue, will check it out! Which version are you using?

abtExp:
Faced The Same Issue : 
 Webpack-cli : 2.0.4 
webpack : 3.11.0

ev1stensberg:
Could you guys upgrade to `webpack-cli 2.0.6` ?

wangtoday:
[...]

PR #274, opened 2018-02-22
misc(bugfix): fix 'webpack-cli init' creating empty module.exports

jevan0307:

bugfix

No

No

**Summary**
Fix issue #271 . Fix the bug that command `webpack-cli init` will create a config file with empty module.exports. 

No

**Other information**

### DIFF
--- a/lib/commands/init.js
+++ b/lib/commands/init.js
@@ -1,7 +1,8 @@
 "use strict";
 
 const npmPackagesExists = require("../utils/npm-packages-exists");
-const creator = require("../init/index").creator;
+const defaultGenerator = require("../generators/init-generator");
+const modifyHelper = require("../utils/modify-config-helper");
 
 /**
  *
@@ -15,7 +16,7 @@ const creator = require("../init/index").creator;
 
 module.exports = function initializeInquirer(pkg) {
 	if (pkg.length === 0) {
-		return creator();
+		return modifyHelper("init", defaultGenerator);
 	}
 	return npmPackagesExists(pkg);
 };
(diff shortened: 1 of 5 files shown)


jevan0307:
@ev1stensberg I have made a commit based on your comment, please check it again.

ev1stensberg:
#277 fixes this. You can reimplement that solution!

jevan0307:
The new implementation has been committed.

btw, I also found 2 points might cause problems:
1. Yeoman will crashed with package.json with no name field. Since package.json created with `webpack-cli init` doesn't have name field, it might be a trouble in some use cases. This problem is similar to [this issue](https://github.com/Polymer/polymer-cli/issues/247).
2. We may have config files with different base names. However, commands except for `init`, such as `add`, always takes `webpack.config.js` as config file and throw an error when it does not exist. Is it reasonable?

Should I create issues for these two?

ev1stensberg:
webpack-cli isn't responsible for initializing new npm/node libs, that's what `npm init` is about. About the second bullet point, yes.

ev1stensberg:
Seems like there's some missing stuff from the branch I showed you. Do you mind if we merge that and  then you can do this issue instead?

 > We may have config files with different base names. However, commands except for init, such as add, always takes webpack.config.js as config file and throw an error when it does not exist. Is it reasonable?

jevan0307:
Did I miss something in your branch? It is fine for me to merge that and close this pr.
I will create a issue for the new problem and create a new pr for it later.

ev1stensberg:
Awesome sauce, thanks buddy! Hit me up if you need guidance.

jevan0307:
@ev1stensberg I think the assertion in `lib/utils/modify-config-helper.js` when the specified file does not exist is necessary for the command like `add`, like what `lib/commands/migrate.js` has done to handle the error. Since #279 may need the error handler, please check it asap.

ev1stensberg:
You can allow people to add args to the command, take care of the arg and pass it down to the helper

Issue #261, opened 2018-02-09
Monorepo and Dependency Strategy for the CLI


ev1stensberg:
The new webpack-cli might be good for some folks but bad for others. When I ( as an FE dev) want to run webpack to compile my architecture I don't want to get more installation time because there's new features like the migration and initialization.

In the next couple of steps of the CLI's lifecycle we're gonna do the following ( like we've already done with webpack )

- Features like init and migrate are going to be published at `webpack-addons` and then we are going to ask the user to install the dependencies for a given feature if he/she hasn't installed it yet- This way, users can benefit from both having decreased bundle size as well as having the option to add utilities like migrate and init. 

Ultimately, what I'd like to see, is a monorepo like the lighthouse team has for chrome-logger, chrome-launcher and similar. This way we can keep the packages in one repo to preserve a compact and clear structure of the CLI.

montogeek:
What is the current structure of the repo, is there any parts that could be extracted into smaller packages?

ev1stensberg:
All the separate features like serve, init, migrate and so on. No separate packages yet, with the exception of the add-ons suite

EugeneHlushko:
Is there a way to contribute? Who is going to create `webpack-addons-xxx` repos under webpack organization?

ev1stensberg:
Gonna take this up with the team later, soon finished with my assignments so that I can actually do something soon. Would be good to discard webpack add-ons and host the packages as a minorepo in webpack-cli

billyjanitsch:
For one-off chores like init and migration, why not make each tool a separate pkg/CLI, and encourage users to run them using `npx`? e.g. `npx webpack-init` or `npx webpack-migrate`. Then users don't incur the cost of including them as dev dependencies.

I don't really see the purpose of a monolithic CLI that doesn't actually includes its features, especially when a great alternative exists via `npx`. It seems like a maintenance burden without any real advantage.

ev1stensberg:
Yes, that's the plan

ev1stensberg:
And while it is a burden, it allows us to be flexible around composition and we also allow people to use the AST transforms directly

ev1stensberg:
This is resolved and included in a newer version of the webpack-cli.

PR #275, opened 2018-02-22
[Don't merge] move to yarn workspaces and lerna

sendilkumarn:
This PR tries to convert the repo into monorepo using yarn workspaces and lerna

Let us have a discussion on this 

@ev1stensberg @fokusferit @ematipico


ev1stensberg:
This might not be ideal, I'm chatting with some folks at chrome to see the best option

ematipico:
Yarn workspaces it's basically a good way to handle lerna multiple projects. Do they can co-exists together?

fokusferit:
I would like to have different dimensions answered here:

- What kind of problem do we want to solve here?
- Is monorepo the only answer to solve that? Maybe there is a "better"/"faster"/"easier" way to solve the problem?
- What are the trade-offs in moving to a monorepo vs. not? Do we really improve/win on different dimensions (speed of development, maintenance, release cycle etc. ?)

I will try to ping some friends from the Go and K8s community as they have experience with monorepo's.

ev1stensberg:
I think we should take a step back and use scoped packages instead. Reopening later, but `lerna` seems like an overkill, w.r.t `Flow`, `packages` and so on compared to having sub-directories with `package.json`s

sendilkumarn:
#261

fokusferit:
Are we now back with monorepo and npx support?

ev1stensberg:
Yep, figured out it was the best option.

sendilkumarn:
Let me open a new PR.

PR #276, opened 2018-02-22
Revert "Show help on no command"

sokra:
Reverts webpack/webpack-cli#252

It shows help way to often, even when config file was provided.

I propose to revert this change until a better solution is found.


ev1stensberg:
no

dhruvdutt:
Seems like this PR changed it. Look at files diff.
image

ev1stensberg:
`ls -la bin/webpack.js`, nothing broke.

PR #277, opened 2018-02-23
ast(init): fix init command

ev1stensberg:

**Summary**

**Other information**

### DIFF
--- a/lib/init/transformations/index.js
+++ b/lib/init/transformations/index.js
@@ -98,7 +98,9 @@ module.exports = function runTransform(webpackProperties, action) {
 			.filter(e => e[1]);
 
 		const ast = j(
-			action ? webpackProperties.configFile : "module.exports = {}"
+			action && action !== "init"
+				? webpackProperties.configFile
+				: "module.exports = {}"
 		);
 		const transformAction = action || null;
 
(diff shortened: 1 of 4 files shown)


lib/init/transformations/index.js
```diff
@@ -111,15 +113,16 @@ module.exports = function runTransform(webpackProperties, action) {
 		})
 			.then(_ => {
 				let configurationName;
-				if (!config.configName) {
+				if (!config.configName && action !== "init") {
```
ematipico: You could save the expression `action !== "init"` in variable
```js
const isActionNotInit = action !== "init";
```

package.json
```diff
@@ -31,13 +31,11 @@
     "travis:integration": "yarn prepare && nyc jest && yarn reportCoverage",
     "travis:lint": "yarn lint && yarn bundlesize",
     "bundlesize": "bundlesize",
-    "commitmsg": "commitlint -e $GIT_PARAMS",
```
ematipico: Why this has been removed?

ematipico:
@ev1stensberg is there an issue to understand what we are trying to fix with this PR?

package.json
ev1stensberg: errors even though commit is correct

lib/init/transformations/index.js
ev1stensberg: its used one time, so why bother?

ev1stensberg:
Yeah, #274

lib/init/transformations/index.js
ematipico: No it is not, I see twice in the code changes.

Issue #258, opened 2018-01-31
configuration validation : strict "Promise" instanceof validation


jbottigliero:
**Do you want to request a *feature* or report a *bug*?**
Feature.

**What is the current behavior?**
webpack/webpack-cli#240 introduces validation which causes `Promise` based configurations to pass due to `instanceof Function` being `true`.

**What is the expected behavior?**
At the point the validation runs, the configuration can never be _just_ a `Function`, it's expected to be an object, an array of objects, or a `Promise`.

**If this is a feature request, what is motivation or use case for changing the behavior?**
Provide stricter validation schema.

**Please mention other relevant information such as the browser version, Node.js version, Operating System and programming language.**

The main concern with adding the above is that it requires webpack/webpack#6430.

The `ajv-keywords` package does not support `"instanceof": "Promise"` in versions `<3.1.0`, which has a `peerDependency` of `ajv@^6.0.0`.

An alternative to updating the package versions would be to add `Promise` as a "custom" constructor [(outlined here)](https://github.com/epoberezkin/ajv-keywords/tree/v2.1.1#instanceof), but this would still have to be implemented in `webpack/webpack` based `ajv` being configured in core, and not directly exposed.

jbottigliero:
I have a branch ready with this change that I will PR once webpack/webpack#6430 is part of a tagged release.

ev1stensberg:
Sounds good

dhruvdutt:
webpack [4.0.0](https://github.com/webpack/webpack/releases/tag/v4.0.0) is out. :fire: 

@jbottigliero Can we make the PR now?

jbottigliero:
@dhruvdutt – opened #298, cheers!

jbottigliero:
#298 was merged to `master` – closing.

PR #298, opened 2018-03-01
[feature] strict Promise configuration validation

jbottigliero:
- updates configuration schema to use `"instanceof": "Promise"` (webpack/webpack-cli#258)

feature/refactoring

Covered by existing tests.

n/a

**Summary**
Previously, validation for `Promise` configurations passed via `"instanceof": "Function"`, with `webpack@4+` (`ajv-keywords@3+`), we can do strict validation on `"instanceof": "Promise"`.

To accommodate this change we have to change the check to `anyOf` due to `Promise` passing `"type": "object"`.

```js
> const Ajv = require('ajv');
undefined
> const ajv = new Ajv;
undefined
> ajv.validate({type:'object'}, new Promise(res => res));
true
```

### DIFF
--- a/schemas/webpackConfigurationSchema.json
+++ b/schemas/webpackConfigurationSchema.json
@@ -1,5 +1,5 @@
 {
-  "oneOf": [
+  "anyOf": [
     {
       "type": "object",
       "description": "A webpack configuration object."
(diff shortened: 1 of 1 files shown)


jbottigliero:
Didn't realize `commitlint`was enabled now... amended my commit message to follow the spec.

EugeneHlushko:
totally legit 👍

PR #309, opened 2018-03-04
Clean up dependencies

### LINKED ISSUE #306: Many devDependencies listed as dependencies

Bug

Many devDependencies of this project are listed as dependencies in package.json such as [prettier](https://github.com/webpack/webpack-cli/blob/3aabdf473b79dcd2238291fee961851df7f6a7cd/package.json#L102) and [codecov](https://github.com/webpack/webpack-cli/blob/3aabdf473b79dcd2238291fee961851df7f6a7cd/package.json#L84). This causes them to be installed when a user installs webpack-cli.

There are also some dependencies that don't appear to be used at all, such as [global](https://github.com/webpack/webpack-cli/blob/3aabdf473b79dcd2238291fee961851df7f6a7cd/package.json#L89).

If a dependency isn't needed at runtime, it shouldn't be listed in dependencies. 

Only the dependencies necessary to run webpack-cli are installed alongside it.

webpack-cli is currently a _huge_ install (~66mb). It would be nice to cut down on the dependencies it requires, especially since this project is now required to interact with webpack via CLI.

### LINKED ISSUE #307: Unpin dependencies where possible

Bug.

Currently the dependencies on log-symbols and yargs are [pinned to exact versions](https://github.com/webpack/webpack-cli/blob/3aabdf473b79dcd2238291fee961851df7f6a7cd/package.json#L98). Is there any reason for this?

(I realize there are reasons not to upgrade yargs to `11.x` but it's currently pinned to a specific `9.x` version, i.e. `9.0.1` rather than `^9.0.1`)

The dependencies on log-symbols and  yargs are declared as `^2.2.0` and `^9.0.1`, respectively.

Generally pinning dependencies is bad for the ecosystem because it prevents deduping between packages.

billyjanitsch:

A general clean-up of dependencies.

**Did you add tests for your changes? If relevant, did you update the documentation?**

No, but I'm happy to if you'd like.

**Summary**

- Remove unnecessary dependency on global (seems to have been introduced unintentionally in #231, probably from running `npm i global pkg` instead of `npm i --global pkg`)
- Move codecov, which is never used at runtime, from dependencies to devDependencies (fixes #306)
- Unpin dependencies on log-symbols and yargs (fixes #307 -- note that this does _not_ upgrade yargs to the latest version, which I know you don't want)
- Remove seemingly unnecessary dependencies on recast and uglifyjs-webpack-plugin (not sure about this one, but I can't find any mentions of these in the source)

Let me know if you'd like me to back out or further discuss any of the changes individually.

No. These are intended not to have any behavioral effect on the CLI.

**Other information**


package.json
```diff
@@ -95,18 +93,16 @@
...
     "p-lazy": "^1.0.0",
     "prettier": "^1.5.3",
-    "recast": "^0.14.4",
     "resolve-cwd": "^2.0.0",
     "supports-color": "^5.2.0",
-    "uglifyjs-webpack-plugin": "^1.2.2",
     "v8-compile-cache": "^1.1.2",
     "webpack-addons": "^1.1.5",
-    "yargs": "9.0.1",
+    "yargs": "^9.0.1",
```
ev1stensberg: @shellscape which version was corrupted?

shellscape: I believe it was 8.x, but I can't recall for sure.

package.json
```diff
@@ -95,18 +93,16 @@
     "listr": "^0.13.0",
     "loader-utils": "^1.1.0",
     "lodash": "^4.17.5",
-    "log-symbols": "2.2.0",
+    "log-symbols": "^2.2.0",
     "mkdirp": "^0.5.1",
     "p-each-series": "^1.0.0",
     "p-lazy": "^1.0.0",
     "prettier": "^1.5.3",
-    "recast": "^0.14.4",
```
ematipico: We don't use `recast` inside the code? :O

billyjanitsch:
@ematipico prettier is [used in lib code](https://github.com/webpack/webpack-cli/blob/3aabdf473b79dcd2238291fee961851df7f6a7cd/lib/utils/run-prettier.js), I think to format webpack configs that it outputs?

package.json
billyjanitsch: No, a recast object is passed around but it looks like it's created by jscodeshift (which has its own dependency on a different version of recast)

ematipico: Great! Thanks

billyjanitsch:
Done! I also refreshed the package-lock.json, but I can revert that if you'd prefer.

PR #325, opened 2018-03-08
chore(ES6): changed 'var' into 'const' where convenient

ematipico:
Refactor, removed last references to `var`

No, but I updated one snapshot

No

**Summary**
Slowly moving to ES6

No


lib/utils/copy-utils.js
```diff
@@ -34,14 +34,14 @@ var generatorCopy = (
...
  */
-var generatorCopyTpl = (
+const generatorCopyTpl = (
 	generator,
 	templateDir,
 	templateData
 ) => /** @param {string} filePath */ filePath => {
-	var sourceParts = templateDir.split(path.delimiter);
+	const sourceParts = templateDir.split(path.delimiter);
 	sourceParts.push.apply(sourceParts, filePath.split("/"));
```
ev1stensberg: Mutable object, but okay..

ematipico: It doesn't get reassigned, so `const` is ok. Otherwise eslint would have failed :)

lib/generate-loader/index.js
```diff
@@ -1,13 +1,13 @@
-var yeoman = require("yeoman-environment");
-var LoaderGenerator = require("../generators/loader-generator").LoaderGenerator;
+const yeoman = require("yeoman-environment");
+const LoaderGenerator = require("../generators/loader-generator").LoaderGenerator;
```
sendilkumarn: can we use this `const { LoaderGenerator } = require("../generators/loader-generator")` ?

lib/generate-plugin/index.js
```diff
@@ -1,13 +1,13 @@
-var yeoman = require("yeoman-environment");
-var PluginGenerator = require("../generators/plugin-generator").PluginGenerator;
+const yeoman = require("yeoman-environment");
+const PluginGenerator = require("../generators/plugin-generator").PluginGenerator;
```
dhruvdutt: Can you change to `const { PluginGenerator }`

lib/utils/ast-utils.test.js
```diff
@@ -132,12 +132,12 @@ var a = { plugs: [] }
 	describe("findVariableToPlugin", () => {
 		it("should find the variable name of a plugin", () => {
 			const ast = j(`
-			var packageName = require('package-name');
-			var someOtherVar = somethingElse;
-			var otherPackage = require('other-package');
+			const packageName = require('package-name');
+			const someOtherconst = somethingElse;
```
dhruvdutt: ~`someOtherconst`~ => `someOtherConst`

lib/utils/copy-utils.js
ev1stensberg: The `.push` method is mutating it, ( even though the push method is immutable ) but that's another thing :D

lib/generate-plugin/index.js
ev1stensberg: Can we use ES6 destructing now? Cool!

Issue #321, opened 2018-03-08
init command shoult use `src/index.js` as entry module by default


montogeek:
Feature

**What is the current behavior?**
`init` command does not provide a default value for entry module.
```
👉  ./node_modules/.bin/webpack init

ℹ INFO For more information and a detailed description of each question, have a look at https://github.com/webpack/webpack-cli/blob/master/INIT.md
ℹ INFO Alternatively, run `webpack(-cli) --help` for usage info.

? Will your application have multiple bundles? No
? Which module will be the first to enter the application? [example: './src/index']
>> Please specify an answer!
```

**What is the expected behavior?**

`init` command should provide a default value `src/index.js` as entry module following webpack itself approach to zero configuration

dhruvdutt:
Can we discuss this over #315 as it's quite related?

ev1stensberg:
Actually we can leave it blank by default, as 0CJS can handle it

ev1stensberg:
Just want to add that prompting that default is `src/index` is good to make sure users understand we're actually setting defaults.

vijaykrishnavanshi:
Made a PR with this change.

vijaykrishnavanshi:
Doing same for multiple bundle case also would be a good idea. 
Thoughts?

EugeneHlushko:
I am not sure you want to add something that is already default, i actually omited it in one of my PR earlier: https://github.com/webpack/webpack-cli/pull/356/files#diff-1b2b84fcd9935ef62f04b2ffd905d2c6R86

PR #335, opened 2018-03-11
misc(init): use `src/index.js` as entry module by default

vijaykrishnavanshi:

This PR adapt init command to use ./src/index as default with proper prompt. According to the discussion on issue #321.

No, but tests pass locally.

N/A

**Summary**

Allow webpack-cli to generate config file without much blocking.
Issue #321 

No

### DIFF
--- a/lib/generators/utils/entry.js
+++ b/lib/generators/utils/entry.js
@@ -1,6 +1,7 @@
 "use strict";
 
 const InputValidate = require("webpack-addons").InputValidate;
+const Input = require("webpack-addons").Input;
 const validate = require("./validate");
 
 /**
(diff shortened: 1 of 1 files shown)


vijaykrishnavanshi:
@ev1stensberg but considering the fact that 0CJS will be configurable in future will this be a good idea.
By removing do you mean no entry param in config?

PlayMa256:
@vijaykrishnavanshi yes. Webpack already defaults that to `./src/index`

ev1stensberg:
No entry, that's correct. If there's 0CJS then there's no point setting it if user ignores entry

ematipico:
Is this PR still valid? As far as I understood there's no need to set any default as we have the 0CJS in place that does the job. Worth closing?

PR #340, opened 2018-03-12
Support --build-delimiter for opt-in output delimiter

### LINKED ISSUE #192: Print distinct end line to support better output error parsing in watch mode

A feature.

The output consists of various information regarding the included bundles, time taken, etc., followed by warnings and errors if there are any, but nothing follows that, indicating that the compilation is fully completed.

\-

Print something unique like `--- Compilation completed. ---` so the output of one cycle can be found between the `Hash: XXXXX` and that line. Optionally a similar line could be added before each compilation (e.g. `--- Starting compilation. ---`).

Support editors and development tools in aggregating and highlighting issues in code.

**Please mention other relevant information such as the browser version, Node.js version, webpack version and Operating System.**

\-

---

This issue was moved from webpack/webpack#6002 by @sokra. Orginal issue was by @brunnerh.

This sounds like a useful opt-in feature.

bitpshr:
Feature

Yes

N/A

**Summary**
This PR introduces a new `--build-delimiter` CLI option to add opt-in support for custom text after  normal build output. This feature is useful for tooling that parses `stdout`, originally arose from webpack/webpack#6002, and resolves #192.

No

**Other information**
I was unsure of the purpose of `bin/process-options.js` as it doesn't appear to be used anywhere. I noticed this file has been recently updated, so I added the CLI option to both `bin/process-options.js` and `bin/webpack.js`.

### DIFF
--- a/bin/process-options.js
+++ b/bin/process-options.js
@@ -98,6 +98,10 @@ module.exports = function processOptions(yargs, argv) {
 			if (bool) outputOptions.cachedAssets = true;
 		});
 
+		ifArg("build-delimiter", function(value) {
+			outputOptions.buildDelimiter = value;
+		});
+
 		if (!outputOptions.exclude && !argv["display-modules"])
 			outputOptions.exclude = [
 				"node_modules",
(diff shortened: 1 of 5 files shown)


bitpshr:
Thanks for another quick review @ev1stensberg and @ematipico.

As far as tooling, here's one slightly contrived use case: we wrote an internal tool that pipes build output as individual Slack messages for a particularly important build that sits in watch mode. This was difficult to do; we either had to use `--json` and manually construct our own Slack message, or just directly pipe `stdout`, which was difficult since `Hash: xxxx` is the only true delimiter that currently exists, and it shows _before_ builds, not after. We ended up using `stdout`, but the logic involved will now be much simpler with a custom delimiter between runs. I'm unaware of major public tools that would use this feature. But still, even beyond tooling, it's nice to be able to more-easily distinguish one build from another with something like `--build-delimiter="\n\n---completed---\n\n"`, for example.

PR #351, opened 2018-03-17
expose mode to webpack.config.js

wood1986:
* Expose `--mode` information to `webpack.config.js`.
* Before this, we have to use `const DEBUG = require("yargs").argv.mode !== "production"`
* After that, we can use the information inside `webpack` such that `webpack.mode.DEVELOPMENT`

See `test/binCases/mode`


bin/convert-argv.js
```diff
@@ -4,9 +4,9 @@ fs.existsSync = fs.existsSync || path.existsSync;
 const interpret = require("interpret");
 const prepareOptions = require("./prepareOptions");
 const webpackConfigurationSchema = require("../schemas/webpackConfigurationSchema.json");
-const validateSchema = require("webpack").validateSchema;
-const WebpackOptionsValidationError = require("webpack")
-	.WebpackOptionsValidationError;
+const webpack = require("webpack");
```
ematipico: You could optimise like this
```js
const { validateSchema,  WebpackOptionsValidationError } = require("webpack");
```

wood1986:
I did sign it
image

michael-ciniawsky:
That's already partially possible when using a `{Function}` in `webpack.config.js` ? 🙃 

**webpack.config.js**
```js
module.exports = (env, argv) => ({
  mode: argv.mode || 'development'
  ...config
})
```

Is this really needed ?

wood1986:
* Your way is no different from `const DEBUG = require("yargs").argv.mode !== "production"`.
* Both ways are asking developers to write some codes or predicate in `webpack.config.js`. 
* We have lots of similar code if we have several `webpack.config.js`.
* I think your suggestion will never be as simple as `webpack.mode.PRODUCTION`.
* In short, I think the predicate statement should stay inside the webpack.

wood1986:
no. They are different. You still have to write the logic/predicate in the `webpack.config.js` which is I want to avoid and simplify the configuration. If you want us to achieve via the function, this means you push the responsibility to developer side.

Think about the app has 5 `webpack.config.js`s for different purposes. Each of them has debug and release mode. Then we have to write the same function to get the mode info. 

I think `webpack` itself should know which mode I am using. `NODE_PROCESS.ENV` is for the actual application code not the webpack.config.js.

ev1stensberg:
Could you post an issue with the proper template filled out? Let's see what the community says, always open to change

PR #356, opened 2018-03-20
cli(init): webpack4 ready

### LINKED ISSUE #350: `webpack-cli init`  - Support for webpack version 4

 - Feature
Creates an old starter template for webpack.config.js

CommonsChunkPlugin is deprecated and would not expect it in the template.
Add a question which version of webpack the user wants to use.

EugeneHlushko:

Make `init` command generate webpack4 ready config file

no

no

**Summary**
Fixes #350 

**Other information**


lib/generators/init-generator.js
```diff
@@ -339,8 +338,7 @@ module.exports = class InitGenerator extends Generator {
 				if (regExpForStyles) {
 					if (this.isProd) {
 						this.configuration.config.topScope.push(tooltip.cssPlugin());
-						// TODO: Replace with regular version once v.4 is out
-						this.dependencies.push("extract-text-webpack-plugin@next");
+						this.dependencies.push("extract-text-webpack-plugin");
```
dhruvdutt: We'll still need to use the pre-release version.

EugeneHlushko: yup just pushed smth to reference issue

EugeneHlushko: will work on it tonight :)

EugeneHlushko:
Yes it does, do you want to keep it simple like this? @ev1stensberg 
```
optimization.splitChunks: {
  chunks: "all",
  minChunks: 2
}
```

ev1stensberg:
Go for it

ev1stensberg:
Keep the commonChunk tip, but explain what `splItChunks` is, and that they can remove that prop if they want to.

lib/generators/utils/tooltip.js
```diff
@@ -17,13 +17,18 @@ module.exports = {
...
+ * SplitChunksPlugin is enabled by default and replaced
+ * deprecated CommonsChunkPlugin. It automatically identifies modules which
+ * should be splitted of chunk by heuristics using module duplication count and
+ * module category (i. e. node_modules). And splits the chunks…
  *
- * https://webpack.js.org/plugins/commons-chunk-plugin/
+ * It is safe to remove "splitChunks" from the generated configuration
+ * and was added as an educational example.
+ *
+ * Documentation page will be available here soon:
```
dhruvdutt: We will need to rephrase this a little once [#1863](https://github.com/webpack/webpack.js.org/pull/1863) gets merged.

EugeneHlushko:
Sorry for late notice, just realized i mixed up `minChunks` with `minSize` for `init` generator :)

EugeneHlushko:
Sorry but i didn't quite get that @ev1stensberg 
Did you mean adding a caching group per entry ?

ev1stensberg:
yep

lib/generators/utils/entry.js
```diff
@@ -55,7 +55,7 @@ module.exports = (self, answer) => {
 					self.prompt([
 						InputValidate(
 							`${entryProp}`,
-							`What is the location of "${entryProp}"? [example: "./src/${entryProp}"]`,
+							`What is the location of "${entryProp}"? [example: './src/${entryProp}']`,
```
EugeneHlushko: This one was showing example in double quotes when other lines use single quotes, changed it. If there was no good reason for it?

lib/generators/init-generator.js
```diff
@@ -375,14 +375,39 @@ module.exports = class InitGenerator extends Generator {
...
+				if (entry && typeof entry === "object") {
+					// if user had provided multiple entry points, we make cacheGroup example
+					// per entry point
+					const entries = Object.keys(entry);
+					if (entries.length > 1) {
+						this.configuration.config.webpackOptions.optimization.splitChunks.cacheGroups = {};
+						entries.map(item => {
+							const newCacheGroup = {
+								chunks: "'all'",
+								// TODO: implement RegExp from entry for test value of caching group
```
EugeneHlushko: Implementing a RegExp based on entry is not working for most cases provided for entry of the example config. It might need work on entry answer parser and optimize transformer. Right now regexp if done similar to scss ones outputs bad examples e.g.
\".\/src\/app'.js
Could use `path` do detect delimiters and `dirname` of entry and then to try putting up a regexp close to what we have in the docs: https://github.com/webpack/webpack.js.org/pull/1863/files#diff-e08b426c523b9305ce571576d5b640d0R165

But then the default scenario when you have `./src/app` and `./src/someOtherEntry` would both result in `./src/app.js` and `./src/someOtherEntry.js` and both dirnames would come put as `src` so that regexp for `cacheGroups[group].test` would be the same `\src\`

EugeneHlushko:
@ev1stensberg i did put up an update but its not very satisfying experience for simple examples like `./src/app` and `./src/otherApp` etc. Pls have a look

lib/generators/utils/tooltip.js
dhruvdutt: The documentation is available now. Can you rephrase this?

EugeneHlushko:
Hi @ev1stensberg 
Did you see my previous comment on this? I can resolve the conflict and apply those changes i suggest vs implementing a smelly RegExp for dynamic caching group as this is done for **example** anyway.

lib/generators/utils/tooltip.js
EugeneHlushko: webpack/webpack.js.org#1955 was merged, I'll rephrase this when resolving conflict

ev1stensberg:
Implement the one that you think is the best, I'll re-review once done

EugeneHlushko:
@ev1stensberg pls do now

lib/generators/init-generator.js
```diff
@@ -378,15 +374,28 @@ module.exports = class InitGenerator extends Generator {
...
-					oneOrMoreEntries.forEach(prop => {
-						this.configuration.config.webpackOptions.plugins.push(
-							createCommonsChunkPlugin(prop)
-						);
-					});
-				}
+				// add splitChunks options for transparency
+				this.configuration.config.topScope.push(tooltip.splitChunks());
+				this.configuration.config.webpackOptions.optimization = {
+					splitChunks: {
```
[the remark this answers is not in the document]
EugeneHlushko: they are defaults from the docs, we also add a comment to that link here:
https://github.com/webpack/webpack-cli/pull/356/files#diff-f722437ab3fa8013b0afe4475d557c7fR27

and here is direct link:  https://webpack.js.org/plugins/split-chunks-plugin/#optimization-splitchunks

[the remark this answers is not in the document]
ematipico: Oh yeah, I meant to ask if you put these links INTO the code as a comment :)
Sorry for misunderstanding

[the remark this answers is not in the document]
EugeneHlushko: Yeah as a quick ref 👍

lib/generators/init-generator.js
```diff
@@ -378,15 +374,29 @@ module.exports = class InitGenerator extends Generator {
...
+						minChunks: 1,
+						// for production name is recommended to be off
+						name: !this.isProd,
+						cacheGroups: {
+							vendors: {
+								test: "/[\\\\/]node_modules[\\\\/]/",
+								priority: -10
+							},
+						},
+						default: {
```
ev1stensberg: This didn't work when trying to build a web app using this PR. Removing `default` makes the app run. I think maybe that you can use the object entries from when the user types what the names of their entries should be to add chunkGroups. Say if I answer yes to the first question at the CLI and type that I've got 2 entries, `app` and `utils`, then you should use that when setting chunk groups in addition to the node_modules for vendor. Makes sense?

lib/generators/utils/entry.js
```diff
@@ -20,7 +20,7 @@ module.exports = (self, answer) => {
 			.prompt([
 				InputValidate(
 					"multipleEntries",
-					"Type the names you want for your modules (entry files), separated by comma [example: 'app,vendor']",
+					"Type the names you want for your modules (entry files), separated by comma [example: app,vendor]",
```
ev1stensberg: Use this to split groups of chunks.

lib/generators/utils/tooltip.js
```diff
@@ -17,13 +17,17 @@ module.exports = {
  *
  */`;
 	},
-	commonsChunk: _ => {
+	splitChunks: _ => {
```
ev1stensberg: Good stuff :3

lib/init/transformations/optimization/optimization.js
```diff
@@ -0,0 +1,53 @@
+"use strict";
+
+const utils = require("../../../utils/ast-utils");
```
ev1stensberg: This is okay for now, GSOC participants will fix the rest of missing stuff =)

PR #358, opened 2018-03-20
--entry should override config.entry

### LINKED ISSUE #155: Overwriting config parameters through CLI

feature

According to my understanding and tests, I can use webpack either in combination with a config file which includes the entry and output paths, or I can pass the entry and output path as CLI arguments. But both alternatives are mutually exclusive, i.e. I cannot overwrite the entry and output paths specified in a config file through CLI parameters, while still using the other specifications in the config file. 

I want to use a webpack.config.js AND overwrite the entry and output paths of that config file using CLI parameters. 

I need to change entry and output paths dynamically and depending on some context while keeping the other entries of an entry file. 

**Please mention other relevant information such as the browser version, Node.js version, webpack version and Operating System.**

---

This issue was moved from webpack/webpack#4895 by @bebraw. Orginal issue was by @fera0013.

bitpshr:
Feature

Yes

Yes

**Summary**
It should be possible to use a configuration file while also overridding specific parameters using CLI options. This is already possible for most configuration options. For example, the CLI will ignore any `output` configured in a project's `webpack.config.js` and instead use the value of `--output` if it's passed. This allows a configuration file to be used while also allowing certain options to be overridden dynamically at runtime. Unfortunately, the current implementation doesn't handle `entry` in this manner. Instead, the CLI will _append_ the value of any `--entry` option to the `entry` configuration in a project's `webpack.config.js` instead of overriding, which [prevents projects](https://github.com/webpack/webpack-cli/issues/155) from dynamically overriding `entry` points using CLI options.

This PR updates the CLI to handle `--entry` the same way as `--output`: any entry points specified via CLI options override `entry` configuration specified in a configuration file, rather than appending to it.

Resolves #155 

**Example**
_webpack.config.js_
```js
module.exports = {
	entry: {
		foo: './some/path.js'
	}
};
```

Should create one `foo` entry based on config file:
```sh
webpack-cli
```

Should override config file and create one `bar` entry:
```sh
webpack-cli --entry bar=./some/path.js
```
_or_
```sh
webpack-cli bar=./some/path.js
```

Eh, sort of. If projects were previously relying on the undocumented functionality where CLI-passed entry points were appended to configuration-defined entry points, they will break with this release. I'd argue that since the previous functionality wasn't explicit or documented that it doesn't warrant a major version bump.

**Other information**
This PR also updates the usage help text printed by the CLI since it wasn't exactly accurate. Using CLI options is not mutually-exclusive to using a configuration file, and an `entry` point can be dynamically defined as either a CLI option or a CLI argument.


ev1stensberg:
@fokusferit this might be implemented already, could you check the binTestCases?

fokusferit:
Also I'm currently thinking about this [line](https://github.com/bitpshr/webpack-cli/blob/50ec0f37da187d1db2cb933a7ea165c92b03df6d/bin/convert-argv.js#L344):

```
			function(name, entry) {
				if (
					typeof options.entry[name] !== "undefined" &&
					options.entry[name] !== null
				) {
					options.entry[name] = [].concat(options.entry[name]).concat(entry);
				} else {
					options.entry[name] = entry;
				}
			}
```

Is this if-else still necessary or valid?

We are forcing that entry is `options.entry = {}`, we also know then that options.entry[name] will always be the else block, right?

fokusferit:
@ev1stensberg you're right, sorry for that. It's here https://github.com/webpack/webpack-cli/tree/master/test/binCases/entry/multi-file ^^

bitpshr:
@ev1stensberg I removed the array check because if we get to [that point in the code](https://github.com/webpack/webpack-cli/pull/358/files#diff-daabbe74e88640829a3ad9e9899efd8cL569), we know that entries were passed as CLI arguments, so we force `options.entry` to be an empty object anyways. The array check should no longer be needed. Good point to mention though: this implementation means that CLI argument entries take precedence over CLI option entries, which in turn take precedence over config file entries. This was intentional, as a user shouldn't ever be passing both CLI arguments _and_ CLI options, e.g. `webpack-cli bar=./some/path.js --entry bar=./another/path.js`. They should just choose one syntax.

@fokusferit I left the conditional check you referenced in the code because that function as a whole appears to be called for every entry provided via CLI options. I figured we should still allow a downstream user to provide multiple paths for the same entry:

```sh
webpack-cli --entry bar=./some/path.js --entry bar=./another/path.js
```

Thanks for the review! Happy to rework any of this, just let me know.

bitpshr:
Thanks @ematipico, @fokusferit, and @ev1stensberg. Let me know if you'd like me to clarify or clean up anything else before landing this.

bitpshr:
@shellscape I'm not sure I follow; most CLI options passed dynamically at runtime override corresponding configuration properties defined in a configuration file. See [`--output`](https://github.com/bitpshr/webpack-cli/blob/50ec0f37da187d1db2cb933a7ea165c92b03df6d/bin/convert-argv.js#L35-L42), [`--context`](https://github.com/bitpshr/webpack-cli/blob/50ec0f37da187d1db2cb933a7ea165c92b03df6d/bin/convert-argv.js#L203-L205), [`--watch`](https://github.com/bitpshr/webpack-cli/blob/50ec0f37da187d1db2cb933a7ea165c92b03df6d/bin/convert-argv.js#L210-L212), [`--mode`](https://github.com/bitpshr/webpack-cli/blob/50ec0f37da187d1db2cb933a7ea165c92b03df6d/bin/convert-argv.js#L333-L335), [--target](https://github.com/bitpshr/webpack-cli/blob/50ec0f37da187d1db2cb933a7ea165c92b03df6d/bin/convert-argv.js#L464-L466), [--devtool](https://github.com/bitpshr/webpack-cli/blob/50ec0f37da187d1db2cb933a7ea165c92b03df6d/bin/convert-argv.js#L486-L488), etc., not to mention all [`output`-related configuration properties](https://github.com/bitpshr/webpack-cli/blob/50ec0f37da187d1db2cb933a7ea165c92b03df6d/bin/convert-argv.js#L406-L450) beyond just `--output`.

While I understand the desire to dynamically add to configuration-file-defined `entry` points via the CLI instead of overriding them, this makes it so it's simply not possible to override `entry` points at all without the use of a second configuration file. If a specific use case requires that entries are dynamically added at runtime via the CLI, _all_ entries will have to be passed via the CLI, not just those that are dynamic.

I disagree with the assertion that this PR trades one set of disadvantages for another: this PR makes dynamically adding to entries slightly more tedious so that overriding entries is at all possible, and does so by handling `entry` like most other CLI configuration options.

shellscape:
I'm not sure you have a full handle on the scope of the CLI options available, because "most" is not a true assertion. By your logic you should then make all flags that modify the `plugins` collection override and wipe away any and all plugins specified in the config.

> this PR makes dynamically adding to entries slightly more tedious so that overriding entries is at all possible

That illustrates my point perfectly - that is literally trading one inconvenience for another. This PR was not forward thinking, and merged with tunnel vision for only one way of doing things, whilst ignoring the other. It's a failure of vision.

bitpshr:
Ah, so we handle any modifications to the `plugins` configuration in an additive manner as well. I see your gripe with the asymmetry there. Granted, that's a slightly contrived piece of configuration to use to highlight your issues with the approach; for example, passing `--hot` has to modify the `plugins` configuration by necessity, but most users would not expect this to override all other plugins. I'd also argue that most users passing a `--plugin` option are doing so specifically to dynamically load an additional plugin, not to override and define a new plugin configuration altogether.

Is `plugins` the only remaining piece of configuration that's added to via associated CLI options instead of being overridden by them? If so, and you see the asymmetric handling of `plugins` compared to `entry` and most other configuration options as problematic, please feel free to open an issue to discuss the larger question of how such CLI options should be handled:

1. Should CLI options always override configuration file options, except for `plugins`? **(current behavior)**
2. Should CLI options always override configuration file options?
3. Should CLI options always append to configuration file options (where applicable)?
4. Should CLI options conditionally override configuration file options based on some flag?
5. ?

PR #360, opened 2018-03-20
fix: checks validity of an identifier

anu-007:
bugfix - adds a check for validity of an identifier

No
`generateIdentifierRegex` is based on [this](https://mathiasbynens.be/notes/javascript-identifiers-es6) article. A similar implementation can be seen in `esprima` [link](https://github.com/jquery/esprima/blob/master/tools/generate-identifier-regex.js).
**Summary**
`generateIdentifierRegex` generates a long regex for invalid keywords which is then checked in `ast-util.js`.

No

**Other information**
No

### DIFF
--- a/lib/utils/ast-utils.js
+++ b/lib/utils/ast-utils.js
@@ -1,4 +1,5 @@
 const hashtable = require("./hashtable");
+const validateIdentifier = require("./validate-identifier");
 
 function safeTraverse(obj, paths) {
 	let val = obj;
@@ -149,8 +150,8 @@ function createIdentifierOrLiteral(j, val) {
 			return createExternalRegExp(j, val);
 		} else {
 			// Use identifier instead
-			// TODO: Check if literalVal is a valid Identifier!
-			return j.identifier(literalVal);
+			if (!validateIdentifier.isKeyword(literalVal) || !validateIdentifier.isIdentifierStart(literalVal) || !validateIdentifier.isIdentifierChar(literalVal))
+				return j.identifier(literalVal);
 		}
 	}
 	return j.literal(literalVal);
--- a/lib/utils/validate-identifier.js
+++ b/lib/utils/validate-identifier.js
@@ -1,0 +1,128 @@
+// This is a trick taken from Esprima. It turns out that, on
+// non-Chrome browsers, to check whether a string is in a set, a
+// predicate containing a big ugly `switch` statement is faster than
+// a regular expression, and on Chrome the two are about on par.
+// https://tc39.github.io/ecma262/#sec-keywords
+function isKeyword(code) {
+	switch (code.length) {
+		case 2:
+			return (code === "if") || (code === "in") || (code === "do");
+		case 3:
+			return (code === "var") || (code === "for") || (code === "new") ||
+				(code === "try") || (code === "let");
+		case 4:
+			return (code === "this") || (code === "case") ||
+				(code === "void") || (code === "with") || (code === "enum") || (code === "true") ||
+				(code === "null") || (code === "eval");
+		case 5:
+			return (code === "while") || (code === "break") || (code === "catch") ||
+				(code === "throw") || (code === "const") || (code === "yield") ||
+				(code === "class") || (code === "super") || (code === "false") || (code === "await");
+		case 6:
+			return (code === "return") || (code === "typeof") || (code === "delete") ||
+				(code === "switch") || (code === "export") || (code === "import") ||
+				(code === "static") || (code === "public") ;
+		case 7:
+			return (code === "default") || (code === "finally") || (code === "extends") ||
+				(code === "private") || (code === "package");
+		case 8:
+			return (code === "function") || (code === "continue") || (code === "debugger");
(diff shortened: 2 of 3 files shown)


anu-007:
@ev1stensberg sorry, I accidentally commit `yarn.lock`. By using this method i.e. generating a regex every time a related test run takes a huge amount of time (see Travis [logs](https://travis-ci.org/webpack/webpack-cli/jobs/356233792)) that is why builds are failing. can you guide me what should I do here, to make build pass successfully? Thanks !!

package.json
```diff
@@ -93,11 +93,13 @@
     "listr": "^0.13.0",
     "loader-utils": "^1.1.0",
     "lodash": "^4.17.5",
+    "lodash.template": "^4.4.0",
```
ematipico: We already have `lodash` installed, so there's no need install `lodash.template`
You could do (just a suggestion, there are tons of way)
```js
const template = require('lodash/template');
```

lib/utils/generate-identifier-regex.js
```diff
@@ -0,0 +1,55 @@
+const regenerate = require("regenerate");
+const template = require("lodash.template");
+
+// Which Unicode version should be used?
+const version = "10.0.0";
```
ematipico: `version` => `UNICODE_VERSION`
The name `version` is too generic

lib/utils/generate-identifier-regex.js
```diff
@@ -0,0 +1,55 @@
...
+
+// Set up a shorthand function to import Unicode data.
+const get = function(what) {
+	return require(`unicode-${ version }/${ what }/code-points.js`);
+};
+
+// Get the Unicode properties needed to construct the ES6 regex.
+const ID_Start = get("Binary_Property/ID_Start");
```
ematipico: We use `camelCase` as naming convention, could rename please?

lib/utils/generate-identifier-regex.js
```diff
@@ -0,0 +1,55 @@
+const regenerate = require("regenerate");
+const template = require("lodash.template");
```
ooflorent: The following would be better:

```js
const template = require("lodash/template")

File: lib/utils/ast-utils.js [hunk 5]
```diff
@@ -149,8 +150,8 @@ function createIdentifierOrLiteral(j, val) {
 			return createExternalRegExp(j, val);
 		} else {
 			// Use identifier instead
-			// TODO: Check if literalVal is a valid Identifier!
-			return j.identifier(literalVal);
+			if (!literalVal.match(generateIdentifierRegex()))
```
ooflorent: The regex will be generated each time this branch is visited. Maybe we could generate it upfront?

lib/utils/generate-identifier-regex.js
```diff
@@ -0,0 +1,55 @@
...
+
+// Get the Unicode properties needed to construct the ES6 regex.
+const ID_Start = get("Binary_Property/ID_Start");
+const ID_Continue = get("Binary_Property/ID_Continue");
+const Other_ID_Start = get("Binary_Property/Other_ID_Start");
+
+const compileRegex = template("/^(?!(?:<%= reservedWords %>)$)" +
+	"(?:<%= identifierStart %>)(?:<%= identifierPart %>)*$/");
+
+module.exports = function generateES6Regex() {
```
ooflorent: Could you rename this to `generateIdentifierRegex` please?

lib/utils/ast-utils.js
anu-007: @ooflorent the generated regex is 14971 characters long and it is not a good practice to use such a huge regex directly also esprima, Acorn etc. validates identifier this way hence I thought it will be right to do not use it directly as it will be difficult to debug and modify codinghorror [blog](https://blog.codinghorror.com/regex-use-vs-regex-abuse/) mentioned it. Apart from this, I have replaced `match` with `test` to increase the [performance](https://jsperf.com/regexp-test-vs-match-m5) but still, it takes too much time and the build fails.Thanks !!

anu-007:
@ev1stensberg I modified the script and drop the use of regexp which cause the slowing down of validation (due to it's large size) I have tried to optimize it but still it takes too much time. The new implementation is based on the way [esprima](https://github.com/jquery/esprima/blob/c833018c78cf49084e1ec217828ad879c2ed1568/src/scanner.ts#L343), [acorn](https://github.com/acornjs/acorn/blob/master/src/identifier.js), [babel](https://github.com/babel/babel/blob/add96d626d98133e26f62ec4c2aeee655bed069a/packages/babylon/src/util/identifier.js#L35) do it.

ev1stensberg:
@ematipico PTAL

anu-007:
@ev1stensberg before merging this PR let me add some unit tests as @ematipico suggested. I will add them ASAP. Thanks !!

anu-007:
@ev1stensberg @ematipico I have added few unit tests, please let me know if more tests are required.Thanks !!

ev1stensberg:
@ematipico PTAL

PR #362, opened 2018-03-22
chore(upgrade): webpack 4.2 and other dependencies

dhruvdutt:


package.json
```diff
@@ -116,20 +116,20 @@
...
+    "jest": "^22.4.3",
+    "jest-cli": "^22.4.3",
     "jsdoc": "^3.5.5",
     "lint-staged": "^7.0.0",
-    "nyc": "^11.4.1",
+    "nyc": "^11.6.0",
     "prettier-eslint-cli": "^4.7.1",
     "schema-utils": "^0.4.5",
-    "webpack": "^4.1.1",
+    "webpack": "^4.2.0",
```
ematipico: This is something that bugs me. Shouldn't we support `^4.0.0` and that's it? If not, why?

montogeek: Why does it bugs you?

ematipico: The `cli` should have a loose dependency from `webpack` as far as I understood. The `cli` uses only some utilities from `webpack`. So I still have some unclear thoughts about how this relation is inside the `cli`. Just personal doubts :)

dhruvdutt: About loose dependency, I think it depends on the peer dependency which is currently `"^4.0.0"`.
https://github.com/webpack/webpack-cli/blob/2222f1d81dbf69ddbec865eb161531e48369410f/package.json#L109-L111

IMO we should keep ***our*** project's dependencies up to date to update the lock file since all npm installation would pick from the versions specified in lock file and not the latest version based on `^` (unless you remove the lock file explicitly before npm installations).

The loose dependency should still carry as it's a peer.

dhruvdutt:
Upgraded `yargs` as well.

package.json
ematipico: Thanks!

PR #363, opened 2018-03-23
cli(init): use extractMiniCSSPlugin

### LINKED ISSUE #374: init command should create config with mini-css-extract-plugin
Feature

The init command creates config with `extract-text-webpack-plugin` to extract CSS into separate files when used in production. However, it should use `mini-css-extract-plugin` since ETWP no longer supports `webpack@4`.

Use `mini-css-extract-plugin` instead of  `extract-text-webpack-plugin` in generated config file.

Supporting the new `MiniCssExtractPlugin`.

jevan0307:

Feature

No

No

**Summary**
The init command always asks "If you want to bundle your CSS files, what will you name the bundle? (press enter to skip)". However, the answer is not used if users answer "no" to the previous question "Are you going to use this in production?".      
This change skips the CSS bundle name question when the config is not used in production.

Also, fix #374 

No

### DIFF
--- a/lib/generators/init-generator.js
+++ b/lib/generators/init-generator.js
@@ -149,9 +149,7 @@ module.exports = class InitGenerator extends Generator {
 				]);
 			})
 			.then(stylingTypeAnswer => {
-				if (!this.isProd) {
-					ExtractUseProps = [];
-				}
+				ExtractUseProps = [];
 				switch (stylingTypeAnswer["stylingType"]) {
 					case "SASS":
 						this.dependencies.push(
@@ -162,20 +160,20 @@ module.exports = class InitGenerator extends Generator {
 						);
 						regExpForStyles = `${new RegExp(/\.(scss|css)$/)}`;
 						if (this.isProd) {
-							ExtractUseProps = `
-								use: [{
-									loader: "css-loader",
+							ExtractUseProps.push(
+								{
+									loader: "'css-loader'",
 									options: {
 										sourceMap: true
 									}
-								}, {
-									loader: "sass-loader",
+								},
+								{
+									loader: "'sass-loader'",
 									options: {
 										sourceMap: true
 									}
-								}],
-								fallback: "style-loader"
-							`;
+								}
+							);
 						} else {
 							ExtractUseProps.push(
 								{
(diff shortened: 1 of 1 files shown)


jevan0307:
@ev1stensberg do you mean `mini-css-extract-plugin`?    
The plugin itself is not responsible for the question init asked, should I keep the changes and migrate to that plugin?

lib/generators/init-generator.js
```diff
@@ -199,20 +197,20 @@ module.exports = class InitGenerator extends Generator {
...
-								}, {
-									loader: "less-loader",
+								},
+								{
+									loader: "'less-loader'",
 									options: {
 										sourceMap: true
 									}
-								}],
-								fallback: "style-loader"
```
ematipico: The fallback into style-loader is not supported anymore

lib/generators/init-generator.js
```diff
@@ -246,28 +244,26 @@ module.exports = class InitGenerator extends Generator {
 						);
 						regExpForStyles = `${new RegExp(/\.css$/)}`;
 						if (this.isProd) {
-							ExtractUseProps = `
-								use: [{
-									loader: "style-loader"
-								},{
-									loader: "css-loader",
+							ExtractUseProps.push(
```
ematipico: Why this part is not a string anymore?

ematipico:
The change of the css was outside of the scope of the issue. Although, since the work is done, could you create an issue for visibility reason please?

lib/generators/init-generator.js
jevan0307: With `extract-text-webpack-plugin` before, this part was used as an argument in `ExtractTextPlugin.extract({ ${ExtractUseProps} })` with `fallback`. Since it could not be represented as a js object, we used string to do it. The usage of `mini-css-extract-plugin` is changed, so we don't need to use string anymore.

lib/generators/init-generator.js
```diff
@@ -326,42 +321,50 @@ module.exports = class InitGenerator extends Generator {
...
+								`new MiniCssExtractPlugin({ filename:'${cssBundleName}.[chunkhash].css' })`
 							);
 						} else {
 							this.configuration.config.webpackOptions.plugins.push(
-								"new ExtractTextPlugin('style.css')"
+								"new MiniCssExtractPlugin({ filename:'style.css' })"
 							);
 						}
 
+						ExtractUseProps.unshift(
```
ev1stensberg: Why are you unshifting here?

jevan0307: MiniCssExtractPlugin should be the last loader applied on the source. unshift() is used to insert MCEP loader at the front of the loaders array.

ev1stensberg: Oki, good stuff

PR #353, opened 2018-03-18
Lerna workspace

sendilkumarn:
Followup and refined #275 

This PR tries to make this webpack-cli repo into monorepo using lerna.

PR #371, opened 2018-03-29
cli(refactor): improve folder structure

ev1stensberg:
Refactoring

Yes
N/A
**Summary**
Improves folder structure and removes boilerplate code.

No

**Other information**
<3


lib/commands/migrate.js
```diff
@@ -1,36 +1,37 @@
...
-
-module.exports = function migrate(
-	currentConfigPath,
-	outputConfigPath,
-	options
-) {
+module.exports = function migrate(...args) {
+	const filePaths = args ? args.slice(2).pop() : null;
+	const currentConfigPath = args.slice(2).length === 1 ? [] : [filePaths];
+	if (!filePaths.length) {
```
dhruvdutt: `(!filePaths || filePaths.length)` since filePath can be undefined.
^In case of string, currently it's an array.

lib/commands/migrate.js
```diff
@@ -1,36 +1,37 @@
...
+		throw new Error("Please specify a path to your webpack config");
+	}
+	const outputConfigPath = path.resolve(process.cwd(), filePaths[0]);
```
dhruvdutt: No need to create redundancy for config paths if both are same.
We can do just one variable with `configPath` instead of `inputConfigPath` and `outputConfigPath`.

dhruvdutt: Also, make sure to update this line as well.
https://github.com/webpack/webpack-cli/blob/0de8a4e9402fe01414407485abf82a1f9eb5c3a9/lib/commands/migrate.js#L45

This might have broken the command as in previous code `currentConfigPath` was a path.resolve and currently, it's an array.

lib/generators/loader-generator.js
```diff
@@ -36,7 +37,7 @@ const LoaderGenerator = webpackGenerator(
 			validate: str => str.length > 0
 		}
 	],
-	path.join(__dirname, "templates"),
+	path.resolve(__dirname, "..", "generate-loader"),
```
dhruvdutt: #369 :broken_heart:

lib/index.js
```diff
@@ -14,50 +12,10 @@ const path = require("path");
...
-		case "info": {
-			return require("./commands/info.js")();
-		}
-		default: {
-			throw new Error(`Unknown command ${command} found`);
-		}
+	if (!command) {
+		throw new Error(`Unknown command ${command} found`);
+	} else if (command == "serve") {
+		return require(`./commands/${command}`).serve();
```
dhruvdutt: Awesome! :heart:

lib/commands/migrate.js
```diff
@@ -1,36 +1,37 @@
...
-*/
-
-module.exports = function migrate(
-	currentConfigPath,
-	outputConfigPath,
-	options
-) {
+module.exports = function migrate(...args) {
+	const filePaths = args ? args.slice(2).pop() : null;
+	const currentConfigPath = args.slice(2).length === 1 ? [] : [filePaths];
```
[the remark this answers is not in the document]
ev1stensberg: we're supporting more than one args

lib/commands/migrate.js
ev1stensberg: its a string now, naming is a bit weird

lib/commands/migrate.js
ev1stensberg: see above

lib/commands/migrate.js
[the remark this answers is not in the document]
dhruvdutt: In `migrate`? What's the second argument and where is it used in the code?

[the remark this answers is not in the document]
ematipico: Better to leave the functionality open to multiple arguments :)

lib/commands/migrate.js
ematipico: Based on line 28, `filePaths` can be `null`. There's need for a second check: `filePaths.length` could throw and error

montogeek:
What about #353 ?

lib/index.js
```diff
@@ -14,50 +12,10 @@ const path = require("path");
...
-		}
-		default: {
-			throw new Error(`Unknown command ${command} found`);
-		}
+	if (!command) {
+		throw new Error(`Unknown command ${command} found`);
+	} else if (command === "serve") {
+		return require(`./commands/${command}`).serve();
 	}
+	return require(`./commands/${command}`)(...args);
```
ematipico: This would throw an error no file found. Are we OK with this? Or do we want to throw a better error message?

lib/commands/init.js
```diff
@@ -9,12 +9,16 @@ const modifyHelper = require("../utils/modify-config-helper");
...
- * @param {Object} pkg - packages included when running the init command
+ * @param {Array} args - array of arguments such as
+ * packages included when running the init command
  * @returns {Function} creator/npmPackagesExists - returns an installation of the package,
  * followed up with a yeoman instance of that if there's packages. If not, it creates a defaultGenerator
  */
 
-module.exports = function initializeInquirer(pkg) {
+module.exports = function initializeInquirer(...args) {
+	const popArgs = args ? args.slice(2).pop() : null;
```
sendilkumarn: this means, `popArgs` ->  Do you mean you want to pop the arguments right. ( more than one)
The 3rd argument will always have all the packages that you want to check or the packages are in argument 3, 4, 5?

lib/commands/migrate.js
sendilkumarn: Bingo! @ematipico Good catch

lib/generators/loader-generator.test.js
```diff
@@ -1,6 +1,6 @@
 "use strict";
 
-const makeLoaderName = require("./loader-generator").makeLoaderName;
+const { makeLoaderName } = require("./loader-generator");
```
sendilkumarn: I love these changes ❤️

lib/index.js
sendilkumarn: check for `require` and then doing this would make sense

ev1stensberg: Yep, we've defined all the commands in `webpack/bin/webpack.js`, rest will the webpack file take care of. What do you mean with check for require?

ev1stensberg:
PTAL. @montogeek lerna will stay, temporary fix

lib/commands/migrate.js
```diff
@@ -1,42 +1,81 @@
...
+						"Migration output path not specified. " +
+						"Do you want to use your existing webpack " +
+						"configuration?",
+					default: "Y"
+				}
+			])
+			.then(ans => {
+				if (!ans["confirmPath"]) {
+					console.log(chalk.red("✖ ︎Migration aborted due no output path"));
+					process.exitCode = 1;
```
dhruvdutt: I think without `return`, this might continue with migration.

lib/index.js
sendilkumarn: `if(require(...))`

lib/commands/migrate.js
```diff
@@ -1,42 +1,81 @@
...
+ * @returns {Function} Runs the migration using the 'runMigrate'
+ * function.
+ */
 
-* @returns {Promise} Runs the migration using a promise that will throw any errors during each transform
-* or output if the user decides to abort the migration
-*/
+module.exports = function migrate(...args) {
+	const filePaths = args ? args.slice(3) : null;
+	if (!filePaths.length) {
```
ematipico: Double check here
```js
if (!filePaths || !filePaths.length)
```

ev1stensberg: Gotcha

ev1stensberg: I actually tested this manually and we don't need that double check. It's going to be an empty array anyway if filepaths is empty

lib/commands/migrate.js
ev1stensberg: no

dhruvdutt: screenshot_20180401_165304

ev1stensberg: `Migration aborted due no output path`

ev1stensberg: https://www.nczonline.net/blog/2014/02/04/maintainable-node-js-javascript-avoid-process-exit/

dhruvdutt: That's true but `exitCode` won't exit the process as it doesn't call the `exit()` method implicitly. To exit the process we'll need to do:
```js
process.exitCode = 1;
process.exit(); // this will exit with exit code 1
```
or
```js
process.exit(1); // this will also exit with exit code 1
```
or add a `return;`.

Without exiting, the process will continue and do the migration even after the chalk error log.

lib/commands/migrate.js
ematipico: So remove the ternary operator or fall  back the assignment to an empty array. It might work but people that see the code might think that is wrong. Or add a comment

lib/commands/migrate.js
```diff
@@ -1,42 +1,81 @@
...
+		throw new Error("Please specify a path to your webpack config");
```
dhruvdutt: Instead of throwing an error which generates a long non-useful (in this case) stack trace, how about using chalk to show a one-line error message and exit the process gracefully? :wink:

dhruvdutt: image

lib/commands/migrate.js
ematipico: @dhruvdutt is right, we need to abort from the chain of promises. But `return` won't help. I'd suggest to throw an error and handle it inside the `catch`

lib/commands/migrate.js
ev1stensberg: Stack traces are good for us when people submit bugs. Console.errors when we're inside a promise

ev1stensberg:
PTAL

lib/commands/migrate.js
ev1stensberg: This looks ugly though, error msg is intuitive enough for us to debug

lib/commands/migrate.js
ev1stensberg: Done

lib/commands/migrate.js
ematipico: Yeah but in here we are intentionally showing an error to the user, not an error due to a bug in our code. I think we should separate the two concerns

dhruvdutt: I'm convinced both ways. :smile_cat:

lib/commands/migrate.js
```diff
@@ -1,42 +1,83 @@
...
 const chalk = require("chalk");
 const diff = require("diff");
 const inquirer = require("inquirer");
 const PLazy = require("p-lazy");
 const Listr = require("listr");
 
-const validate = require("webpack").validate;
-const WebpackOptionsValidationError = require("webpack")
-	.WebpackOptionsValidationError;
+const { validate } = require("webpack");
```
sendilkumarn: nit: can we merge this together?

lib/index.js
nveenjain: Can't we put this `require` request in `try catch` block? and in catch block `throw new Error` stating that command not found, otherwise the error thrown will be that module was not found.

PR #372, opened 2018-03-29
WIP: refactor: standardize command files

PlayMa256:
Refactor command files and how they are invoked.

Refactor

**Summary**
I recently have seen polymer CLI and how they structured their cli and also saw @ev1stensberg tweet saying that this would be a good refactor to have, so i gave it a shot.

The problem is i just saw the pr list and found #371 which basically does a little bit of this. 

I would like to know about you guys, if this is desirable. From my POV, this looks way clear than what it was before.

It's just more of a "what you guys think? feedback please!!" then a proper PR.

I appreciate any feedback. If it is not desirable, i delete this without any problems.

Again: **WIP** and it still has things to `refactor/rename/move` to make everything more organized.

No.

### DIFF
--- a/bin/webpack.js
+++ b/bin/webpack.js
@@ -44,7 +44,9 @@
 
 	if (NON_COMPILATION_CMD) {
 		// eslint-disable-next-line
-		require("../lib/index")(NON_COMPILATION_CMD, process.argv);
+		const WebpackCLI = require("../lib/index")
+		const instance = new WebpackCLI();
+		instance.run(NON_COMPILATION_CMD, process.argv);
 		return;
 	}
 
--- a/lib/commands/add.js
+++ b/lib/commands/add.js
@@ -11,6 +11,12 @@ const modifyHelper = require("../utils/modify-config-helper");
  *
  */
 
-module.exports = function add() {
-	return modifyHelper("add", defaultGenerator);
+module.exports = class AddCommand {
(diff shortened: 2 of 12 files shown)


ev1stensberg:
I don't think we need class instances to run explicit commands. With polymer it makes sense, cause the entire suite is based on yeoman, our isn't.

PlayMa256:
Alright, thanks!
Deleting it!

ev1stensberg:
Thanks for the PR though! Can I point you to another issue?

PlayMa256:
Absolutly!

ev1stensberg:
Could you have a go at this? https://github.com/webpack/webpack-cli/issues/152

PlayMa256:
Sure! Never worked with that, but its worth learning :smile:

I'm going to have a look at that later, i will post there if i have any doubts! 

Btw, did you found out if there is a better option other then spawning another node process, to add that new feature? From further researches, that is basically the way to do. Mocha already does that too. I posted a link to their cli. Let me know what you think and i will proceed the development there!

ev1stensberg:
No I haven't. Can you link to the file mocha does this in?

PlayMa256:
Here you have it!
https://github.com/mochajs/mocha/blob/master/bin/mocha

ev1stensberg:
You can try to do the same

Issue #289, opened 2018-02-26
[Feature]: Increase memory with a flag / Support Node Flags


damianobarbati:
Could `webpack-cli` handle the memory limit increase with a flag?

Right now we have to manually use:
```node --max_old_space_size=4096 node_modules/webpack/bin/webpack.js```

A simple flag would be very nice instead:
```webpack --increase-memory 4096```

Or multiple custom node options like:
```webpack --node.increase-memory 4096 --node.trace-deprecation --node.experimental-modules```

PlayMa256:
I'm down to grab this, but i'm not really sure where do i add the flags.
Is this similar to [this](https://github.com/webpack/webpack-cli/blob/master/bin/convert-argv.js#L29)?

montogeek:
@PlayMa256 No, it is part of the process args, not process environment variables

montogeek:
You could do something like `argv["max_old_space_size"] = 4096` somewhere.
I believe this shouldn't be part of webpack, it should be handled by the user.

PlayMa256:
I expressed myself in a wrong way, sorry for that.
I understand it's not a process args, but i meant to do this the same way that the production mode is handled.
Anyways. You answered my question!
Thank you @montogeek

PlayMa256:
I was having a look how to implement those flags, and i've reached a fork road. I'm not able to override node flag options when a process already started, so i was thinking... spawning a child process passing down v8 flags would be a cumbersome idea or is there a way (i actually dont know any...) to update v8 flags when a process already started?

What do you all think?

ev1stensberg:
Getting back to you on this. This might not be possible without invoking `node ./path/to/webpack`

PlayMa256:
is there a way to run a single test only? A single binary test?

ematipico:
Not at the moment :(

ev1stensberg:
This impl can be used with #15 . We keep track of pId and restart a process once it's done ( I think )

damianobarbati:
Can npx used to accomplish the result? i.e: `npx -n=' --max_old_space_size=4096'`
Not sure how to inspect max memory limit for current process in js.

evenstensberg:
Yes if npm supports injection of args. We cannot use this as a fix to us, as the flags needs to be supported by default not by running npx

PR #377, opened 2018-04-07
cli(node): Add node flags to CLI

PlayMa256:
Feature

Test added, i'm not really sure if that it is enough. If it is not, let me know. Since node has quite a few flags, i don't think testing each of them is a good thing.

To be updated. Just double checking: this update has to go in the webpack docs, or internal docs?

**Summary**
This is a feature implementation based on #289. 
This pr introduces a way to pass down node flags to execute the cli. To be able to differenciate these flags, a prefix is added to them. E.g. `--node.max_old_value=2048`. At the end, this allows us to accept every single node flag, without having to compare every single one.

no


ematipico:
This is awesome! There's a problem with your commit message, that's the reason of the broken build. Try to amend your commit and remove the "." (full stop)

PlayMa256:
Ok! I was not sure why it was breaking though. I'm fixing it.

Enviado a partir do VMware Boxer

Em 07/04/2018 14:26, Emanuele <notifications@github.com> escreveu:

This is awesome! There's a problem with your commit message, that's the reason of the broken build. Try to amend your commit and remove the "." (full stop)

ematipico:
Is your branch rebased from `master`? I don't get why appveyor build is failing

PlayMa256:
Me neither... let me try once again rebasing this, i had already. Just a second.

PlayMa256:
@ematipico now it should be updated. By some reason when i ran `git rebase master` it was not finding differences... anyways. Let's just wait for the cis to run. I'm going to keep my eyes on this.

PlayMa256:
@ematipico it is going to fail again though! As you can [see](http://prntscr.com/j2wyq9) it has been rebased with master.

ev1stensberg:
Regarding the rationale around accessing node flags in webpack. This is good for us, cause people use webpack inside Docker, CI's etc. It's pretty nice when you can tap directly into v8 or nodeJS for profiling or perf optimizations. I'm in favor for this.

ev1stensberg:
@bmeurer thoughts?

montogeek:
@ev1stensberg If the can pass flags to `webpack` bin, can as well pass them to `node`.

ev1stensberg:
@montogeek not necessarily. The process won't pick the flags up

montogeek:
@ev1stensberg How so?

PlayMa256:
What @montogeek is trying to say is: if someone can call `npm run build` || `npm run something` they can call: `node --max_old_memory=1024....` as well.

boneskull:
See https://github.com/nodejs/node/pull/19335

bmeurer:
I don't have a strong opinion here, but I tend to agree with @PlayMa256.

boneskull:
Mocha does this same thing.  Why?  The ability to pass flags directly to the executable is a better user experience.

ematipico:
I really like this feature and as @boneskull  pointed out, the user experience improves

montogeek:
@tj What you think about this change? You did it in mocha

PlayMa256:
After further analysis on appveyor, i found it is [running out of memory](https://ci.appveyor.com/project/ev1stensberg/webpack-cli/build/590#L599). I just could not see where it is breaking, since it barely started running tests.

ematipico:
@PlayMa256, you could try an experiment. Run tests with current branch and see the amount of memory used. Run tests from master and see the amount memory used. If they differ too much, the problem you be into the code of this PR.

PlayMa256:
@ematipico i've ran the tests and compared the heap consumption for each of them, i've compared master and this branch, both are almost the same (considering the marging of error...)

I removed the tests i added. Let me see how ci behaves.

PlayMa256:
After further investigation, i could not find from where this possible "memory leak" is coming from. I gathered the amount of memory being used by each test, from master and from this branch, both tests are the same. Locally the tests run perfect, without any problem. 

At first i suspected that every single command you ran was going though [this](https://github.com/webpack/webpack-cli/pull/377/files#diff-877a99a662829398e8136ee938cf1c93R37) if statement, but endedup up not being true. Could someone give me a hand on this? I ran out of ideas on how to debug this properly.

bin/process-node-flags.js
```diff
@@ -0,0 +1,36 @@
+const spawn = require("cross-spawn");
+const path = require("path");
+
+module.exports = function(argv) {
```
ematipico: Could you add some jsdoc comments here?

bin/webpack.js
```diff
@@ -32,6 +32,13 @@
 		"info"
 	];
 
+	const nodeFlags = process.argv.filter((arg) => arg.includes("node."));
+
+	if (nodeFlags.length) {
+		require("./process-node-flags")(process.argv);
+		return;
```
ematipico: Why there's a `return` here?

PlayMa256: Same reason as used [here](https://github.com/webpack/webpack-cli/blob/master/bin/webpack.js#L46). If we don't exit, yargs will warn saying these flags are unknown and the normal cli flow will follow, with this return, the other file can strip those flags, spawn another webpack-cli process and work as intended.

PlayMa256: Tests added. Let me know if they aren't enough. I've tried to add as many flags as possible.

ev1stensberg: Okay great, thanks!

ematipico:
@ev1stensberg do you have any idea why appveyour goes out of memory during the tests?

Issue #389, opened 2018-04-09
Update transformation: uglifyJsPlugin


bitpshr:
**Do you want to request a *feature* or report a *bug*?**
Feature

**What is the current behavior?**
The current transformation supports migrating a webpack configuration from `v1` to `v2`. 

**What is the expected behavior?**
The transformation should support migrating a webpack configuration from `v3` to `v4`.

**If this is a feature request, what is motivation or use case for changing the behavior?**
After [some discussion](https://github.com/webpack/webpack-cli/issues/373) the decision was made to update our migration transformations to actively support migrating only from the previous version to the current version, which currently means `v3 - v4`. Migrating older versions such as `v1` or `v2` will no longer be supported.

This feature request is to update the `uglifyJsPlugin` transformation accordingly.

PlayMa256:
i'm working on this!

ematipico:
Thanks!

PlayMa256:
Quick question though, i could not find in the docs.

If i want to customize how the minification should work (add parameters, etc), i should use `optimization.minimizer: [new Uglify({...opts}]`?

montogeek:
@PlayMa256 Yes

PlayMa256:
@montogeek thanks, good to know where i can find things! now i can go straight there.

Issue #388, opened 2018-04-09
Update transformation: resolve


bitpshr:
**Do you want to request a *feature* or report a *bug*?**
Feature

**What is the current behavior?**
The current transformation supports migrating a webpack configuration from `v1` to `v2`. 

**What is the expected behavior?**
The transformation should support migrating a webpack configuration from `v3` to `v4`.

**If this is a feature request, what is motivation or use case for changing the behavior?**
After [some discussion](https://github.com/webpack/webpack-cli/issues/373) the decision was made to update our migration transformations to actively support migrating only from the previous version to the current version, which currently means `v3 - v4`. Migrating older versions such as `v1` or `v2` will no longer be supported.

This feature request is to update the `resolve` transformation accordingly.

PlayMa256:
This is for v2->v3, which is also present in [here](https://webpack.js.org/migrate/3/#resolve-root-resolve-fallback-resolve-modulesdirectories). Issue can be closed. Also, folder can be deleted (?)

bitpshr:
PR submitted for the removal of this transformation.

Issue #386, opened 2018-04-09
Update transformation: removeDeprecatedPlugins


bitpshr:
**Do you want to request a *feature* or report a *bug*?**
Feature

**What is the current behavior?**
The current transformation supports migrating a webpack configuration from `v1` to `v2`. 

**What is the expected behavior?**
The transformation should support migrating a webpack configuration from `v3` to `v4`.

**If this is a feature request, what is motivation or use case for changing the behavior?**
After [some discussion](https://github.com/webpack/webpack-cli/issues/373) the decision was made to update our migration transformations to actively support migrating only from the previous version to the current version, which currently means `v3 - v4`. Migrating older versions such as `v1` or `v2` will no longer be supported.

This feature request is to update the `removeDeprecatedPlugins` transformation accordingly.

PlayMa256:
Plugins to be removed: 
- [ ] webpack.optimize.CommonsChunkPlugin
- [ ] webpack.optimize.UglifyJsPlugin (which is done by #401 )

ematipico:
* `CommonsChunkPlugin` is not deprecated but it's been changed in favor of `splitChunks`, which is covered already in another PR
* not deprecated but changed

I don't think we have plugins have been totally removed.

PlayMa256:
Yep, if forgot we have a PR open for split chunks too. I just updated my comment. We can close this i guess

Issue #382, opened 2018-04-09
Update transformation: bannerPlugin


bitpshr:
**Do you want to request a *feature* or report a *bug*?**
Feature

**What is the current behavior?**
The current transformation supports migrating a webpack configuration from `v1` to `v2`. 

**What is the expected behavior?**
The transformation should support migrating a webpack configuration from `v3` to `v4`.

**If this is a feature request, what is motivation or use case for changing the behavior?**
After [some discussion](https://github.com/webpack/webpack-cli/issues/373) the decision was made to update our migration transformations to actively support migrating only from the previous version to the current version, which currently means `v3 - v4`. Migrating older versions such as `v1` or `v2` will no longer be supported.

This feature request is to update the `bannerPlugin` transformation accordingly.

ematipico:
Is this changed from webpack 3 to 4? I don't think so

PlayMa256:
@ematipico Nop, it is not from v3 to v4. 
[This](https://github.com/webpack/webpack/blob/fde018300aa52262c384e937c408d5dd97d62951/test/configCases/plugins/banner-plugin/webpack.config.js#L15) example was updated a long time ago. Seems that this is from v1->v2.

PR #390, opened 2018-04-09
Update extractTextPlugin transformation

### LINKED ISSUE #373: Migrate and init should move to mini-css-extract-plugin
Feature

At the moment, when migrating or initiating a new configuration `ExtractTextPlugin` it still  uses `ExtractTextPlugin`. `ExtractTextPlugin` should become `MiniCssExtractPlugin`

The `cli` should ask it the user wants to migrate to v3 or v4 of webpack. If v4, it should migrate to `MiniCssExtractPlugin`. The init feature should use `MiniCssExtractPlugin` as well.

Supporting the new `MiniCssExtractPlugin`

bitpshr:
Feature

Yes

N/A

**Summary**
After [some discussion](https://github.com/webpack/webpack-cli/issues/373) the decision was made to update our migration transformations to actively support migrating only from the previous version to the current version, which currently means `v3 - v4`. Migrating older versions such as `v1` or `v2` will no longer be supported. This PR begins this work by updating the `extractTextPlugin` transformation. Resolves #373.

_Note:_ `v3` configurations that contain multiple `ExtractTextPlugin` instances are not yet supported due to a [limitation](https://github.com/webpack-contrib/mini-css-extract-plugin/issues/45) with `MiniCssExtractPlugin`.

Yes. This PR changes `ExtractTextPlugin` migration behavior and produces a `v4`-specific configuration.

**Other information**
This PR only includes updates to the `extractTextPlugin` transformation. The following issues have been opened for the remaining transformations:

- #382
- #383
- #384
- #385
- #386
- #387
- #388
- #389


lib/migrate/extractTextPlugin/__testfixtures__/extractTextPlugin.input.js
```diff
@@ -1,16 +1,26 @@
...
 
 module.export = {
     module: {
         rules: [
             {
                 test: /\.css$/,
-                use: ExtractTextPlugin.extract('style-loader', 'css-loader')
+                use: ExtractTextPlugin.extract({
+                    use: 'css-loader',
+                    fallback: ['style-loader', { loader: 'test-object-loader' }],
```
bitpshr: I tried to fully demonstrate that string, object, and array loader configurations all work via this test case.

lib/migrate/extractTextPlugin/extractTextPlugin.js
```diff
@@ -1,58 +1,145 @@
...
+		prodLoaders.push(loader, ...switchedLoader);
+		currentFallbackLoader && pathSwitch(currentFallbackLoader, switchedFallbackLoader => {
+			devLoaders.push(...switchedFallbackLoader, ...switchedLoader);
+		});
+	});
+
+	// Add new loader
+	loaderPath && Object.assign(loaderPath, {
+		value: currentFallbackLoader ?
+			j.conditionalExpression(
```
bitpshr: `MiniCssExtractPlugin` does not support `fallback` and `disabled` options. Instead, users should manually swap out `MiniCssExtractPlugin.loader` with the loader that was normally used as a `fallback`, which is what we're doing with this ternary. More information can be found [here](https://github.com/webpack-contrib/mini-css-extract-plugin/issues/88#issuecomment-3786856530).

PlayMa256:
I found some problems. I had the current configuration (that was not created by me and i think it is a good "point out of the curve", since people can still have things like this: https://gist.github.com/PlayMa256/0a51b41e1b0eba910869764f378bf932

This was the generated config: https://gist.github.com/PlayMa256/587b2d3890c9c96e3f3dbe567c8ecfce

* It has 2 empty variables, which were kept in the plugins list.
* `MiniCssPlugin` has null parameters. I think it is better to remove them other then having them null.
* Is there a way to remove the `extractCSS.extract` statements and replace with `MiniCssExtractPlugin.loader`?

bitpshr:
Good catch and feedback. On it now.

PlayMa256:
Well, the null problems seems to be gone. In the example i sent you (which is kind absurd i admit), it still [have](https://gist.github.com/PlayMa256/db9dc93e2ae5d376e41e25cca5b108af) (`extractSCSS.extract...`). We can't control that much, neither handle the problem in this case, where there are `fallbacks` and such.

Could you create a test for multiple loaders in a rule?
I mean:
```
[
          'style-loader',
          'css-loader',
          ExtractTextPlugin...
]
```
A scenario similar to this? That could show problems if it fails.

bitpshr:
@PlayMa256 I updated the logic so options are only passed if they're defined and `null` is never passed. I also updated the tests to verify cases when `ExtractTextPlugin.extract` is an argument to a function call (like your example, `['css-hot-loader'].concat(ExtractTextPlugin.extract(...))`, and cases when `ExtractTextPlugin.extract` is a member of an array (like your multiple loaders scenario above.) All are working correctly.

However, after digging into multiple-instance support a bit more, I think we may need to only support migrating configurations with a single instance of `ExtractTextPlugin` until true multiple-instance support lands in `MiniCssExtractPlugin`. See webpack-contrib/mini-css-extract-plugin#45. Until that happens, it's not possible to translate per-instance options such as `filename`. We could technically translate `extractCss.extract`-type call statements to `ExtractTextPlugin.extract` statements, but it wouldn't accomplish much since `MiniCssExtractPlugin` only supports one definition in the `plugins` array for now.

PlayMa256:
Oh, got it! I did not know about this problem, i just found this tricky `webpack` config and it could be a nice way to debug this feature :). No problem not supporting this, we could insert this disclaimer in the docs. I'm about to go home, as i arrive and have time to test, i can give you more feedback if necessary! Good job @billyjanitsch !!

ev1stensberg:
@ematipico this is a nice PR for you to test & to get more knowledge of the `migrate` internals. Easy testable as well.

bitpshr:
Thanks, I moved most functionality to `ast-utils` and made it generic so other transformations can use them going forward. This should be ready for another review when you have time @ev1stensberg @ematipico.

lib/migrate/extractTextPlugin/extractTextPlugin.js
```diff
@@ -1,58 +1,111 @@
...
- * Transform for ExtractTextPlugin arguments. Consolidates arguments into single options object.
+ * Transform for ExtractTextPlugin. Replaces it with MiniCssExtractPlugin and translates
+ * configuration where possible.
  *
  * @param {Object} j - jscodeshift top-level import
  * @param {Node} ast - jscodeshift ast to transform
  * @returns {Node} ast - jscodeshift ast
  */
-
 module.exports = function(j, ast) {
```
ev1stensberg: named functions

lib/utils/ast-utils.js
```diff
@@ -651,6 +651,155 @@ function createFunctionWithArguments(j, p, name) {
...
+
+/**
+ *
+ * Finds and removes a node for a given plugin name. If the plugin
+ * is the last in the plugins array, the array is also removed.
+ *
+ * @param {any} j — jscodeshift API
+ * @param {Node} node - Node to start search from
+ * @param {String} pluginName - name of the plugin to remove
+ * @param {Node | Void} - path to the root webpack configuration object if plugin is found
```
ev1stensberg: Missing return

lib/utils/ast-utils.js
```diff
@@ -651,6 +651,155 @@ function createFunctionWithArguments(j, p, name) {
...
+		});
+	return found;
+}
+
+/**
+ *
+ * Calls an function with a path wrapped in an array, using path.elements if available
+ *
+ * @param {NodePath} path - path to wrap in array
+ * @param {Function} action - called with array path
```
ev1stensberg: Missing return

lib/migrate/extractTextPlugin/extractTextPlugin.js
```diff
@@ -1,58 +1,111 @@
...
-
-		const literalArgs = args.filter(p => utils.isType(p, "Literal"));
-		if (literalArgs && literalArgs.length > 1) {
-			const newArgs = j.objectExpression(
-				literalArgs.map((p, index) =>
-					utils.createProperty(j, index === 0 ? "fallback" : "use", p.value)
-				)
-			);
-			p.value.arguments = [newArgs];
+	const devLoaders = [];
```
ev1stensberg: Can you abstract all this into more functions that can be reused, like find value "filename" for a given prop

bitpshr: I'm not sure I follow; we already abstracted functionality into simple, single-purpose functions like `getPluginOptionValue` and `getPluginLoaderValue`. If we start getting even more abstract than this, we risk targeting nodes that aren't actually the nodes we care about. For example, someone could define `filename` completely outside of the webpack configuration object, so we need to be very careful.

The functions I already pulled out (`createPluginByName`, `removePluginByName`, `getLoaderOptionValue`, `getPluginOptionValue`, and `removeRequire`) are reusable and safe.

lib/utils/ast-utils.js
bitpshr: Added `@returns {Void}`

bitpshr:
@ev1stensberg thanks for the reviews. Everything has been addressed except for the following comment:

> Could you extract these into more functions that could be reused?

What type of functions were you imagining? The functions I already added (`createPluginByName`, `removePluginByName`, `getLoaderOptionValue`, `getPluginOptionValue`, and `removeRequire`) are reusable and safe, and as abstract as any other function in `ast-utils`.

bitpshr:
Thanks @ematipico. Could you please see my [comment](https://github.com/webpack/webpack-cli/pull/390#issuecomment-381354455) above and take another look when you have a chance @ev1stensberg? Thanks in advance!

lib/migrate/extractTextPlugin/extractTextPlugin.js
```diff
@@ -1,58 +1,111 @@
...
+	requireExists && addRequire(j, ast, "MiniCssExtractPlugin", "mini-css-extract-plugin");
+
+	// Remove old plugin
+	pluginExists = removePluginByName(j, ast, "ExtractTextPlugin");
+
+	// Add new plugin
+	pluginExists && createPluginByName(j, ast, "MiniCssExtractPlugin", loaderOptions);
+
+	// Remove old loader
+	ast.find(j.Identifier, { name: "ExtractTextPlugin" })
```
ev1stensberg: This could be abstracted into an util, maybe you can make it general so that we can reuse it for later?

lib/migrate/extractTextPlugin/extractTextPlugin.js
```diff
@@ -1,58 +1,111 @@
...
+	currentLoader && pathToArray(currentLoader, switchedLoader => {
+		const loader = j.memberExpression(j.identifier("MiniCssExtractPlugin"), j.identifier("loader"), false);
+		prodLoaders.push(loader, ...switchedLoader);
+		currentFallbackLoader && pathToArray(currentFallbackLoader, switchedFallbackLoader => {
+			devLoaders.push(...switchedFallbackLoader, ...switchedLoader);
+		});
+	});
+
+	// Add new loader
+	const loader = currentFallbackLoader ?
```
ev1stensberg: This might be another util, `addLoader`

lib/migrate/extractTextPlugin/extractTextPlugin.js
```diff
@@ -1,58 +1,111 @@
...
+		j.conditionalExpression(
```
ev1stensberg: It might be too specific, is there another way of adding a loader?

bitpshr: I created an `addConditionalLoader` utility function per your comment above. I think it'll be useful in the long run, as it makes it easy to add loaders based on `NODE_ENV`.

bitpshr:
Thanks @ev1stensberg. This should be as generic as possible now. Let me know your thoughts.

bitpshr:
Thanks @ev1stensberg. Rebased.

Issue #391, opened 2018-04-09
[Feature] Read the types of mode available from webpack


ematipico:
**Do you want to request a *feature* or report a *bug*?**
Feature

**What is the current behavior?**
At the moment the available type of `--mode` are hardcoded

**What is the desired behavior?**
The `cli` should retrieve/read the available type of `--mode` from `webpack`

**If this is a feature request, what is motivation or use case for changing the behavior?**
To make the `cli` less hardcoded and flexible

PlayMa256:
I can do that. They are found at the schemas file, right? I found them there, just want to make sure it is the right place.

PlayMa256:
I'm just restesting another pr, i will do that!

PR #392, opened 2018-04-09
Retrieve information for CLI option from webpack schema options file

PlayMa256:
Feature/Refactor

Not needed (?)

**Summary**
As requested in #391 , reading choices for mode directly from `webpack` options schema.

No.

### DIFF
--- a/bin/config-yargs.js
+++ b/bin/config-yargs.js
@@ -7,6 +7,11 @@ const RESOLVE_GROUP = "Resolving options:";
 const OPTIMIZE_GROUP = "Optimizing options:";
 const INIT_GROUP = "Initialization:";
 
+const availableModes = () => {
+	const webpackOptionsSchema = require("webpack/schemas/WebpackOptions.json");
+	return webpackOptionsSchema.properties.mode.enum;
+};
+
 module.exports = function(yargs) {
 	yargs
 		.help("help")
@@ -88,7 +93,7 @@ module.exports = function(yargs) {
 			},
 			mode: {
 				type: "string",
-				choices: ["development", "production", "none"],
+				choices: availableModes(),
 				describe: "Mode to use",
 				group: CONFIG_GROUP,
 				requiresArg: true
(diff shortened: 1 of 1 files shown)


bin/config-yargs.js
```diff
@@ -7,6 +7,11 @@ const RESOLVE_GROUP = "Resolving options:";
 const OPTIMIZE_GROUP = "Optimizing options:";
 const INIT_GROUP = "Initialization:";
 
+const availableModes = () => {
+	const webpackOptionsSchema = require("webpack/schemas/WebpackOptions.json");
```
dhruvdutt: Can we keep this statement on top of the file rather than doing inline require?

montogeek:
That is not going to work, that file exports an `ajv` JSON Schema https://www.npmjs.com/package/ajv, not just an array of options.

PlayMa256:
@montogeek are you sure? i tried to give cli any other mode, it complained and answered correctly. You can see [here](http://prntscr.com/j30iwf)

bin/config-yargs.js
bitpshr: If we move the `require` to the top of the file, perhaps with a less-redundant name like `optionsSchema`, why not just reference `optionsSchema.properties.mode.enum` directly in the `mode` object below? It would save the unnecessary indirection of creating a function that's only called in one place:

```js
const optionsSchema = require("webpack/schemas/WebpackOptions.json");

// ...

mode: {
	choices: optionsSchema.properties.mode.enum,
	// ...
}
```

PlayMa256: @bitpshr i'm actually doing that now. I refactored all that file to get all the informations from webpack schema. I will be updating it ASAP.

ematipico:
We could add those missing properties later on

shellscape:
@ev1stensberg no thoughts here. webpack-dev-server is dead to me (all hail webpack-serve), and I'm working on a cli alternative 😄

ev1stensberg:
haha okay good stuff!

PR #403, opened 2018-04-13
fix: output bundle file name for single entry

anu-007:
bugfix

no
no
**Summary**
According to [docs](https://webpack.js.org/configuration/output/#output-filename) for a single entry point, output bundle file should use a static name but on generating `config` file using `init` it uses a file named `[name].bundle.js` it should simply be called `bundle.js`.generated `config` [link](https://gist.github.com/anu-007/8c0384ff66f1efc2e25a99d701420e22#file-webpack-config-js-L18)

no

### DIFF
--- a/lib/generators/init-generator.js
+++ b/lib/generators/init-generator.js
@@ -102,7 +102,7 @@ module.exports = class InitGenerator extends Generator {
 					};
 				} else {
 					this.configuration.config.webpackOptions.output = {
-						filename: "'[name].bundle.js'"
+						filename: "'[name].[chunkhash].js'"
 					};
 				}
 				if (outputTypeAnswer["outputType"].length) {


PlayMa256:
@ev1stensberg should we generate a "generic" config, or a more optimized one? I mean, we could use `[name].[hash].js` for cache improvements?

bitpshr:
I don't think this is a bug. The documentation states that the output path _can_ be a static name for a single entry point, but it doesn't have to be. Our current `[name].bundle.js` output path is totally valid, but to @PlayMa256's point, a more-generic cache-busting path like `[name].[hash].js` would be an improvement worth considering.

anu-007:
@PlayMa256 @bitpshr thanks for the review and I completely agree with you we can use more generic output bundle name. I am thinking of using something like `bundle.[chunkhash].js` please let me know if it's ok. thanks !!

bitpshr:
I prefer `[name].[hash].js` Like @PlayMa256 suggested but it’s really just preference. I like that the bundle name is both cache-busting and derived from the original entry filenams. `bundle.[hash].js` would also be fine though.

anu-007:
@bitpshr updated.

PR #405, opened 2018-04-15
Remove v1 -> v2-specific transformations

### LINKED ISSUE #383: Update transformation: loaderOptionsPlugin
Feature

The current transformation supports migrating a webpack configuration from `v1` to `v2`. 

The transformation should support migrating a webpack configuration from `v3` to `v4`.

After [some discussion](https://github.com/webpack/webpack-cli/issues/373) the decision was made to update our migration transformations to actively support migrating only from the previous version to the current version, which currently means `v3 - v4`. Migrating older versions such as `v1` or `v2` will no longer be supported.

This feature request is to update the `loaderOptionsPlugin` transformation accordingly.

### LINKED ISSUE #384: Update transformation: loaders
Feature

The current transformation supports migrating a webpack configuration from `v1` to `v2`. 

The transformation should support migrating a webpack configuration from `v3` to `v4`.

After [some discussion](https://github.com/webpack/webpack-cli/issues/373) the decision was made to update our migration transformations to actively support migrating only from the previous version to the current version, which currently means `v3 - v4`. Migrating older versions such as `v1` or `v2` will no longer be supported.

This feature request is to update the `loaders` transformation accordingly.

### LINKED ISSUE #388: Update transformation: resolve

bitpshr:
Feature

Yes

N/A

**Summary**
The `loaders`, `loaderOptionsPlugin`, and `resolve` transformations are specific to v1 -> v2 upgrades and as such are no longer needed. This PR removes these transformations and updates migrations test snapshots accordingly.

Resolves #383
Resolves #384
Resolves #388

Yes. This PR removes previously-existing transformations and aims to produce a v4-compatible configuration instead of a v2-compatible configuration.


ev1stensberg:
no. That would raise an incredible amount of tech debt in the length.

bitpshr:
Thanks for the thoughts @ev1stensberg @dhruvdutt. It sounds like we have three options:

1. Remove the code altogether. The transformation is firmly v1 -> v2 specific, and git provides historical record of the code if we ever need to reference it or use it in the future.
2. Leave the code in-place so the transformation continues to be tested and maintained, but remove it from the [transforms object](https://github.com/webpack/webpack-cli/blob/master/lib/migrate/index.js#L15-L25) since it's no longer relevant.
3. Do nothing; leave the code in place and leave the transformation active.

I personally vote for option 1. While there's technically no true harm in leaving the code in-place, it does add both the bundle size and maintenance burden for functionality that is no longer relevant or supported. Leaving the code in place also misleads anyone who is reading our source code. I feel that if our transformations are now v3 -> v4, we should be uniform across the board.

Regardless of whether we go with option 1 or option 2, I don't think 3 is a valid path: we should not be migrating v1 configurations anymore, especially only for specific transformations.

dhruvdutt:
@bitpshr I would say let's keep it till we get the migration for 3 to 4 to a stable point and then we can clean up. :)

ematipico:
I personally think that we should move on and stop maintaining code for old versions of `webpack`, especially version 1. If in the future there would be bugs, we can always go back and make a patch. Versioning system are made for these kind of situations.

bitpshr:
Thanks for the input everyone. Let me know if you'd like anything cleaned up on this.

bitpshr:
Thanks @ev1stensberg and @ematipico. Rebased.

PR #408, opened 2018-04-18
cli(mode): set mode to none if none is set after defaults

### LINKED ISSUE #407: Warning message says --mode=none is supported, but using --mode=none produces error

Bug

Running without `--mode` results in a warning that includes a message about using `--mode=none`:

```sh
$ webpack
...
WARNING in configuration
The 'mode' option has not been set, webpack will fallback to 'production' for this value. Set 'mode' option to 'development' or 'production' to enable defaults for each environment.
You can also set it to 'none' to disable any default behavior. Learn more: https://webpack.js.org/concepts/mode/
```

Yet running with `--mode=none` produces the `--help` text, followed by this message:

```sh
$ webpack --mode=none
...
Invalid values:
  Argument: mode, Given: "none", Choices: "development", "production"
```

``` sh
$ npm init --yes
$ npm i -D webpack webpack-cli
$ mkdir src
$ touch src/index.js
$ webpack
$ webpack --mode=none
```

Either:

1. Do not have a message claiming support for `--mode=none` or
2. Add support for `--mode=none`

**Please paste the results of `webpack-cli info` here, and mention other relevant information such as programming language.**

```
$ webpack-cli info

  System:
    OS: Windows 10
    CPU: x64 Intel(R) Core(TM) i7-7700K CPU @ 4.20GHz
  Binaries:
    Node: 9.11.1
    Yarn: 1.5.1
    npm: 5.8.0
  Browsers: {}
  npmPackages:
    webpack:
      wanted: ^4.6.0
      installed: 4.6.0
    webpack-cli:
      wanted: ^2.0.14
      installed: 2.0.14
```

ev1stensberg:

Bugfix

Y
Y
**Summary**
Sets mode to `none` if no mode is set after all mode options has been set.

N/A

No

**Other information**
Fixes #407

### DIFF
--- a/bin/convert-argv.js
+++ b/bin/convert-argv.js
@@ -40,7 +40,9 @@ module.exports = function(...args) {
 		argv["output-filename"] = path.basename(output);
 		argv["output-path"] = path.dirname(output);
 	}
-
+	if(!argv.mode) {
+		argv.mode = "none";
+	}
 	let configFileLoaded = false;
 	let configFiles = [];
 	const extensions = Object.keys(interpret.extensions).sort(function(a, b) {


ev1stensberg:
@rossipedia https://github.com/webpack/webpack-cli/blob/master/bin/config-yargs.js#L90-L95 also has this, new version hasn't been published yet.

bin/convert-argv.js
```diff
@@ -40,7 +40,9 @@ module.exports = function(...args) {
 		argv["output-filename"] = path.basename(output);
 		argv["output-path"] = path.dirname(output);
 	}
-
+	if(!argv.mode) {
```
PlayMa256: can't this be set as `default: 'none'` in the [options](https://github.com/webpack/webpack-cli/blob/master/bin/config-yargs.js#L90-L95)? Same effect.

ev1stensberg: The two statements above are relying on it not to be set, so that's why it's placed here

PlayMa256: The same does not happen when you set `default: none` via `yargs`? If not set, it gets the default value? (just curiosity btw)

dhruvdutt: Yup. I think @PlayMa256 is right. Config phase happens first.

ev1stensberg: Okay gotcha, will rebase the tomorrow, internet here is pretty bad

ev1stensberg: Actually I don't think we should set default as none, webpack will warn us, that's intuitive and it informs the user about the choices they're making

PR #409, opened 2018-04-18
fix(init): support non-string options for single entry

### LINKED ISSUE #397: [BUG] Single entry should support adding path, process and non-string options
`Which module will be the first to enter the application? [example: './src/index']` should allow us to write `path.join` and similar, like we can do with [multiple entries](https://github.com/webpack/webpack-cli/blob/master/lib/generators/utils/entry.js#L35-L46). 

1. Run `webpack-cli init`
2. Type `path.join(__dirname, 'this.js')

A valid entry using `path`, `process` or similar should be visible in the webpack config.

BugFix

**Please paste the results of `webpack-cli info` here, and mention other relevant information such as programming language.**

:-s

--- 

### This issue is reserved for first time contributors, only. Any PR submitted by previous contributors will not get review nor acceptance.

Saberre:

Bugfix.

No, there wasn't an existing test suite for `lib/generators/utils/entry.js`.

No.

**Summary**

Fixes #397. Add same input processing code to single entry choice. There were three same code blocks, so I refactored it to a separate function. I'm not sure about adding a utility function to the top of `lib/genereators/utils/entry.js`. If it's better to move the function somewhere, please let me know!

No.

**Other information**

### DIFF
--- a/lib/generators/utils/entry.js
+++ b/lib/generators/utils/entry.js
@@ -3,6 +3,20 @@
 const InputValidate = require("webpack-addons").InputValidate;
 const validate = require("./validate");
 
+function normalizeEntryPoint(entryPoint) {
+	if (
+		entryPoint.charAt(0) !== "(" &&
+		entryPoint.charAt(0) !== "[" &&
+		!entryPoint.includes("function") &&
+		!entryPoint.includes("path") &&
+		!entryPoint.includes("process")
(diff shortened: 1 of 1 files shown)


ev1stensberg:
Seems like you've just abstracted the function used for multiple entries ( which is great <3 )

PR #410, opened 2018-04-18
fix: remove reference to specific version during migration

PlayMa256:
bugfix

Not needed

**Summary**
Migration feedback was referencing a specific version of webpack. I kept that generic.

No

### DIFF
--- a/lib/commands/migrate.js
+++ b/lib/commands/migrate.js
@@ -178,9 +178,15 @@ function runMigration(currentConfigPath, outputConfigPath) {
 							);
 						}
 					}
+					console.log(chalk.green(`\n✔︎ New webpack config file is at ${outputConfigPath}.`));
 					console.log(
 						chalk.green(
-							`\n ✔︎ New webpack config file is at ${outputConfigPath}`
+							"✔︎ Heads up! Updating to the latest version could contain breaking changes."
+						)
+					);
+					console.log(
+						chalk.green(
+							"✔︎ Plugin and loader dependencies may need to be updated."
 						)
 					);
 				});
--- a/lib/commands/migrate.js
+++ b/lib/commands/migrate.js
@@ -180,7 +180,7 @@ function runMigration(currentConfigPath, outputConfigPath) {
 					}
 					console.log(
 						chalk.green(
-							`\n ✔︎ New webpack v2 config file is at ${outputConfigPath}`
+							`\n ✔︎ New webpack config file is at ${outputConfigPath}`
 						)
 					);
 				});


lib/commands/migrate.js
```diff
@@ -180,7 +180,7 @@ function runMigration(currentConfigPath, outputConfigPath) {
 					}
 					console.log(
 						chalk.green(
-							`\n ✔︎ New webpack v2 config file is at ${outputConfigPath}`
+							`\n ✔︎ New webpack config file is at ${outputConfigPath}`
 						)
```
PlayMa256: Since how plugins and loaders works changed from v3->v4, i think we could add a message down below, when everything is finished, telling the user to update all plugins/loaders. What do you all think?

dhruvdutt: Definitely, we should add that.

PlayMa256:
Updated.

lib/commands/migrate.js
```diff
@@ -180,7 +180,17 @@ function runMigration(currentConfigPath, outputConfigPath) {
 					}
 					console.log(
 						chalk.green(
-							`\n ✔︎ New webpack v2 config file is at ${outputConfigPath}`
+							`\n✔︎ New webpack config file is at ${outputConfigPath}`
+						)
+					);
+					console.log(
+						chalk.green(
+							"✔︎ Consider updating your plugins and loaders!"
```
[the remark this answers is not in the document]
PlayMa256: Consider updating your dependencies (loaders and plugins) (???). What do you suggest?

lib/commands/migrate.js
```diff
@@ -178,9 +178,15 @@ function runMigration(currentConfigPath, outputConfigPath) {
 							);
 						}
 					}
+					console.log(chalk.green(`✔︎ New webpack config file is at ${outputConfigPath}.`));
```
bitpshr: We lost the new line before these messages and are now printing unnecessary newlines between each one. We should really just clean this up and add these statements into a single `console.log` with a template string (this avoids four unnecessary function calls):

```js
console.log(chalk.green(`
✔︎ New webpack config file is at ${outputConfigPath}.
✔︎ Updating to the latest version can contain breaking changes.
✔︎ Plugin and loader dependencies may also need to be updated.
`));
```

Alternatively we can keep them separate like they are now:

```js
console.log(chalk.green(`\n✔ New webpack config file is at ${outputConfigPath}.`));
console.log(chalk.green("✔ Updating to the latest version can contain breaking changes."));
console.log(chalk.green("✔ Plugin and loader dependencies may also need to be updated."));
```

lib/commands/migrate.js
```diff
@@ -178,9 +178,15 @@ function runMigration(currentConfigPath, outputConfigPath) {
 							);
 						}
 					}
+					console.log(chalk.green(`\n✔︎ New webpack config file is at ${outputConfigPath}.`));
 					console.log(
 						chalk.green(
-							`\n ✔︎ New webpack v2 config file is at ${outputConfigPath}`
+							"✔︎ Heads up! Updating to the latest version can contain breaking changes."
```
ev1stensberg: Could

lib/commands/migrate.js
```diff
@@ -178,9 +178,15 @@ function runMigration(currentConfigPath, outputConfigPath) {
...
+						)
+					);
+					console.log(
+						chalk.green(
+							"✔︎ Plugin and loader dependencies may also need to be updated."
```
ev1stensberg: Drop the also

ematipico:
@PlayMa256 Could you rebase your branch from master please?

PlayMa256:
@ematipico done

PR #423, opened 2018-05-01
cli(symlinks): always defer to global

ev1stensberg:

Removes preference of local imports as they result in undefined behaviour.
No need
N/A
**Summary**

No

**Other information**


flyyang:
is this released?

ev1stensberg:
It was, but it didnt work well. Doing a refactor that will fix this in aug

PR #433, opened 2018-05-05
Move to scoped packages

sendilkumarn:


sendilkumarn:
I too had the same idea, but I think it is better to have the core as the base package, like users then just install `webpack-cli`

ev1stensberg:
So it lives under `webpack-cli`?

sendilkumarn:
yeah

ev1stensberg:
Closing in favor for #434

PR #434, opened 2018-05-06
chore(cli): move to lerna and scoped packages

ev1stensberg:
@sendilkumarn your pr is pretty outdated, let's continue on this branch. 

- [ ] Install deps at correct package

- [ ] Link them up

- [ ] Make tests and linting pass

- [ ] Test coverage should be collected from all repos

- [ ] Commands should not use dependencies, we should prompt users to install it if it's not installed and every action should come from the package itself


ev1stensberg:
Tried to migrate the repo :p Anything in particular I did wrong?

sendilkumarn:
Just shocked by this, Glad that I saw the PR before I started to merge things 👍 I will have a look at it this week.

ev1stensberg:
Gotcha. I'm not as proficient in Lerna as you, but this is the folder structure I propose. Independent packages, top scope is completely empty and allows users to use the CLI without a huge noldus modulus. 

Does it look ok? I think monorepo becomes hard for tests and the CI w.r.t collecting coverage and making sure each package do their jam

cli.js
```diff
@@ -15,7 +15,7 @@
 	}
 
 	require("v8-compile-cache");
-	const ErrorHelpers = require("./errorHelpers");
+	const ErrorHelpers = require("./utils/errorHelpers");
```
sendilkumarn: I don't understand this change. It is wrong, it should be `webpack-cli-utils` right?

ematipico:
@sendilkumarn @ev1stensberg why don't you create an ad-hoc branch and merge everything there? Maybe the `next` branch. Once the PR is ready we can merge the branch straight to master. It's a breaking change (a big one), this solution would help to mitigate disruptions.

sendilkumarn:
@ematipico the problem is any merge that happens to master it will be completely broken and literally redo everything. I already created 3 long PRs and plus one from @ev1stensberg 

Can we move all the dev into the `next` 🤔

ematipico:
I don't see where's the problem, I am surprised you didn't do it before 😅

cli.js
ev1stensberg: No, this is outer-scope utils for the core CLI files. Maybe naming convention is bad?

ev1stensberg:
Yep, dev into next. I'll rebase this one

ev1stensberg:
@sendilkumarn rebased the most recent master to dev branch. If we publish on next, will it work?

PR #447, opened 2018-05-10
cli(node): add node-options handler

PlayMa256:
Feature

Yes

**Summary**
Add possibility to pass node options to the cli.

no


PlayMa256:
@ev1stensberg done.

PlayMa256:
Althought monitoring process did not showed anything problematic, i've seem some memory problems when tests runs... i'm having a look at that.

bin/webpack.js
```diff
@@ -34,6 +34,13 @@
 		"info"
 	];
 
+	const hasNodeFlags = process.argv.filter(arg => arg.includes("node."));
+
+	if (hasNodeFlags.length) {
+		require("./process-node-options")(process.argv);
+		return;
+	}
```
Connormiha: There is no need `filter` to check it.
```js
const hasNodeFlags = process.argv.filter(arg => arg.includes("node."));

if (hasNodeFlags.length) {
```
=>
```js
const hasNodeFlags = process.argv.some(arg => arg.includes("node."));

if (hasNodeFlags) {
```

ev1stensberg:
@PlayMa256 any update on this? Is it ready for a review?

PlayMa256:
Sorry @ev1stensberg , i had no time to look at this last week. By the way, do you have any ideas why the tests are consuming more memory and failing at appvoyer? The process management seems right, they aren't being kept alive when the children has terminated. 

To be able to implement that, the other approach i had implemented (with a different file invoking `webpack.js`) would be better, since we would need to use `watchpack` or something like that to look for changes on the config file and restart the process. I'm suggesting the other approach because that would be a lot more organized than this one.

ev1stensberg:
Try to commit a new commit with rebase against the next branch, should be better now

.gitignore
```diff
@@ -30,3 +30,6 @@ lerna-debug.log
 
 # Yeoman file
 .yo-rc.json
+
+# Yarn lock file
+yarn.lock
```
dhruvdutt: This has been already merged in #455. Please rebase.

PlayMa256: i rebased =/

PlayMa256: I'll try to see what i can get with this... rebasing again...

ematipico: Rebased from `next`?

PlayMa256: Yes, when i do git rebase upstream/next it says it has nothing to update. Neither a diff shows any difference. I've also copied next to my fork and it says the same.

PlayMa256: @ematipico @dhruvdutt reverted everything all the way from the rebase, where my branch started. Now git managed to do the rebase correctly and i only replayed all the commands...

PR #451, opened 2018-05-18
cli(init): Better defaults

ev1stensberg:

N/A
N/A
**Summary**
Fixes empty entry on scaffold error and sets better defaults.
No

**Other information**


packages/generators/utils/module.js
```diff
@@ -10,10 +10,13 @@
 module.exports = _ => {
 	return {
 		test: `${new RegExp(/\.js$/)}`,
-		exclude: "/node_modules/",
+		include: ["path.resolve(__dirname, 'src')"],
```
ev1stensberg: @sokra just wondering. If the entry folder is different, then this would need to change, right?

packages/generators/init-generator.js
```diff
@@ -29,7 +29,14 @@ module.exports = class InitGenerator extends Generator {
...
 		super(args, opts);
 		this.isProd = false;
-		this.dependencies = ["webpack", "webpack-cli", "uglifyjs-webpack-plugin"];
+		(this.usingDefaults = false),
+		(this.dependencies = [
+			"webpack",
+			"webpack-cli",
+			"uglifyjs-webpack-plugin",
+			"babel-plugin-syntax-dynamic-import",
+			"path"
```
dhruvdutt: Isn't `path` core Node.js module?

dhruvdutt: Yes, but these are packages that would get installed by npm/yarn, right? Definition part would be in [topScope](https://github.com/webpack/webpack-cli/blob/next/packages/generators/init-generator.js#L69).

packages/generators/init-generator.js
```diff
@@ -93,20 +100,25 @@ module.exports = class InitGenerator extends Generator {
...
 					this.configuration.config.webpackOptions.output = {
 						filename: "'[name].[chunkhash].js'"
 					};
 				}
 				if (outputTypeAnswer["outputType"].length) {
 					outputPath = outputTypeAnswer["outputType"];
 				}
-				this.configuration.config.webpackOptions.output.path = `path.resolve(__dirname, '${outputPath}')`;
+				if (!this.usingDefaults) {
+					this.configuration.config.webpackOptions.output.path = `path.resolve(__dirname, '${outputPath}')`;
```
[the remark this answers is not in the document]
ev1stensberg: I don't know. It's not a super-big generator, so it might be a premature optimization?

.appveyor.yml
```diff
@@ -16,6 +16,7 @@ install:
   - npm ci
   - npm install --global codecov
   - npm install eslint-plugin-node@latest --save-dev
+  - npm install eslint-plugin-prettier@latest --save-dev
```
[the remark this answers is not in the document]
ev1stensberg: It was fetched from the rebase I think

ev1stensberg:
@sendilkumarn what do you mean with fetch from config file?

packages/generators/init-generator.js
ematipico: Yes exactly. `path` needs to be removed

packages/generators/init-generator.js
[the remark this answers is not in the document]
ematipico: I would say that we **have to think about optimisation** in advance

packages/generators/init-generator.js
```diff
@@ -394,9 +406,13 @@ module.exports = class InitGenerator extends Generator {
 				done();
 			});
 	}
+	/*
```
ematipico: Is this commented code? If so, it can be removed

packages/utils/recursive-parser.js
```diff
@@ -36,7 +36,9 @@ module.exports = function recursiveTransform(j, ast, value, action, key) {
 			} else if (key === "merge") {
 				utils.parseMerge(j, p, value);
 			} else {
-				utils.addProperty(j, p, key, value);
+				if (value) {
```
ematipico: Why not an `else if`?

ev1stensberg: old commit

packages/generators/init-generator.js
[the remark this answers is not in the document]
ev1stensberg: 🤔

packages/utils/scaffold.js
```diff
@@ -43,15 +43,23 @@ module.exports = function runTransform(webpackProperties, action) {
...
 			initActionNotDefined
 				? webpackProperties.configFile
 				: "module.exports = {}"
 		);
 		const transformAction = action || null;
-
 		return pEachSeries(transformations, f => {
-			return astTransform(j, ast, config.webpackOptions[f], transformAction, f);
+			if (f === "merge" || f === "topScope") {
+				return astTransform(j, ast, f, config[f], transformAction);
```
dhruvdutt: This is a breaking change in astTranform signature. Fixed in #462

ev1stensberg: Nope. `f` is key, `config[f] / config.webpackOptions[f]` is key

PR #452, opened 2018-05-19
cli(color): don't use supports-color on non-tty

ev1stensberg:
Bugfix

N/A
N/A
**Summary**

Discards supports-color on non-tty output.

No

**Other information**
???!?


ev1stensberg:
Process-options is not used anymore, so that's removed

bin/cli.js
```diff
@@ -79,7 +79,9 @@ For more information, see https://webpack.js.org/api/cli/.`);
 			type: "boolean",
 			alias: "colors",
 			default: function supportsColor() {
-				return require("supports-color").supportsColor;
+				if (process.stdout.isTTY === true) {
```
ematipico: The PR says to not use supports-color. So why here it's the opposite?

ev1stensberg: ` don't use supports-color on non-tty`

ematipico: Yeah I see my mistake :D

ryancastle:
@ev1stensberg FWIW I believe this prevents colors from being possible in some meaningful use cases. Although I also realise there may not be a good resolution.

We run `webpack --watch` and `webpack-dev-server` inside docker containers. Colors are no longer easily achievable on `webpack --watch`. Our usage is non-interactive, but colors make a **big** difference to the usability. It seems that `webpack-dev-server` still might use the old "supports-color" or similar, because it works fine.

We also can't allocate a TTY using `docker run -t`, because the behaviour of those containers changes.

ev1stensberg:
Hi @ryancastle . Is there any way for you do set `process.stdout.isTTY` in a ENV file/invocation through shell? If not, let me know and I'll add an option for you to use together with this

ryancastle:
Added this early in `webpack.config.js` which resolved it. Let's hope it's not a fib. :)

`process.stdout.isTTY = true;`

ev1stensberg:
Awesome! Happy to help out any way possible 👍

jayjun:
@ev1stensberg Phoenix (Elixir web framework) starts watchers directly as child processes, so `process.stdout.isTTY` is `undefined`.

Phoenix 1.4 (the next version) will support webpack [officially](https://github.com/phoenixframework/phoenix/blob/ca86ede492659bf3624338265cec0b08f284989e/installer/templates/phx_umbrella/apps/app_name_web/config/dev.exs#L18) and it’s a bummer colors stopped working.

Can this be changed so passing `--colors` or `--no-color` overrides whatever `process.stdout.isTTY` is?

ev1stensberg:
Yes

jayjun:
Passing `FORCE_COLOR=1` as an environment variable works and, in my opinion, way less hackish than `process.stdout.isTTY = true` but these are still an unnecessary hacks.

supports-color already has logic to [detect TTY](https://github.com/chalk/supports-color/blob/bcd82c998094535f3462a3e37856852938ccbb37/index.js#L50) and [`--color` overrides](https://github.com/chalk/supports-color/blob/bcd82c998094535f3462a3e37856852938ccbb37/index.js#L8-L17) properly.

I suggest we remove the `process.env.isTTY` check and let support-color do its thing.

ev1stensberg:
@jayjun That has been bothering me, supports-color test env vars for the environments but still it fails, tested this myself before the patch. I'm going to reimplement this in the new webpack-cli version which is probably coming next month or so.

PR #453, opened 2018-05-19
cli(symlinks): Fix paths

ev1stensberg:
Fixes error that says "cannot find module webpack / webpack-cli"

N/A, let's see the CI results
N/A
**Summary**
Fixes the error where webpack and the CLI have different locations and one of the deps will throw. Tried unshifting `module.paths`, so that the global one will be the first to be looked up, but it fails even with regular require statements. The fix is to assign the webpack dep to a variable in the process in order to reference it correctly for further use.

No

**Other information**


bin/cli.js
```diff
@@ -15,6 +15,17 @@
 	}
 
 	require("v8-compile-cache");
+
+	// try local module, fallback to global
+	try {
```
ev1stensberg: @sokra you might have to do the same thing at `webpack/bin`once we publish the next branch to master. Path lookups are failing, and adjusting them according to their global/local paths aren't helping...

PR #456, opened 2018-05-20
ast(parser): add

dhruvdutt:
Feature

Yes

No

**Summary**
Initial WIP for adding AST support for add commands.

- [x] Create util function for update AST node
- [x] Add support for all node types (right-hand side)
- [x] For nested objects, only update the required key, not replace the entire object (keep existing nested object keys uneffected)
- [x] Add tests

No


ev1stensberg:
Split these up to PR's for Add, update and remove respectively.

packages/utils/recursive-parser.js
```diff
@@ -4,38 +4,59 @@ const utils = require("./ast-utils");
...
+				"name"
+			]) === "exports"
+			);
+		})
+		.filter(p => p.value.properties);
+
 	if (node.size() !== 0) {
-		// push to existing key
-		return ast;
+		// select node with existing key
```
ev1stensberg: Abstract this

packages/utils/recursive-parser.js
```diff
@@ -4,38 +4,59 @@ const utils = require("./ast-utils");
...
+				// update property/replace
+					j(p).replaceWith(
+						j.property(
+							"init",
+							j.literal(key),
+							j.literal(value)
+						)
+					);
+				} else if (action === "remove") {
+					j(p).remove();
```
ev1stensberg: Not enough, just worry about the update feature for now

packages/utils/recursive-parser.js
dhruvdutt: Updated to use `findRootNodesByName` util method.

packages/utils/recursive-parser.js
```diff
@@ -4,38 +4,51 @@ const utils = require("./ast-utils");
...
+		.filter(p => p.value.properties);
+
 	if (node.size() !== 0) {
-		// push to existing key
-		return ast;
+		// select node with existing key
+		return utils.findRootNodesByName(
+			j, root, key
+		)
+			.forEach(p => {
```
ematipico: What is `p`? Could we choose a better name?

packages/utils/recursive-parser.js
```diff
@@ -4,38 +4,51 @@ const utils = require("./ast-utils");
 
 module.exports = function recursiveTransform(j, ast, value, action, key) {
 	const node = utils.findRootNodesByName(j, ast, key);
+
+	// get module.exports prop
+	const root = ast
+		.find(j.ObjectExpression)
+		.filter(p => {
```
ematipico: Same here

packages/utils/ast-utils.js
```diff
@@ -429,6 +429,62 @@ function addProperty(j, p, key, value) {
...
+ *
+ * Recursively updates an object/property to a node
+ * @param {any} j — jscodeshift API
+ * @param {Node} p - AST node
+ * @param {String} key - key of a key/val object
+ * @param {Any} value - Any type of object
+ * @returns {Node} - the updated ast
+ */
+
+function updateProperty(j, p, key, value) {
```
ematipico: Also here, `p` means nothing. We could call it `propertyToAdd`

packages/utils/ast-utils.js
```diff
@@ -429,6 +429,62 @@ function addProperty(j, p, key, value) {
...
+	let valForNode;
```
ematipico: This can be moved after the `if`. There's no need to declare it if we exist from the function

packages/utils/ast-utils.js
```diff
@@ -429,6 +429,62 @@ function addProperty(j, p, key, value) {
...
+	if (!p) {
+		return;
+	}
+	if (Array.isArray(value)) {
+		let arrExp = j.arrayExpression([]);
```
ematipico: `arrExp` can a different name. Maybe `valueForNodeAsArray`? Or something like that means something

packages/utils/ast-utils.js
```diff
@@ -429,6 +429,62 @@ function addProperty(j, p, key, value) {
...
+			"value",
+			"value",
+		]) && p.value.value.type === "ArrayExpression") {
+			arrExp = p.value.value;
+		}
+		value.forEach(val => {
+			addProperty(j, arrExp, null, val);
+		});
+		valForNode = arrExp;
+	} else if (
```
ematipico: Could we have comment here where we explain what kind of check we're doing? That `__paths` sounds like an internal thing, so it would be better to explain why

packages/utils/ast-utils.js
```diff
@@ -429,6 +429,62 @@ function addProperty(j, p, key, value) {
...
+		Object.keys(value).forEach(prop => {
+			addProperty(j, objectExp, prop, value[prop]);
+		});
+		valForNode = objectExp;
+	} else {
+		valForNode = createIdentifierOrLiteral(j, value);
+	}
+	let pushVal;
+	if (key) {
+		pushVal = j.property("init", j.identifier(key), valForNode);
```
ematipico: When do you have this case?

packages/utils/ast-utils.js
dhruvdutt: I think we can move it below `if (!p) { return; }` check, but not inside if block as it's getting assigned in every _if else_ condition which will make variable declarations redundant.

packages/utils/ast-utils.js
dhruvdutt: @ev1stensberg This check is to avoid AST nodes during the recursion, right? So, this will pass only object values from webpack config and avoid AST node objects.

packages/utils/ast-utils.js
dhruvdutt: Almost every case. What happens is that we compute the right-hand part of a particular property (for instance _development_ or _production_) and store it in `valForNode` variable. We then push this value to the key (_mode_ in this case). Similarly, `valForNode` can be a Literal (_mode_), ArrayExpression (_plugins_) or ObjectExpression(_output_).

The else part of this check happens when we recursively add something. For instance, we have an array of objects (_rules_). We will recursively add every object to an ArrayExpression. In this case, there is no `key`. We just have to keep adding objects inside the array.

packages/utils/ast-utils.js
ematipico: > I think we can move it below if (!p) { return; } check
Yes, that's what I meant

packages/utils/ast-utils.js
ematipico: Yes, I get that. I am more up for names that mean something based on the context of the operation. Just that.

packages/utils/recursive-parser.js
ematipico: But `p` has a different meaning based on where it is used. We have JSDoc that helps us already. At least let's give something to the developers, so when they land to the code they know what that variable is meant for

packages/utils/ast-utils.js
ematipico: Thanks :smile:

packages/utils/ast-utils.js
ev1stensberg: If a node has `__paths`, it's a regexp which cannot be resolved into a regular literal or identifier

packages/utils/ast-utils.js
ev1stensberg: Maybe a comment will do, if you follow common sense we check if the value is an array exp, so we want to create an array node

packages/utils/recursive-parser.js
ev1stensberg: p is alright

packages/utils/recursive-parser.js
```diff
@@ -4,38 +4,51 @@ const utils = require("./ast-utils");
...
-				);
-			})
-			.filter(p => p.value.properties);
 		return root.forEach(p => {
 			if (key === "topScope") {
 				utils.parseTopScope(j, p, value);
 			} else if (key === "merge") {
 				utils.parseMerge(j, p, value);
 			} else {
+				// init, add new property
```
ev1stensberg: comment right after conditional if you're gonna comment it. Also, I think it's clear what we are doing here

packages/utils/ast-utils.js
```diff
@@ -429,6 +429,62 @@ function addProperty(j, p, key, value) {
...
+		value.forEach(val => {
+			addProperty(j, arrExp, null, val);
+		});
+		valForNode = arrExp;
+	} else if (
+		typeof value === "object" &&
+		!(value.__paths || value instanceof RegExp)
+	) {
+		let objectExp = j.objectExpression([]);
+		if (safeTraverse(p, [
```
ev1stensberg: You'll have to check the entire key/val operation. Not sure if this is enough, let me test

packages/utils/ast-utils.js
ev1stensberg: First rule of recursion is to have an exit condition, but for this function it needs to exit gracefully, as it's expected to return a node

packages/utils/ast-utils.js
ev1stensberg: p is fine, it's an ast node, we don't really care about the type of the node as long as we're updating its properties ( see the name of the function )

packages/utils/recursive-parser.js
dhruvdutt: This part would be used in _add_ command as well in cases when a property doesn't exist. For instance, when someone tries to add `mode` property which isn't present in config, `node.size()` will be 0 and this `addProperty()` would be called. That's why want to make super clear that this part of the code is getting used in _init_ and _add_.
In cases when `mode` is already present and we're overwriting it, `updateProperty()` will be called. Basically, _add_ and _update_ commands would work same.

packages/utils/ast-utils.js
dhruvdutt: I've added some tests. Let me know if you run into any issues here.

packages/utils/__testfixtures__/fixture-2.input.js
```diff
@@ -0,0 +1,9 @@
+const webpack = require("webpack");
+
+module.exports = {
+	entry: {
+        ed: 'index.js',
```
sendilkumarn: really `ed Sheeran` ha ha 😄

ev1stensberg: He's a good singer though! 😆

packages/utils/recursive-parser.js
ev1stensberg: Maybe we can just have three options: `add` (renamed from init), `update` and `remove`

ev1stensberg:
@dhruvdutt could you do a rebase?

dhruvdutt:
@ev1stensberg Rebased. PTAL.

dhruvdutt:
Rebased. PTAL @ev1stensberg

packages/utils/recursive-parser.test.js
```diff
@@ -25,3 +25,55 @@ defineTest(
...
+				test: new RegExp(/\.(js|vue)$/),
+				loader: "'eslint-loader'",
+				enforce: "'pre'",
+				include: ["customObj", "'Stringy'"],
+				options: {
+					formatter: "'someOption'"
+				}
+			}
+		],
+		nice: "':)'",
```
ematipico: Could you also adda a `Promise` to the test? Just to make sure that we cover every case
```js
{
  foo: Promise.resolve()
}
```

packages/utils/recursive-parser.test.js
```diff
@@ -25,3 +25,55 @@ defineTest(
 	},
 	"init"
 );
+
```
ematipico: How does it behave if you tried to add property that already exists? Could you define a test for this case please?

packages/utils/ast-utils.js
ematipico: Worth leaving a comment inside the code

packages/utils/recursive-parser.test.js
ev1stensberg: ^ Otherwise this would just be the `init` feature

dhruvdutt: `entry` was already present in the input fixture.  I've added `mode` as well to make it clearer.

packages/utils/recursive-parser.test.js
dhruvdutt: Added! :+1:

manualt/webpack.config.js
```diff
@@ -0,0 +1,15 @@
+const webpack = require('webpack');
```
ev1stensberg: Remove this folder please

packages/utils/recursive-parser.test.js
```diff
@@ -21,7 +21,61 @@ defineTest(
...
 		],
 		nice: "':)'",
+		foo: "Promise.resolve()",
 		man: "() => duper"
 	},
 	"init"
 );
+
+// Tests for _add_ command
+defineTest(
```
ev1stensberg: This is nice. Just have one test case with a lot of properties, that will hold and tests would run faster

ev1stensberg: No need to comment this, the action specifies what the test is supposed to do

packages/utils/ast-utils.js
```diff
@@ -507,6 +507,62 @@ function parseMerge(j, ast, value, action) {
...
+ */
+
+function updateProperty(j, p, key, value) {
+	if (!p) {
+		return;
+	}
+	let valForNode;
+	if (Array.isArray(value)) {
+		let arrExp = j.arrayExpression([]);
+		if (safeTraverse(p, [
```
ev1stensberg: These specific parts could be abstracted. One function to  have a sanity check for the node, return `p.value.value.type` from the function and use that. i.e `safeTraverseAndGetType(p) <string>`

packages/utils/ast-utils.js
```diff
@@ -507,6 +507,62 @@ function parseMerge(j, ast, value, action) {
...
+			"value",
+			"value",
+		]) && p.value.value.type === "ArrayExpression") {
```
ev1stensberg: Same here, see function comment.

packages/utils/recursive-parser.js
```diff
@@ -2,41 +2,54 @@
 
 const utils = require("./ast-utils");
 
-module.exports = function recursiveTransform(j, ast, key, value, action) {
+module.exports = function recursiveTransform(j, ast, value, action, key) {
```
ev1stensberg: don't change this. `key`, `value` is correct order.

packages/utils/recursive-parser.js
```diff
@@ -2,41 +2,54 @@
...
-		// push to existing key
-		return ast;
+		// select node with existing key
+		return utils.findRootNodesByName(
+			j, root, key
+		)
+			.forEach(p => {
+				if (action === "add") {
+				// update property/replace
+					j(p).replaceWith(
```
ev1stensberg: Looks good. Only problem is that we need to find a way to mark anonymous objects / values and ask if the user wants to update those inside the `add` generator.

ev1stensberg:
Make sure to change those function calls and revert the `runTransform` signature to the way it was. Remove the `init` file in `packages/init`, that is unused. Also, make sure to tackle point no. 4

packages/utils/ast-utils.js
```diff
@@ -14,8 +14,9 @@ function safeTraverse(obj, paths) {
 	return val;
 }
 
-function safeTraverseAndGetType(p, traversalArray) {
-	return safeTraverse(p, traversalArray).type;
+function safeTraverseAndGetType(p) {
```
ev1stensberg: Jsdocs and tests needed

packages/utils/ast-utils.js
```diff
@@ -1,5 +1,6 @@
 const validateIdentifier = require("./validate-identifier");
 
+// Traverse safely over a path object for array for paths
```
ev1stensberg: This is not valid JSDoc format

packages/utils/ast-utils.js
```diff
@@ -14,6 +15,7 @@ function safeTraverse(obj, paths) {
 	return val;
 }
 
+// Traverse safely and return `type` for path object with value.value property
```
ev1stensberg: Not a valid JSdoc format

packages/utils/ast-utils.test.js
```diff
@@ -210,4 +210,58 @@ const a = { plugs: [] }
 			expect(j(require).toSource()).toMatchSnapshot();
 		});
 	});
+
+	describe("safeTraverse", () => {
+		it("should safe traverse", () => {
+			const VALUE = "a value";
```
ev1stensberg: better naming here

packages/utils/ast-utils.test.js
```diff
@@ -210,4 +210,58 @@ const a = { plugs: [] }
...
+				value: {
+					value: VALUE
+				}
+			};
+			const require = utils.safeTraverse(p, ["value", "value"]);
+			expect(require).toEqual(VALUE);
+		});
+
+		it("should safe traverse thrice", () => {
+			const VALUE = "a value";
```
ev1stensberg: this should match a realistic case, not `value`, `value`, `value`

packages/utils/ast-utils.js
```diff
@@ -382,26 +389,37 @@ function getRequire(j, constName, packagePath) {
  * @param {Node} p - AST node
  * @param {String} key - key of a key/val object
  * @param {Any} value - Any type of object
+ * @param {Boolean} update - Flag to check if node needs to update
  * @returns {Node} - the created ast
  */
 
-function addProperty(j, p, key, value) {
-	let valForNode;
+function addProperty(j, p, key, value, update = false) {
```
ev1stensberg: This should be action, not `update`. Adjust to general cases, not the ones that fits your needs now. It might haunt you l8ter...

packages/utils/ast-utils.js
```diff
@@ -14,6 +21,17 @@ function safeTraverse(obj, paths) {
 	return val;
 }
 
+/**
+ *
+ * Traverse safely and return `type` for path object with value.value property
+ * @param {Node} path - pathNode
```
ev1stensberg: change to AST node in description

packages/utils/ast-utils.js
```diff
@@ -415,6 +444,11 @@ function addProperty(j, p, key, value) {
 	} else {
 		pushVal = valForNode;
 	}
+
+	// incase of _add_ or while updating a property,
+	// we only return the generated pushVal which will be replace the node path
```
ev1stensberg: keep this, discard the other comment

packages/utils/ast-utils.test.js
```diff
@@ -210,4 +210,144 @@ const a = { plugs: [] }
...
+				.find(j.ObjectExpression);
+
+			root.forEach(p => {
+				utils.addProperty(j, p, "entry", propertyValue);
+			});
+
+			expect(ast.toSource()).toMatchSnapshot();
+		});
+
+		it("add entry property while add", () => {
```
ev1stensberg: `using` not `while`

packages/utils/ast-utils.test.js
```diff
@@ -210,4 +210,144 @@ const a = { plugs: [] }
...
+					}
+				}
+			};
+			const require = utils.safeTraverseAndGetType(p);
+			expect(require).toEqual(false);
+		});
+	});
+
+	describe("addProperty", () => {
+		it("add entry property while init", () => {
```
ev1stensberg: `while` -> `using`

packages/utils/ast-utils.test.js
```diff
@@ -210,4 +210,146 @@ const a = { plugs: [] }
...
+		it("should safe traverse and get type", () => {
+			const NODE_TYPE = "NodeType";
+			const p = {
+				value: {
+					value: {
+						type: NODE_TYPE
+					}
+				}
+			};
+			const require = utils.safeTraverseAndGetType(p);
```
ev1stensberg: vague naming, please change to descriptive names

packages/utils/ast-utils.test.js
```diff
@@ -210,4 +210,146 @@ const a = { plugs: [] }
...
+			const NODE_TYPE = "NodeType";
+			const p = {
+				ed: {
+					sheeran: {
+						type: NODE_TYPE
+					}
+				}
+			};
+			const require = utils.safeTraverseAndGetType(p);
+			expect(require).toEqual(false);
```
ev1stensberg: this is not a valid test, you're testing the node against itself. Give it a real path to check. `p`, `foo`, `bar`.

packages/utils/ast-utils.test.js
```diff
@@ -210,4 +210,146 @@ const a = { plugs: [] }
...
+		});
+	});
+
+	describe("addProperty", () => {
+		it("add entry property while init", () => {
+			const ast = j("module.exports = {}");
+			const propertyValue = {
+				objects: "are",
+				super: [
```
ev1stensberg: we can trim this down so it doesn't get overbloated

packages/utils/ast-utils.test.js
```diff
@@ -210,4 +210,146 @@ const a = { plugs: [] }
...
+					foo: Promise.resolve(),
+					man: () => duper
+				}
+			}`);
+			const propertyValue = {
+				objects: "are not",
+				super: [
+					"op",
+					{
+						test: new RegExp(/\.(wasm|c)$/),
```
ev1stensberg: less props

packages/utils/ast-utils.test.js
```diff
@@ -213,14 +213,16 @@ const a = { plugs: [] }
...
 	describe("safeTraverse", () => {
 		it("should safe traverse", () => {
-			const VALUE = "a value";
+			const type = {
+				type: "NodeType"
+			};
 			const p = {
 				value: {
-					value: VALUE
+					value: type
```
ev1stensberg: change the props to `foo`, `bar`, we're not testing props, we're testing the AST values. This would confuse people looking at the tests

packages/utils/ast-utils.test.js
```diff
@@ -213,14 +213,16 @@ const a = { plugs: [] }
...
 				}
 			};
 			const require = utils.safeTraverse(p, ["value", "value"]);
-			expect(require).toEqual(VALUE);
+			expect(require).toEqual(type);
```
ev1stensberg: type is not descriptive, give it a mock name, like `testObject`

packages/utils/ast-utils.test.js
```diff
@@ -210,4 +210,122 @@ const a = { plugs: [] }
...
+		it("should safe traverse and return false if not found", () => {
+			const NODE_TYPE = "NodeType";
+			const p = {
+				foo: {
+					bar: {
+						type: NODE_TYPE
+					}
+				}
+			};
+			const typeValue = utils.safeTraverseAndGetType(p);
```
ev1stensberg: se the prev comment on a proper test case

dhruvdutt: We are hardcoding `value.value` inside [`safeTraverseAndGetType()`](https://github.com/webpack/webpack-cli/pull/456/files#diff-0214ffef8038112039348669c8ba1baaR31), so the test has to be in that shape only.
Should we add an optional second arg for pathValues inputs and fall back to `values.values` when second arg is not present?

PR #461, opened 2018-05-23
ast(parser): command remove

dhruvdutt:
Feature

No

No

**Summary**
peek 2018-06-27 20-33

[EDIT] Forgot to show that if there are more than one key in an object, it will keep other keys and if there are is one key in an object then it will delete the whole object.

peek 2018-06-27 20-41

No

Closes #500

### DIFF
--- a/packages/generators/remove-generator.js
+++ b/packages/generators/remove-generator.js
@@ -1,3 +1,122 @@
 const Generator = require("yeoman-generator");
+const path = require("path");
+const fs = require("fs");
+const { List } = require("@webpack-cli/webpack-scaffold");
 
-module.exports = class RemoveGenerator extends Generator {};
+const PROP_TYPES = require("@webpack-cli/utils/prop-types");
+
+/**
+ *
+ * Generator for removing properties
+ * @class	RemoveGenerator
+ * @extends	Generator
+ * @returns	{Void} After execution, transforms are triggered
+ *
+ */
+
+module.exports = class RemoveGenerator extends Generator {
+	constructor(args, opts) {
+		super(args, opts);
+		this.configuration = {
+			config: {
+				webpackOptions: {},
+			}
+		};
+
+		let configPath = path.resolve(process.cwd(), "webpack.config.js");
+		const webpackConfigExists = fs.existsSync(configPath);
+		if (!webpackConfigExists) {
+			configPath = null;
+			// end the generator stating webpack config not found or to specify the config
+		}
+		this.webpackOptions = require(configPath);
+	}
+
+	getPropTypes() {
+		return Object.keys(this.webpackOptions);
+	}
+
+	getModuleLoaders() {
+		if (this.webpackOptions.module && this.webpackOptions.module.rules) {
+			return this.webpackOptions.module.rules.map(rule => rule ? rule.loader : null);
+		}
+	}
+
+	prompting() {
+		const done = this.async();
+		let propValue;
+
+		return this.prompt([
+			List(
+				"propType",
+				"Which property do you want to remove?",
+				Array.from(this.getPropTypes())
+			)
+		])
+			.then(({ propType }) => {
+				if (!PROP_TYPES.has(propType)) {
+					console.log("Invalid webpack config prop");
+					return;
+				}
+
+				propValue = this.webpackOptions[propType];
+				if (typeof propValue === "object") {
+					if (Array.isArray(propValue)) {
+						return this.prompt([
+							List(
+								"keyType",
+								`Which key do you want to remove from ${propType}?`,
+								Array.from(propValue)
+							)
+						]).then(({ keyType }) => {
+							this.configuration.config.webpackOptions[propType] = [ keyType ];
+						});
+					} else {
+						return this.prompt([
+							List(
+								"keyType",
+								`Which key do you want to remove from ${propType}?`,
+								Array.from(Object.keys(propValue))
+							)
+						])
+							.then(({ keyType }) => {
+								if (propType === "module" && keyType === "rules") {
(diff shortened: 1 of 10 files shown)


packages/generators/remove-generator.js
```diff
@@ -1,3 +1,46 @@
...
+		};
+	}
+
+	prompting() {
+		const done = this.async();
+
+		return this.prompt([
+			List(
+				"actionType",
+				"What property do you want to remove?",
```
ev1stensberg: Which would probably be better here

packages/generators/remove-generator.js
```diff
@@ -1,3 +1,46 @@
...
+				Array.from(PROP_TYPES.keys())
+			)
+		])
+			.then(actionTypeAnswer => {
```
ev1stensberg: You'll need to prompt for sub-props too. Check out how the add generator is built

packages/utils/recursive-parser.js
```diff
@@ -9,9 +9,41 @@ module.exports = function recursiveTransform(j, ast, key, value, action) {
...
 	if (node.size() !== 0) {
-		// push to existing key
-		return ast;
+		// select node with existing key
+		return utils.findRootNodesByName(
+			j, root, key
+		)
+			.forEach(p => {
+				if (action === "remove") {
+					// remove property
```
ev1stensberg: needs to be recursive

packages/generators/remove-generator.js
ematipico: Agree

packages/generators/remove-generator.js
dhruvdutt: I think in case of removal, sub-props would need webpack config to be read. For instance, in case of `plugins`, before we prompt user which plugin he wants to remove, we need to read all plugins from the webpack config file.

In case of add, it works as we prompt all options (or plugins that can be installed) and then the user selects which plugin to install.

If we want to go for sub-prompts then it would need the reading of webpack config file. That's why I've kept it simple for now and removed the key that is selected directly.

ev1stensberg: could you try to impl the read from config? It's quite nice

packages/utils/recursive-parser.js
```diff
@@ -39,6 +39,9 @@ module.exports = function recursiveTransform(j, ast, key, value, action) {
 			if (action === "add") {
 				// update property/replace
 				j(p).replaceWith(utils.addProperty(j, p, key, value, action));
+			} else if (action === "remove") {
+				// remove property
+				j(p).remove();
```
ev1stensberg: is this enough?

packages/generators/remove-generator.js
```diff
@@ -1,3 +1,105 @@
...
+				} else {
+					this.webpackOptions[propType] = null;
+				}
+			})
+			.then(() => {
+				done();
+			});
+	}
+
+	writing() {
```
ev1stensberg: Looking at this, we can abstract the add generator into a base generator and pass in the name ( action ), and use that instead if the behaviour is the same

dhruvdutt: The logic for removal is quite different and simple. It just keeps on checking type of `value` or right-hand part and prompts for removal of keys.
I think there is no common logic between add and remove generator other than saving of yeoman rc file inside `writing()`.

packages/utils/recursive-parser.js
dhruvdutt: Definitely not enough. Working on AST parser for removal..

ev1stensberg: How is it going? What have you done today?

packages/generators/remove-generator.js
ev1stensberg: Ok :)

ev1stensberg:
Let's go for the recursive approach. Feel free to do a benchmark and reopen this if the non-recursive gives better performance. :)

dhruvdutt:
Actually, the tslint issue was resolved in another TS PR. Anyway, I've added the same fix to this PR as well.

packages/utils/recursive-parser.test.js
```diff
@@ -2,41 +2,98 @@
...
+	"fixture-3",
+	"entry",
+	{
+		a: null,
+	}
+);
+
+defineTest(
+	__dirname,
+	"remove",
```
ev1stensberg: You can just do one deep test for remove with multiple props

packages/info/index.ts
```diff
@@ -1,11 +1,12 @@
 import * as envinfo from "envinfo";
+import * as process from "process";
 
 /**
  * Prints debugging information for webpack issue reporting
  */
 
 export default async function info() {
-	console.log(
+	process.stdout.write(
```
ev1stensberg: is this still throwing console error with the rule enabled?

dhruvdutt: No, it would show the warning. We can decide on one standard convention to follow between `console` or `process`.

packages/utils/recursive-parser.test.js
dhruvdutt: This approach is non-recursive and it doesn't drill down deep inside props.
Recursive drilling is only needed for the creation of nodes.

We do cover every case as demonstrated in the GIF. :)
The tests have to be individual. We can add more tests if needed. I think one-two test for each prop type should be good.

ematipico:
@dhruvdutt you can rebase now and the TS lint fix should be ready :)

Next time, please do a cherry-pick

dhruvdutt:
> A test with a deep property and this is lgtm from me

I've added tests for `resolve.alias` property.

Issue #463, opened 2018-05-24
Bug: `--display-entrypoints=false` does not hide entries


sfger:
**Describe the bug**

 `webpack --watch --progress --display-error-details --hide-modules=true --display-chunks=false --display-entrypoints=false` is not work as expected

**What is the current behavior?**

No matter true or false when I give a value of --display-entrypoints, the entrypoints in  the stdout output always shows!  

**To Reproduce**
1. find a project use webpack-cli and open it in the cmd
2. Input the order: webpack --watch --progress --display-error-details --hide-modules=true --display-chunks=false --display-entrypoints=false
3. See the stdout output

**Expected behavior**
--display-entrypoints=false : the entrypoints do not shows

**Screenshots**
![image](https://user-images.githubusercontent.com/2604977/40461180-ab25af22-5f3c-11e8-8af7-e9355bc70c64.png)

**Please paste the results of `webpack-cli info` here, and mention other relevant information*

**Additional context**
I debuged the code in bin/webpack.js:
just change the code from:
				ifArg("display-entrypoints", function(bool) {
					if (bool) outputOptions.entrypoints = true;
				});
to:
				ifArg("display-entrypoints", function(bool) {
					outputOptions.entrypoints = !!bool;
				})
It will work well:)

ev1stensberg:
Could you submit a PR?

PR #464, opened 2018-05-24
fix(entry): fix a bug when `--display-entrypoints=false` does not hide entries

sfger:
Fix bug: https://github.com/webpack/webpack-cli/issues/463

Yes
No
**Summary**
Fix bug: https://github.com/webpack/webpack-cli/issues/463

NO

**Other information**

### DIFF
--- a/bin/webpack.js
+++ b/bin/webpack.js
@@ -357,7 +357,7 @@ For more information, see https://webpack.js.org/api/cli/.`);
 				});
 
 				ifArg("display-entrypoints", function(bool) {
-					outputOptions.entrypoints = !!bool;
+					outputOptions.entrypoints = bool;
 				});
 
 				ifArg("display-reasons", function(bool) {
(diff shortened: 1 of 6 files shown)


ematipico:
Just out of curiosity, what was the value of `bool` when you debugged the code?

bin/webpack.js
```diff
@@ -357,7 +357,7 @@ For more information, see https://webpack.js.org/api/cli/.`);
 				});
 
 				ifArg("display-entrypoints", function(bool) {
-					if (bool) outputOptions.entrypoints = true;
+					outputOptions.entrypoints = !!bool;
```
ev1stensberg: Double boolean trick is enough to flip it correctly. Used it myself, but it might be some pitfalls, so might be better to do a strict equal instead `if (bool === true)`

sfger: Yes, The trick is redundant。
The param:`bool` is true or false anyway!

ematipico: How come? I think there's something fishy here, otherwise a simple `if (bool)` would be enough, unless `bool` has values different from `true` or `false`, that's why I asked what's its value

ematipico:
Could we add a test?

sfger:
@webpack-bot:
Sure~ That's a good way!

PR #478, opened 2018-06-02
require validateSchema and optionsValidationError from webpack

juliantellez:

### Bug Fix: 

The new version of webpack-cli adds webpack to the global process object. 
This causes applications such as webpack-dev-derver to break since it assumes that `process.webpackModule` is available in the current process.
This pr simply reverts this assumption.

**Summary**

**Other information**

### DIFF
--- a/bin/convert-argv.js
+++ b/bin/convert-argv.js
@@ -4,9 +4,9 @@ fs.existsSync = fs.existsSync || path.existsSync;
 const interpret = require("interpret");
 const prepareOptions = require("./prepareOptions");
 const webpackConfigurationSchema = require("./webpackConfigurationSchema.json");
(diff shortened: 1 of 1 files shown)


ev1stensberg:
Is there any way to keep webpack on the process variable? Global vs local installations was failing cause local/global paths wasn't resolving correctly.

ev1stensberg:
Thanks! Gotta do some more work on this before we can submit the fix!

juliantellez:
Thanks @ev1stensberg ! glad to have put in my two pennies worth 👍

ev1stensberg:
Helped us triage the bug in order to fix that quickly, thanks again @juliantellez !

Issue #479, opened 2018-06-02
[v3.0.0] Wrong workspace setup causing a yarn installation warning on user-end


AviVahl:
**Describe the bug**

When installing `webpack-cli@v3.0.0` using `yarn`, one sees the following warning:
`warning Workspaces can only be enabled in private projects.`

The warning only occurs if `webpack-cli` is being resolved for the first time (meaning it is not yet in the `yarn.lock` file).

**To Reproduce**
```
mkdir new-project
cd new-project
yarn init -y
yarn add webpack webpack-cli --dev
```
See warning.

**Expected behavior**
No such warning should show. I thought there was an error with **my own** workspace (which uses `webpack-cli`).

**Additional context**
`webpack-cli` should be just another package inside packages. It should have its own `package.json` without the "workspaces" field, which causes the warning on user-end.

The root package.json should be `"private": true` and contain the `"workspaces"` key. It should not be published.

Using latest `yarn@1.7.0`.

ev1stensberg:
Could you do a PR?

AviVahl:
Changing the entire project setup and moving stuff around requires better level of familiarity with the current setup/scripts/lifecycle/ci/tooling.

Instead, I made a PR to just remove the `"workspaces"` key, as it is not really used. It should fix the warning on user end.

ev1stensberg:
okay great. That's what I was hoping for as well. CLI shouldn't be a private package. Thanks for having a look!

AviVahl:
Just to clarify, the broader intention was to move the webpack-cli to a `cli` folder inside packages, with its own (non-private) package.json.

In a `yarn` workspace project, the root package.json is just a common, private, one.

This is just not a yarn workspace. Its a `lerna bootstrap`ed one.

ev1stensberg:
CC-ing @sendilkumarn , he knows more about this than me. And yes, if the package remains `webpack-cli`, I'm in favor.

iMobs:
Alternatively, `"npmClient": "yarn"` and `"useWorkspaces": true` could be set in `lerna.json` and `"private": true` set in `package.json` to leverage yarn workspaces hoisting node modules.

jmckenney:
> Alternatively, `"npmClient": "yarn"` and `"useWorkspaces": true` could be set in `lerna.json` and `"private": true` set in `package.json` to leverage yarn workspaces hoisting node modules.

I wish this was stackoverflow and we could vote this to the top.

PR #481, opened 2018-06-02
fix: Remove workspaces key, as yarn workspaces is not in use

AviVahl:
Removes the `"workspaces"` key from the main package.json 

Nope, as no functionality has changed

Not relevant.

**Summary**
this monorepo uses `lerna bootstrap` with `npm` as a dependency manager, so the "workspaces" key is not used.

fixes https://github.com/webpack/webpack-cli/issues/479

No.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,9 +21,6 @@
     "module",
     "bundler"
   ],
-  "workspaces": [
-    "packages/*"
-  ],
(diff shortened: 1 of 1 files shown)


sendilkumarn:
Thanks for sending this in 🎉  @AviVahl

Issue #493, opened 2018-06-04
webpack-cli info does not work anymore (pending promise)


DanielRuf:
**Describe the bug**

`webpack-cli info` does not resolve the promise

**What is the current behavior?**

```
webpack-cli info
Promise { <pending> }
```

PR #494, opened 2018-06-04
fix: use async / await for info command

DanielRuf:
This PR fixes https://github.com/webpack/webpack-cli/issues/493

No

**Summary**
Currently `webpack-cli info` fails as it returns a promise which is not directly resolved.

Probably requires a Node.js version which supports async / await.

**Other information**

### DIFF
--- a/packages/info/index.js
+++ b/packages/info/index.js
@@ -6,9 +6,9 @@ const envinfo = require("envinfo");
  * Prints debugging information for webpack issue reporting
  */
 
-module.exports = function info() {
+module.exports = async function info() {
 	console.log(
-		envinfo.run({
+		await envinfo.run({
 			System: ["OS", "CPU"],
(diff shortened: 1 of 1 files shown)


DanielRuf:
> Async functions are not supported yet on Node 6.0.0. (node/no-unsupported-features)
9
> module.exports = async function info() {
> Async functions are not supported yet on Node 6.0.0. (node/no-unsupported-features)
11
> await envinfo.run({

So just 8+ is supported as I already thought.

tabrindle:
🤦‍♂️ sorry about this.

DanielRuf:
> 🤦‍♂️ sorry about this.

Why are you sorry and for what?

tabrindle:
I thought I had gone back and made PRs to upgrade everyone using older versions of envinfo already. Turns out I hadn't!

PR #499, opened 2018-06-08
ast(parser): fix bug with literals

dhruvdutt:
Bugfix

Yes

**Summary**
In case of generation of literal nodes, the final string value inside webpack.config.js doesn't yield quotes. This fixes the bug.

Here's a small demo of adding `mode` before and after the fix:
image

No

**Other information**

### DIFF
--- a/packages/utils/ast-utils.js
+++ b/packages/utils/ast-utils.js
@@ -432,6 +432,7 @@ function addProperty(j, p, key, value, action) {
 		});
 		valForNode = objectExp;
 	} else {
+		value = `'${value}'`;
 		valForNode = createIdentifierOrLiteral(j, value);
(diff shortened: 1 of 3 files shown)


packages/utils/ast-utils.js
```diff
@@ -432,6 +432,7 @@ function addProperty(j, p, key, value, action) {
 		});
 		valForNode = objectExp;
 	} else {
+		value = `'${value}'`;
```
ev1stensberg: this is incorrect. The reason why your tests are failing on remove is because you haven't double-quoted your strings.

ev1stensberg:
Add generator needs to be the one to use double quotes, not ast transform.

PR #500, opened 2018-06-08
[recursive] ast(parser): remove

dhruvdutt:

**Summary**
Add support for remove command.

**Other information**


packages/utils/ast-utils.js
```diff
@@ -436,6 +436,7 @@ function addProperty(j, p, key, value, action) {
 		});
 		valForNode = objectExp;
 	} else {
+		value = `'${value}'`;
```
dhruvdutt: Some of the code like this is a part of PR #499. I'll rebase and cleanup once it gets merged.

ev1stensberg: do it in this PR. Your code is obviously wrong when you see that you've changed snapshots signature.

packages/utils/scaffold.js
```diff
@@ -38,7 +38,7 @@ module.exports = function runTransform(webpackProperties, action) {
 	const webpackConfig = Object.keys(webpackProperties).filter(p => {
 		return p !== "configFile" && p !== "configPath";
 	});
-	const initActionNotDefined = action && action !== "init" ? true : false;
+	const addActionDefined = action && action === "add";
```
ev1stensberg: this needs a fallback value and
it will break the init feature as well.

packages/utils/scaffold.js
```diff
@@ -50,7 +50,7 @@ module.exports = function runTransform(webpackProperties, action) {
 			transformations.push("merge");
 		}
 		const ast = j(
-			initActionNotDefined
+			addActionDefined
```
ev1stensberg: Why are you changing this btw?

packages/utils/scaffold.js
dhruvdutt: Added the fallback value. Actually, we can get away without fallback value as `action` won't be undefined ever. If it's not equal to "add", it `addActionDefined` would false. We will only need fallback value if `action` is undefined. 
Also, I don't see how `init` will break. It will continue to work.

packages/utils/scaffold.js
dhruvdutt: Because in cases other than add (init and remove), we're starting from "module.exports = {}" and not reading existing config. We're regenerating the whole config file by using `addProperty`. The generator would directly clear the things to remove and the newly generated config file would be created based on this it.

packages/utils/ast-utils.js
dhruvdutt: If we want to handle this in add generator, it's easy as we can do at when we generate/assign literal string but if we think about remove generator which is essentially copying the whole config then it won't be possible to find and reassign every string literal in the webpack config object we read from user file.

ev1stensberg: Move this to generator, AST should not be concerned with this.

packages/generators/remove-generator.js
```diff
@@ -1,3 +1,119 @@
 const Generator = require("yeoman-generator");
+const path = require("path");
```
ev1stensberg: `const { resolve } = require('path');`

packages/generators/remove-generator.js
```diff
@@ -1,3 +1,119 @@
...
+			}
+		};
+
+		let configPath = path.resolve(process.cwd(), "webpack.config.js");
+		const webpackConfigExists = fs.existsSync(configPath);
+		if (!webpackConfigExists) {
+			configPath = null;
+			// end the generator stating webpack config not found or to specify the config
+		}
+		this.webpackOptions = require(configPath);
```
ev1stensberg: This is neat, but it will throw on packages that uses plugins. Maybe a JSON serialize would be better. If you did the same approach with `add`, the strings from the config needs to be padded in order for the ast to construct correctly. 

It's unclear what the comment is pointing to, what is it suppose to comment?

packages/generators/remove-generator.js
```diff
@@ -1,3 +1,119 @@
...
+		this.configuration.config.webpackOptions = this.webpackOptions;
+	}
+
+	getPropTypes() {
+		return Object.keys(this.webpackOptions);
+	}
+
+	getModuleLoaders() {
+		if (this.webpackOptions.module && this.webpackOptions.module.rules) {
+			return this.webpackOptions.module.rules.map(rule => rule ? rule.loader : null);
```
ev1stensberg: Add a filter here cause some props might be null and therefore we shouldn't add more than needed to the yeoman file. Don't you think?

packages/generators/remove-generator.js
```diff
@@ -1,3 +1,119 @@
...
+		return this.prompt([
+			List(
+				"propType",
+				"Which property do you want to remove?",
+				Array.from(this.getPropTypes())
+			)
+		])
+			.then(({ propType }) => {
+				if (!PROP_TYPES.has(propType)) {
+					console.log("Invalid webpack config prop");
```
ev1stensberg: This is weak logging, provide a more descriptive message

packages/generators/remove-generator.js
```diff
@@ -1,3 +1,119 @@
...
+									])
+										.then(({ loader }) => {
+											this.webpackOptions.module.rules =
+											this.webpackOptions.module.rules
+												.filter((rule) => rule.loader !== loader);
+										});
+								} else {
+									// remove the complete prop object if there is only one key
+									if (Object.keys(this.webpackOptions[propType]).length <= 1) {
+										delete this.webpackOptions[propType];
```
ev1stensberg: Lucky that we aren't in strict mode :-P

packages/utils/__snapshots__/ast-utils.test.js.snap
```diff
@@ -12,29 +12,29 @@ exports[`utils addProperty add entry property using add 1`] = `
                   }],
 
                   man: () => duper,
-                  objects: are,
+                  objects: 'are',
```
ev1stensberg: See what I mentioned about the string stuff. This is invalid syntax for the ast construction.

packages/utils/recursive-parser.js
```diff
@@ -39,6 +39,8 @@ module.exports = function recursiveTransform(j, ast, key, value, action) {
 			if (action === "add") {
 				// update property/replace
```
ev1stensberg: should have been `update/replace` property

dhruvdutt:
Closed in favour of #461.

PR #501, opened 2018-06-09
cli(add): support for webpack config file

### LINKED ISSUE #279: Specify config file with webpack-cli add
Feature

We may have webpack config files with different base names. However, commands except for `init`, such as `add`, always takes `webpack.config.js` as the processed file. It will be better if we can specify the path to config file for these commands.    
Affected commands: `add`(current version), `update`, `remove`(future)

Take command `add` as an example:
```
$ webpack-cli add <webpack_config_file>
```
To add components to the specified config file.
This usage is similar to existing command `migrate`.

**Please mention other relevant information such as the browser version, Node.js version, Operating System and programming language.**

dhruvdutt:
Feature

No

Yes

**Summary**
Add support to specify webpack config file with args, default to webpack.config.js

- Specifying config file: `webpack-cli add webpack.dev.js`

image

- Default config file: `webpack-cli add`

image

- Non-existential file: `webpack-cli add aaahha.js`

image

Similar usage for `update`, `remove`.

No

Fixes #279


packages/add/index.js
```diff
@@ -11,6 +11,13 @@ const modifyHelper = require("@webpack-cli/utils/modify-config-helper");
  *
  */
 
-module.exports = function add() {
-	return modifyHelper("add", defaultGenerator);
+module.exports = function add(...args) {
+	const DEFAULT_WEBPACK_CONFIG_FILENAME = "webpack.config.js";
+
+	const filePaths = args.slice(3);
```
ev1stensberg: Why are you slicing 3 here?

packages/utils/resolve-packages.js
```diff
@@ -3,7 +3,7 @@
 const path = require("path");
 const chalk = require("chalk");
 
-const modifyConfigHelper = require("./modify-config-helper");
+const modifyConfig = require("./modify-config-helper");
```
ev1stensberg: don't see a reason to change this

dhruvdutt: For consistency.
We're using `modifyConfig` everywhere across at codebase other than this place.

Please check this: https://github.com/webpack/webpack-cli/search?q=return+modifyHelper&unscoped_q=return+modifyHelper

packages/add/index.js
dhruvdutt: Because first 3 args would be node path, webpack-cli path, add command.
4th item in the array would be everything after add command like the custom config path/name. So, we're slicing first three items to get the config path.

We can't directly use the last element of `args` because that would be "add" if the user doesn't mention any arg after add.

packages/utils/resolve-packages.js
ev1stensberg: Use the previous convention 👍

packages/add/index.js
ev1stensberg: Oki. let me check it out later to make sure its working 👍

packages/utils/resolve-packages.js
dhruvdutt: Updated all occurrences to use the previous convention.

PR #502, opened 2018-06-12
cli(prompts): Add autocomplete support for config props

dhruvdutt:
Feature

**Summary**
peek 2018-06-12 20-07

**Other information**

### DIFF
--- a/packages/generators/add-generator.js
+++ b/packages/generators/add-generator.js
@@ -2,10 +2,7 @@ const Generator = require("yeoman-generator");
 const glob = require("glob-all");
 const path = require("path");
 const inquirerAutoComplete = require("inquirer-autocomplete-prompt");
-const Confirm = require("@webpack-cli/webpack-scaffold").Confirm;
-const List = require("@webpack-cli/webpack-scaffold").List;
-const Input = require("@webpack-cli/webpack-scaffold").Input;
-const AutoComplete = require("@webpack-cli/webpack-scaffold").AutoComplete;
+const { AutoComplete, Confirm, Input, List } = require("@webpack-cli/webpack-scaffold");
 
 const webpackSchema = require("./utils/optionsSchema.json");
 const webpackDevServerSchema = require("webpack-dev-server/lib/optionsSchema.json");
@@ -66,13 +63,11 @@ const traverseAndGetProperties = (arr, prop) => {
  */
 const searchProps = (answers, input) => {
 	input = input || "";
-	return new Promise(resolve => {
-		resolve(
-			PROPS.filter(food =>
-				food.toLowerCase().includes(input.toLowerCase())
-			)
-		);
-	});
+	return new Promise.resolve(
+		PROPS.filter(food =>
+			food.toLowerCase().includes(input.toLowerCase())
+		)
+	);
 };
 
 /**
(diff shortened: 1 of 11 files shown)


packages/generators/add-generator.js
```diff
@@ -1,9 +1,11 @@
 const Generator = require("yeoman-generator");
 const glob = require("glob-all");
 const path = require("path");
+const inquirerAutoComplete = require("inquirer-autocomplete-prompt");
 const Confirm = require("@webpack-cli/webpack-scaffold").Confirm;
 const List = require("@webpack-cli/webpack-scaffold").List;
 const Input = require("@webpack-cli/webpack-scaffold").Input;
+const AutoComplete = require("@webpack-cli/webpack-scaffold").AutoComplete;
```
ematipico: Trivial: at this point we could refactor with destructoring 
```js
const { Confirm, List, Input, AutoComplete }= require("@webpack-cli/webpack-scaffold")

File: packages/generators/add-generator.js [hunk 2]
```diff
@@ -14,6 +16,8 @@ const getPackageManager = require("@webpack-cli/utils/package-manager")
 const npmExists = require("@webpack-cli/utils/npm-exists");
 const entryQuestions = require("./utils/entry");
 
+const PROPS = Array.from(PROP_TYPES.keys());
```
ematipico: `.keys` already returns an array. Why did you use `Array.from`?

packages/generators/add-generator.js
```diff
@@ -50,6 +54,27 @@ const traverseAndGetProperties = (arr, prop) => {
...
+ *
+ * @param {Object} answers	Prompt answers object
+ * @param {String} input	Input search string
+ *
+ * @returns {Promise} Returns promise which resolves to filtered props
+ *
+ */
+const searchProps = (answers, input) => {
+	input = input || "";
+	return new Promise(resolve => {
```
ematipico: Just a suggestion: 

```js
return Promise.resolve(
  PROPS.filter(food =>
    food.toLowerCase().includes(input.toLowerCase())
  )
)

File: packages/generators/add-generator.js (hunk 2 above)
Comment (dhruvdutt) [reply packages/generators/add-generator.js]: When using a [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set), it returns an Iterator instead of array. Hence, we have to use `Array.from` method.

File: packages/generators/add-generator.js [hunk 4]
```diff
@@ -66,13 +63,11 @@ const traverseAndGetProperties = (arr, prop) => {
...
 const searchProps = (answers, input) => {
 	input = input || "";
-	return new Promise(resolve => {
-		resolve(
-			PROPS.filter(food =>
-				food.toLowerCase().includes(input.toLowerCase())
-			)
-		);
-	});
+	return new Promise.resolve(
```
ematipico: Without `new`

packages/generators/add-generator.js
ematipico: Aaah it was a `Set`! Fair enough, thanks!

packages/generators/add-generator.js
dhruvdutt: Fixed. Thanks!

dhruvdutt:
Yes. I'll start with TS. :tada:

PR #510, opened 2018-06-14
chore(greenkeeper): support for lockfiles

### LINKED ISSUE #509: Bug: Greenkeeper should update package.lock
Currently, greenkeeper does not push a generated package.lock when submitting a PR. It would be nice if it did so we don't have to worry about that.

dhruvdutt:

**Summary**
Greenkeeper supports lockfile updates starting from version 2.
[Ref](https://blog.greenkeeper.io/greenkeeper-support-for-lockfiles-9d9c282efcbe).

**Other information**
Fixes #509

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,7 @@ install:
   - npm ci
   - npm install -g codecov
   - npm install -g eslint
-  - npm install -g greenkeeper-lockfile@1
+  - npm install -g greenkeeper-lockfile@2
   - lerna bootstrap
 
 before_script: greenkeeper-lockfile-update
@@ -62,6 +62,6 @@ jobs:
     - stage: NPM release
       if: branch = master
       node_js: "lts/*"
-      script: 
+      script:
         - npm run changelog
         - npm run travis-deploy-once "npm run semantic-release"


xPaw:
Why was this single PR released as a separate version? I find it rather annoying that every small change is triggering it's own release now, causing bots like greenkeeper or renovate to make new PRs with senseless releases.

ev1stensberg:
Our intention is to automate releases. You can safely use any version without expecting any changes (breaking ones left out). How does this affect you?

xPaw:
Automated releases for every commit somewhat defeats the purposes of having releases.

It affects me because it's unnecessary noise in my project's pull requests when the renovate bot has to make a new PR for every release this project gets.

astorije:
Note that this is me even more difficult to follow as the release notes (that greenkeeper and renovate pick up) are always empty, so every new release requires more research than most npm packages out there.

ev1stensberg:
Based on research I thought semantic release did that by default without having to use a plugin. Would it make your situation better if we added changelog on publish, or are you all swaying for removing it?

astorije:
> Based on research I thought semantic release did that by default without having to use a plugin.

I'm not sure what you mean by this.
This is what a report from Renovate for the v3.0.7 release looks like, for instance: https://github.com/thelounge/thelounge/pull/2559

astorije:
GitHub release notes would help (I don't know if Renovate extracts from changelogs, I doubt it), but I agree with @xPaw that releasing for things like CI configuration generates a lot of noise across the thousands of projects that consume `webpack-cli`, especially since the Travis configuration is ignored in the `.npmignore` file.
This means that https://unpkg.com/webpack-cli@3.0.6/ and https://unpkg.com/webpack-cli@3.0.7/ are exactly equivalent.

One thing you could do to improve automation is only release if files that are not filtered out by `.npmignore` have changed.

ev1stensberg:
ok, thanks for the elaboration.

travi:
So that you don't hear only one opinion, i want to say thanks for releasing often with semantic release. Putting the decision of when to update on the consumer is a far better situation than queueing up a bunch of unreleased changes in master and making us wait for them to be released.

ev1stensberg:
I see your point. I'm in favor for having it. Thinking about fast publishing and getting new versions out there is important, but so is user experience. We want the tool to be up-to-date but still think about user interactions. If it is a lot of work for consumers to update their dependencies, it creates a lot of noise and it makes them loose productivity, then semantic is not the right tool. 

As far as I'm concerned, it's a big productivity for us, but people using the CLI are using more time to keep deps updated. The CLI doesn't come with "Important" changes that has to be updated, so it might be better to do a manual publish on webpack-cli (in my opinion)

travi:
one thing to consider might be how you've configured semantic-release. it looks like this commit was marked as `chore`. by default semantic-release does not publish a new version for `chore`. if you don't want that type of change, you could still keep publishes automated, but adjust your semantic-release config to only publish user-facing changes

PR #511, opened 2018-06-14
feat(typescript): setup base infra

dhruvdutt:
**What kind of change does this PR introduce?**
Feature

**Did you add tests for your changes?**

**If relevant, did you update the documentation?**

**Summary**
Add initial TypeScript support for `@webpack-cli/init` package.
Use `npm run build` and `npm run watch` for dev.

**Does this PR introduce a breaking change?**

**Other information**
#245

PR #518, opened 2018-06-17
feat(init): add typescript support

dhruvdutt:

**Summary**

**Other information**
Done on top of #511


packages/init/package.json
```diff
@@ -14,5 +14,14 @@
...
     "jscodeshift": "^0.5.0",
     "p-each-series": "^1.0.0"
+  },
+  "devDependencies": {
+    "@types/node": "^10.3.3",
+    "@types/p-each-series": "^1.0.0",
+    "typescript": "^2.9.2"
+  },
+  "scripts": {
+    "build": "tsc",
```
ev1stensberg: This should go in webpack-cli `package.json`

dhruvdutt: Updated.

dhruvdutt: Actually, this package is not recognizing `@types/node` if I run `tsc` from root. Still trying to trace what is the issue.

dhruvdutt: For temp. testing purpose please run `tsc` or `npx tsc` from `packages/init` path.

ev1stensberg: No temp fixes on this :) Resolve the issue rather than creating hacky solutions that might break later

packages/init/init.ts
```diff
@@ -16,12 +16,12 @@ const propTypes = require("@webpack-cli/utils/prop-types");
...
  * @param	{Object} transformObject 	- An Object with all transformations
  * @param	{Object} config 			- Configuration to transform
- * @returns {Object} - An Object with the transformations to be run
+ * @returns {Array} - An array with the transformations to be run
  */
 
-function mapOptionsToTransform(config) {
-	return Object.keys(config.webpackOptions).filter(k => propTypes.has(k));
-}
+const mapOptionsToTransform = (config): string[] =>
```
ematipico: Why it's been changed into a `const`?

dhruvdutt: No particular reason. We can decide on some conversion and follow same all across the code.

ematipico: Sounds good to me! Always function unless there are some scope related things (`this.`)?

ematipico: Why?

ev1stensberg: Is it mutated? I recall not

ematipico: What?

ematipico: No the function is pure

ev1stensberg: Then const is good :)

ematipico: Are you basically saying that pure functions should be written with `const puppy = () =>{}` and impure functions should be written with `function puppy() {}`?

ev1stensberg: If the car is modified after declaration then using let is better.

ematipico: OK unless there's a misunderstanding, I just give up

dhruvdutt: Hey, guys. I think there is some misunderstanding. :)
Reassigning variables and purity of functions are different concepts. Reassignment is about mutability and purity is about input (the arguments) and output (return value) of a function.

ematipico: I know, my point was why `function ()` was changed to `const pp = () => {}`. If so, let's just agree on a convention on when having one or the other. I don't care if it's `const` or `let` :no_good_man: . But that's OK, no need to go further

packages/init/init.ts
```diff
@@ -0,0 +1,94 @@
...
+ *
+ * Runs the transformations from an object we get from yeoman
+ *
+ * @param	{Object} webpackProperties 	- Configuration to transform
+ * @param	{String} action 			- Action to be done on the given ast
+ * @returns {Promise} - A promise that writes each transform, runs prettier
+ * and writes the file
+ */
+
+export default function runTransform(webpackProperties, action: string): any {
```
ev1stensberg: We should make this an interface

ematipico: Nice yes, but it has to become a class in order to create an interface. Or you talking of `webpackProperties`?

ev1stensberg:
It will be troublesome to maintain.

ematipico:
That's how you have to work with a monorepo, it's the design. And now that we are implementing typescript, there are more reasons to follow a consistent architectural design

ev1stensberg:
Would be better for us to be able to compile all packages at once. We can create a bash script for it. It's not much work. cd to each package and run npm install + compile then publish

dhruvdutt:
Hey guys, I think @ematipico is right. It's always a good practice for each mono repos to have it's independent build and compiler but in our case, we need individual build setup along with a root level. The current setup should also work from root level. I've fixed the root compilation step. PTAL. :) The PR looks good to go.

ematipico:
> I disagree. A utility script is not hard to implement and I guess this is a matter of being lazy or not ;)

It's not a matter of being lazy, it's a matter of how the tool that has been chose work.

> We can create a bash script for it. It's not much work. cd to each package and run npm install + compile then publish

We can do it in every way we want, this is the not problem.

The matter is not how you kick off the build, but how to **configure the different packages with `typescript`**
What it is not possible to do it to have just one `tsconfig.json` at root level and from there build all the packages. That's it. Unless there's a magical way (please find one, I haven't find it), it's not possible with just one configuration and starting from the root folder. `typescript` cannot find all the packages. Have a look at polymer; all their packages have their own configuration and their own build step.

This is a monorepo and it has to be treated as it is.

ev1stensberg:
Also check out lighthouse

ematipico:
Done but internal packages are not built with typescript, as far as I could see.

After deprecation and porting of the internal commands (add, remove, etc) they will continue to reside on different packages and they will have to have their own build process, regardless. That's one more reason to keep separated build steps.

dhruvdutt:
@ematipico Can you please pull this locally and review if it works? It should now work from the root as well as package level. I've configured it such that both should work fine so even if in future when we move the packages, their independent build command should work fine. :)

ev1stensberg:
webpack CLI is not going to be depreciated but we're merging command logic with webpack command. I think we should make a script to compile all packages.

ematipico:
I tried to explain that the problem is not how you do the build process. It's how you set-up each package to communicate to each other. 

Say there's a bug on the `add` package. You'd rather build ALL the packages instead of build just a single one? Please list the benefits of your point of view compared to the other.

ev1stensberg:
You can chain scripts, it's easier for us to publish packages and to symlink them.

ematipico:
> You can chain scripts, it's easier for us to publish packages and to symlink them.

Again, how you make the build of the packages it's not a problem and it's not **the** problem that I am highlighting here.

I am saying that packages have to communicate by themselves as packages. And also the consuming of those has to happen as packages. That's the best design on the long run. I don't accept this kind of approach or design just because of it would troublesome to maintain: could you please give _us_ some _reason_ why it would be _troublesome_?

ev1stensberg:
@ematipico Are you willing to manually go to a directory and compile that package on tests and publishing? Lerna has a bootstrap command which symlinks our dependencies, and I don't see the reason for making it harder to publish and to generate a release. FWIW, what I'm saying is what we already are doing with jest. Jest is a dev-dependency at root and will go to each pacakge and run the `test` script. It's not any different.

Since there are compat issues with typescript compiling sub-folders, we'll need to make a script for building each pkg as far as I've understood and tested.

I don't see your point on compiling every package at package-level. The commands are the same and the packages are released all at once, and the generated files will be at their correct paths. When commands and structure are the same, its a good sign for abstraction, which is this case.

> I am saying that packages have to communicate between themselves as packages. And also the consuming of those has to happen as packages

I don't see this being relative to why I want to have a utility script, it's not relevant. Packages will still work as before.

> I don't accept this kind of approach or design just because of it would troublesome to maintain: could you please give us some reason why it would be troublesome?

The reason for using lerna is for better maintainability. I will accept less tech-debt and automation over manually doing things at any time when it doesn't have any effect on the end user.

ematipico:
Probably I am the one that is not able to communicate. So I will just show the following:

screen shot 2018-06-23 at 20 53 40

This is *what* I don't want to see inside a package. Does it make sense?

ev1stensberg:
If it's outside the scope of the package it should require from the scope package indeed

package.json
```diff
@@ -100,6 +100,7 @@
     "inquirer": "^6.0.0",
     "interpret": "^1.1.0",
     "loader-utils": "^1.1.0",
+    "p-each-series": "^1.0.0",
```
ev1stensberg: don't add this at root level

packages/init/init.ts
```diff
@@ -0,0 +1,96 @@
...
+					? webpackProperties.configPath
+					: path.join(process.cwd(), configurationName);
+
+				const source: any = ast.toSource({
+					quote: "single",
+				});
+
+				runPrettier(outputPath, source);
+			})
+			.catch((err) => {
```
ev1stensberg: Error/Object type

packages/generators/types/Yeoman.js
```diff
@@ -0,0 +1,22 @@
+// interface IRunEnv extends Object {
```
ev1stensberg: ????

packages/generators/types/Yeoman.ts
```diff
@@ -0,0 +1,25 @@
+// interface IRunEnv extends Object {
+// 	on: (event: string, callbackFn: Function) => void;
```
ev1stensberg: ???

PR #519, opened 2018-06-18
feat(migrate): add typescript support

dhruvdutt:

**Summary**

**Other information**

All tests are passing except this snapshot change:

image

It's happening with [LoaderOptionsPlugin](https://webpack.js.org/plugins/loader-options-plugin/) => `minimize` property.


packages/migrate/index.ts
```diff
@@ -152,19 +167,20 @@ function runMigration(currentConfigPath, outputConfigPath) {
...
-					runPrettier(outputConfigPath, result, err => {
+					runPrettier(outputConfigPath, result, (err: object) => {
 						if (err) {
 							throw err;
 						}
 					});
 
 					if (answer["confirmValidation"]) {
-						const webpackOptionsValidationErrors = validate(
+						const webpackOptionsValidationErrors: any = validate(
```
ev1stensberg: Use of any

packages/migrate/migrate.ts
```diff
@@ -0,0 +1,112 @@
...
+import resolveTransform from "./resolve/resolve";
+import removeJsonLoaderTransform from "./removeJsonLoader/removeJsonLoader";
+import uglifyJsPluginTransform from "./uglifyJsPlugin/uglifyJsPlugin";
+import loaderOptionsPluginTransform from "./loaderOptionsPlugin/loaderOptionsPlugin";
+import bannerPluginTransform from "./bannerPlugin/bannerPlugin";
+import extractTextPluginTransform from "./extractTextPlugin/extractTextPlugin";
+import noEmitOnErrorsPluginTransform from "./noEmitOnErrorsPlugin/noEmitOnErrorsPlugin";
+import removeDeprecatedPluginsTransform from "./removeDeprecatedPlugins/removeDeprecatedPlugins";
+
+interface ItransformsObject {
```
ev1stensberg: This is good stuff 😍

packages/migrate/migrate.ts
```diff
@@ -0,0 +1,112 @@
...
+	loadersTransform: object;
+	resolveTransform: object;
```
ev1stensberg: Actually these are modules, maybe either declare them as modules ( if possible )

packages/migrate/migrate.ts
```diff
@@ -0,0 +1,112 @@
...
+
+	const ast: any = jscodeshift(source);
+	const recastOptions: object = Object.assign(
+		{
+			quote: "single"
+		},
+		options
+	);
+	transforms =
+		transforms || Object.keys(transformations).map(k => transformations[k]);
```
ev1stensberg: missing here?

packages/migrate/migrate.ts
```diff
@@ -0,0 +1,112 @@
...
+
+	return pEachSeries(transforms, (f: Function) => f(ast, source))
+		.then(_ => {
```
ev1stensberg: missing?

packages/migrate/migrate.ts
```diff
@@ -0,0 +1,112 @@
...
+			return ast.toSource(recastOptions);
+		})
+		.catch(err => {
```
ev1stensberg: missing

packages/migrate/index.ts
dhruvdutt: I think we didn't decide if we can or cannot use `any`. Also, I think we shouldn't enforce types completely. Leaving types at some places implicitly means `any`.

packages/migrate/migrate.ts
ematipico: That was the pattern followed so far where if we didn't use anything from the resolved promise, it was decided to use `_ =>` instead of `() =>`.
Now that there's typescript, I am more against this pattern. What's `_`? What's his meaning? We put `any` just because we want to leave `_`? I think we should drop it

packages/migrate/migrate.ts
dhruvdutt: You mean [TS modules](https://www.typescriptlang.org/docs/handbook/modules.html)? Also, can you please show an example of what would be inside each module?

packages/migrate/migrate.ts
dhruvdutt: I also think `() =>` is better. Should I replicate this in all occurrences?

custom_typing/NodePath.ts
```diff
@@ -0,0 +1,32 @@
+interface NodeKey extends Node {
```
ematipico: I have always worked with prefixing the interfaces with capital `I`. What should we do here? I would follow the same pattern

packages/migrate/migrate.ts
ematipico: Let's see how @ev1stensberg feels about it

ematipico: `void` is when your function returns nothing. I am not sure if it's possible to assign it to a parameter... never seen it. Maybe typescript supports it

ev1stensberg: If void is supported, keep underscore, if not, remove it

dhruvdutt:
Hey guys, PTAL at the PR description snapshot failure. I tried looking into the PR diff but it doesn't seem to contain any behavioural changes. I also tried replacing the TS generated `loaderOptionsPlugins.js` with current master => [`loaderOptionsPlugins.js`](https://github.com/webpack/webpack-cli/blob/master/packages/migrate/loaderOptionsPlugin/loaderOptionsPlugin.js) but it results in same issue.

ev1stensberg:
`optimization.minimize` should only be used when uglifyJSPlugin is present, in this case, it should not move the debug prop there. You'll need to fix that

packages/migrate/index.ts
```diff
@@ -74,129 +91,141 @@ module.exports = function migrate(...args) {
...
-					runPrettier(outputConfigPath, result, err => {
+					runPrettier(outputConfigPath, result, (err: object) => {
 						if (err) {
 							throw err;
 						}
 					});
 
-					if (answer["confirmValidation"]) {
+					if (answer.confirmValidation) {
 						const webpackOptionsValidationErrors = validate(
```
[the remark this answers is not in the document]
dhruvdutt: That's actually a default tslint rule. The reason being TS cannot check if properties exist on an object (in this case `answer`) if we use `["property"]` convention.

For instance:
```js
var answer: object; // declaration
answer["confirm"];  // usage
```
Above syntax would pass the TS compiler, but using `answer.confirm` won't pass compiler unless we define type for `answer` object.

Hence, it's good to follow this convention. I've followed this convention everywhere across all packages.

packages/migrate/index.ts
```diff
@@ -74,129 +91,141 @@ module.exports = function migrate(...args) {
...
-							"✔︎ Plugin and loader dependencies may need to be updated."
-						)
+							"✔︎ Plugin and loader dependencies may need to be updated.",
+						),
 					);
 				});
 		})
-		.catch(err => {
-			const errMsg = "\n ✖ ︎Migration aborted due to some errors: \n";
+		.catch((err: object) => {
```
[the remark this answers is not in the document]
dhruvdutt: It doesn't return anything. Shouldn't it be `void`? Specifying any other type other than `void` or `any`

packages/migrate/namedModulesPlugin/namedModulesPlugin.ts
```diff
@@ -12,19 +14,20 @@ const {
...
  */
-module.exports = function(j, ast) {
+export default function(j: IJSCodeshift, ast: INode): INode {
 	// Remove old plugin
-	const root = findAndRemovePluginByName(j, ast, "webpack.NamedModulesPlugin");
+	const root: INode = findAndRemovePluginByName(j, ast, "webpack.NamedModulesPlugin");
 
 	// Add new optimizations option
-	root &&
+	if (root) {
```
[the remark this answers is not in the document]
dhruvdutt: `tslint` finds this as unused expression and results in linting error. IMHO it is a confusing syntax as well. It's better to be slighly imperative in such cases and define _if_ block.

image

We are returning `ast` and nothing from this statement. If we were returning from same statement it wouldn't have been an issue in first place.

packages/migrate/index.ts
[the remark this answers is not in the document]
ev1stensberg: oki 👍

packages/migrate/index.ts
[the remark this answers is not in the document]
ev1stensberg: that's weird, I assume promises `then` and `catch` will yield in a promise result. Go for void

packages/migrate/bannerPlugin/bannerPlugin.ts
```diff
@@ -10,11 +12,11 @@ const utils = require("@webpack-cli/utils/ast-utils");
  * @returns {Node} ast - jscodeshift ast
  */
 
-module.exports = function(j, ast) {
+export default function(j: IJSCodeshift, ast: INode): INode {
 	return utils
 		.findPluginsByName(j, ast, ["webpack.BannerPlugin"])
-		.forEach(path => {
-			const args = path.value.arguments; // any node
+		.forEach((path: INode) => {
```
ev1stensberg: returns void

packages/migrate/index.ts
```diff
@@ -1,18 +1,31 @@
...
+		validate: Function;
+		WebpackOptionsValidationError: any;
+	};
+	stdout: {
+		write: Function;
+	};
+	exitCode: number;
+};
 
 const { validate } = process.webpackModule;
```
ev1stensberg: this needs to change to using `require('webpack'), bad rebase?

packages/migrate/index.ts
```diff
@@ -74,129 +91,141 @@ module.exports = function migrate(...args) {
...
 		},
 		{
-			title: "Migrating config to newest version",
-			task: ctx => {
-				const transformations = require("./migrate").transformations;
+			task: (ctx: INode): string | void | Listr | Promise<any> => {
 				return new Listr(
-					Object.keys(transformations).map(key => {
-						const transform = transformations[key];
+					Object.keys(transformations).map((key: string) => {
```
ev1stensberg: missing return type on obj.keys?

packages/migrate/index.ts
```diff
@@ -74,129 +91,141 @@ module.exports = function migrate(...args) {
...
 						]);
 					} else {
 						console.log(chalk.red("✖ Migration aborted"));
 					}
 				})
-				.then(answer => {
-					if (!answer) return;
+				.then((answer: {
+					confirmValidation: boolean;
+				}) => {
```
ev1stensberg: missing return on promise?

packages/migrate/loaders/loaders.ts
```diff
@@ -52,43 +54,43 @@ module.exports = function(j, ast) {
...
 	 */
 
-	const createArrayExpressionFromArray = function(path) {
-		const value = path.value;
+	const createArrayExpressionFromArray = (path: INode): INode => {
+		const value: INode = path.value;
 		// Find paths with `loaders` keys in the given Object
-		const paths = value.properties.filter(prop =>
-			prop.key.name.startsWith("loader")
+		const paths: INode[] = value.properties.filter((prop: INode) =>
```
ev1stensberg: missing return type?

packages/migrate/loaders/loaders.ts
```diff
@@ -52,43 +54,43 @@ module.exports = function(j, ast) {
...
-					})
+					}),
 				);
-				//	If the value is String of loaders like 'style!css'
+				// 	If the value is String of loaders like 'style!css'
 			} else if (pair.value.type === j.Literal.name) {
 				// Replace it with Array expression of loaders
-				const literalValue = pair.value;
+				const literalValue: INode = pair.value;
 				pair.value = j.arrayExpression(
```
ev1stensberg: missing return?

packages/migrate/loaders/loaders.ts
```diff
@@ -134,8 +136,8 @@ module.exports = function(j, ast) {
 	 * @returns {Boolean} hasLoaderQueryString - whether the loader object contains a query string
 	 */
 
-	const findLoaderWithQueryString = p => {
-		return p.value.properties.reduce((predicate, prop) => {
+	const findLoaderWithQueryString = (p: INode): boolean => {
+		return p.value.properties.reduce((predicate: boolean, prop: INode) => {
```
ev1stensberg: missing return?

packages/migrate/loaders/loaders.ts
```diff
@@ -171,19 +173,19 @@ module.exports = function(j, ast) {
 	 * @returns {Node} p - object expression with a `loaders` object and appropriate `enforce` properties
 	 */
 
-	const fitIntoLoaders = p => {
-		let loaders;
-		p.value.properties.map(prop => {
+	const fitIntoLoaders = (p: INode): INode => {
+		let loaders: INode = null;
+		p.value.properties.map((prop: INode) => {
```
ev1stensberg: missing return?

packages/migrate/loaders/loaders.ts
```diff
@@ -171,19 +173,19 @@ module.exports = function(j, ast) {
...
 			const keyName = prop.key.name;
 			if (keyName === "loaders") {
 				loaders = prop.value;
 			}
 		});
-		p.value.properties.map(prop => {
+		p.value.properties.map((prop: INode) => {
```
ev1stensberg: missing return?

packages/migrate/loaders/loaders.ts
```diff
@@ -171,19 +173,19 @@ module.exports = function(j, ast) {
...
 			const keyName = prop.key.name;
 			if (keyName !== "loaders") {
-				const enforceVal = keyName === "preLoaders" ? "pre" : "post";
-				prop.value.elements.map(elem => {
+				const enforceVal: string = keyName === "preLoaders" ? "pre" : "post";
+				prop.value.elements.map((elem: INode) => {
```
ev1stensberg: missing return?

packages/migrate/loaders/loaders.ts
```diff
@@ -207,10 +209,10 @@ module.exports = function(j, ast) {
 	 * @returns {Node} ast - jscodeshift ast
 	 */
 
-	const prepostLoaders = () =>
+	const prepostLoaders = (_?: void): INode =>
 		ast
 			.find(j.ObjectExpression)
-			.filter(p => utils.findObjWithOneOfKeys(p, ["preLoaders", "postLoaders"]))
+			.filter((p: INode) => utils.findObjWithOneOfKeys(p, ["preLoaders", "postLoaders"]))
```
ev1stensberg: missing return?

packages/migrate/loaders/loaders.ts
```diff
@@ -219,31 +221,31 @@ module.exports = function(j, ast) {
 	 * @returns {Node} ast - jscodeshift ast
 	 */
 
-	const loadersToRules = () =>
+	const loadersToRules = (_?: void): INode =>
 		ast
 			.find(j.Identifier)
 			.filter(checkForLoader)
-			.forEach(p => (p.value.name = "rules"));
+			.forEach((p: INode) => (p.value.name = "rules"));
```
ev1stensberg: returns void

packages/migrate/loaders/loaders.ts
```diff
@@ -219,31 +221,31 @@ module.exports = function(j, ast) {
...
 
-	const loadersToArrayExpression = () =>
+	const loadersToArrayExpression = (_?: void): INode | void =>
 		ast
 			.find(j.ObjectExpression)
-			.filter(path => utils.findObjWithOneOfKeys(path, ["loader", "loaders"]))
+			.filter((path: INode) => utils.findObjWithOneOfKeys(path, ["loader", "loaders"]))
 			.filter(
-				path =>
+				(path: INode) =>
```
ev1stensberg: missing return?

packages/migrate/loaders/loaders.ts
```diff
@@ -219,31 +221,31 @@ module.exports = function(j, ast) {
...
 	 *
 	 * @returns {Node} ast - jscodeshift ast
 	 */
 
-	const loadersToArrayExpression = () =>
+	const loadersToArrayExpression = (_?: void): INode | void =>
 		ast
 			.find(j.ObjectExpression)
-			.filter(path => utils.findObjWithOneOfKeys(path, ["loader", "loaders"]))
+			.filter((path: INode) => utils.findObjWithOneOfKeys(path, ["loader", "loaders"]))
```
ev1stensberg: missing return?

packages/migrate/loaders/loaders.ts
```diff
@@ -219,31 +221,31 @@ module.exports = function(j, ast) {
...
 						"parent",
 						"parent",
 						"node",
 						"key",
-						"name"
-					]) === "rules"
+						"name",
+					]) === "rules",
 			)
 			.forEach(createArrayExpressionFromArray);
```
ev1stensberg: would be nice to have type check on this if possible

packages/migrate/loaders/loaders.ts
```diff
@@ -293,10 +295,10 @@ module.exports = function(j, ast) {
 	 * @returns {Node} ast - jscodeshift ast
 	 */
 
-	const loaderWithQueryProp = () =>
+	const loaderWithQueryProp = (_?: void): INode =>
 		ast
 			.find(j.Identifier)
-			.filter(p => p.value.name === "query")
+			.filter((p: INode): boolean => p.value.name === "query")
```
ev1stensberg: ret missin

packages/migrate/loaders/loaders.ts
```diff
@@ -327,21 +329,21 @@ module.exports = function(j, ast) {
...
+			const keyName: string = prop.key.name;
 			if (keyName === "options") {
 				options = prop;
 			}
 		});
 
 		if (options) {
 			p.value.properties = p.value.properties.filter(
-				prop => prop.key.name !== "options"
+				(prop: INode) => prop.key.name !== "options",
```
ev1stensberg: missing return?

packages/migrate/migrate.ts
```diff
@@ -0,0 +1,116 @@
...
+export const transformSingleAST = (
+	ast: object,
+	source: string,
+	transformFunction: (jscodeshift: object, ast: object, source: string) => object)
+	: pLazy<{}> => {
+
+	return new pLazy((
+		resolve: (value?: {} | PromiseLike<{}>) => void,
+		reject: (reason?: object) => void,
+	) => {
```
ev1stensberg: plazy needs a return type

packages/migrate/migrate.ts
```diff
@@ -0,0 +1,116 @@
...
+		options,
+	);
+	transforms =
+		transforms || Object.keys(transformations).map((k: string) => transformations[k]);
+
+	return pEachSeries(transforms, (f: Function) => f(ast, source))
+		.then((value: Function[]): string | PromiseLike<string> => {
+			return ast.toSource(recastOptions);
+		})
+		.catch((err: object) => {
```
ev1stensberg: missing ret type

packages/migrate/namedModulesPlugin/namedModulesPlugin.ts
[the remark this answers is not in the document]
ev1stensberg: ok

packages/migrate/outputPath/outputPath.ts
```diff
@@ -9,66 +11,69 @@ const utils = require("@webpack-cli/utils/ast-utils");
...
 
-module.exports = function(j, ast) {
-	const literalOutputPath = ast
+export default function(j: IJSCodeshift, ast: INode): INode | void {
+
+	const literalOutputPath: INode = ast
 		.find(j.ObjectExpression)
 		.filter(
-			p =>
+			(p: INode) =>
```
ev1stensberg: missing return?

packages/migrate/outputPath/outputPath.ts
```diff
@@ -9,66 +11,69 @@ const utils = require("@webpack-cli/utils/ast-utils");
...
 				utils.safeTraverse(p, ["parentPath", "value", "key", "name"]) ===
-				"output"
+				"output",
 		)
 		.find(j.Property)
 		.filter(
-			p =>
+			(p: INode) =>
```
ev1stensberg: missing return?

packages/migrate/outputPath/outputPath.ts
```diff
@@ -9,66 +11,69 @@ const utils = require("@webpack-cli/utils/ast-utils");
...
 					utils.safeTraverse(p, ["value", "init", "callee", "name"]) ===
-					"require"
+					"require",
 			)
 			.filter(
-				p =>
+				(p: INode): boolean =>
 					utils.safeTraverse(p, ["value", "init", "arguments"]) &&
-					p.value.init.arguments.reduce((isPresent, a) => {
+					p.value.init.arguments.reduce((isPresent: boolean, a: INode) => {
```
ev1stensberg: missing return type?

packages/migrate/removeJsonLoader/removeJsonLoader.ts
```diff
@@ -9,24 +11,25 @@ const utils = require("@webpack-cli/utils/ast-utils");
...
 
-	function removeLoaderByName(path, name) {
-		const loadersNode = path.value.value;
+	function removeLoaderByName(path: INode, name: string): void {
+		const loadersNode: INode = path.value.value;
+
 		switch (loadersNode.type) {
 			case j.ArrayExpression.name: {
-				let loaders = loadersNode.elements.map(p => {
+				const loaders: string[] = loadersNode.elements.map((p: INode) => {
```
ev1stensberg: missing ret type for map

packages/migrate/resolve/resolve.ts
```diff
@@ -7,21 +9,23 @@
  * @returns {Node} ast - jscodeshift ast
  */
 
-module.exports = function transformer(j, ast) {
-	const getRootVal = p => {
-		return p.node.value.properties.filter(prop => prop.key.name === "root")[0];
+export default function transformer(j: IJSCodeshift, ast: INode): INode[] | void {
+
+	const getRootVal = (p: INode): INode => {
+		return p.node.value.properties.filter((prop: INode) => prop.key.name === "root")[0];
```
ev1stensberg: missing ret type

packages/migrate/resolve/resolve.ts
```diff
@@ -7,21 +9,23 @@
...
 			return prop.key.name === "root" ? index : rootIndex;
 		}, -1);
 	};
 
-	const isModulePresent = p => {
-		const modules = p.node.value.properties.filter(
-			prop => prop.key.name === "modules"
+	const isModulePresent = (p: INode): INode | false => {
+		const modules: INode[] = p.node.value.properties.filter(
+			(prop: INode) => prop.key.name === "modules",
```
ev1stensberg: missing ret on filter

packages/migrate/resolve/resolve.ts
```diff
@@ -31,42 +35,49 @@ module.exports = function transformer(j, ast) {
...
 	};
 
 	return ast
 		.find(j.Property)
-		.filter(p => {
+		.filter((p: INode): boolean => {
 			return (
 				p.node.key.name === "resolve" &&
-				p.node.value.properties.filter(prop => prop.key.name === "root")
+				p.node.value.properties.filter((prop: INode) => prop.key.name === "root")
```
ev1stensberg: here, missing ret

packages/migrate/types/NodePath.ts
```diff
@@ -0,0 +1,89 @@
...
+	filter?: (p: (p: INode) => boolean) => INode[] | any;
+	find?: (objectExpression: object, filterExpression?: object) => INode;
+	forEach?: (p: (p: INode) => void) => void;
+	get?: (property: string) => INode;
+	remove?: (_?: void) => void;
+	nodes?: (_?: void) => INode[];
+	pop?: (_?: void) => INode;
+	key?: {
+		name: string;
+		value: INode | string | any;
```
ev1stensberg: if the key resolves to any there's not much need of the interface

packages/migrate/types/NodePath.ts
```diff
@@ -0,0 +1,89 @@
...
+	name?: string;
+	object?: object;
+	parent?: INode;
+	properties?: INode[];
+	property?: INode;
+	prune?: Function;
+	replaceWith?: (objectExpression: object) => INode;
+	size?: (_?: void) => number;
+	type?: string;
+	value?: INode | string | any;
```
ev1stensberg: same here

packages/migrate/types/NodePath.ts
```diff
@@ -0,0 +1,89 @@
...
+	toSource?: (object: {
+		quote?: string,
+	}) => string;
+	source: string;
+	ast: INode;
+	rules?: any[];
```
ev1stensberg: this is not any array. It has specific types

packages/migrate/types/NodePath.ts
```diff
@@ -0,0 +1,89 @@
...
+	BlockStatement: IExpressionObject;
+	Program: IExpressionObject;
+	filters: {
+		VariableDeclarator: {
+			requiresModule: Function,
+		},
+	};
+}
+
+export type valueType = string | number | boolean | any[] | INode | null;
```
ev1stensberg: `any[]` is bad

packages/migrate/index.ts
dhruvdutt: This is same as `master`.

https://github.com/webpack/webpack-cli/blob/617af0bddb5140c07b1a0079b377311b10dd027d/packages/migrate/index.js#L11

packages/migrate/loaders/loaders.ts
dhruvdutt: Where exactly? We're already doing type checking for `createArrayExpressionFromArray` function expression.

packages/migrate/loaders/loaders.ts
dhruvdutt: Where exactly is it missing?

packages/migrate/loaders/loaders.ts
dhruvdutt: Where exactly is it missing?

packages/migrate/migrate.ts
dhruvdutt: Where exactly?

packages/migrate/index.ts
ev1stensberg: okay, ,then you should submit a bugfix fixing this to webpack

dhruvdutt: Sorry, I don't understand what we need to fix in ***webpack***?

Should the change be?
```js
`const { validate } = require('webpack'); `
```

packages/migrate/types/NodePath.ts
dhruvdutt: I know. If we're keeping `INode | string`, then we'll have to do `typeof` check everywhere in the codebase. There are more than 50 places (like below) where we would have to do these checks. :disappointed: 

image

packages/migrate/types/NodePath.ts
dhruvdutt: I know. If we're keeping `INode | string`, then we'll have to do `typeof` check everywhere in the codebase. There are more than 100 places (like below) where we would have to do these checks. :disappointed: 

image

ematipico: Any chance that dynamic types could help in this case?

dhruvdutt: @ematipico Can you please give an example how it would work in this case?

ematipico: I mean that maybe we have to intentionally use the `any` type because of the fact this particular case we want a soft typing... We are relying on the dynamic nature of JavaScript

dhruvdutt: Yeah. I've done soft typing whenever we're using/assigning it. :100:

packages/migrate/index.ts
ev1stensberg: const { validate } = require('webpack')

packages/migrate/migrate.ts
ev1stensberg: in the catch block

packages/migrate/migrate.ts
```diff
@@ -0,0 +1,116 @@
...
+
+	const ast: INode = jscodeshift(source);
+	const recastOptions: object = Object.assign(
+		{
+			quote: "single",
+		},
+		options,
+	);
+	transforms =
+		transforms || Object.keys(transformations).map((k: string) => transformations[k]);
```
ev1stensberg: map here is missing return type

packages/migrate/loaders/loaders.ts
ev1stensberg: in filter

packages/migrate/loaders/loaders.ts
```diff
@@ -195,7 +197,7 @@ module.exports = function(j, ast) {
 		});
 		if (loaders) {
 			p.value.properties = p.value.properties.filter(
-				prop => prop.key.name === "loaders"
+				(prop: INode) => prop.key.name === "loaders",
```
ev1stensberg: missing return in filter

packages/migrate/loaders/loaders.ts
```diff
@@ -103,25 +105,25 @@ module.exports = function(j, ast) {
...
 
-	const createLoaderWithQuery = p => {
-		let properties = p.value.properties;
-		let loaderValue = properties.reduce(
-			(val, prop) => (prop.key.name === "loader" ? prop.value.value : val),
-			""
+	const createLoaderWithQuery = (p: INode): INode => {
+		const properties: INode[] = p.value.properties;
+		const loaderValue: string = properties.reduce(
+			(val: string, prop: INode) => (prop.key.name === "loader" ? prop.value.value : val),
```
ev1stensberg: missing ret type in reduce

packages/migrate/index.ts
```diff
@@ -74,129 +95,144 @@ module.exports = function migrate(...args) {
...
 				})
-				.then(answer => {
-					if (!answer) return;
+				.then((answer: {
+					confirmValidation: boolean;
+				}): void => {
+					if (!answer) { return; }
 
-					runPrettier(outputConfigPath, result, err => {
+					runPrettier(outputConfigPath, result, (err: object) => {
```
ev1stensberg: runprettier missing ret type

packages/migrate/index.ts
```diff
@@ -74,129 +95,144 @@ module.exports = function migrate(...args) {
...
-			task: ctx => {
-				const transformations = require("./migrate").transformations;
+			task: (ctx: INode): string | void | Listr | Promise<any> => {
 				return new Listr(
-					Object.keys(transformations).map(key => {
-						const transform = transformations[key];
+					Object.keys(transformations).map((key: string): {
+						task: (_?: void) => string;
+						title: string;
+					} => {
```
ev1stensberg: missing ret for map here

packages/migrate/index.ts
```diff
@@ -74,129 +95,144 @@ module.exports = function migrate(...args) {
...
 						if (err) {
 							reject(err);
 						}
 						try {
-							const jscodeshift = require("jscodeshift");
+							const jscodeshift: IJSCodeshift = require("jscodeshift");
 							ctx.source = content;
 							ctx.ast = jscodeshift(content);
 							resolve();
 						} catch (err) {
```
ev1stensberg: missing types

packages/migrate/index.ts
```diff
@@ -74,129 +95,144 @@ module.exports = function migrate(...args) {
...
 		{
-			title: "Reading webpack config",
-			task: ctx =>
-				new PLazy((resolve, reject) => {
+			task: (ctx: INode): string | void | Listr | Promise<{}> =>
+				new pLazy((
+						resolve: (value?: object) => void,
+						reject: (reason?: string | object, err?: object) => void,
+					) => {
 					fs.readFile(currentConfigPath, "utf8", (err, content) => {
```
ev1stensberg: fs read missing types

packages/migrate/extractTextPlugin/extractTextPlugin.ts
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+export default function(j: IJSCodeshift, ast: INode): void | INode {
+	const changeArguments = (path: INode): INode => {
+		const args: INode[] = path.value.arguments;
+
+		const literalArgs: INode[] = args.filter((p: INode): boolean => utils.isType(p, "Literal"));
+		if (literalArgs && literalArgs.length > 1) {
+			const newArgs: object = j.objectExpression(
+				literalArgs.map((p: INode, index: number) =>
```
ev1stensberg: map missing type

packages/migrate/extractTextPlugin/extractTextPlugin.ts
```diff
@@ -0,0 +1,60 @@
...
+ * @param {String} pluginName - name of the plugin
+ * @returns {Boolean} isPluginInvocation - whether `node` is the invocation of the plugin denoted by `pluginName`
+ */
+
+function findInvocation(j: IJSCodeshift, path: INode, pluginName: string): boolean {
+	let found: boolean = false;
+	found =
+		j(path)
+			.find(j.MemberExpression)
+			.filter((p: INode) => p.get("object").value.name === pluginName)
```
ev1stensberg: filter missing type

packages/migrate/loaders/loaders.ts
dhruvdutt: No. We're returning `path` at end of the block. Same as previous behavior.

packages/migrate/index.ts
dhruvdutt: Yeah. We are using const `WebpackOptionsValidationError` from webpack as well. So,
```js
const { validate, WebpackOptionsValidationError } = require('webpack'); 
```

packages/migrate/migrate.ts
dhruvdutt: `(err: object)`

packages/migrate/index.ts
dhruvdutt: Oh, I think you missed it. It's already present. The object with properties `task` and `title` is the return type.

```js
.map((key: string): {
    task: (_?: void) => string;
    title: string;
}

File: packages/migrate/index.ts (hunk 46 above)
Comment (dhruvdutt) [reply packages/migrate/index.ts]: TS doesn't allow adding types in `catch` block of _try()_ _catch()_.
```
[ts] Catch clause variable cannot have a type annotation.
```

We're only adding error object in `catch` block of promises.

ev1stensberg: weird. Okay

PR #521, opened 2018-06-18
chore(add-cmd): add typescript support

dhruvdutt:

**Summary**

**Other information**


packages/info/index.ts
```diff
@@ -5,6 +5,7 @@ import envinfo from "envinfo";
  */
 
 export default async function info() {
+	/* tslint:disable-next-line */
```
ev1stensberg: Why are you disabling this? console.error rule? Use process instead

packages/add/index.ts
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Is called and returns a scaffolding instance, adding properties
+ *
+ * @param	{String[]} args - array of arguments such as
+ * @returns {Function} modifyConfigHelper - A helper function that uses the action
+ * 	we're given on a generator
+ *
+ */
+
+export default function add(...args: string[]): any {
```
ematipico: It should return a function, correct?

ev1stensberg: yes

PR #523, opened 2018-06-19
feat(scaffold): add typescript support

dhruvdutt:

**Summary**

**Other information**


packages/webpack-scaffold/index.ts
```diff
@@ -0,0 +1,121 @@
...
+
+export function createArrowFunction(value: Function): string {
+	return `() => '${value}'`;
+}
+
+export function createRegularFunction(value: Function): string {
+	return `function () {\n return '${value}'\n}`;
+}
+
+export function createDynamicPromise(arr: Function[]): string {
```
ev1stensberg: not always an array, could be arr or string

packages/webpack-scaffold/index.ts
```diff
@@ -0,0 +1,121 @@
...
+	type: string;
+	name: string;
+	message: string;
+}
+
+interface IList extends IBaseObj {
+	choices?: string[];
+}
+
+interface IInput extends IBaseObj {
```
ev1stensberg: better convention here. `Inq_Input` etc maybe

packages/webpack-scaffold/index.ts
```diff
@@ -0,0 +1,121 @@
+import jscodeshift from "jscodeshift";
+
+interface IBaseObj {
```
ematipico: I think we should change the name. `IBaseObj` sounds too generic

packages/webpack-scaffold/index.ts
ematipico: What "Inq" stands for?

dhruvdutt: I think Inquirer?

packages/webpack-scaffold/index.ts
dhruvdutt: This interface would be scoped to this file so I thought to name it that way. What do you suggest?

ematipico: `IScaffoldBase` or `IScaffoldBaseObject`? Something like this. What do you think?

PR #527, opened 2018-06-25
[WIP] Refactor bin folder

evenstensberg:
- [x] Class based instances for packages

- [x] Set up structure

- [x] Use logger module

- [x] Flags are validated correctly

- [x] Flags are grouped
- [x]  Better abstraction level for adapter layer

- [x] Flags are converted to their respective properties mapped to an webpackOptions object
- [x] Hook up logic with compilation

- [ ] 0CJS story experiment on next branch

- [ ] `interactive` command

- [ ] `extend` option
- [ ] fill options


ematipico:
Could you please fill the issue description with the relevant information of what this is PR for, why, etc.? 
Also, could you create an issue?

ev1stensberg:
There's a couple of issues on this already. Read on `general improvements` tickets and the old PR that was submitted on this. It's work in progress :)

ematipico:
Yes, I know. But for clarity and transparency we want to fill the whole template to give information to the community. Please fill the template with relevant inforamtion

shellscape:
Please be mindful not to wholesale copy elements of structure and/or content from [`webpack-command`](https://github.com/webpack-contrib/webpack-command). The recent 3ced7e5 comment adds files to `'lib/gorups` that exactly match [`lib/flags`](https://github.com/webpack-contrib/webpack-command/tree/master/lib/flags) in `webpack-command`.

ev1stensberg:
There's opportunities to split work here without us conflicting work. If someone could update the tests to be more effective that would be nice.

ematipico:
What  can I take that is not assigned yet?

lib/webpack-cli.js
```diff
@@ -0,0 +1,77 @@
...
+				return res;
+			})
+			.filter(e => e);
+	}
+
+	normalizeGroup(group) {
+		return this.arrayToObject(group);
+	}
+
+	async run(args, yargsOptions) {
```
ematipico: I get errors from `eslint` saying that this is not supported, because we support, as min version, `6.11.5`, which doesn't have `async`/`await`

ev1stensberg:
@ematipico can you refactor tests to run in parallel as a normal jest suite? See webpack command for similar implementation

ematipico:
Is there a reason why some functions have been implemented with `async`/`await`? Are we planning to do a major node bump? I checkout your branch and had lot of eslint errors due to unsupported features.

ev1stensberg:
Because they are async. Is async/await v7+?

evenstensberg:
Just a quick update here. Infra is okay, need to make each option group return

 ```js
{
 options: {...} // sent to compiler
webpackOutputOptions: {} // sent to adjust output etc like stats
proccessingErrors: [] // array indicating if there's something wrong with the option
}
```

lib/webpack-cli.js
evenstensberg: Hm that is bad, will look into it

PR #532, opened 2018-06-27
feat(generators): add typescript support

dhruvdutt:

**Summary**

**Other information**

Not ready for review yet.


packages/generators/add-generator.ts
```diff
@@ -1,19 +1,24 @@
...
-const path = require("path");
-const inquirerAutoComplete = require("inquirer-autocomplete-prompt");
-const { AutoComplete, Confirm, Input, List } = require("@webpack-cli/webpack-scaffold");
+import Generator from "yeoman-generator";
 
-const webpackSchema = require("./utils/optionsSchema.json");
-const webpackDevServerSchema = require("webpack-dev-server/lib/optionsSchema.json");
-const PROP_TYPES = require("@webpack-cli/utils/prop-types");
+import glob from "glob-all";
+import inquirerAutoComplete from "inquirer-autocomplete-prompt";
```
ev1stensberg: `import autoComplete`

packages/generators/add-generator.ts
```diff
@@ -61,12 +66,12 @@ const traverseAndGetProperties = (arr, prop) => {
...
  *
  */
-const searchProps = (answers, input) => {
+const searchProps = (answers: object, input: string): Promise<string[]> => {
 	input = input || "";
 	return Promise.resolve(
-		PROPS.filter(food =>
-			food.toLowerCase().includes(input.toLowerCase())
-		)
+		PROPS.filter((food: string) =>
```
ev1stensberg: Why is this named food? ^_^

packages/generators/add-generator.ts
```diff
@@ -79,29 +84,39 @@ const searchProps = (answers, input) => {
...
 	}
 
-	prompting() {
-		let done = this.async();
-		let action;
-		let self = this;
-		let manualOrListInput = action =>
-			Input("actionAnswer", `What do you want to add to ${action}?`);
+	public prompting() {
+		const done: any = this.async();
```
ev1stensberg: This is a promise that resolves to a boolean if I remember right

packages/generators/add-generator.ts
```diff
@@ -79,29 +84,39 @@ const searchProps = (answers, input) => {
...
+		let action: string;
+		const self: this = this;
+		let manualOrListInput: any = (promptAction: string) =>
+			Input("actionAnswer", `What do you want to add to ${promptAction}?`);
 
 		// first index indicates if it has a deep prop, 2nd indicates what kind of
-		let isDeepProp = [false, false];
+		const isDeepProp: any[] = [false, false];
```
ev1stensberg: a bit naive to use `any` here when you can see the type

packages/generators/add-generator.ts
```diff
@@ -125,44 +142,46 @@ module.exports = class AddGenerator extends Generator {
...
 							this.configuration.config.item = action;
 						});
 				}
-				let temp = action;
+				const temp: string = action;
 				if (action === "resolveLoader") {
 					action = "resolve";
 				}
-				const webpackSchemaProp = webpackSchema.definitions[action];
+				const webpackSchemaProp: any = webpackSchema.definitions[action];
```
ev1stensberg: avoid using `any` here, just check the prop instead

packages/generators/add-generator.ts
```diff
@@ -198,25 +217,25 @@ module.exports = class AddGenerator extends Generator {
 					}
 				}
 				// WDS has its own schema, so we gonna need to check that too
-				const webpackDevserverSchemaProp =
+				const webpackDevserverSchemaProp: any =
```
ev1stensberg: Avoid using `any` here

packages/generators/add-generator.ts
```diff
@@ -421,49 +448,57 @@ module.exports = class AddGenerator extends Generator {
...
 											done();
-										}
+										},
 									);
 								} else {
 									// We got the schema prop, we've correctly prompted it, and can add it directly
 									this.configuration.config.item = action + "." + isDeepProp[1];
 									this.configuration.config.webpackOptions[action] = {
-										[isDeepProp[1]]: deepPropAns.deepProp
+										[isDeepProp[1]]: deepPropAns.deepProp,
```
ev1stensberg: Not sure how dynamic types are in typescript, but I think it's possible to reassign types or to make it `oneOf`

ev1stensberg: oneOf as in it could be one of many types, but not `any`.

packages/generators/addon-generator.ts
```diff
@@ -0,0 +1,83 @@
...
+			);
+
+			copyFiles.forEach(this.copy);
+			copyTemplateFiles.forEach(this.copyTpl);
+		}
+
+		public install() {
+			this.npmInstall(["webpack-defaults", "bluebird"], {
+				"save-dev": true,
+			}).then(() => {
```
ev1stensberg: could you add `void` here as param and `_` as argument?

packages/generators/plugin-generator.ts
```diff
@@ -28,12 +29,10 @@ const PluginGenerator = addonGenerator(
 		"test/functional.test.js.tpl",
 		"examples/simple/src/index.js.tpl",
 		"examples/simple/src/lazy-module.js.tpl",
-		"examples/simple/src/static-esm-module.js.tpl"
+		"examples/simple/src/static-esm-module.js.tpl",
 	],
 	["src/_index.js.tpl", "examples/simple/_webpack.config.js.tpl"],
-	gen => ({ name: _.upperFirst(_.camelCase(gen.props.name)) })
+	(gen: any) => ({ name: _.upperFirst(_.camelCase(gen.props.name)) }),
```
ev1stensberg: gen is a specific instance, not `any`.

packages/generators/loader-generator.ts
```diff
@@ -47,13 +48,10 @@ const LoaderGenerator = addonGenerator(
 		"examples/simple/webpack.config.js.tpl",
 		"examples/simple/src/index.js.tpl",
 		"examples/simple/src/lazy-module.js.tpl",
-		"examples/simple/src/static-esm-module.js.tpl"
+		"examples/simple/src/static-esm-module.js.tpl",
 	],
 	["src/_index.js.tpl"],
-	gen => ({ name: gen.props.name })
+	(gen: any) => ({ name: gen.props.name }),
```
ev1stensberg: gen is not `any`, it has a specific type.

packages/generators/types/yeoman-generator.d.ts
```diff
@@ -0,0 +1,261 @@
...
+        destinationPath(...path: string[]): string;
+		determineAppname(): void;
+		env: {
+			adapter: {
+				promptModule: {
+					registerPrompt: (promptName: string, promptModule: object) => void;
+				}
+			}
+		};
+        getCollisionFilter(): (output: any) => void;
```
ev1stensberg: any is not the case here either

packages/generators/types/yeoman-generator.d.ts
```diff
@@ -0,0 +1,261 @@
...
+        hookFor(name: string, config: IHookConfig): void;
+        option(name: string, config?: IYeomanGeneratorOption): void;
+        rootGeneratorName(): string;
+        run(args?: any): void;
```
ev1stensberg: any is a specific type, object or array.

packages/generators/types/yeoman-generator.d.ts
```diff
@@ -0,0 +1,261 @@
...
+        once(event: string, listener: Function): this;
+        removeListener(event: string, listener: Function): this;
+        removeAllListeners(event?: string): this;
+        setMaxListeners(n: number): this;
+        getMaxListeners(): number;
+        listeners(event: string): Function[];
+        emit(event: string, ...args: any[]): boolean;
+        listenerCount(type: string): number;
+
+        async(): any;
```
ev1stensberg: async is probably a promise

packages/generators/types/yeoman-generator.d.ts
```diff
@@ -0,0 +1,261 @@
...
+        prompt(opt: IPromptOptions | IPromptOptions[], callback: (answers: any) => void): void;
+        prompt<T>(opt: IPromptOptions | IPromptOptions[]): PromiseLike<T>;
+        log(message: string) : void;
+        npmInstall(packages: string[], options?: any, cb?: Function): any;
```
ev1stensberg: this is not `any`

packages/generators/types/yeoman-generator.d.ts
```diff
@@ -0,0 +1,261 @@
...
+        installDependencies(options?: IInstallDependencyOptions): void;
+        spawnCommand(name: string, args?: string[], options?: Object): void;
+        spawnCommandSync(name: string, args?: string[], options?: Object): void;
+
+        appname: string;
+        gruntfile: IGruntFileStatic;
+        options: { [key: string]: any };
```
ev1stensberg: not any here

packages/generators/types/yeoman-generator.d.ts
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    interface IPromptOptions{
+        type?: string;
+        name: string;
+        message: string | ((answers: Object) => string);
+        choices?: any[] | ((answers: Object) => any);
+        default?: string | number | string[] | number[] | ((answers: Object) => (string | number | string[] | number[]));
+        validate?: ((input: any) => boolean | string);
+        filter?: ((input: any) => any);
```
ev1stensberg: insecure about the use of `any` here

packages/generators/types/yeoman-generator.d.ts
```diff
@@ -0,0 +1,261 @@
...
+        store?: boolean;
+    }
+
+    interface IGruntFileStatic {
+        loadNpmTasks(pluginName: string): void;
+        insertConfig(name:string, config:any):void;
+        registerTask(name:string, tasks:any):void;
+        insertVariable(name:string, value:any):void;
+        prependJavaScript(code:string):void;
+    }
```
ev1stensberg: these any values should be more type specific

ematipico:
Please update the template with the relevant information and link this PR to the original issue

packages/generators/add-generator.ts
dhruvdutt: Yeah. I tried `boolean[]` and `Array<object | boolean>` but both won't work because we also assign the values as boolean, string, object, array in different scopes. Some of the examples:

https://github.com/webpack/webpack-cli/blob/51dc005a800cd25e19f986adc6ff46436cc425b4/packages/generators/add-generator.js#L412

https://github.com/webpack/webpack-cli/blob/51dc005a800cd25e19f986adc6ff46436cc425b4/packages/generators/add-generator.js#L443

packages/generators/add-generator.ts
dhruvdutt: You mean something like?
```
const done: Promise<boolean>
```

Also, if it's a `done` is a promise, the compiler would fail all calls like `done()`. I think `done` is a function. Something like this should work:
```
const done: () => void | boolean
```

packages/generators/add-generator.ts
ev1stensberg: https://stackoverflow.com/questions/40914716/dynamic-typescript-typing

packages/generators/add-generator.ts
ev1stensberg: It's a promise http://yeoman.io/authoring/running-context.html

packages/generators/add-generator.ts
dhruvdutt:
```js
const isDeepProp: [boolean, string | boolean | object | object[]]
```

^This is a valid type declaration which allows to control specifically which element should be of which type. Even after declaring the second element to be one of the boolean, string, object or array of objects, the compiler fails. I think the complex reassignment logic we're doing inside the generator is making the compiler's type prediction to fail.

TS compiler works for simple logic like doing `typeof var === "string"` and then using it as string, the compiler infers it to be a `string` even if the type was `string | boolean ...other types`.

packages/generators/add-generator.ts
dhruvdutt: > If the asynchronous API you’re relying upon doesn’t support promises, then you can rely on the legacy this.async() way. Calling this.async() will return a function to call once the task is done.

It's not a promise. It could be something like this:
```js
const done: () => void | boolean
```

packages/generators/add-generator.ts
ev1stensberg: maybe add a interface for this if it's possible

packages/generators/plugin-generator.ts
ev1stensberg: y, run the generator with a console log in between the return arg to test

packages/generators/remove-generator.js
```diff
@@ -13,110 +13,91 @@ const PROP_TYPES = require("@webpack-cli/utils/prop-types");
...
+                webpackOptions: {},
+            },
+        };
+        let configPath = path.resolve(process.cwd(), "webpack.config.js");
+        const webpackConfigExists = fs.existsSync(configPath);
+        if (!webpackConfigExists) {
+            configPath = null;
+            // end the generator stating webpack config not found or to specify the config
+        }
+        this.webpackOptions = require(configPath);
```
ev1stensberg: you'll need to do a json read on this, require's will try to resolve dependencies, which we don't need to install in order to run the ast transformation

packages/generators/remove-generator.ts
```diff
@@ -0,0 +1,131 @@
...
+										};
+									}
+								}
+							});
+					}
+				} else {
+					this.configuration.config.webpackOptions[propType] = null;
+				}
+			})
+			.then(() => {
```
ev1stensberg: void here `_: void`

packages/generators/remove-generator.ts
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	public getModuleLoadersNames(): string[] {
+		if (this.webpackOptions.module && this.webpackOptions.module.rules) {
+			return this.webpackOptions.module.rules.map((rule: any) => rule ? rule.loader : null);
+		}
+	}
+
+	public prompting() {
+		const done: () => void | boolean = this.async();
```
ev1stensberg: `(_)`

packages/generators/types/json-loader.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module "*.json" {
+	const value: any;
```
ev1stensberg: this is not `any`, bool | num | object | array

packages/generators/utils/entry.ts
```diff
@@ -0,0 +1,107 @@
...
+/**
+ *
+ * Prompts for entry points, either if it has multiple or one entry
+ *
+ * @param	{Object} self 	- A variable holding the instance of the prompting
+ * @param	{Object} answer - Previous answer from asking if the user wants single or multiple entries
+ * @returns	{Object} An Object that holds the answers given by the user, later used to scaffold
+ */
+
+export default function entry(self: any, answer: {
```
ev1stensberg: generator instance on `self`

packages/generators/utils/entry.ts
```diff
@@ -0,0 +1,107 @@
...
+					validate,
+				),
+			])
+			.then((multipleEntriesAnswer: {
+				multipleEntries: string,
+			}) => {
+				const webpackEntryPoint: object = {};
+				entryIdentifiers = multipleEntriesAnswer.multipleEntries.split(",");
+
+				function forEachPromise(entries, fn) {
```
ev1stensberg: missing types

packages/generators/utils/module.ts
```diff
@@ -0,0 +1,27 @@
+interface IModule {
+	include: string[];
+	loader: string;
+	options: {
+		plugins: string[];
+		presets: any[];
```
ev1stensberg: string

packages/webpack-scaffold/index.ts
```diff
@@ -11,7 +11,7 @@ interface IInquirerList extends IScaffoldBaseObject {
 }
 
 interface IInquirerInput extends IScaffoldBaseObject {
-	validate?: Function;
+	validate?: (input: any) => string | boolean;
```
ev1stensberg: input is string

packages/webpack-scaffold/index.ts
```diff
@@ -95,7 +95,7 @@ export function Input(name: string, message: string): IInquirerInput {
 	};
 }
 
-export function InputValidate(name: string, message: string, cb: Function): IInquirerInput {
+export function InputValidate(name: string, message: string, cb?: (input: any) => string | boolean): IInquirerInput {
```
ev1stensberg: input is string

packages/generators/types/yeoman-generator.d.ts
```diff
@@ -0,0 +1,55 @@
...
+		public spawnCommand(name: string, args?: string[], options?: Object): void;
+	}
+
+	interface IPromptOptions {
+		type?: string;
+		name: string;
+		message: string | ((answers: Object) => string);
+		choices?: any[] | ((answers: Object) => any);
+		default?: string | number | string[] | number[] | ((answers: Object) => (string | number | string[] | number[]));
+		validate?: ((input: any) => boolean | string);
```
ev1stensberg: string as input

packages/generators/remove-generator.ts
dhruvdutt: If you do that then we'll have to pass something when we call `done()` at the end of generators.

ev1stensberg: oh okay, let's keep it then (:

.gitignore
```diff
@@ -33,3 +33,6 @@ lerna-debug.log
 
 # Yarn lock file
 yarn.lock
+
+# Custom typings
+custom_typing/*.js
```
ev1stensberg: types

packages/generators/add-generator.ts
```diff
@@ -79,29 +85,39 @@ const searchProps = (answers, input) => {
...
 
-module.exports = class AddGenerator extends Generator {
+export default class AddGenerator extends Generator {
+	private dependencies: string[];
+	private configuration: {
+		config: {
+			configName?: string,
+			topScope?: string[],
+			item?: string;
+			webpackOptions?: any,
```
ev1stensberg: object

packages/generators/add-generator.ts
```diff
@@ -79,29 +85,39 @@ const searchProps = (answers, input) => {
...
-		let done = this.async();
-		let action;
-		let self = this;
-		let manualOrListInput = action =>
-			Input("actionAnswer", `What do you want to add to ${action}?`);
+	public prompting() {
+		const done: () => void | boolean = this.async();
+		let action: string;
+		const self: this = this;
+		let manualOrListInput: any = (promptAction: string) =>
```
ev1stensberg: in either way this returns an inquirer instance/promise

packages/generators/add-generator.ts
```diff
@@ -125,44 +143,46 @@ module.exports = class AddGenerator extends Generator {
...
 					])
-						.then(entryTypeAnswer => {
+						.then((entryTypeAnswer: {
+							entryType: boolean,
+						}) => {
 							// Ask different questions for entry points
 							return entryQuestions(self, entryTypeAnswer);
 						})
-						.then(entryOptions => {
+						.then((entryOptions: object) => {
```
ev1stensberg: You should be consistent with the types like you have done on previous functions

packages/generators/add-generator.ts
```diff
@@ -125,44 +143,46 @@ module.exports = class AddGenerator extends Generator {
...
 							this.configuration.config.webpackOptions.entry = entryOptions;
 							this.configuration.config.item = action;
 						});
 				}
-				let temp = action;
+				const temp: string = action;
```
ev1stensberg: it's temp for a reason

packages/generators/add-generator.ts
```diff
@@ -125,44 +143,46 @@ module.exports = class AddGenerator extends Generator {
...
 				 */
-				let defOrPropDescription = webpackSchemaProp
+				let defOrPropDescription: object = webpackSchemaProp
 					? webpackSchemaProp.properties
 					: webpackSchema.properties[action].properties
 						? webpackSchema.properties[action].properties
 						: webpackSchema.properties[action].anyOf
 							? webpackSchema.properties[action].anyOf.filter(
-								p => p.properties || p.enum
+								(p) => p.properties || p.enum,
```
ev1stensberg: object

packages/generators/addon-generator.ts
```diff
@@ -24,54 +25,59 @@ const copyUtils = require("@webpack-cli/utils/copy-utils");
...
-			return this.prompt(prompts).then(props => {
+	return class AddOnGenerator extends Generator {
+		private props: {
+			name?: string;
+		};
+		private copy: (value: string, index: number, array: string[]) => void;
+		private copyTpl: (value: string, index: number, array: string[]) => void;
+
+		public prompting() {
+			return this.prompt(prompts).then((props) => {
```
ev1stensberg: type missing

packages/generators/init-generator.ts
```diff
@@ -25,92 +23,109 @@ const tooltip = require("./utils/tooltip");
...
-module.exports = class InitGenerator extends Generator {
+export default class InitGenerator extends Generator {
+	private isProd: boolean;
+	private usingDefaults: boolean;
+	private dependencies: string[];
+	private configuration: {
+		config: {
+			configName?: string,
+			topScope?: string[],
+			webpackOptions?: any,
```
ev1stensberg: object

packages/generators/loader-generator.ts
ev1stensberg: ^ do a manual console.log to test the type of this

packages/generators/plugin-generator.ts
```diff
@@ -13,13 +14,13 @@ const addonGenerator = require("./addon-generator");
...
-			name: "name",
-			message: "Plugin name",
 			default: "my-webpack-plugin",
 			filter: _.kebabCase,
-			validate: str => str.length > 0
-		}
+			message: "Plugin name",
+			name: "name",
+			type: "input",
+			validate: (str) => str.length > 0,
```
ev1stensberg: types missing

packages/generators/plugin-generator.ts
ev1stensberg: see prev comment

packages/generators/remove-generator.ts
```diff
@@ -0,0 +1,131 @@
...
+ * @returns	{Void} After execution, transforms are triggered
+ *
+ */
+
+export default class RemoveGenerator extends Generator {
+	private configuration: {
+		config: {
+			configName?: string,
+			topScope?: string[],
+			webpackOptions?: any,
```
ev1stensberg: object

packages/generators/remove-generator.ts
```diff
@@ -0,0 +1,131 @@
...
+		this.webpackOptions = fs.readFileSync(configPath, "utf8");
+	}
+
+	public getPropTypes(): string[] {
+		return Object.keys(this.webpackOptions);
+	}
+
+	public getModuleLoadersNames(): string[] {
+		if (this.webpackOptions.module && this.webpackOptions.module.rules) {
+			return this.webpackOptions.module.rules.map((rule: any) => rule ? rule.loader : null);
```
ev1stensberg: any is false

packages/generators/types/index.ts
```diff
@@ -0,0 +1,6 @@
+export interface ISchemaProperties {
+	additionalProperties?: boolean;
+	definitions?: any;
+	properties?: any;
```
ev1stensberg: get the types of these

packages/generators/types/json-loader.d.ts
ev1stensberg: ^

packages/generators/types/yeoman-generator.d.ts
```diff
@@ -0,0 +1,55 @@
...
+				},
+			},
+		};
+		public composeWith(namespace: string, options: any, settings?: IComposeSetting): YeomanGeneratorBase;
+		public destinationRoot(rootPath?: string): string;
+		public destinationPath(...path: string[]): string;
+		public run(args?: any, callback?: Function): void;
+		public runInstall(packager: string, dependencies: string[], options?: object): void;
+		public on(event: string, listener: Function): this;
+		public async(): any;
```
ev1stensberg: not any

packages/generators/types/yeoman-generator.d.ts
```diff
@@ -0,0 +1,55 @@
...
+			set: (setProperty: string, setValue: object) => void;
+		};
+		public env: {
+			adapter: {
+				promptModule: {
+					registerPrompt: (promptName: string, promptModule: object) => void;
+				},
+			},
+		};
+		public composeWith(namespace: string, options: any, settings?: IComposeSetting): YeomanGeneratorBase;
```
ev1stensberg: options isn't any

packages/generators/types/yeoman-generator.d.ts
```diff
@@ -0,0 +1,55 @@
...
+		public destinationRoot(rootPath?: string): string;
+		public destinationPath(...path: string[]): string;
+		public run(args?: any, callback?: Function): void;
```
ev1stensberg: array or obj on args, check docs on that

packages/generators/types/yeoman-generator.d.ts
```diff
@@ -0,0 +1,55 @@
...
+		public runInstall(packager: string, dependencies: string[], options?: object): void;
+		public on(event: string, listener: Function): this;
+		public async(): any;
+		public prompt(opt: IPromptOptions | IPromptOptions[], callback: (answers: any) => void): void;
+		public prompt<T>(opt: IPromptOptions | IPromptOptions[]): PromiseLike<T>;
+		public log(message: string): void;
+		public npmInstall(packages?: string[] | string, options?: any, cb?: Function): Promise<>;
```
ev1stensberg: object is the type of options

packages/generators/types/yeoman-generator.d.ts
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	// tslint:disable-next-line
+	class NamedBase extends YeomanGeneratorBase implements INamedBase {
+		constructor(args: string | string[], options: any);
+	}
+
+	// tslint:disable-next-line
+	class Base extends NamedBase implements IBase {
+		public static extend(protoProps: IQueueProps, staticProps?: any): YeomanGeneratorBase;
```
ev1stensberg: check type on the missing type of `any`

packages/generators/utils/entry.ts
```diff
@@ -12,27 +11,33 @@ const validate = require("./validate");
...
 
-module.exports = (self, answer) => {
-	let entryIdentifiers;
-	let result;
-	if (answer["entryType"] === true) {
+export default function entry(self, answer: {
+	entryType: boolean;
+}): any {
+	let entryIdentifiers: string[];
+	let result: any;
```
ev1stensberg: 🤔 ...

packages/generators/utils/entry.ts
```diff
@@ -12,27 +11,33 @@ const validate = require("./validate");
...
+				entryIdentifiers = multipleEntriesAnswer.multipleEntries.split(",");
+
+				function forEachPromise(entries: string[], fn: (entryProp: string) => Promise<{} | void>) {
+					return entries.reduce((promise: Promise<{}>, prop: string) => {
+						const trimmedProp: string = prop.trim();
+
+						return promise.then((n: object) => {
 							if (n) {
-								Object.keys(n).forEach(val => {
+								Object.keys(n).forEach((val) => {
```
ev1stensberg: missing type

packages/generators/utils/tooltip.ts
```diff
@@ -7,17 +6,19 @@
  * @returns {Object} An Object that consists of tooltip methods to be invoked
  */
 
-module.exports = {
-	uglify: _ => {
+export default {
+	cssPlugin: (): string => {
```
ev1stensberg: `(_:void)`

packages/generators/utils/tooltip.ts
```diff
@@ -7,17 +6,19 @@
...
+ * one of your module entries!
  *
- * https://github.com/webpack-contrib/uglifyjs-webpack-plugin
+ * https://github.com/webpack-contrib/mini-css-extract-plugin
  *
  */`;
 	},
-	splitChunks: _ => {
+
+	splitChunks: (): string => {
```
ev1stensberg: void

packages/generators/utils/tooltip.ts
```diff
@@ -31,17 +32,8 @@ module.exports = {
...
- * use css modules that will be moved into a separate CSS file instead of inside
- * one of your module entries!
- *
- * https://github.com/webpack-contrib/mini-css-extract-plugin
- *
- */`;
-	},
-	postcss: _ => {
+
+	postcss: (): string => {
```
ev1stensberg: void param

packages/generators/utils/tooltip.ts
```diff
@@ -56,5 +48,15 @@ module.exports = {
  * https://github.com/jonathantneal/precss
  *
  */`;
-	}
+	},
+
+	uglify: (): string => {
```
ev1stensberg: void as a param

packages/info/README.md
```diff
@@ -0,0 +1,24 @@
+# webpack-cli info
```
ev1stensberg: This shouldn't be here, is from another pr

packages/info/index.ts
```diff
@@ -1,19 +1,18 @@
-"use strict";
-
-const envinfo = require("envinfo");
+import * as envinfo from "envinfo";
```
ev1stensberg: this is from another pr

packages/generators/utils/tooltip.ts
dhruvdutt: https://github.com/webpack/webpack-cli/blob/617af0bddb5140c07b1a0079b377311b10dd027d/packages/generators/init-generator.js#L338

packages/generators/utils/tooltip.ts
dhruvdutt: What to pass while calling the method?

https://github.com/webpack/webpack-cli/blob/617af0bddb5140c07b1a0079b377311b10dd027d/packages/generators/init-generator.js#L381

packages/generators/utils/tooltip.ts
dhruvdutt: > What to pass while calling the method?

packages/generators/utils/tooltip.ts
dhruvdutt: > What to pass while calling the method?

ev1stensberg: _?: void

packages/generators/utils/tooltip.ts
ev1stensberg: _?: void

packages/generators/utils/tooltip.ts
ev1stensberg: _?: void

ematipico: Can we have some kind of eslint rule about this underscore? I keep see inconsistency and honestly I don't like it. But this is a project and I don't mind its rules but where is the rule for this underscore? Developers don't know why (me included) it's there.

packages/generators/add-generator.ts
dhruvdutt: Okay, so what are you stating? Sorry, I don't understand.

ev1stensberg: it's named temp for a reason, it get's mutated I assume?

packages/generators/addon-generator.ts
```diff
@@ -24,54 +26,57 @@ const copyUtils = require("@webpack-cli/utils/copy-utils");
...
  * @returns {Generator} A class extending Generator
  */
-function addonGenerator(
-	prompts,
-	templateDir,
-	copyFiles,
-	copyTemplateFiles,
-	templateFn
+export default function addonGenerator(
+	prompts: any[],
```
ev1stensberg: My bet is that these are array of functions or promises, debug it to check

packages/generators/init-generator.ts
```diff
@@ -381,39 +406,40 @@ module.exports = class InitGenerator extends Generator {
...
 			this.configuration.config.topScope.push(
 				tooltip.uglify(),
 				"const UglifyJSPlugin = require('uglifyjs-webpack-plugin');",
-				"\n"
+				"\n",
 			);
 		}
-		const packager = getPackageManager();
+		const packager: string = getPackageManager();
 		const opts = packager === "yarn" ? { dev: true } : { "save-dev": true };
```
ev1stensberg: missing type

packages/generators/add-generator.ts
dhruvdutt: It isn't getting mutated. It would have anyway caused linting error if we were reassigning it.

packages/generators/remove-generator.ts
```diff
@@ -0,0 +1,136 @@
...
+			},
+		};
+
+		let configPath: string = path.resolve(process.cwd(), "webpack.config.js");
+		const webpackConfigExists: boolean = fs.existsSync(configPath);
+		if (!webpackConfigExists) {
+			configPath = null;
+			// end the generator stating webpack config not found or to specify the config
+		}
+		this.webpackOptions = fs.readFileSync(configPath, "utf8");
```
ev1stensberg: I think you'll need to do a JSON parse too, this will be a utf8 encoded string

packages/generators/types/index.ts
```diff
@@ -0,0 +1,37 @@
+export interface ISchemaProperties {
+	additionalProperties?: boolean;
+	definitions?: object;
+	properties?: object;
+	type?: string;
+}
+
+export interface IWebpackOptions {
```
ev1stensberg: These types need to be written as they will be, all of them are strings in the ast types, but here they are specified. Check webpack validation schema for definition of types

packages/generators/types/json-loader.d.ts
ev1stensberg: CC @TheLarkInn

packages/generators/types/yeoman-generator.d.ts
```diff
@@ -0,0 +1,89 @@
...
+	public props: {
+		name: string;
+	};
+	public composeWith(namespace: string, options?: object, settings?: IComposeSetting): YeomanGeneratorBase;
+	public destinationRoot(rootPath?: string): string;
+	public destinationPath(...path: string[]): string;
+	public run(target: string, options?: object, done?: Function): IRunEnv;
+	public runInstall(packager: string, dependencies: string[], options?: object): void;
+	public on(event: string, listener: Function): this;
+	public async(): () => void | boolean;
```
ev1stensberg: is `_? void` possible here?

packages/generators/types/yeoman-generator.d.ts
```diff
@@ -0,0 +1,89 @@
...
+		public props: {
+			name: string;
+		};
+		public composeWith(namespace: string, options?: object, settings?: IComposeSetting): YeomanGeneratorBase;
+		public destinationRoot(rootPath?: string): string;
+		public destinationPath(...path: string[]): string;
+		public run(target: string, options?: object, done?: Function): IRunEnv;
+		public runInstall(packager: string, dependencies: string[], options?: object): void;
+		public on(event: string, listener: Function): this;
+		public async(): () => void | boolean;
```
ev1stensberg: is `_?:void` possible here?

tslint.json
```diff
@@ -15,6 +15,10 @@
       ],
       "no-console": {
         "severity": "warning"
+      },
+      "max-line-length": {
+        "options": { "limit": 120 },
```
ev1stensberg: let's keep this to 70 and warn when it's over limit

packages/generators/types/index.ts
dhruvdutt: You mean everything from this [file](https://github.com/webpack/webpack/blob/master/schemas/WebpackOptions.json)?

packages/generators/remove-generator.ts
dhruvdutt: Yeah. This will return a string. Doing `JSON.parse()` on that string would result in error because the string contents aren't in JSON convertable format as it would be `module.exports = { ... }`. Even if we somehow read it in some JS variable, we would have to read object at `module.exports`.

I think the best way is to do `require(configPath)`. Same as we were doing earlier. It would directly read the webpack config object present at `module.export = { ... }`

packages/generators/add-generator.ts
ev1stensberg: alright. Nice

packages/generators/remove-generator.ts
ev1stensberg: Yes you are right on the part on requiring the correct module, but when we do a require it will try to resolve dependencies we might not have in the repo and access variables that we might not need to bother with. It will result in a lot of errors for us and that's not good. 

Try to implement this using a JSON read instead. It is fully possible. To access the variables you'll need to do something like module['entry'] etc

packages/generators/types/index.ts
ev1stensberg: Yes that's right

ev1stensberg: Base your values on that file, avoid using `any`

packages/generators/remove-generator.ts
dhruvdutt: > it will try to resolve dependencies we might not have in the repo.

`require` doesn't always / only resolve npm dependencies. It's also a [standard convention for importing modules](https://nodejs.org/api/modules.html#modules_modules). If it's a file path like in this case, it won't try to resolve dependencies but simply import the module.exports object.

> It will result in a lot of errors for us and that's not good.

We were already using this convention in the existing codebase (master).

https://github.com/webpack/webpack-cli/blob/617af0bddb5140c07b1a0079b377311b10dd027d/packages/utils/modify-config-helper.js#L71

https://github.com/webpack/webpack-cli/blob/617af0bddb5140c07b1a0079b377311b10dd027d/packages/generators/add-generator.js#L326

dhruvdutt: Can we have some third opinion @ematipico ?

ev1stensberg: In that case, it should be changed. @dhruvdutt either way you do this, make sure not to import a file directly, as it would resolve the dependencies. I assume the closest thing you'll get to a error-free case is a stringified and parsed json

ev1stensberg:
Needs a rebase :)

.gitignore
```diff
@@ -33,6 +33,3 @@ lerna-debug.log
 
 # Yarn lock file
 yarn.lock
-
-# Custom typings
```
ev1stensberg: Missing types folder

packages/generators/remove-generator.ts
```diff
@@ -14,46 +15,57 @@ const PROP_TYPES = require("@webpack-cli/utils/prop-types");
...
 			List(
 				"propType",
 				"Which property do you want to remove?",
-				Array.from(this.getPropTypes())
-			)
+				Array.from(this.getPropTypes()),
+			),
 		])
-			.then(({ propType }) => {
+			.then(({ propType }: { propType: string }) => {
```
ev1stensberg: returns promise

packages/generators/remove-generator.ts
```diff
@@ -66,42 +78,44 @@ module.exports = class RemoveGenerator extends Generator {
 							List(
 								"keyType",
 								`Which key do you want to remove from ${propType}?`,
-								Array.from(propValue)
-							)
-						]).then(({ keyType }) => {
+								Array.from(propValue),
+							),
+						]).then(({ keyType }: { keyType: string }) => {
```
ev1stensberg: returns promise

packages/generators/remove-generator.ts
```diff
@@ -66,42 +78,44 @@ module.exports = class RemoveGenerator extends Generator {
...
 							List(
 								"keyType",
 								`Which key do you want to remove from ${propType}?`,
-								Array.from(Object.keys(propValue))
-							)
+								Array.from(Object.keys(propValue)),
+							),
 						])
-							.then(({ keyType }) => {
+							.then(({ keyType }: { keyType: string }) => {
```
ev1stensberg: returns promise

packages/generators/remove-generator.ts
```diff
@@ -66,42 +78,44 @@ module.exports = class RemoveGenerator extends Generator {
...
+											Array.from(this.getModuleLoadersNames()),
+										),
 									])
-										.then(({ rule }) => {
-											const loaderIndex = this.getModuleLoaders().indexOf(rule);
-											const loader = this.webpackOptions.module.rules[loaderIndex];
-											this.configuration.config.webpackOptions.module = {
-												rules: [ loader ]
-											};
+										.then(({ rule }: { rule: string }) => {
```
ev1stensberg: returns promise

packages/generators/remove-generator.ts
```diff
@@ -111,12 +125,12 @@ module.exports = class RemoveGenerator extends Generator {
 					this.configuration.config.webpackOptions[propType] = null;
 				}
 			})
-			.then(() => {
+			.then((_: void) => {
```
ev1stensberg: returns promise

packages/generators/utils/entry.ts
```diff
@@ -51,16 +60,16 @@ module.exports = (self, answer) => {
...
-							validate
-						)
-					])
-				).then(entryPropAnswer => {
-					Object.keys(entryPropAnswer).forEach(val => {
+							validate,
+						),
+					]),
+				).then((entryPropAnswer: object) => {
+					Object.keys(entryPropAnswer).forEach((val: string) => {
```
ev1stensberg: returns void

packages/generators/utils/entry.ts
```diff
@@ -51,16 +60,16 @@ module.exports = (self, answer) => {
...
 							`What is the location of "${entryProp}"? [example: ./src/${entryProp}]`,
-							validate
-						)
-					])
-				).then(entryPropAnswer => {
-					Object.keys(entryPropAnswer).forEach(val => {
+							validate,
+						),
+					]),
+				).then((entryPropAnswer: object) => {
```
ev1stensberg: returns promise

packages/generators/utils/entry.ts
```diff
@@ -82,10 +91,12 @@ module.exports = (self, answer) => {
...
 					"singularEntry",
-					"Which module will be the first to enter the application? [default: ./src/index]"
-				)
+					"Which module will be the first to enter the application? [default: ./src/index]",
+				),
 			])
-			.then(singularEntryAnswer => {
+			.then((singularEntryAnswer: {
+				singularEntry: string,
+			}) => {
```
ev1stensberg: returns promise

packages/generators/utils/plugins.ts
```diff
@@ -1,14 +1,11 @@
...
  *
  * Callable function with the initial plugins
  *
- * @param {Void} _ - void value
  * @returns {Function} An function that returns an array
  * that consists of the uglify plugin
  */
 
-module.exports = _ => {
+export default function(): string[] {
```
ev1stensberg: `(_?:void)`

packages/webpack-scaffold/index.ts
```diff
@@ -1,17 +1,23 @@
 import * as jscodeshift from "jscodeshift";
 
-interface IScaffoldBaseObject {
-	type: string;
+export interface IScaffoldBaseObject {
+	type?: string;
 	name: string;
 	message: string;
+	choices?: ((answers: Object) => any) | any[];
```
ev1stensberg: returns array of strings

packages/generators/utils/plugins.ts
ematipico: It's not a promise, so `_` should be removed. Can we have a eslint rule?

ev1stensberg: `_` on empty params please http://po-ru.com/diary/rubys-magic-underscore/

packages/generators/utils/entry.ts
ev1stensberg: Every promise returns a promise https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then

dhruvdutt: Take a look at this code (from `master`) where we are returning non-Promise / string.

https://github.com/webpack/webpack-cli/blob/617af0bddb5140c07b1a0079b377311b10dd027d/packages/generators/utils/entry.js#L88-L97

There are many such examples where we aren't returning Promise. What should we do in such cases?

packages/generators/utils/plugins.ts
ematipico: That's Ruby, that treats them differently, what's your point?
Also, underscore is used as temporary placeholder in libraries such as Ramda, where it has a different meaning. My point is, what's the heff meaning of this underscore in this project? Where's documented? Why there's no eslint rule to tell me that the developer has to put in certain cases?

packages/generators/utils/entry.ts
ematipico: I think there's some confusion here. The `.then`  returns a promise but the actual callback called inside the `.then` returns whatever we want. 

So if we have 

```js

Const p = promise.then(_ => 5)
```
p is of the type `Promise<Number>` and the callback returns a type of `Number`.

To clarify, the callback doesn't return a promise

ev1stensberg: Okay. Emanuelle cleared it out. Add types for those though ;)

packages/generators/utils/plugins.ts
ev1stensberg: Responded this internally. Needs an issue to be in the desc of the `update docs` issue

packages/generators/add-generator.ts
```diff
@@ -41,9 +49,9 @@ function replaceAt(string, index, replace) {
  * @returns	{Boolean} hasProp - Boolean indicating if the property
  * is present
  */
-const traverseAndGetProperties = (arr, prop) => {
-	let hasProp = false;
-	arr.forEach(p => {
+const traverseAndGetProperties = (arr: object[], prop: string): boolean => {
+	let hasProp: boolean = false;
+	arr.forEach((p: object) => {
```
ev1stensberg: returns void

packages/generators/add-generator.ts
```diff
@@ -61,12 +69,12 @@ const traverseAndGetProperties = (arr, prop) => {
...
  *
  */
-const searchProps = (answers, input) => {
+const searchProps = (answers: object, input: string): Promise<string[]> => {
 	input = input || "";
 	return Promise.resolve(
-		PROPS.filter(food =>
-			food.toLowerCase().includes(input.toLowerCase())
-		)
+		PROPS.filter((prop: string) =>
```
ev1stensberg: missing ret type for filter

packages/generators/add-generator.ts
```diff
@@ -176,19 +204,19 @@ module.exports = class AddGenerator extends Generator {
 						}
 						// TODO: manually implement stats and devtools like sourcemaps
 					} else if (hasEnumProp) {
-						const originalPropDesc = defOrPropDescription[0].enum;
+						const originalPropDesc: object = defOrPropDescription[0].enum;
 						// Array -> Object -> Merge objects into one for compat in manualOrListInput
 						defOrPropDescription = Object.keys(defOrPropDescription[0].enum)
-							.map(p => {
+							.map((p: string) => {
```
ev1stensberg: missing ret on map

packages/generators/add-generator.ts
```diff
@@ -176,19 +204,19 @@ module.exports = class AddGenerator extends Generator {
...
 									{
-										[originalPropDesc[p]]: "noop"
-									}
+										[originalPropDesc[p]]: "noop",
+									},
 								);
 							})
-							.reduce((result, currentObject) => {
-								for (let key in currentObject) {
+							.reduce((result: object, currentObject: object) => {
```
ev1stensberg: missing return for reduce

packages/generators/add-generator.ts
```diff
@@ -286,54 +318,55 @@ module.exports = class AddGenerator extends Generator {
...
 				if (action === "plugins") {
-					const pluginExist = glob
+					const pluginExist: string = glob
 						.sync([
 							"node_modules/webpack/lib/*Plugin.js",
-							"node_modules/webpack/lib/**/*Plugin.js"
+							"node_modules/webpack/lib/**/*Plugin.js",
 						])
-						.map(p =>
+						.map((p: string) =>
```
ev1stensberg: map missing return

packages/generators/add-generator.ts
```diff
@@ -286,54 +318,55 @@ module.exports = class AddGenerator extends Generator {
...
 							p
 								.split("/")
 								.pop()
-								.replace(".js", "")
+								.replace(".js", ""),
 						)
 						.find(
-							p => p.toLowerCase().indexOf(answerToAction.actionAnswer) >= 0
+							(p: string) => p.toLowerCase().indexOf(answerToAction.actionAnswer) >= 0,
```
ev1stensberg: missing return type

packages/generators/add-generator.ts
```diff
@@ -286,54 +318,55 @@ module.exports = class AddGenerator extends Generator {
...
-						const pluginsSchemaPath = glob
+						const pluginsSchemaPath: string = glob
 							.sync([
 								"node_modules/webpack/schemas/plugins/*Plugin.json",
-								"node_modules/webpack/schemas/plugins/**/*Plugin.json"
+								"node_modules/webpack/schemas/plugins/**/*Plugin.json",
 							])
 							.find(
-								p =>
+								(p: string) =>
```
ev1stensberg: missing return

packages/generators/add-generator.ts
```diff
@@ -286,54 +318,55 @@ module.exports = class AddGenerator extends Generator {
...
 								? require(resolvePluginsPath)
 								: null;
-							let pluginsSchemaProps = ["other"];
+							let pluginsSchemaProps: string[] = ["other"];
 							if (pluginSchema) {
 								Object.keys(pluginSchema)
-									.filter(p => Array.isArray(pluginSchema[p]))
-									.forEach(p => {
-										Object.keys(pluginSchema[p]).forEach(n => {
+									.filter((p: string) => Array.isArray(pluginSchema[p]))
```
ev1stensberg: missing return type

packages/generators/add-generator.ts
```diff
@@ -286,54 +318,55 @@ module.exports = class AddGenerator extends Generator {
...
+									.forEach((p: string) => {
+										Object.keys(pluginSchema[p]).forEach((n: string) => {
```
ev1stensberg: missing ret on foreach x2

packages/generators/add-generator.ts
```diff
@@ -372,39 +409,40 @@ module.exports = class AddGenerator extends Generator {
...
-								console.error(
-									answerToAction.actionAnswer,
-									"doesn't exist on NPM or is built in webpack, please check for any misspellings."
-								);
-								process.exit(0);
-							}
-						});
+						npmExists(answerToAction.actionAnswer)
+							.then((p: string) => {
+								if (p) {
```
ev1stensberg: missing ret on npmexists

packages/generators/addon-generator.ts
```diff
@@ -24,54 +26,57 @@ const copyUtils = require("@webpack-cli/utils/copy-utils");
...
+		public default() {
+			const currentDirName: string = path.basename(this.destinationPath());
 			if (currentDirName !== this.props.name) {
 				this.log(`
 				Your project must be inside a folder named ${this.props.name}
 				I will create this folder for you.
 				`);
-				mkdirp(this.props.name);
-				const pathToProjectDir = this.destinationPath(this.props.name);
+				mkdirp(this.props.name, (err: object) => {
```
ev1stensberg: missing ret type

packages/generators/addon-generator.ts
```diff
@@ -24,54 +26,57 @@ const copyUtils = require("@webpack-cli/utils/copy-utils");
...
 			copyTemplateFiles.forEach(this.copyTpl);
 		}
 
-		install() {
+		public install() {
 			this.npmInstall(["webpack-defaults", "bluebird"], {
-				"save-dev": true
-			}).then(() => {
+				"save-dev": true,
+			}).then((_: void) => {
```
ev1stensberg: missing ret type

packages/generators/loader-generator.ts
```diff
@@ -29,13 +30,13 @@ function makeLoaderName(name) {
...
-			name: "name",
-			message: "Loader name",
 			default: "my-loader",
 			filter: makeLoaderName,
-			validate: str => str.length > 0
-		}
+			message: "Loader name",
+			name: "name",
+			type: "input",
+			validate: (str: string) => str.length > 0,
```
ev1stensberg: missing ret type

packages/generators/loader-generator.ts
```diff
@@ -47,13 +48,10 @@ const LoaderGenerator = addonGenerator(
 		"examples/simple/webpack.config.js.tpl",
 		"examples/simple/src/index.js.tpl",
 		"examples/simple/src/lazy-module.js.tpl",
-		"examples/simple/src/static-esm-module.js.tpl"
+		"examples/simple/src/static-esm-module.js.tpl",
 	],
 	["src/_index.js.tpl"],
-	gen => ({ name: gen.props.name })
+	(gen: IYeoman) => ({ name: gen.props.name }),
```
ev1stensberg: missing ret type

packages/generators/plugin-generator.ts
```diff
@@ -13,13 +14,13 @@ const addonGenerator = require("./addon-generator");
...
-			name: "name",
-			message: "Plugin name",
 			default: "my-webpack-plugin",
 			filter: _.kebabCase,
-			validate: str => str.length > 0
-		}
+			message: "Plugin name",
+			name: "name",
+			type: "input",
+			validate: (str: string) => str.length > 0,
```
ev1stensberg: missing return type

packages/generators/plugin-generator.ts
```diff
@@ -28,12 +29,10 @@ const PluginGenerator = addonGenerator(
 		"test/functional.test.js.tpl",
 		"examples/simple/src/index.js.tpl",
 		"examples/simple/src/lazy-module.js.tpl",
-		"examples/simple/src/static-esm-module.js.tpl"
+		"examples/simple/src/static-esm-module.js.tpl",
 	],
 	["src/_index.js.tpl", "examples/simple/_webpack.config.js.tpl"],
-	gen => ({ name: _.upperFirst(_.camelCase(gen.props.name)) })
+	(gen: IYeoman) => ({ name: _.upperFirst(_.camelCase(gen.props.name)) }),
```
ev1stensberg: missing return type

packages/generators/remove-generator.ts
```diff
@@ -0,0 +1,138 @@
...
+	public getPropTypes(): string[] {
+		return Object.keys(this.webpackOptions);
+	}
+
+	public getModuleLoadersNames(): string[] {
+		if (typeof this.webpackOptions === "object") {
+			if (this.webpackOptions.module && this.webpackOptions.module.rules) {
+				return this.webpackOptions.module.rules.map((rule: {
+					loader: string;
+				}) => rule ? rule.loader : null);
```
ev1stensberg: missing ret type

packages/generators/utils/entry.ts
```diff
@@ -12,27 +15,33 @@ const validate = require("./validate");
...
+				entryIdentifiers = multipleEntriesAnswer.multipleEntries.split(",");
+
+				function forEachPromise(entries: string[], fn: (entryProp: string) => Promise<{} | void>) {
+					return entries.reduce((promise: Promise<{}>, prop: string) => {
+						const trimmedProp: string = prop.trim();
+
+						return promise.then((n: object) => {
 							if (n) {
-								Object.keys(n).forEach(val => {
+								Object.keys(n).forEach((val: string) => {
```
ev1stensberg: missing return type

packages/generators/utils/module.ts
```diff
@@ -0,0 +1,36 @@
...
+}
+
+/**
+ *
+ * Returns an module.rule object that has the babel loader if invoked
+ *
+ * @returns {Function} A callable function that adds the babel-loader with env preset
+ */
+export default function(): IModule {
+	return {
```
ev1stensberg: returns obj

packages/generators/utils/validate.ts
```diff
@@ -8,10 +6,10 @@
  * @returns {String | Boolean } Returns truthy if its long enough
  * Or a string if the user hasn't written anything
  */
-module.exports = value => {
+export default function validate(value: string): string | boolean {
 	const pass = value.length;
```
ev1stensberg: missing type on const

packages/generators/add-generator.ts
dhruvdutt: `npmExists` returns a Promise whose types are added in _/utils_ PR. This should be implicitly covered when we merge all PRs. :)

packages/generators/addon-generator.ts
dhruvdutt: We aren't storing whatever `mkdirp` returns then why and where to add types?

packages/generators/remove-generator.ts
dhruvdutt: Type is already present as an object.

packages/generators/utils/module.ts
dhruvdutt: Yeah! `interface IModule extends Object` :)

PR #559, opened 2018-08-01
allow array value for --ouput-library

### LINKED ISSUE #557: --output-library does not accept arrays
**Describe the bug**

Unable to pass an array-value to `--output-library`.

```
webpack --output-library-target window --output-library key1 key2 ./src/main.js
webpack --output-library-target window --output-library key1 --output-library key2 ./src/main.js
```

At least one of these formats should pass `["key1", "key2"]` to the `output.library` option. However, only one value can be passed.

output is currently something like
```
window.key2=...
```

**Expected behavior**

should be
```
window.key1=window.key1||{},window.key1.key2=...
```

Hoten:
Fixes #557.


ematipico:
Hi @Hoten, thanks for your PR!
I wonder if you could add another test where we pass just one argument as `--output-library`? To make sure we don't break existing functionality (not tested, as far as I can see).

bin/convert-argv.js
```diff
@@ -441,7 +441,10 @@ module.exports = function(...args) {
 
 		ifArg("output-library", function(value) {
 			ensureObject(options, "output");
-			options.output.library = value;
+			if (typeof options.output.library === "undefined") {
```
ev1stensberg: I think you can do `if (!options.output.library)` here

Hoten:
Done.

ev1stensberg:
Hi @Hoten , could you re-open in a new PR? Things like this needs a better review before merging. @dhruvdutt shouldn't really merge PR's that are unapproved by all members as he is a student for us during the summer..

dhruvdutt:
My bad. Let's take a thorough relook. Can you please open a new PR @Hoten?

Hoten:
Opened a new PR.

PR #562, opened 2018-08-02
Output lib array opt

### LINKED ISSUE #557: --output-library does not accept arrays

Hoten:
Fixes #557


Hoten:
I just left for vacation and don't have access to my machine. And to be frank, I find the extra hoop here to make a simple contribution to be cumbersome. Is there a simple way within GitHubs workflow that you could make these sorts of (to people outside of the project maintenance mindset, annoying) changes?

ematipico:
Don't worry, we can rewrite the message of the commit while we do the merge onto develop!

bin/convert-argv.js
```diff
@@ -441,7 +441,10 @@ module.exports = function(...args) {
 
 		ifArg("output-library", function(value) {
 			ensureObject(options, "output");
-			options.output.library = value;
+			if (!options.output.library) {
```
ev1stensberg: Default value could be an empty array from `yargs`

test/binCases/output/output-library-many/stdin.js
```diff
@@ -0,0 +1,12 @@
+"use strict";
+
+module.exports = function testAssertions(code, stdout, stderr) {
+	expect(code).toBe(0);
+	expect(stdout).toEqual(expect.anything());
+	expect(stdout[5]).toContain("bundle.js");
+	expect(stdout[7]).toMatch(/index\.js.*\{0\}/);
+	expect(stderr).toHaveLength(0);
+
+	const output = require("fs").readFileSync(require("path").join(__dirname, "../../../js/bin/output/output-library-many/bundle.js"), "utf-8");
```
ev1stensberg: import and destruct at top level instead

test/binCases/output/output-library-single/test.opts
```diff
@@ -0,0 +1,5 @@
+./index.js
+-o ../../../js/bin/output/output-library-single/bundle.js
```
ev1stensberg: Don't think this is needed, what's the syntax in the other tests?

bin/convert-argv.js
Hoten: Just tried this, does not work. Default value is not set if a value is given via the CLI - it isn't an initial value. However I did replace this code w/ `ensureArray`

test/binCases/output/output-library-many/stdin.js
Hoten: done

test/binCases/output/output-library-single/test.opts
Hoten: removed. only the original output test used this flag, I just copied from there.

Hoten:
Addressed comments, pushed changes + squashed to one commit.

Hoten:
I also wanted a way to quickly run just my tests, so I introduced an env filter for binCases

```
BIN_TEST_CASES_GREP=output npx jest
```

test/binCases/output/output-library-single/stdin.js
```diff
@@ -0,0 +1,14 @@
...
+const fs = require("fs");
+const path = require("path");
+
+module.exports = function testAssertions(code, stdout, stderr) {
+	expect(code).toBe(0);
+	expect(stdout).toEqual(expect.anything());
+	expect(stdout[7]).toMatch(/index\.js.*\{0\}/);
+	expect(stderr).toHaveLength(0);
+
+	const output = fs.readFileSync(path.join(__dirname, "../../../js/bin/output/output-library-single/main.js"), "utf-8");
```
ev1stensberg: this should probably be in an own variable

test/binCases/output/output-library-many/stdin.js
```diff
@@ -0,0 +1,14 @@
...
+const fs = require("fs");
+const path = require("path");
+
+module.exports = function testAssertions(code, stdout, stderr) {
+	expect(code).toBe(0);
+	expect(stdout).toEqual(expect.anything());
+	expect(stdout[7]).toMatch(/index\.js.*\{0\}/);
+	expect(stderr).toHaveLength(0);
+
+	const output = fs.readFileSync(path.join(__dirname, "../../../js/bin/output/output-library-many/main.js"), "utf-8");
```
ev1stensberg: could you abstract this a lil better? Quite long line and slashes that could be trimmed down using `path.resolve`

test/binCases/output/output-library-many/stdin.js
```diff
@@ -0,0 +1,14 @@
...
+	expect(output).toContain("window.key1=window.key1||{},window.key1.key2=function");
```
ev1stensberg: Maybe match this assertion using the array indice and regexes

test/binCases/output/output-library-single/stdin.js
ev1stensberg: See prev comment 👍

test/binCases/output/output-library-many/stdin.js
```diff
@@ -0,0 +1,14 @@
+"use strict";
+
+const fs = require("fs");
```
ev1stensberg:
```js
const {readFileSync} = require('fs')
const {join} = require('path')


const output = readFileSync(join(__dirname, __dirname, "..", "..", ".., "js", "bin", "output", output-library-many", "main.js"), "utf-8")
```

test/binCases/output/output-library-many/stdin.js
ev1stensberg: `expect(stdout[myIndices]).toMatch(/window.key1=window.key/*\{0\}/);`

test/binCases/output/output-library-single/stdin.js
ev1stensberg: se prev comment

test/binCases/output/output-library-single/stdin.js
```diff
@@ -0,0 +1,14 @@
...
+	expect(output).toContain("window.key1=function");
```
ev1stensberg: see prev comment

ev1stensberg:
Seems like tests aren't failing. @sokra is this valid?

sokra:
They ain't failing...?

ev1stensberg:
Did so locally on mine, but saw the CI afterwards. Was weird..

PR #573, opened 2018-08-24
chore(ci): remove commitlint

dhruvdutt:
CI change

N/A

N/A

**Summary**

Failing the CI on writing improper commit message is a very aggressive policy for new contributors. While merging PRs, we should ideally squash all the commits and edit the commit message following the git cz convention.

No

**Other information**


dhruvdutt:
I can't find a way to run a script only on `master` branch commit. If we specify `master` branch in the config, it also runs those script when a PR has `base` branch as master.

More info [here](https://docs.travis-ci.com/user/conditional-builds-stages-jobs/).

dhruvdutt:
Precommit hooks and failing the CI just because of an invalid commit message is very aggressive for new contributors.

We aren't removing git-cz, we're just removing the lint part. We can just clean pre-commit hooks and commitlint from the CI checks during PRs.

When a new contributor comes, while merging the PR we’ll write the commit message following the commitlint guideline.

ev1stensberg:
It's the linting part I want to keep.  

> There’s different reasons to keep code linting and I think the biggest ones that I want to elevate are that they allow us to write meaningful changelogs, allow people that browse our code to know better what each commit is doing and lastly it allows us to browse the git history more efficiently

ematipico:
This totally makes Even and totally agree with you, we have lot of tools to manage this part. Although the management of the commits **doesn't have to hit the contributors, especially the first ones**, we cannot lift this burden on them. 

We already have a code base that is difficult to tackle, we cannot make the small contributions even more difficult.

People that manage the project are aware of the rules and have tools to be safe:
- committint on pre release
- squash and rewrite commit message during merging
- CI on master

But our burden cannot be user's burden.

package.json
```diff
@@ -33,6 +33,7 @@
     "bundlesize": "bundlesize",
     "changelog": "conventional-changelog --config ./build/changelog-generator/index.js --infile CHANGELOG.md --same-file",
     "clean:all": "rimraf node_modules package-lock.json packages/*/node_modules packages/*/package-lock.json",
+    "commit": "git cz",
```
ematipico: I guess we want to keep the command

Issue #577, opened 2018-09-03
Bug: No types declarations of packages


rishabh3112:
**Describe the bug**
Couldn't import packages in typescript file

**What is the current behavior?**
After importing `webpack-scaffold` in a typescript file.
I cant use any of its prompt functions like `Confirm` , `Input`, etc. as it doesn't have any types declaration 

**Expected behavior**
Being able to import packages and use its functions and other methods

**Screenshots**
![untitled](https://user-images.githubusercontent.com/15812317/44983987-ce8c6f80-af98-11e8-86eb-fd91848543f5.png)

**Additional context**
simple fix is to set `"declaration": true` in `ts-config.json` file under complierOptions  and add folllowing in `package.json`
```json
types: '<filename>.d.ts'
```
refers: sendilkumarn/webpack-scaffold-pwa#59

ematipico:
Fixed in #578

rishabh3112:
When will this ship to npm @ematipico?

ev1stensberg:
If emanuelle has write access to npm he can do it now

PR #578, opened 2018-09-03
Types for packages

rishabh3112:
refers #577


ematipico:
Do they actually exist those files?

rishabh3112:
They don't that's why this PR is for @ematipico

rishabh3112:
Have set `'declaration': true` in `tsconfig.json`
when building it will create a declaration file named `<filename>.d.ts`. and `index.ts` was the file whose type declaration has to be there in the package so that's why setting types to `index.d.ts`.

rishabh3112:
If the whole codebase has been shifted to typescript
then can I remove `'allowJS': true ` from `tsconfig.json`
@ematipico

ematipico:
Unfortunately we still need it I think, probably for the tests? I think @dhruvdutt can answer your question in a better way

dhruvdutt:
Some parts of the codebase like the entrypoint folder is still in JS but this shouldn't affect the TS compilation process. We can remove probably `allowJs` after verifying the sanity. We have kept it because the TS migration happened incrementally.

About the `index.d.ts` file, I don't think it makes sense for all packages. IMO it only makes sense for `webpack-scaffold`. Maybe `webpack-utils` as well. We'll need to invest and verify the sanity of auto-generated interfaces as well. I'll probably take a look into this. Thanks for pointing it out. 🤗 

Also, just a friendly note to not override the PR template. You don't need to fill it completely, you can ignore the questions/fields which aren't relevant to your PR but please keep the template consistent. 🙂

ev1stensberg:
> About the index.d.ts file, I don't think it makes sense for all packages. IMO it only makes sense for webpack-scaffold.

Why?

dhruvdutt:
@ev1stensberg: `index.d.ts` file makes sense only for `webpack-scaffold` because that's the only package we expose outside / people can install, import and consume this packages inside their applications/packages.
Rest of all packages are getting used internally inside TypeScript itself. No need to create interfaces for that unless we intend to expose them outside for users to import and consume. Do we?

@rishabh3112: You can definitely work on this. I think it needs some inputs from my end, but I'm always open to review/suggest. :)

rishabh3112:
Okay, @dhruvdutt.
So should I rebase and export declarations it for `webpack-scaffold` only?

rishabh3112:
But I think exporting declarations would be better as it will allow scaling of the CLI for new packages (that would be written in TS) by reusing packages already there.

ematipico:
Thinking about it, we should export the types of every package. If need to import an interface from `utils`, doing 

```js
import { IInterface } from '../packages/utils/myUtil'
```

Is not correct, instead we should do

```js
import { IInterface } from '@webpack-cli/utils'
```

So yeah, I think this is the correct way of doing things. Doesn't matter if the package will be used internally or externally. 
I think inside your code base we still use the first signature.

dhruvdutt:
@ematipico In our codebase, we follow

```js
import { IInterface } from '@webpack-cli/utils'
```

@rishabh3112 Can you check if internal packages get TS definitions implicitly since they're communicating in TS itself? We only need `index.d.ts` file only when we're importing and consuming these API from outside, right?

ematipico:
```
import { IInterface } from '@webpack-cli/utils'
```
^ this is from the outside :)

rishabh3112:
Yes, @dhruvdutt, but packages may develop outside this repo too.
Nevertheless, I which direction we should go now?

dhruvdutt:
Cool. Let's keep it then. @rishabh3112 Go ahead. :shipit: 
Why can't we see `index.d.ts` files in PR diff? Isn't it added to git?

rishabh3112:
They are auto-generated thus no need to add to the repo @dhruvdutt.
**But remember to build before it got published on npm.**

rishabh3112:
I am finding a way for that @ev1stensberg 
May file copying while building works ?

ev1stensberg:
Maybe a solution that doesn’t require so much configuration would do good. Feel free to test out your solution, if it’s clean and is working then we’ll have a look 🙏

rishabh3112:
Would like to have a review now 😃

rishabh3112:
Is this solution fine @ev1stensberg or need to find some other solution.

packages/tsconfig.declarations.json
```diff
@@ -0,0 +1,6 @@
+{
```
ematipico: I think it's better to keep all the tsconfigs at the same level. Is there a reason why it's here? Would that be possible to move it inside the root folder?

rishabh3112:
As it is used when declaration of ts files is needed (i.e. ts files in packages) @ev1stensberg

ev1stensberg:
Could it be done without the suffix?

rishabh3112:
Yes.
What name do you want then?

ematipico:
Maybe `tsconfig.packages.json`? What do you guys think?

rishabh3112:
Why this PR was closed @ematipico?

ematipico:
Apologies! I might have hit the wrong button

ematipico:
> Could it be done without the suffix?

@ev1stensberg we need to have a specific configuration only for those packages that need to expose the interfaces, so we need to keep separated configurations.

About the naming, yes let's call it with `packages`, it helps to understand what's the configuration **for**. Other than that, I am really happy with this PR!

ev1stensberg:
Oki :)

rishabh3112:
Do I need to rebase here for commitlint to pass or you may reference the latest commit @ematipico?

dhruvdutt:
@rishabh3112 No need to stress about the commitlint. We'll rewrite the message during merge. We got it covered.

Can you please fix the conflicts?

ematipico:
> Can you please fix the conflicts?

PR is good to go :) no conflicts

rishabh3112:
As far as i can see it says  `branch has no conflicts with base branch` @dhruvdutt

dhruvdutt:
screen shot 2018-09-21 at 6 26 07 pm

PR #603, opened 2018-09-21
chore(fix): fix clean all script

sendilkumarn:

build related change 

NA
NA
**Summary**
Remove package-lock.json

No
**Other information**
NIL

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "build": "tsc",
     "bundlesize": "bundlesize",
     "changelog": "conventional-changelog --config ./build/changelog-generator/index.js --infile CHANGELOG.md --same-file",
-    "clean:all": "rimraf node_modules package-lock.json packages/*/node_modules packages/*/package-lock.json",
+    "clean:all": "rimraf node_modules packages/*/node_modules",
     "docs": "typedoc",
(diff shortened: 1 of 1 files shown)


ematipico:
Shouldn't we remove the clean:all command from the bootstrap?

ematipico:
Why?

sendilkumarn:
Bootstrap will help to link dependencies in the repo together and it is advisable to remove node modules and install dependencies when the link happens.

dhruvdutt:
@sendilkumarn Can you please link to docs or blog for a detailed explanation? I'm also slightly sceptical about the removal of node_modules.

sendilkumarn:
Okay here is the brief overview, I cannot find the exact ticket (or) issue where it was discussed. 

running `npm install` across packages / in the root cause some symlinks & installed dependencies across packages in `lerna`. This causes dependencies to break when you run `npm i` multiple times or even with concurrency in `npm i`

On the other hand, `npm i` will remove any `npm link` that is already in that package. As a matter of fact `lerna` depends mostly on the linked npm packages.

So it is advised to clean up `node_modules`, before running `bootstrap`. Take any package that is using lerna for example, it does clean the dependencies and recreate them. 

I am not sure, what is the loss here, of course when you are bootstrapping, you might need to install the packages (freshly).

PR #604, opened 2018-09-21
fix tslint warnings

sendilkumarn:

linting changes
NA
NA
**Summary**
we can switch off console.log warnings and removed that line length issue 

NA
**Other information**
NIL

### DIFF
--- a/tslint.json
+++ b/tslint.json
@@ -13,9 +13,7 @@
       "ban-types": [
         false
       ],
-      "no-console": {
-        "severity": "off"
-      }
+      "no-console": [true, "log"]
     },
     "rulesDirectory": [],
     "linterOptions": {
--- a/tslint.json
+++ b/tslint.json
@@ -14,7 +14,7 @@
         false
       ],
       "no-console": {
-        "severity": "warning"
+        "severity": "off"
       }
     },
     "rulesDirectory": [],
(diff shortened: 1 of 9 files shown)


tslint.json
```diff
@@ -14,13 +14,7 @@
         false
       ],
       "no-console": {
-        "severity": "warning"
-      },
-      "max-line-length": {
-        "options": {
-          "limit": 120
```
dhruvdutt: I think we should retain line length.

ematipico: I agree too. No need to remove it

tslint.json
```diff
@@ -14,13 +14,7 @@
         false
       ],
       "no-console": {
-        "severity": "warning"
```
ematipico: I would change the configuration to allow only `console.error`. And we should delete the `console.log` across the board

ev1stensberg: samesies

tslint.json
ev1stensberg: samesies

sendilkumarn: Isn't this part of `tslint:recommended` https://github.com/palantir/tslint/blob/237fd905a9fb3c9ca9325d2da6e8145e4a44fe51/src/configs/recommended.ts#L63

PR #584, opened 2018-09-13
Feature/package test ts

ematipico:
**What kind of change does this PR introduce?**
First typescript test 

**Did you add tests for your changes?**
Yes 

**If relevant, did you update the documentation?**
No

**Summary**
Added first typescript test, really small

**Does this PR introduce a breaking change?**
No

**Other information**

PR #605, opened 2018-09-22
fix: resolve references in schema

sokra:
avoid accessing definitions directly

bugfix

existing test should be fine

not relevant

**Summary**
refactor the way of accessing webpack's schema to not create a internal dependency on the exact structure of the schema.

No

**Other information**
see also https://github.com/webpack/webpack/pull/7232

### DIFF
--- a/bin/config-yargs.js
+++ b/bin/config-yargs.js
@@ -1,4 +1,58 @@
 const optionsSchema = require("./optionsSchema.json");
+
+const nestedProperties = ["anyOf", "oneOf", "allOf"];
+
+const resolveSchema = schema => {
+	let current = schema;
+	if (schema && typeof schema === "object" && "$ref" in schema) {
+		const path = schema.$ref.split("/");
+		for (const element of path) {
+			if (element === "#") {
+				current = optionsSchema;
+			} else {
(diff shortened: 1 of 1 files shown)


ev1stensberg:
@ematipico any reason for the fix prefix failing?

bin/config-yargs.js
```diff
@@ -1,4 +1,58 @@
...
+		const path = schema.$ref.split("/");
+		for (const element of path) {
+			if (element === "#") {
+				current = optionsSchema;
+			} else {
+				current = current[element];
+			}
+		}
+	}
+	return schema;
```
sendilkumarn: Is this `current`?

ematipico:
> @ematipico any reason for the fix prefix failing?

Still broken, will fix in another PR

dhruvdutt:
@ematipico Same CI issue is happening for #602 as well. Seems to be something related to `conventional-changelog-lint-config-cz`.

Could be happening because due to removal of `cz-conventional-changelog-lint` package.

ematipico:
Yes, let's discuss it here #584

PR looks good to me!

PR #620, opened 2018-09-30
chore: update contributing docs, add script

### LINKED ISSUE #619: DOCS: contributing setup: rimraf dep needed before running npm bootstrap
**Describe the bug**

If the user does not have ``rimraf`` installed before running ``npm bootstrap``, 
``sh: 1: rimraf: not found`` error is thrown. 

This dependency should be specified in https://github.com/webpack/webpack-cli/blob/master/CONTRIBUTING.md.

dhruvdutt:
Docs and guide update

N/A

Yes

**Summary**
Simplifies setup process, add `postinstall` script to bootstrap packages.

No

**Other information**
Closes #619

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,7 +33,7 @@ In case you are suggesting a new feature, we will match your idea with our curre
 * `git clone <your-clone-url> && cd webpack-cli`
 
 ### Setup with npm
-* Install the dependencies and bootstrap the packages: `npm run bootstrap`
+* Install the dependencies and bootstrap the packages: `npm install`
 * Run the tests with: `npm test`
 
 ### Setup with yarn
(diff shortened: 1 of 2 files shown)


manu-chroma:
Hey, I am just curious that if we run ``bootstrap`` as the ``postinstall`` script, 
won't ``bootstrap`` (using ``clean:all``) cleanup all the packages we just installed and re-install them again? 
I understand it will use the local cache 2nd time around but still why the redundancy?

ev1stensberg:
To run bootstrap one needs to do that to prevent lerna from making a corrupt installation from what I understood

SimenB:
Tried to run install in this repo, and the fact that `postinstall` calls `bootstrap` which calls `install` which calls `postinstall` when done, means the installation is currently in an infinite loop.

ev1stensberg:
Fixed now, sorry for the inconvenience

dhruvdutt:
Sorry for the inconvenience, guys. Should have cleaned up the script.

Thank you very much @ev1stensberg for fixing it. 💙

ev1stensberg:
I didnt fix it, a rm rf needs to be done I think, was a quick fix, but 2 lines with same code is redundant, should probably make a «clean» script and use that instead

PR #622, opened 2018-09-30
Refactor npmPackagesExists() function

styfle:
Refactoring `npmPackagesExists()` function

No, there is no API change

No

**Summary**

This makes the code more readable and avoids checking the boolean twice

No 

**Other information**


ematipico:
@ev1stensberg could you explain? I don't see how: an error is always thrown

ev1stensberg:
The initial idea is to run this on multiple packages, so you can combine different scaffolds. I don't know if it still works ( might need some tweaking ).

How is this handled now? How is the error thrown? I think we got a setup for it. We need to look at potential errors this might raise.

PR #621, opened 2018-09-30
chore(tests): PoC on new way of running integration tests

SimenB:
**What kind of change does this PR introduce?**
An alternate way of running integration tests that are based on multiple test files, instead of one large one. This allows you to use Jest's filtering and watch mode.

**Did you add tests for your changes?**
Heh...

**If relevant, did you update the documentation?**
N/A

**Summary**
See https://github.com/facebook/jest/issues/7076

This is a quick start. We can do more for e.g. options, we can make it async instead of sync etc.. This is also missing your nyc wrapper, should be easy enough to re-add, though (I want https://github.com/facebook/jest/issues/5274 to be a thing...)

**Does this PR introduce a breaking change?**

Nope

**Other information**
This is very much in progress. I don't know if you wanna migrate all tests before merging, or do it piece by piece. I'm happy to help out, but I think you should set it up with the abstractions you want yourself. And the correct abstraction often reveal themselves as you migrate more and more tests.

PR #632, opened 2018-10-09
[WIP] Better test suite for binary tests

ematipico:

Refactor of the test bin cases in order to run them singularly. Also, it implements snapshot tests

Well, kinda :)

Not yet, probably I will once everything is ready

**Summary**
So, taken from #621 I started the refactor of the bin tests. 
As suggested, I removed from the stream of the output everything that has the Hash, timestamps and versions. This is because they change at every build/version change of webpack, causing the fail of the snapshot test.

Also, as suggested, we could change the way we pass the arguments from being an array to being an object. Feedbacks are welcom

**Other information**
Let's have some feedback


ematipico:
Can't push on a branch that is not mine :(

ematipico:
#635 which continues Simen's PR

PR #635, opened 2018-10-10
Better test suite for binary tests

ematipico:

Refactor of the test bin cases in order to run them singularly. Also, it implements snapshot tests

Well, kinda :)

Not yet, probably I will once everything is ready

**Summary**
So, taken from #621 I started the refactor of the bin tests. 
As suggested, I removed from the stream of the output everything that has the Hash, timestamps and versions. This is because they change at every build/version change of webpack, causing the fail of the snapshot test.

Also, as suggested, we could change the way we pass the arguments from being an array to being an object. Feedbacks are welcom

**Other information**
Let's have some feedback


ev1stensberg:
Would be nice if we can remove async timeouts and try to make jest decide it

ematipico:
That's the thing. In these tests we are running the watcher and I followed the same pattern of before, where we kill the process after a while, collect the result and after that we run the assertions. Do you have a better solution?

test/testUtils.js
```diff
@@ -0,0 +1,65 @@
...
+}
+
+function runWatch(testCase, args = []) {
+	const cwd = path.resolve(testCase);
+
+	const outputPath = path.resolve(testCase, "bin");
+	const argsWithOutput = args.concat("--output-path", outputPath);
+
+	return new Promise(resolve => {
+		// try {
```
SimenB: You can do `Promise.resolve().then(() => execa(...))....` to avoid doing `new Promise`. Not sure if it's better or not

test/binCases/watch/single-config/single-config.test.js
```diff
@@ -0,0 +1,30 @@
+"use strict";
+
+const { runWatch } = require("../../../testUtils");
+jest.setTimeout(10000);
+
+test("single-config", done => {
+	runWatch(__dirname, [
```
SimenB: you can return the promise instead of using `done`, then jest will wait for it

ematipico:
Hi @SimenB , thanks for your feedback!! Really appreciated! I tried to follow your suggestions but still we got issue inside the CI. Probably there's something that I am still missing...

So, I kill the process when there's some output (like you suggested where you wait for some kind of error). When there's output, we kill the process.

Inside jests tests, I return the promise and in theory jest should wait for it. Although on Travis it keeps failing with a timeout error, so Jest it's not waiting properly... any tips?

SimenB:
Jest has a timeout of 5 seconds. If your test takes longer than that, you have to increase the timeout.

Also note that you have to remove the `done` callback argument when you return the promise. If your callback takes an argument, jest will wait for that argument to be invoked

ev1stensberg:
What Simen proposed was something I was looking for. As we don't need to wait in a fixed interval for each compilation to finish. An ETA of each compilation would be < 15s and I think 60 as a fixed limit for any compilation would be useless when we can await for output. 

What would be ideal is not to spawn processes in sequence, but in parallel, as each test is isolated

ematipico:
@ev1stensberg Yes it would be super awesome, but on Travis it looks like that the process takes A LOT print out some output and because of that the test fails :(

I tried now to increase the jest timeout to 30 seconds and see how it goes.

SimenB:
Oh btw, make sure to set `maxWorkers` on Travis of you haven't already.

ematipico:
What's missing is finding right way to make the watch tests working on the CI.

ev1stensberg:
If we can edit the code used by the test to trigger watch we can do it programatically. Read and write to the js file and return after the compilation is done. Or we can make use of async await. Or all above :)

SimenB:
CI still does `npm test`, it doesn't run where you've defined your `maxWorkers`. Fixing that should speed up the build

ev1stensberg:
@ematipico do you have access to travis, or do you need it?

evenstensberg:
@simenB does this look ok?

evenstensberg:
Env variables for CIs needs to be adjusted but is there anything else?

test/binCases/config-name/found-many/found-many.test.js
```diff
@@ -0,0 +1,27 @@
...
+		"--output-chunk-filename",
+		"[id].chunk.js",
+		"--target",
+		"async-node",
+		"--mode",
+		"production"
+	]);
+
+	expect(code).toBe(0);
+	expect(stdout).toEqual(expect.anything());
```
SimenB: Unnecessary assertion

test/binCases/config-name/found-many/found-many.test.js
```diff
@@ -0,0 +1,27 @@
...
+	expect(stdout).toContain("./index2.js");
+	expect(stdout).toContain("./index3.js");
+	expect(stderr).toHaveLength(0);
```
SimenB: `toBe('')`? Or `toEqual`

test/binCases/configFile/plugins-precedence/__snapshots__/plugins-precedence.test.js.snap
```diff
@@ -0,0 +1,13 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`plugins-precedence 1`] = `
+"  Asset       Size  Chunks             Chunk Names
+null.js  966 bytes       0  [emitted]  null
+Entrypoint null = null.js
+[0] ./index.js 265 bytes {0} [built]
+[1] ./ok.js 23 bytes {0} [built]
+
+WARNING in configuration
```
SimenB: In Jest we have a helper called `extractSummary` which might be nice here. Separates `summary` from `rest`

test/binCases/watch/info-verbosity-off/info-verbosity-off.test.js
```diff
@@ -0,0 +1,30 @@
+"use strict";
+
+const { runWatch } = require("../../../testUtils");
+
+test("info-verbosity-off", () => {
+	return runWatch(__dirname, [
```
SimenB: async-await?

ematipico: We still support node 6 and we don't use Babel. I would have used already if I could

SimenB: Jest comes with babel support out of the box (that's how code coverage works), so you can just add `babel-preset-env` and use it in tests if you want, even if you don't want to transpile your source code. Up to you, though 🙂

evenstensberg: Preset env is doable, but will it throw on node v6 and below?

SimenB: nah, just use `node: 'current'` as `target`

ematipico: I wouldn't mind to use babel for your test suite honestly

evenstensberg:
Does anyone know how we can avoid the jest timeout invocation in Travis and Azure?

SimenB:
Increase the timeout? CI is generally really slow unless you pay money for more cpu/ram

evenstensberg:
There we go. Could you elaborate on `extractSummary`?

ematipico:
We can have that extractSummary utility later. I'd like to get this done. We should increase the timeout to 60 seconds as before

SimenB:
`extractSummary` in Jest is just a function that splits out the summary from everything else in `stdout`, allowing you to make more granular assertions about the output that's unique to a test. Mostly cosmetics, so no need to wait for it to merge this PR 🙂

SimenB:
:tada:

PR #650, opened 2018-10-18
feat(gui): foundation for webpack-cli

rishabh3112:
**What kind of change does this PR introduce?**

Foundation for UI feature
**Did you add tests for your changes?**
No
**If relevant, did you update the documentation?**
Yes
**Summary**
Create the foundation for `ui`

**Does this PR introduce a breaking change?**

**Other information**
refer README.md for more info
refers #647
Closes #647

Issue #647, opened 2018-10-16
[UI] Decide foundation of the project


ematipico:
Before starting implementing features, there's need to agree on the tools to use to build the UI tool.

We have to remember that this is a tool that will interact A LOT with the `webpack-cli` itself and some of the packages that we currently maintain. For instances, if we want to add a new loader, once the user decided which loader to add, we can run the command `webpack-cli add` or calling the method straightaway.

So there are two parts that to be developed, the server and the client.

Here's my 2cents:
- Server: `express`
- Frontend: `create-react-app` or if typescript, using babel for compilation. I don't trust the typescript compiler (and people are moving away from it)

About the styling of the site, I don't have particular inclination but usually I tend to choose solutions that are javascript-free (like `bulma`), but the I don't mind other choices.

For testing, we can stick with Jest.

Surely I have missed a lot, so feel free to add suggestions

ev1stensberg:
I think at minimum req would be to use webpack dev server and that tool suite.

renatoagds:
Agreed with @ev1stensberg. we should use webpack-dev-server. other point, I would like to follow with `material-ui` as our styling lib. since we're going in top of React, we get an lib that offer less work, providing the components, will reduce the amount of work.

For tests I would like to add `enzyme` too, so we could test our components.

About the tool itself, IMO, we should follow it as a CLI mirror, just reproducing the commands that we already have in CLI, based in that, our tasks could be:

- [ ] Project Structure - Setup `create-react-app`, and generate our in final of this, we should have the `create-react-app` initial screen running.
- [ ] Init - A screen, that will reproduce the same as `webpack-cli init` (same questions, etc).
- [ ] Add - A screen, that will reproduce the same as `webpack-cli add`.
- [ ] Info - A screen, that will reproduce the same as `webpack-cli info`.

I think we could have this as our v1/MVP of Webpack GUI. Am I missing something?

rishabh3112:
So, Are we starting off with JS first 🤔

ematipico:
I am not a big supporter of `enzyme` because it's too much dependent to the React versions. So there's the risk that can't upgrade React too frequently because `enzyme` might not support the new changes and can't keep up. Also the APIs are too wide. I am more keen to use the simple `ReactDOM` or `react-testing-library`.

[...]

PR #651, opened 2018-10-18
feat(ui): foundation for webpack-cli ui

rishabh3112:

GUI for webpack-cli
No
Yes
**Summary**

**Other information**
refer README.md under `ui/` to get started contributing
refers #647 
closes #650


packages/ui/public/README.md
```diff
@@ -0,0 +1,2567 @@
+This project was bootstrapped with [Create React App](https://github.com/facebook/create-react-app).
```
dhruvdutt: Clean this file.

packages/add/tsconfig.json
```diff
@@ -1,3 +1,3 @@
-{
-  "extends": "../../tsconfig.packages.json"
-}
+{
```
dhruvdutt: What's this diff? I don't understand.

package.json
```diff
@@ -148,6 +148,6 @@
     "typedoc-plugin-monorepo": "^0.1.0",
     "typescript": "^2.9.2",
     "webpack": "^4.19.1",
-    "webpack-dev-server": "^3.1.8"
+    "webpack-dev-server": "^3.1.9"
```
dhruvdutt: Please don't mix up separate concerns in a single PR. I'll do with the typescript upgrade.

greenkeeper.json
```diff
@@ -0,0 +1,22 @@
+{
+  "groups": {
+    "default": {
+      "packages": [
```
dhruvdutt: Rebase from master. This commit has been already merged.

package.json
rishabh3112: Later version of webpack dev server was needed for ui so upgraded it 👀

packages/ui/public/README.md
rishabh3112: Okay 👍

greenkeeper.json
rishabh3112: I rebased from webpack-cli master but can't figure out why this commit pops up.

packages/add/tsconfig.json
rishabh3112: May be a `LF`/ `CRLF` issue

package.json
rishabh3112: Nothing related to typescript here

dhruvdutt: I meant about version upgrading. Oh, if it's needed then alright. 👍

rishabh3112:
Is that structure fine @ematipico?

packages/ui/app.js
```diff
@@ -0,0 +1,18 @@
+var express = require("express");
```
ematipico:
```suggestion
const express = require("express");
```

packages/ui/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "@webpack-cli/ui",
+  "version": "0.0.0",
+  "private": true,
+  "scripts": {
+    "start": "node ./bin/www"
+  },
+  "dependencies": {
+    "cookie-parser": "~1.4.3",
```
ematipico: Ideally we could add `body-parser`. For the next step :)

packages/ui/bin/www
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/env node
+
+/**
+ * Module dependencies.
+ */
+
+var app = require('../app');
```
ematipico:
```suggestion
const app = require('../app');
```

ematipico:
Could you try to understand why Travis fails by not finding a package?

rishabh3112:
We have to update `travis-ci` script to install packages in `packages/ui/public`
@ematipico if you say I will do that here itself 😀

packages/ui/package.json
rishabh3112: You mean add it as dependency 🤔

packages/ui/app.js
rishabh3112: Sure will do every where possible in this file 👍

packages/ui/bin/www
rishabh3112: And here too

.travis.yml
```diff
@@ -36,6 +36,7 @@ install:
   - npm install -g eslint
   - npm install -g greenkeeper-lockfile@2
   - lerna bootstrap
+  - cd packages/ui/public && npm ci && cd ../../
```
rishabh3112:
```suggestion
  - cd packages/ui/public && npm ci && cd ../../../
```

rishabh3112:
I can't get `travis` now. Maybe its due to travis is looking for packages for `packages/ui/public` in root `./node_modules` not in `./packages/ui/public/node_modules`.
🤔

ematipico:
I think we have to change the folder structure here, we cannot have nested `package.json` unless we make them as a real package. So we have two solutions

1. We make the ui frontend and the ui server two separate packages straight under the folder `packages` and we could call them `ui-server` and `ui-gui`. This will be the smoother solution but the integration between the two of them could be hard.

2. Having a flat structure. Just one `package.json` where we install all the libraries we need (CRA, express, etc.). Having a `src` folder where we are going to store all our files.

The build will consist in one step plus one optional:
- CRA build which will store static files
- the optional could be "building" the server code but I don't thing it's necessarary

```
package.json
app.js <== this will kick off the server
src/
==> gui/ index.js (entry point)
==> server/ index.js (entry point)
```

rishabh3112:
According to me @ematipico 
Creating separate packages for UI's server and GUI under `webpack-cli` doesn't seem to be a good solution as the `user` of the UI would have to install both to get UI working. Complicating workflow for development as well as usage.

Secondly, Flattening the structure will kill the purpose of working on the `server` and the `gui` independently.
**Consider case:**
Someone has to test server He/She would also have to build the frontend too but frontend may be breaking at that point of time thus obstructing work of those working on `server`

vice-versa may happen too.

Therefore I had gone for nested one but seems to be we have to either find some solution for present scenario or *adopt any of your suggestion* or find something new.

ematipico:
> Creating separate packages for UI's server and GUI under webpack-cli doesn't seem to be a good solution as the user of the UI would have to install both to get UI working. Complicating workflow for development as well as usage.

I think we are missing the true potential of monorepo. The user doesn't need to do anything if we play it correctly. We will have our main package called `ui` which will have two deps inside the the `package.json`

```json
"dependencies": {
  "@webpack-cli/ui-server": "0.1.0",
  "@webpack-cli/ui-gui": "0.1.0"
}
```

And the user won't need to install anything other than the main `@webpack-cli/ui`. 
This will allow us to work on server and gui without breaking each other and we can release separated fixes.

Also, building them together won't be too much of a problem I think. What do you think?

rishabh3112:
That's great then !!
Then we can have three packages named `ui-server`, `ui-gui` and `ui`.

Also, this will increase the scope of this project as then one can exploit the API made under `ui-server` for more use cases (like an electron app)!!
So, an LTS version maintained for `ui-server` would be a great addition

> Also, building them together won't be too much of a problem I think. What do you think?

Building them together will happen ultimately in each solution available.

We can work for this solution than 👍

ematipico:
> Can we work for this solution than @ematipico?

Sure! Wanna try to apply these changes that we just discussed using this PR?

rishabh3112:
Yes

rishabh3112:
Yes

packages/ui/app.js
```diff
@@ -0,0 +1,18 @@
+const express = require("express");
+const path = require("path");
+const cookieParser = require("cookie-parser");
```
renatoagds: @rishabh3112 why we're using cookies?

packages/ui/README.md
```diff
@@ -0,0 +1,26 @@
+# webpack-cli ui
+Act as GUI for `webpack-cli`
+
+## Project Structure
+The `frontend` and `backend` are segregated. Backend will provide REST API which frontend will exploit by means of `ajax json` calls.
+
+#### Backend Structure (base: `./`)
+1. `./app.js` : main entry for the package and server
+2. `./routes/`: will contain endpoints for features like
```
renatoagds: @rishabh3112 I would suggest use `api` instead `routes`, when I read `routes` it looks like we're doing SSR.

packages/ui/bin/www
```diff
@@ -0,0 +1,90 @@
...
+ * Event listener for HTTP server "listening" event.
+ */
+
+function onListening() {
+  var addr = server.address();
+  var bind = typeof addr === 'string'
+    ? 'pipe ' + addr
+    : 'port ' + addr.port;
+  debug('Listening on ' + bind);
+}
```
renatoagds: @rishabh3112 we could move all this `var` to `const`.

renatoagds:
@ematipico @rishabh3112 agreed with the solution about an flat `package.json`, just to check more about it, we'll create an packages inside `packages/ui`, and then create `ui-server` and `ui-gui`?

packages/ui/README.md
gustavo-depaula: in backend, the `routes` nomenclature is pretty common and standard, ex: https://expressjs.com/pt-br/guide/routing.html

packages/ui/bin/www
bhavya9107: Okay I will @renatoagds 😃
P.S. Its a express generator template which I would edit once foundation has been laid.

packages/ui/README.md
renatoagds: @gustavo-depaula it suggest you use an route but not suggest that you use `/routes/*`. so, in terms of an `api` is common to do the routes against `<ip>/api/*`.

rishabh3112: I feel @renatoagds is correct here.
But we should have a index route too that will serve react app. So, I think we can have both of them !!
api for json calls
routes for serving html

packages/ui/app.js
rishabh3112: Its a express generator template. Will update after foundation is merged and we are working on server !!

Maybe we can use cookies too!! for storing previous responses by user for init

packages/ui/bin/www
rishabh3112: Will do here too 👍

rishabh3112:
travis 🤦‍♂️
@ematipico getting same error as before 😅

rishabh3112:
Nevertheless is this structure seems good ?

packages/ui-server/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "@webpack-cli/ui-server",
+  "version": "0.0.0",
+  "private": true,
+  "main": "app.js",
+  "scripts": {
+    "start": "node ./bin/www"
```
ematipico:
```json
 "publishConfig": {
    "access": "public"
  }
```
is missing

packages/ui/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "@webpack-cli/ui",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
```
ematipico:
```json
 "publishConfig": {
    "access": "public"
  }
```
is missing

packages/ui/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "@webpack-cli/ui",
+  "version": "0.1.0",
+  "private": true,
```
ematipico:
```suggestion
```

packages/ui-server/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "@webpack-cli/ui-server",
+  "version": "0.0.0",
+  "private": true,
```
ematipico:
```suggestion

```

packages/ui/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "@webpack-cli/ui",
```
ematipico: I would call this package `@webpack-cli/ui-gui`

packages/ui/package.json
```diff
@@ -0,0 +1,27 @@
...
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
+    "react": "^16.5.2",
+    "react-dom": "^16.5.2",
+    "react-scripts": "2.0.5",
+    "@webpack-cli/ui-server": "0.0.0"
```
ematipico: The server should be a `devDependency`

packages/webpack-scaffold/tsconfig.json
```diff
@@ -1,3 +1,3 @@
-{
-  "extends": "../../tsconfig.packages.json"
-}
+{
```
ematipico: If you can, would you be able to revert these changes that don't really contain any real change?

rishabh3112:
Having a ui-gui seems to be a redundant package
1. As `ui-gui` would  requires `ui-server` as dependency
2. we are just packing another package with both as dependency thus `ui-gui` would be identical to `ui`.

Hope you get my point 😅 @ematipico

packages/webpack-scaffold/tsconfig.json
rishabh3112: Will try 👍

ematipico:
Hey @rishabh3112 , I totally see your point and that's what I thought at the very beginning but after a while I thought: 

"The `ui` packages should be the final package that will be delivered with the new frontend and backend attached together. The `ui` package will be responsible to build the gui and making sure that the server will serve the static files" 

"Also, the `gui` and the `server` should be in separated packages because we should be free to develop separately, this means that `server` should be able to serve any static file that we want, let's just give an **input** and that's it. On the other way around, also the `gui` should be send payloads to a server that is set as **input**. So the frontend develop, if they want, can set their own fake express server just for the sake of sending some payload"

"Also, having the two packages in two different folders, would allow us to link our own local version of the packages (AKA `npm link`), helping the developer experience."

`ui/package.json`
```json
"dependencies": {
  "@webpack-cli/ui-server": "0.1.0",
  "@webpack-cli/ui-gui": "0.1.0"
}
```

`ui-gui/package.json` (so we can use `npm link`)
```json
"devDependencies": {
  "@webpack-cli/ui-server": "0.1.0",
}
```

`ui-server/package.json` (so we can use `npm link`)
```json
"devDependencies": {
  "@webpack-cli/ui-gui": "0.1.0"
}
```

ematipico:
Hey @rishabh3112 , I totally see your point and that's what I thought at the very beginning but after a while I thought: 

"The `ui` packages should be the final package that will be delivered with the new frontend and backend attached together. The `ui` package will be responsible to build the gui and making sure that the server will serve the static files" 

"Also, the `gui` and the `server` should be in separated packages because we should be free to develop separately, this means that `server` should be able to serve any static file that we want, let's just give an **input** and that's it. On the other way around, also the `gui` should be send payloads to a server that is set as **input**. So the frontend develop, if they want, can set their own fake express server just for the sake of sending some payload"

"Also, having the two packages in two different folders, would allow us to link our own local version of the packages (AKA `npm link`), helping the developer experience."

`ui/package.json`
```json
"dependencies": {
  "@webpack-cli/ui-server": "0.1.0",
  "@webpack-cli/ui-gui": "0.1.0"
}
```

`ui-gui/package.json` (so we can use `npm link`)
```json
"devDependencies": {
  "@webpack-cli/ui-server": "0.1.0",
}
```

`ui-server/package.json` (so we can use `npm link`)
```json
"devDependencies": {
  "@webpack-cli/ui-gui": "0.1.0"
}
```

ematipico:
Hey @rishabh3112 , I totally see your point and that's what I thought at the very beginning but after a while I thought: 

"The `ui` packages should be the final package that will be delivered with the new frontend and backend attached together. The `ui` package will be responsible to build the gui and making sure that the server will serve the static files" 

"Also, the `gui` and the `server` should be in separated packages because we should be free to develop separately, this means that `server` should be able to serve any static file that we want, let's just give an **input** and that's it. On the other way around, also the `gui` should be send payloads to a server that is set as **input**. So the frontend develop, if they want, can set their own fake express server just for the sake of sending some payload"

"Also, having the two packages in two different folders, would allow us to link our own local version of the packages (AKA `npm link`), helping the developer experience."

`ui/package.json`
```json
"dependencies": {
  "@webpack-cli/ui-server": "0.1.0",
  "@webpack-cli/ui-gui": "0.1.0"
}
```

`ui-gui/package.json` (so we can use `npm link`)
```json
"devDependencies": {
  "@webpack-cli/ui-server": "0.1.0",
}
```

`ui-server/package.json` (so we can use `npm link`)
```json
"devDependencies": {
  "@webpack-cli/ui-gui": "0.1.0"
}
```

ematipico:
Hey @rishabh3112 , I totally see your point and that's what I thought at the very beginning but after a while I thought: 

"The `ui` packages should be the final package that will be delivered with the new frontend and backend attached together. The `ui` package will be responsible to build the gui and making sure that the server will serve the static files" 

"Also, the `gui` and the `server` should be in separated packages because we should be free to develop separately, this means that `server` should be able to serve any static file that we want, let's just give an **input** and that's it. On the other way around, also the `gui` should be send payloads to a server that is set as **input**. So the frontend develop, if they want, can set their own fake express server just for the sake of sending some payload"

"Also, having the two packages in two different folders, would allow us to link our own local version of the packages (AKA `npm link`), helping the developer experience."

`ui/package.json`
```json
"dependencies": {
  "@webpack-cli/ui-server": "0.1.0",
  "@webpack-cli/ui-gui": "0.1.0"
}
```

`ui-gui/package.json` (so we can use `npm link`)
```json
"devDependencies": {
  "@webpack-cli/ui-server": "0.1.0",
}
```

`ui-server/package.json` (so we can use `npm link`)
```json
"devDependencies": {
  "@webpack-cli/ui-gui": "0.1.0"
}
```

ematipico:
Hey @rishabh3112 , I totally see your point and that's what I thought at the very beginning but after a while I thought: 

"The `ui` packages should be the final package that will be delivered with the new frontend and backend attached together. The `ui` package will be responsible to build the gui and making sure that the server will serve the static files" 

"Also, the `gui` and the `server` should be in separated packages because we should be free to develop separately, this means that `server` should be able to serve any static file that we want, let's just give an **input** and that's it. On the other way around, also the `gui` should be send payloads to a server that is set as **input**. So the frontend develop, if they want, can set their own fake express server just for the sake of sending some payload"

"Also, having the two packages in two different folders, would allow us to link our own local version of the packages (AKA `npm link`), helping the developer experience."

`ui/package.json`
```json
"dependencies": {
  "@webpack-cli/ui-server": "0.1.0",
  "@webpack-cli/ui-gui": "0.1.0"
}
```

`ui-gui/package.json` (so we can use `npm link`)
```json
"devDependencies": {
  "@webpack-cli/ui-server": "0.1.0",
}
```

`ui-server/package.json` (so we can use `npm link`)
```json
"devDependencies": {
  "@webpack-cli/ui-gui": "0.1.0"
}
```

rishabh3112:
Thats neat @ematipico 👍.

I will do that in next commit then.
After that we need a thorough discussion like that would be endpoints and what sort of payload will frontend send.

P.S. 5 duplicate messages!! it literally filled my inbox😂

renatoagds:
@rishabh3112 sorry! I didn't answered you before you made your commits. But, about the `server` structure, I think we could keep `routes` as main folder, and inside it we could have `api` folder and `index.js` file, following what we have here: https://github.com/gothinkster/node-express-realworld-example-app

what you think? 👍

rishabh3112:
Yeah I like that idea as that way our folder structure will reassemble routing!!
Seems to me a great idea 👍 @renatoagds.
Will do that in next commit.

packages/ui/package.json
rishabh3112: Will do separately as `ui` is moved to `ui-gui` 😅

packages/webpack-scaffold/tsconfig.json
rishabh3112: I updated line feed from `LF` to `CRLF` but no luck

packages/ui-gui/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "@webpack-cli/ui",
```
ematipico:
```suggestion
  "name": "@webpack-cli/ui-gui",
```

ematipico: This is breaking the build I think

rishabh3112: done

ematipico:
Looks like that CRA doesn't support lerna/workspaces https://github.com/facebook/create-react-app/issues/1333

rishabh3112:
Hmm, Any workaround in your mind? if not then I think we can set up `react` without the help of CRA or use `vue-cli` to generate initial scaffolding.

ematipico:
Both `vue-cli` scaffold and `create-react-app` have a lot of stuff inside and it might be difficult for the community to understand and keep up with their internals. The latter already has everything ready for react.

Usually my first choice is to have my own configuration if the cli/scaffolder doesn't help me. 

But I'm also fine to eject from CRA

rishabh3112:
So, then I eject `create-react-app`. Or build from scratch? @ematipico

ematipico:
We can build it from scratch I'd say. For now we can have some basic stuff, no need to put too much stuff such as eslint plugins, etc. I think just having a package with the main packages would be fine, so we can close this PR once for all!

rishabh3112:
Done but wanted to know should I setup `vue` or `react` or decide it later? @ematipico

ev1stensberg:
Use preact please

renatoagds:
@ ev1stensberg any reason for preact?

ev1stensberg:
It’s smaller and we are planning to integrate cli  in preact

ematipico:
We can discuss on a separate thread. I'm happy to merge this

rishabh3112:
Okay, we would then discuss it at #647. If there anything left here, let me know.

PR #653, opened 2018-10-23
Upgrade: TypeScript, Yeoman Generator 

dhruvdutt:
Major version upgrades

N/A

N/A

**Summary**
Fix breaking changes for `yeoman-generator`

No

**Other information**

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2276,9 +2276,9 @@
       }
     },
     "@types/glob": {
-      "version": "7.1.0",
-      "resolved": "https://registry.npmjs.org/@types/glob/-/glob-7.1.0.tgz",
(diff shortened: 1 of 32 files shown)


ematipico:
When I was playing around `yeoman-generator.ts`, I noticed there are some types that are missing. Fox example `IComposeSettings`. I wonder why `tsc` doesn't catch it... Could you please fix that?

dhruvdutt:
I took the initial definitions from [polymer-cli](https://github.com/Polymer/polymer-cli/blob/master/custom_typings/yeoman-generator.d.ts).
We will need to update lot of methods and make `tsc` config more aggressive on checks. I think we can do that in a separate PR as it will take a while to review as well. Let's merge this for now.

PR #671, opened 2018-11-02
Webpack GUI: Preact setup

renatoagds:
Adding the GUI setup, based in [preact-cli](https://github.com/developit/preact-cli), using `material` theme.


packages/ui-gui/package.json
```diff
@@ -1,28 +1,38 @@
...
-    "serve": "webpack-dev-server --config=webpack.prod.js"
+    "start": "if-env NODE_ENV=production && npm run -s serve || npm run -s dev",
+    "build": "preact build",
+    "serve": "preact build && preact serve",
+    "dev": "preact watch",
+    "lint": "eslint src"
   },
-  "license": "MIT",
+  "eslintConfig": {
+    "extends": "eslint-config-synacor",
```
ematipico: Why this config? I am curios, I've never seen it

packages/ui-gui/package.json
```diff
@@ -1,28 +1,38 @@
...
-    "webpack-dev-server": "^3.1.10"
+    "eslint": "^4.5.0",
+    "eslint-config-synacor": "^1.1.0",
+    "if-env": "^1.0.0",
+    "preact-cli": "^2.0.0"
   },
-  "dependencies": {}
+  "dependencies": {
+    "preact": "^8.2.1",
+    "preact-compat": "^3.17.0",
```
ematipico: Compact is not needed if we don't use react

packages/ui-gui/src/routes/404/index.js
```diff
@@ -0,0 +1,21 @@
...
+export default class NotFound extends Component {
+	render() {
+		return (
+			<div class={`${style.home} page`}>
+				<Card>
+					<div class={style.cardHeader}>
+						<h2 class=" mdc-typography--title">404! Page not found.</h2>
+					</div>
+					<div class={style.cardBody}>
+						Looks like the page you are trying to access, doesn't exist.
```
ematipico: You have to escape the `'`

packages/ui-gui/package.json
```diff
@@ -1,28 +1,38 @@
...
-    "babel-preset-env": "^1.7.0",
-    "css-loader": "^1.0.0",
-    "mini-css-extract-plugin": "^0.4.4",
-    "node-sass": "^4.9.3",
-    "sass-loader": "^7.1.0",
-    "style-loader": "^0.23.1",
-    "webpack": "^4.21.0",
-    "webpack-cli": "^3.1.2",
-    "webpack-dev-server": "^3.1.10"
+    "eslint": "^4.5.0",
```
ematipico: We should add `@webpack-cli/ui-server` to the `devDependencies `

rishabh3112: Not really I think,  one can use `ui` for testing the gui as a whole.

While `ui-gui` can use `webpack-dev-server` to test while development.

ematipico: Yes, of course. But we also need to have a command to bootstrap the real server. Somewhere the GUI will call some rest API to fetch some data

rishabh3112: What I was saying was that rest API server can be ran using `webpack-cli ui` this way we wouldn't be cluttering `ui-gui`. Rest depends on what majority feels 😅

packages/ui-gui/package.json
renatoagds: @ematipico this was generated by `preact-cli` booilerplate.

packages/ui-gui/package.json
renatoagds: @ematipico 👍

packages/ui-gui/src/style/index.css
```diff
@@ -0,0 +1,57 @@
...
+	-webkit-font-smoothing: antialiased;
+	-moz-osx-font-smoothing: grayscale;
+}
+
+/* fallback */
+@font-face {
+  font-family: 'Material Icons';
+  font-style: normal;
+  font-weight: 400;
+  src: url(https://fonts.gstatic.com/s/materialicons/v29/2fcrYFNaTjcS6g4U3t-Y5UEw0lE80llgEseQY3FEmqw.woff2) format('woff2');
```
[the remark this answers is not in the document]
renatoagds: @rishabh3112 sure! I think we could that in the future, since this is just the bootstrap generated from preact-cli, we'll probably change a lot of things here.

packages/ui-gui/package.json
renatoagds: @ematipico agreed with you, but install will fail since we don't have it published yet. so, for now I'll keep it out and we add in future 👍

renatoagds:
@ematipico jut added build folder in `gitignore`, can we go ahead with this or we have more points?

packages/ui-gui/package.json
```diff
@@ -1,28 +1,35 @@
...
-    "babel-preset-env": "^1.7.0",
-    "css-loader": "^1.0.0",
-    "mini-css-extract-plugin": "^0.4.4",
-    "node-sass": "^4.9.3",
-    "sass-loader": "^7.1.0",
-    "style-loader": "^0.23.1",
-    "webpack": "^4.21.0",
-    "webpack-cli": "^3.1.2",
-    "webpack-dev-server": "^3.1.10"
+    "if-env": "^1.0.0",
```
evenstensberg: Crossenv here instead if this package is responsible for env

ematipico: Also the `README.md` of `if-env`, they recommend to install `per-env`. Mmmmh 😅

renatoagds: @evenstensberg @ematipico replaced `if-env` for `cross-env`.

packages/ui-gui/.gitignore
```diff
@@ -0,0 +1,67 @@
...
+# Output of 'npm pack'
+*.tgz
+
+# Yarn Integrity file
+.yarn-integrity
+
+# dotenv environment variables file
+.env
+
+.DS_Store
```
evenstensberg: Could you add a comment here as well of what ds store is?

packages/ui-gui/package.json
```diff
@@ -1,28 +1,35 @@
...
   "name": "@webpack-cli/ui-gui",
   "version": "0.0.1",
+  "license": "MIT",
   "description": "GUI for webpack-cli ui",
-  "main": "index.js",
+  "main": "src/index.js",
   "scripts": {
-    "build": "webpack --config=webpack.prod.js",
-    "serve": "webpack-dev-server --config=webpack.prod.js"
+    "start": "cross-env NODE_ENV=production && npm run -s serve || npm run -s dev",
```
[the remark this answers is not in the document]
rishabh3112: If I get your question,
`&&` has higher preference over `||`
There for if `NODE_ENV=production` is true then serve will run if `NODE_ENV=production` is false then due to `&&` serve will not run instead `dev` will run.

packages/ui-gui/package.json
renatoagds: @evenstensberg @ematipico thinking more about, `cross-env` is not what we expect using here, since `if-env` do a `if/else` to run different command based in which env was defined, that's not exactly what `cross-env` is proposed to do, I'll change that for `per-env` as suggested in `if-env` repo.

packages/ui-gui/package.json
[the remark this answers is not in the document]
renatoagds: @dhruvdutt that's exactly what @rishabh3112, the only point here is that we're using `if-env`, since it's deprecated, we changed for `cross-env`, but `cross-env` isn't build to do this kind of `if/else`, like a said here: https://github.com/webpack/webpack-cli/pull/671#discussion_r237272276 ... Based in that, I'll update change for `per-env`, since it's proposed in `if-env` repo.

packages/ui-gui/README.md
```diff
@@ -1,10 +1,23 @@
 # webpack-cli ui-gui
+
 GUI for `webpack-cli ui`
 
 #### Package Structure (base: `./`)
+
 1. `./`: Backend will serve from root folder
```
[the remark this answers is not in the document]
renatoagds: @rishabh3112 what exactly you propose to change here?

renatoagds:
@evenstensberg @ematipico can you look again? I made the changes to use `per-env` and add comment explaining .DS_Store

packages/ui-gui/README.md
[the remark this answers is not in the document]
rishabh3112: https://github.com/webpack/webpack-cli/blob/995a9110340b0d99e29bf63596dfc3a2815cc8da/packages/ui/bin/www#L7

Update the path passed to ui-server to location where complied index.html would be located.

rishabh3112:
@dhruvdutt currently its just a setup with preact. We haven't implemented anything yet. So, IMO we can take it later maybe in another PR.

evenstensberg:
+1 on that @rishabh3112

renatoagds:
+1 with @rishabh3112

packages/ui-gui/README.md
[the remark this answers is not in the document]
renatoagds: @rishabh3112 makes sense. I'll take a look at this one.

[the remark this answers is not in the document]
renatoagds: @rishabh3112 done. updated with the new path.

renatoagds:
@evenstensberg could go ahead and merge?

ematipico:
@renatoagds please next time make sure to have commit messages that are compliant to our guidelines. As you can see the build fails because of that. Thanks for your PR!!

manishupadhyaya:
@emitipico anything I can contribute to implementing any design to code ?

PR #682, opened 2018-11-05
docs: correct use of individual package in node env

hemal7735:
README docs

**Did you add tests for your changes?** NA

**Summary**
I found out while playing around with the examples given in the doc.
We transpile the `ts` files using babel which converts 

`export default [func/variables]` 
to 
``` bash
Object.defineProperty(exports, "__esModule", { value: true });
exports.default = default_1
```

**Does this PR introduce a breaking change?** No

### DIFF
--- a/packages/add/README.md
+++ b/packages/add/README.md
@@ -16,7 +16,7 @@ To run the scaffolding instance programmatically, install it as a dependency. Wh
 
 ### Node
 ```js
-const add = require("@webpack-cli/add");
+const add = require("@webpack-cli/add").default;
 add();
 ```
 
--- a/packages/generate-loader/README.md
+++ b/packages/generate-loader/README.md
@@ -17,7 +17,7 @@ To run the package programmatically, install it as a dependency. When using the 
 ### Node
 
 ```js
-const generateLoader = require("@webpack-cli/generate-loader");
+const generateLoader = require("@webpack-cli/generate-loader").default;
 generateLoader();
 ```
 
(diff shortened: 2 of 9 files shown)


hemal7735:
@dhruvdutt can you review please?

ematipico:
Although this is a bug!
In a nodejs environment we don't have the concept of `.default` because we just use `CommonJS` file resolution!

hemal7735:
How will we fix the TypeScript compilation for that? Isn't this the work of babel? 
In `CommonJS` we don't have the concept of `.default`, however we do have the concept of exporting more than one things, and in this case, it is a simple default.
In package - `serve`, we export many things, in that case, we will need to use this, or `serve` will need the same structure as rest of the packages.

ematipico:
> basarat.gitbooks.io/typescript/docs/tips/defaultIsBad.html

That's an opinion and everything is relative. That's not the issue here.

@hemal7735 No, we use TypeScript for compiling the code. What I would expect from the compilation is something like this:

Before
```js
export default something() {}
```

After 
```js
module.exports = something() {}
```

Instead, TypeScript does something dumb and consider `default` as the name of the configure when instead it's a particular syntax.

hemal7735:
> @hemal7735 No, we use TypeScript for compiling the code. What I would expect from the compilation is something like this:
> 
> Before
> 
> ```js
> export default something() {}
> ```
> 
> After
> 
> ```js
> module.exports = something() {}
> ```
> 
> Instead, TypeScript does something dumb and consider `default` as the name of the function when instead it's a particular syntax.
> Probably there's a configuration for that.

It will still do if we have both `default` and `named` exports. But yeah, I got your point that it should do 
```js
module.exports = something() {}
```
if default is the only export. Maybe it's a feature request for both babel and tsc 😅

PR #683, opened 2018-11-05
feat(tests): add extractHash support in test-infra

### LINKED ISSUE #672: [Feature] Add extractSummary to test infrastructure
It would be good to have more granular assertions using `extractSummary` similar to [Jest](https://github.com/facebook/jest/blob/7b7fd013508de475e2151e50847373713fc97c7e/e2e/Utils.js#L139)

Implement a utility to use `extractSummary` like this: 

https://github.com/jbdemonte/jest/blob/9a90cfee8779ead918d1379a8a9fa33f25030d76/e2e/__tests__/test_name_pattern_skipped.test.js#L19

N/A

**Additional context**
N/A

hemal7735:

This PR adds support for extracting hash after running test. [Jest](https://github.com/facebook/jest/blob/master/e2e/__tests__/test_name_pattern_skipped.test.js#L19) is already doing it. We are trying to mimic it here, for our needs.
**Did you add tests for your changes?** Yes

**Summary**

This PR tries to address the feature-request issue #672 
Closes #672 

**Does this PR introduce a breaking change?** No

### DIFF
--- a/test/test-utils.test.js
+++ b/test/test-utils.test.js
@@ -4,13 +4,13 @@ const { extractHash } = require("./testUtils");
 
 describe("extractHash functionality", () => {
 	it("should throw Error if there is empty string", () => {
-		let stdout = "";
+		const stdout = "";
 
 		expect(() => extractHash(stdout)).toThrowError();
 	});
 
 	it("should throw Error if there is no hash value", () => {
-		let stdout = `Entrypoint main = main.js
+		const stdout = `Entrypoint main = main.js
         [./src/index.js] 2.52 KiB {main} [built] [failed] [1 error]
         ERROR in ./src/index.js
         Module build failed (from ./node_modules/babel-loader/lib/index.js):
@@ -20,7 +20,7 @@ describe("extractHash functionality", () => {
 	});
 
 	it("should throw Error if hash count and config count do not match", () => {
-		let stdout = `Hash: c6fe1f550a76720b1eaec6fe1f550a76720b1eae
+		const stdout = `Hash: c6fe1f550a76720b1eaec6fe1f550a76720b1eae
         Version: webpack 4.25.0
         Child index1:
             Hash: c6fe1f550a76720b1eae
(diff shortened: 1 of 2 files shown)


hemal7735:
👍  thanks, @ematipico for the clarification. I will revisit the problem here.

hemal7735:
I've added support for extractHash. I'm kinda confused regarding the return value of extractSummary functionality. Implementation purely depends upon what info we want to extract to validate.
Can you folks guide me here?

ematipico:
Yes, the original PR lacks of spec and I do apologize for that.

As extract summary we can do basically what that function I pointed out does.

We should strip out any timestamp, hash and information that is dynamic.

hemal7735:
@ematipico any thoughts over `extractHash` functionality presented in the PR?

ematipico:
It would be nice to have some test on some mock string, to make sure it will work

hemal7735:
@ematipico  added relevant tests

test/test-utils.test.js
```diff
@@ -0,0 +1,103 @@
+"use strict";
+
+const { extractHash } = require("./testUtils");
+
+describe("extractHash functionality", () => {
+	it("should throw Error if there is empty string", () => {
+		let stdout = "";
```
evenstensberg: any reason for using let instead of const?

test/testUtils.js
```diff
@@ -75,4 +75,59 @@ function removeTimeStrings(stdout) {
 		.filter(line => !line.includes("Built at"))
 		.join("\n");
 }
-module.exports = { run, runWatch };
+
+function extractHash(stdout) {
```
evenstensberg: Could you add JSDocs to this?

test/test-utils.test.js
hemal7735: I could not find any usecase of `const` here. Can you suggest some so that I can understand the change better?

test/testUtils.js
hemal7735: sure 👍

test/test-utils.test.js
ematipico: Conventionally some people you use `let` when the variable gets reassigned. If it doesn't, `const` it's the default choice

hemal7735: @ematipico @evenstensberg used const wherever needed.

hemal7735:
@evenstensberg I would love to do that, however not here, in the separate PR. 💯 
I believe each PR should have single reversible logical change, which can define what is done in the PR. This was regarding adding `extractHash` support, so we should stick to it.
I can create an issue to track this.
Again, these are my thoughts. I would leave it up to the rest of the community. 😄

evenstensberg:
lgtm here

Issue #649, opened 2018-10-17
[UI] Setting up server project


ematipico:
The UI project will bootstrap a server that will serve the static resources that will be compiled from a different task.

The server, also, will be responsible of detecting all the requests coming from the UI and execute the different commands like
- creating a new webpack configuration file
- modify existing ones
- ideally execute the builds givin a specific webpack 

The `webpack-cli` should already have all the APIs ready for these kind of operations. If not, a refactor will needed in order to expose to achieve such features.

The goal of this is issue is to have a simple and good foundation of a server that we can bootstrap and ready to receive commands from the UI.

LakshSingla:
I would like to take up this issue. For the starters, I will make `init` functionality as laid out in the `packages/ui-server` package. How should one go about creating `init` for a non default scaffold ?

LakshSingla:
@ematipico @evenstensberg  I am stuck on creating the `init` functionality. To achieve the creating of `webpack` config file, I was using `child_process` module, and communicating via the `@webpack-cli/init` using its `spawn` function. But there is difficulty in synchronisation of input and ouput. ( I am creating an output string and directly dumping it onto the child process's `stdin`. I would retry it once more by having seperate calls to the `stdin.write()` of child process. Is there any other alternative I can try (without refactoring other packages), or am I approaching this issue wrongly ??

ematipico:
Would you like to create a PR with the code so we can give some direction and try to understand the problem?

I don't think you need to use a raw `child_process`. You can use `execa` for that (we use it already for our tests).

LakshSingla:
@ematipico Sure thing. Lemme use `execa` in my code and I will send a PR.

LakshSingla:
@ematipico Please review the PR and confirm if I am on the right track. It is not the cleanest way to achieve the above task hence I am a little skeptical if I have understood the aim correctly. I will make the necessary changes as soon as possible so that we are making some real progress on this goal.

rishabh3112:
@webpack/cli-team We can now start developing endpoints. Anyone willing to collaborate on init endpoint with me.

renatoagds:
@rishabh3112 I would like to help the init endpoint with you 👍

PR #685, opened 2018-11-06
[WIP] `init` endpoint of ui-server

LakshSingla:
Feature

No

No

**Summary**
Created the route for `POST: init` for default scaffold. Does not work as expected due to issues in `stdin` of the child process spawned.

#649 

No

**Other information**
None

### DIFF
--- a/packages/ui-server/app.js
+++ b/packages/ui-server/app.js
@@ -3,6 +3,7 @@ const path = require("path");
 const cookieParser = require("cookie-parser");
 const logger = require("morgan");
 const indexRouter = require("./routes/index");
+const apiRouter = require("./routes/api");
 
 module.exports = function(staticFolder) {
 	const app = express();
--- a/packages/ui-server/app.js
+++ b/packages/ui-server/app.js
@@ -3,7 +3,6 @@ const path = require("path");
 const cookieParser = require("cookie-parser");
 const logger = require("morgan");
 const indexRouter = require("./routes/index");
-const apiRouter = require("./routes/api");
 
 module.exports = function(staticFolder) {
 	const app = express();
--- a/packages/ui-server/bin/www/index.js
+++ b/packages/ui-server/bin/www/index.js
@@ -1,10 +1,12 @@
 const path = require('path');
+const weblog = require('webpack-log');
+
 const app = require('../../app.js');
 
+const log = weblog({ name: 'uis'});
 const port = process.env.PORT || 8000;
+const staticFolder = path.join('..', '..', 'static');
 
-console.log(port);
-
-app(path.join('..', '..', 'static')).listen(port, () => {
-	console.log(`server started on 127.0.0.1:${port}`);
+app(staticFolder).listen(port, () => {
+	log.info(`server started on 127.0.0.1:${port}`);
 });
--- a/packages/ui-server/bin/www/index.js
+++ b/packages/ui-server/bin/www/index.js
@@ -3,6 +3,8 @@ const app = require('../../app.js');
 
 const port = process.env.PORT || 8000;
 
+console.log(port);
+
 app(path.join('..', '..', 'static')).listen(port, () => {
 	console.log(`server started on 127.0.0.1:${port}`);
 });
--- a/packages/ui-server/bin/www/index.js
+++ b/packages/ui-server/bin/www/index.js
@@ -1,0 +1,8 @@
+const path = require('path');
+const app = require('../../app.js');
+
+const port = process.env.PORT || 8000;
+
+app(path.join('..', '..', 'static')).listen(port, () => {
+	console.log(`server started on 127.0.0.1:${port}`);
+});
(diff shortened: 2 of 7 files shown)


packages/ui-server/app.js
```diff
@@ -3,6 +3,7 @@ const path = require("path");
 const cookieParser = require("cookie-parser");
 const logger = require("morgan");
 const indexRouter = require("./routes/index");
+const apiRouter = require("./routes/api");
```
rishabh3112: Redundant import
```suggestion

```

packages/ui-server/routes/index.js
```diff
@@ -3,7 +3,9 @@ const router = express.Router();
 
 /* GET home page. */
 router.get("/", function(req, res, next) {
-	res.render("index", { title: "Express" });
+	res.json({ title: "Express" });
 });
 
+router.use(require('./api'));
```
rishabh3112: use `/api` as the base route for the server calls.
We would be using `/api/init` for json calls to render `/init` and same for other endpoints.
```suggestion
router.use('/api',require('./api'));
```

rishabh3112:
Sign CLA too

LakshSingla:
@rishabh3112 Thanks for the suggestions. I have made the required changes. But I am still stuck on how to call the actual `init` process. I have used `execa` to create a child process and passed in the options to its `stdin` (which I receive in the request). But I am facing an issue because of the way I am passing the `stdin` due to which all output is dumped into the child process at once and it cannot read it for subsequent queries.
webpack-init-1

rishabh3112:
I think we should refactor and change interface of `webpack-scaffold` all together so as to avoid running child process. A configuration generating file which will take answers to questions as json input and generates file.

This way debugging in future would be better.Let have others viewpoint. 
cc @evenstensberg @ematipico

LakshSingla:
@rishabh3112 I was thinking the same thing, to provide an alternative way of programmatically passing in the parameters for `init`.

LakshSingla:
@rishabh3112 Should I start with the refactoring? Also what all needs to be refactored? I would begin with the `webpack init` command located at `@webpack-cli/init`.

ematipico:
Yes, inquirier logic and generator logic have to be separated in order to achieve this PR.

The server will just pass an object to the generator, which will create the webpack file for us. The object will contain the same set of answers.

Because of the fact that we are using typescript, I think there could be an interface that design the object answers that we can pass to the generator. If there isn't, we should create it.

I think to make things easier, using typescript for the server might be a better fit because of the typings...

LakshSingla:
@ematipico Should I open a seperate issue regarding this issue? Also, given that the generator logic is in place `@webpack-cli/packages/generators/index-generator.ts`, I propose that we move the inquirer logic to another file.

rishabh3112:
I think better would be that every generator exports:
1. `cli`: would be **default** class
2. `ui`: would be processing Object.

and also a function that would return a multi-dimensional array having **questions**.

Then for passing Object we can:
```js
import { ui } from "@webpack-cli/init"
...
let input: Object; // or we may have a interface here instead of object
// taking input from server
ui(input);
...
```
Review approch @ematipico I would do necessary updations then. Also, @evenstensberg didn't gave review for design as he said to me he will be. If you have suggestions you may let me know. After that I would design other pages as well.

LakshSingla:
@rishabh3112 Thanks for help with the PR! As soon as you are done with the necessary refactoring, lemme know so that I can make the required changes.

LakshSingla:
@rishabh3112 Any updates in refactoring ??

evenstensberg:
No worries

packages/ui-server/bin/www/index.js
```diff
@@ -0,0 +1,10 @@
+const path = require('path');
+const app = require('../../app.js');
+
+const port = process.env.PORT || 8000;
+
+console.log(port);
```
evenstensberg: remove log here

packages/ui-server/bin/www/index.js
```diff
@@ -0,0 +1,10 @@
...
+
+app(path.join('..', '..', 'static')).listen(port, () => {
```
evenstensberg: extract to var

packages/ui-server/bin/www/index.js
```diff
@@ -0,0 +1,10 @@
...
+	console.log(`server started on 127.0.0.1:${port}`);
```
evenstensberg: use webpack logger

LakshSingla: Made the necessary changes

rishabh3112:
#700

fokusferit:
hey @rishabh3112 any update on this PR ?

fokusferit:
Just to understand this correctly, this one can be close as @rishabh3112 continue the work at #756 right?

wizardofhogwarts:
Yes.

evenstensberg:
Could you rebase this?

evenstensberg:
As per meetings, we're decided that UI should be third party

evenstensberg:
Thank you so much for taking the time to create this Pull Request, you've helped shape the way we're going with webpack(cli) ❤️

Issue #689, opened 2018-11-13
Feature: Find config recursively


GMartigny:
**Is your feature request related to a problem? Please describe.**
Unlike shellscape/webpack-command (formerly webpack-contrib/webpack-command) configuration file is not searched upward in the directory tree. When using monorepo, I would like to set one configuration's file at the root and not for each sub-modules.

**Describe the solution you'd like**
Being able to define only one webpack configuration file for all sub-modules of a mono-repo.

**Describe alternatives you've considered**
I could use shellscape/webpack-command, but I would like to stay within the official repo.
I could set symbolic links, but it feels overkills.

**Additional context**
File `./webpack.config.js` exists.
Scripts in `./packages/sub/package.json`:
```
"build:cli": "webpack-cli --mode=production",
"build:com": "webpack-command --mode=production"
```
Running `lerna run build:cli` fail to read `./webpack.config.js`, but `lerna run build:com` works.

This feature could change existing behavior, but it seems closer to what could be expected. (works alike `git config`)

lakatostamas:
Hi!
Can I send a PR for this?

dhruvdutt:
Absolutely. 👍

PR #693, opened 2018-11-18
Feature/find config recursively

lakatostamas:
Feature.
Yes.
Should I open a PR in this repo: https://github.com/webpack/webpack.js.org ?
**Summary**
Issue: https://github.com/webpack/webpack-cli/issues/689
I used a third-party lib to find the config file recursively. This lib exactly does what we need, but if you think we shouldn't use a third-party for this I remove it and implement it by myself. I checked the "old" solution in `config-loader` and it used `cosmiconfig` but I think `findup-sync` is more suitable for us in this repository. (https://github.com/webpack-contrib/config-loader/blob/859799105c17d89a481b3849f75497ed4f656348/lib/load.js#L63)

No (I guess)

**Other information**
-


hemal7735:
@lakatostamas can you also add some negative test-cases to check what happens when it can not find the file?

lakatostamas:
Thanks for the review!

@hemal7735 Sure, I will add negative cases.

@evenstensberg I think I've added one test which is testing the implementation:
here is the test file (2 level deep: `packages/foo`): `test/binCases/config-location/find-recursively/packages/foo/find-recursively.test.js`
and here is the `webpack.config.js`: `test/binCases/config-location/find-recursively/webpack.config.js`

I think this will perform the following checks:
1, `test/binCases/config-location/find-recursively/packages/foo/`
2, `test/binCases/config-location/find-recursively/packages/`
3, `test/binCases/config-location/find-recursively/` (`webpack.config.js` found)

Should I modify this test or it's completely irrelevant?

Sorry if I totally missed the point, this is my first contribution to any webpack-related package.

bin/convert-argv.js
```diff
@@ -80,12 +69,25 @@ module.exports = function(...args) {
...
+			.map(function(filename) {
+				return extensions.map(function(ext) {
+					return filename.concat(ext);
+				});
+			})
+			.reduce(function(a, i) {
+				return a.concat(i);
+			}, []);
+
+		for (i = 0; i < defaultConfigFileNames.length; i++) {
```
hemal7735: We will need some logic change here. Let's see compare both implementation.
**existing:** try to find file from  `["webpack.config", "webpackfile"]` in cwd.

**current-modification:** 
1. try to find file `webpack.config` in up-fashion.
2. if found then return, if not go to step#3
3. try to find file `webpackfile` in up-fashion.
4. if found then return

Do we see the problem here? Even though `webpackfile` is there one level above, it would not be discovered, but 2 level up `webpack.config` would be given chance first.  
I would want both files to get a fair chance to be discovered. @evenstensberg @ematipico your thoughts on this?

evenstensberg: top-down priority is the best

hemal7735: @evenstensberg I did not get you. Can you give any example?

lakatostamas: @hemal7735 I think you are right. I implemented a solution with (dynamically created) RegExp, but I didn't merge it into this PR. I also added `webpackfile` related tests. Could you please check it? https://github.com/lakatostamas/webpack-cli/compare/1f2b7e2..727a59a

bin/convert-argv.js
```diff
@@ -6,6 +6,7 @@ const prepareOptions = require("./prepareOptions");
 const webpackConfigurationSchema = require("./webpackConfigurationSchema.json");
 const validateSchema = require("webpack").validateSchema;
 const WebpackOptionsValidationError = require("webpack").WebpackOptionsValidationError;
+const findup = require("findup-sync");
```
ematipico: Is there some other package as alternative? This one is 161.9
 kilobyte minified (https://bundlephobia.com/result?p=findup-sync@2.0.0)

bin/convert-argv.js
```diff
@@ -80,12 +69,25 @@ module.exports = function(...args) {
 		const configArgList = Array.isArray(argv.config) ? argv.config : [argv.config];
 		configFiles = configArgList.map(mapConfigArg);
 	} else {
-		for (i = 0; i < defaultConfigFiles.length; i++) {
-			const webpackConfig = defaultConfigFiles[i].path;
-			if (fs.existsSync(webpackConfig)) {
+		const defaultConfigFileNames = ["webpack.config", "webpackfile"]
```
ematipico: This is a good chance to move `"webpack.config"` and `"webpackfile"` inside two constants

bin/convert-argv.js
ematipico: Yeah both files should always have the same priority at each path of directory structure.

test/binCases/config-location/not-found/not-found.test.js
```diff
@@ -0,0 +1,20 @@
...
+		"[id].chunk.js",
+		"--target",
+		"async-node",
+		"--mode",
+		"production",
+	]);
+	expect(code).toBe(0);
+	expect(stdout).toEqual(expect.anything());
+	expect(stdout).toContain("./src/index.js");
+	expect(stderr).toHaveLength(0);
```
ematipico: Apart the `./src/index.js`, is there anything else that we should test in the specific (anything() is too generic)? Also, it is worth to have a snapshot test?

test/binCases/config-location/find-recursively/packages/foo/find-recursively.test.js
```diff
@@ -0,0 +1,20 @@
+"use strict";
```
ematipico: This test is not clear to me. Why test itself is nested while the configuration file is at top level? Shouldn't be the other way around?

```
- testfile.test.js
--- folder
-------- folder
-------------- webpack.config.js
-------------- src/index.js (this is 0CJS)
```

bin/convert-argv.js
lakatostamas: @ematipico Yes, this package is large (findup-sync), but this one supports RegExp. I think with this solution the config files would be discovered with equal chance and the code will be more simplified. (I already implemented it, commit: https://github.com/lakatostamas/webpack-cli/commit/710e04a)
I have found an alternative, but it doesn't support RegExp (find-up: https://bundlephobia.com/result?p=find-up@3.0 ) It would be complicated to implement to find config files with an equal chance on every level without RegExp. (but with this PR it would be easy to implement: https://github.com/sindresorhus/find-up/pull/28 )
Which solution do you recommend?
1, Go with the large npm lib which supports RegExp
2, Go with the small npm lib which only accepts a string. In this case, I think we have to implement a solution which stops on every level and tries both patterns.
3, Implement our own solution without a third party (with RegExp)
4, any other solution?

I would like to clarify this before I fix the other issues.

test/binCases/config-location/find-recursively/packages/foo/find-recursively.test.js
lakatostamas: Maybe the title of the PR is misleading but in the issue, the `package.json` (like the test file here) is nested and the `webpack.config.js` is on the top level. It should work like `git config` detection. Do I misunderstand the issue? (https://github.com/webpack/webpack-cli/issues/689#issue-380157084)

ematipico: My mistake! I misunderstood the requirements! Now it's more clear :)

bin/convert-argv.js
ematipico: Actually this link shows some better results 

https://packagephobia.now.sh/result?p=findup-sync

It looks to be smaller than expected!

test/binCases/config-location/not-found/not-found.test.js
evenstensberg: `expect(stdout).toEqual(expect.anything());` could be stripped

test/binCases/config-location/find-recursively/packages/foo/find-recursively.test.js
```diff
@@ -0,0 +1,20 @@
...
+		"[name].js",
+		"--output-chunk-filename",
+		"[id].chunk.js",
+		"--target",
+		"async-node",
+		"--mode",
+		"production"
+	]);
+	expect(code).toBe(0);
+	expect(stdout).toEqual(expect.anything());
```
evenstensberg: You should use the `extractHash` utility.

test/binCases/config-location/find-recursively/packages/foo/find-recursively.test.js
```diff
@@ -0,0 +1,20 @@
...
+
+test("find-recursively", () => {
+	const { code, stdout, stderr } = run(__dirname, [
+		"--output-filename",
+		"[name].js",
+		"--output-chunk-filename",
+		"[id].chunk.js",
+		"--target",
+		"async-node",
+		"--mode",
```
evenstensberg: remove prod, delays test case

test/binCases/config-location/set-explicitly/set-explicitly.test.js
```diff
@@ -0,0 +1,24 @@
...
+		"./webpack.dev.js",
+		"--config-name",
+		"foo",
+		"--output-filename",
+		"[name].js",
+		"--output-chunk-filename",
+		"[id].chunk.js",
+		"--target",
+		"async-node",
+		"--mode",
```
evenstensberg: remove prod

test/binCases/config-location/set-explicitly/webpack.dev.js
```diff
@@ -0,0 +1,8 @@
+var path = require("path");
+
+module.exports = [
+	{
+		name: "foo",
```
evenstensberg: this is wrong

test/binCases/config-location/find-recursively/webpack.config.js
```diff
@@ -0,0 +1,7 @@
+var path = require("path");
+
+module.exports = [
```
evenstensberg: export object instead

test/binCases/config-location/set-implicitly/set-implicitly.test.js
```diff
@@ -0,0 +1,20 @@
...
+
+test("set-implicitly", () => {
+	const { code, stdout, stderr } = run(__dirname, [
+		"--output-filename",
+		"[name].js",
+		"--output-chunk-filename",
+		"[id].chunk.js",
+		"--target",
+		"async-node",
+		"--mode",
```
evenstensberg: remove prod

test/binCases/config-location/set-implicitly/webpack.config.js
```diff
@@ -0,0 +1,8 @@
+var path = require("path");
+
+module.exports = [
+	{
+		name: "foo",
```
evenstensberg: invalid

test/binCases/config-location/webpackfile-config/packages/foo/webpackfile-config.test.js
```diff
@@ -0,0 +1,21 @@
...
+
+test("find-recursively", () => {
+	const { code, stdout, stderr } = run(__dirname, [
+		"--output-filename",
+		"[name].js",
+		"--output-chunk-filename",
+		"[id].chunk.js",
+		"--target",
+		"async-node",
+		"--mode",
```
evenstensberg: remove prod

test/binCases/config-location/webpackfile-config/packages/foo/webpackfile.js
```diff
@@ -0,0 +1,7 @@
+var path = require("path");
+
+module.exports = [
```
evenstensberg: At this level, add a `webpack.config.js` file as well, it shouldn't break the test if implemented correctly.

bin/convert-argv.js
```diff
@@ -80,15 +69,15 @@ module.exports = function(...args) {
...
-		for (i = 0; i < defaultConfigFiles.length; i++) {
-			const webpackConfig = defaultConfigFiles[i].path;
-			if (fs.existsSync(webpackConfig)) {
-				configFiles.push({
-					path: webpackConfig,
-					ext: defaultConfigFiles[i].ext
-				});
-				break;
-			}
+		const webpackConfigFileRegExp = `(webpack.config|webpackfile)(${extensions.join("|")})`;
```
evenstensberg: `configFileRegExp` is sufficient

evenstensberg: add a constant, so we can add more later, `const defaultConfigFiles = ["webpack.config", "webpackfile"]` and join them in the regexp instead like the last node in the regexp

test/binCases/config-location/webpackfile-config/packages/foo/webpackfile.js
lakatostamas: If I add a `webpack.config.js` in the same level as `webpackfile.js` it will break the test because if they are on the same level then `webpack.config.js` will be used. I've added a `webpack.config.js` two level above: https://github.com/webpack/webpack-cli/pull/693/files#diff-3f5d33891236c7320bd882814444bca0R1

test/binCases/config-location/find-recursively/packages/foo/find-recursively.test.js
lakatostamas: @evenstensberg @ematipico So should I use `extractHash` here? I think I fixed the broken test cases except this one.

PR #703, opened 2018-12-03
Move to typescript

sendilkumarn:

**Summary**

**Other information**

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "transform": {
       "^.+\\.(ts)?$": "ts-jest"
     },
-    "testRegex": "/__tests__/.*\\.(test.js|test.ts)$",
+    "testRegex": "/__tests__/.*\\.test.ts$",
     "moduleFileExtensions": [
       "ts",
       "js",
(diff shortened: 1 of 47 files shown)


package.json
```diff
@@ -72,7 +72,7 @@
     "transform": {
       "^.+\\.(ts)?$": "ts-jest"
     },
-    "testRegex": "/__tests__/.*\\.(test.js|test.ts)$",
```
ematipico: Because of that, all the bin cases test don't run anymore. We need to restore

test/testUtils.ts
```diff
@@ -1,28 +1,38 @@
 "use strict";
+import * as execa from "execa";
```
ematipico: This file has to stay a .js. It's used to test bin cases where we have only .js files. I don't think it's a good thing to mix ts and js.

Bin cases tests real world examples where js files are consumed by webpack. I don't see any point to move them to ts

ematipico: I wasn't aware of that part. we should update the documentation!

PR #710, opened 2018-12-09
tests: improve webpack integration tests

### LINKED ISSUE #705: Improve webpack integration tests
Inside `binCases` folder, we run a bunch of integrations test where we make sure that the output on webpack is correct against some specific arguments passed via CLI.

At the moment, when we get the output from a `webpack` process, we remove from it hash, timestamps, etc. and we provide this output to the test case.

This prevent from doing further test cases against hashes for example.

This is the function that strips the output BEFORE being consumed by the test

https://github.com/webpack/webpack-cli/blob/master/test/testUtils.js#L66-L77

The plan would be to create a second utility, ideally called `extractSummary` where we only extract the real content (where we have information about the modules parsed by `webpack`) and use it inside all the tests that at least have the snapshot tests. 

Of course, we will have to remove that `removeTimeStrings`.

The final code should something like this
```js
// my-test.test.js

const { run, extractSummary } = require("../../../testUtils");

test("function-promise", () => {
	const { code, stdout, stderr } = run(__dirname, [
		"--config",
		"./webpack.config.js",
		"--target",
		"async-node",
		"--mode",
		"production"
	]);

	expect(code).toBe(0);
    const summary = extractSummary(stdout)
	expect(summary).toEqual(expect.anything());
	expect(summary).toContain("entry.bundle.js");
	expect(stderr).toHaveLength(0);
	expect(summary).toMatchSnapshot();
});

```

N/A

**Additional context**
N/A

hemal7735:

improvement over current integration tests mentioned in issue #705 **Improve webpack integration tests**

NA

NA

**Summary**

No

**Other information** 
Closes #705 
Tasks:

- [X]  extractSummary from individual tests rather than doing it in test-runner function
- [x]  use extractHash wherever applicable

### DIFF
--- a/test/binCases/watch/info-verbosity-off/info-verbosity-off.test.js
+++ b/test/binCases/watch/info-verbosity-off/info-verbosity-off.test.js
@@ -16,7 +16,7 @@ const copyFilePath = path.resolve(__dirname, copyFile);
 // create copy of "index.js" => "index_copy.js"
 beforeEach(() => {
 	// fs.copyFileSync was added in Added in: v8.5.0
-	// We should migrate it once we stop support for v6.x
+	// We should refactor the below code once our minimal supported version is v8.5.0
 	fs.createReadStream(fileToChangePath).pipe(fs.createWriteStream(copyFilePath));
 });
 
--- a/test/binCases/watch/info-verbosity-off/info-verbosity-off.test.js
+++ b/test/binCases/watch/info-verbosity-off/info-verbosity-off.test.js
@@ -3,10 +3,9 @@
 jest.setTimeout(10E6);
 /* eslint-disable node/no-unsupported-features  */
 /* eslint-disable node/no-unsupported-features/es-syntax  */
-
-const { extractSummary, extractHash, appendDataIfFileExists, runAndGetWatchProc } = require("../../../testUtils");
 const fs = require("fs");
 const path = require("path");
+const { extractSummary, extractHash, appendDataIfFileExists, runAndGetWatchProc } = require("../../../testUtils");
 
 const fileToChange = "index.js";
 const copyFile = "index_copy.js";
--- a/test/binCases/watch/info-verbosity-off/info-verbosity-off.test.js
+++ b/test/binCases/watch/info-verbosity-off/info-verbosity-off.test.js
@@ -1,12 +1,12 @@
 "use strict";
 
-jest.setTimeout(10E6);
+jest.setTimeout(10e6);
 /* eslint-disable node/no-unsupported-features  */
 /* eslint-disable node/no-unsupported-features/es-syntax  */
 
-const { runWatch } = require("../../../testUtils");
+const { runWatch, extractSummary } = require("../../../testUtils");
 
-test("info-verbosity-off", async(done) => {
+test("info-verbosity-off", async done => {
 	const result = await runWatch(__dirname, [
 		"--entry ",
 		"./index.js",
@@ -23,11 +23,14 @@ test("info-verbosity-off", async(done) => {
 		"none"
 	]);
 	const { stdout, stderr } = result;
-	expect(stdout).toEqual(expect.anything());
-	expect(stdout).toContain("");
-	expect(stdout).not.toContain("webpack is watching the files…");
+
+	const summary = extractSummary(stdout);
+
+	expect(summary).toEqual(expect.anything());
+	expect(summary).toContain("");
+	expect(summary).not.toContain("webpack is watching the files…");
 
 	expect(stderr).toHaveLength(0);
-	expect(stdout).toMatchSnapshot();
+	expect(summary).toMatchSnapshot();
 	done();
 });
--- a/test/binCases/watch/info-verbosity-off/info-verbosity-off.test.js
+++ b/test/binCases/watch/info-verbosity-off/info-verbosity-off.test.js
@@ -85,7 +85,6 @@ test.only("info-verbosity-off", async done => {
 
 	webpackProc.stderr.on("data", error => {
 		// fail test case if there is any error
-		expect(true).toBe(false);
-		done();
+		done(error);
 	});
 });
--- a/test/binCases/watch/info-verbosity-off/info-verbosity-off.test.js
+++ b/test/binCases/watch/info-verbosity-off/info-verbosity-off.test.js
@@ -5,18 +5,15 @@ jest.setTimeout(10E6);
 /* eslint-disable node/no-unsupported-features/es-syntax  */
 const fs = require("fs");
 const path = require("path");
-const { extractSummary, extractHash, appendDataIfFileExists, runAndGetWatchProc } = require("../../../testUtils");
+const { extractSummary, extractHash, appendDataIfFileExists, runAndGetWatchProc, copyFile } = require("../../../testUtils");
 
 const fileToChange = "index.js";
-const copyFile = "index_copy.js";
 const fileToChangePath = path.resolve(__dirname, fileToChange);
-const copyFilePath = path.resolve(__dirname, copyFile);
+var copyFilePath;
 
 // create copy of "index.js" => "index_copy.js"
 beforeEach(() => {
-	// fs.copyFileSync was added in Added in: v8.5.0
-	// We should refactor the below code once our minimal supported version is v8.5.0
-	fs.createReadStream(fileToChangePath).pipe(fs.createWriteStream(copyFilePath));
+	copyFilePath = copyFile(__dirname, fileToChange);
 });
 
 afterEach(() => {
--- a/test/binCases/watch/info-verbosity-off/info-verbosity-off.test.js
+++ b/test/binCases/watch/info-verbosity-off/info-verbosity-off.test.js
@@ -21,8 +21,8 @@ beforeEach(() => {
 
 afterEach(() => {
 	try {
-		// deleting the file as it is modified by the test
 		// subsequent test-case runs won't pass as snapshot is not matched
+		// hence, deleting the file as it is modified by the test
 		fs.unlinkSync(fileToChangePath);
 	} catch (e) {
 		console.warn("could not remove the file:" + fileToChangePath + "\n" + e.message);
@@ -31,9 +31,7 @@ afterEach(() => {
 	}
 });
 
-// It is modifying the index.js
-// Which breaks the test-cases second time
-test.only("info-verbosity-off", async done => {
+test("info-verbosity-off", async done => {
 	var webpackProc = runAndGetWatchProc(__dirname, [
 		"--entry ",
 		"./index.js",
@@ -85,6 +83,6 @@ test.only("info-verbosity-off", async done => {
 
 	webpackProc.stderr.on("data", error => {
 		// fail test case if there is any error
-		done(error);
+		done(error.toString());
 	});
 });
--- a/test/binCases/watch/info-verbosity-off/info-verbosity-off.test.js
+++ b/test/binCases/watch/info-verbosity-off/info-verbosity-off.test.js
@@ -4,10 +4,38 @@ jest.setTimeout(10e6);
 /* eslint-disable node/no-unsupported-features  */
 /* eslint-disable node/no-unsupported-features/es-syntax  */
 
-const { runWatch, extractSummary } = require("../../../testUtils");
+const { extractSummary, extractHash, appendDataIfFileExists, runAndGetWatchProc } = require("../../../testUtils");
+const fs = require("fs");
+const path = require("path");
 
-test("info-verbosity-off", async done => {
-	const result = await runWatch(__dirname, [
+const fileToChange = "index.js";
+const copyFile = "index_copy.js";
+const fileToChangePath = path.resolve(__dirname, fileToChange);
+const copyFilePath = path.resolve(__dirname, copyFile);
+
+// create copy of "index.js" => "index_copy.js"
+beforeEach(() => {
+	// fs.copyFileSync was added in Added in: v8.5.0
+	// We should migrate it once we stop support for v6.x
+	fs.createReadStream(fileToChangePath).pipe(fs.createWriteStream(copyFilePath));
+});
+
+afterEach(() => {
+	try {
+		// deleting the file as it is modified by the test
+		// subsequent test-case runs won't pass as snapshot is not matched
+		fs.unlinkSync(fileToChangePath);
+	} catch (e) {
+		console.warn("could not remove the file:" + fileToChangePath + "\n" + e.message);
+	} finally {
+		fs.renameSync(copyFilePath, fileToChangePath);
+	}
+});
+
+// It is modifying the index.js
+// Which breaks the test-cases second time
+test.only("info-verbosity-off", async done => {
+	var webpackProc = runAndGetWatchProc(__dirname, [
 		"--entry ",
 		"./index.js",
 		"--config",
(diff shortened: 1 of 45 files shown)


hemal7735:
@ematipico task-2 is still in-progress, but do suggest if this is the way we want to do for `extractHash`

hemal7735:
Thank you for review @ematipico . Who else can review this PR?

ematipico:
Maybe @evenstensberg could suggest where to test the hashes

evenstensberg:
As if they are valid hashes, and yea that could be a fix. I was thinking about trying to reduce overhead, so we make sure every hash is different if we can for instance run watch and make a inline change to the bundle and try to watch the changes seeing that the hash is changing

hemal7735:
what is the downside of using `snapShot` for assertion?

evenstensberg:
It provides thrash files

ematipico:
> As if they are valid hashes

I think we should drop this for now (for this PR). Checking the validity of a hash means to run the same algorithm that webpack does, I think. Checking if there are numbers and/or letters is too reductive. 

But as you suggested we could have different solutions for checking the hashes

evenstensberg:
Yes, checking if the hash has changed is enough

hemal7735:
> Yes, checking if the hash has changed is enough

In that case we can call `extractHash` method to get the hash-object and compare it with snapshot, which will tell us that the current hash is what we got in the previous run.
What are your thoughts @ematipico @evenstensberg ?

evenstensberg:
Don't use snapshots please, it obfuscates the integration tests

hemal7735:
> if we can for instance run watch and make a inline change to the bundle and try to watch the changes seeing that the hash is changing

@ematipico @evenstensberg I might sound too naive here, but I am lost here to find the solution which can work.

https://github.com/webpack/webpack-cli/blob/master/test/testUtils.js#L34-L64 is where `runWatch` function is defined, which basically just runs the child process and output results.

**Question:**
I can make a change to file programmatically by appending some comment, but how will I get the updated the result back?

evenstensberg:
invoke webpack again perhaps by spawning a process?

hemal7735:
> invoke webpack again perhaps by spawning a process?

Yes, I thought of that, too as it serves the purpose of finding the changed hash. ✅ 
But would that be ideal as we are running webpack again instead of watching it?

hemal7735:
@evenstensberg is this change decent enough for hash assertion? https://github.com/webpack/webpack-cli/pull/710/commits/b8a32b758717ed58e707dcb83bc9c63c789e885b

evenstensberg:
No, you need to create a unique child process and invoke that explicly within the process instead of spawning a new webpack instance

hemal7735:
I'm sorry, I didn't get you. Can you provide some example?

evenstensberg:
1. Spawn a process
2. Change the file
3. Watch for changes

ematipico:
This approach is not doable, in general. How can we modify a file a second time without breaking the CI each time? The solution suggested by Even doesn't work

hemal7735:
> This approach is not doable, in general. How can we modify a file a second time without breaking the CI each time? The solution suggested by Even doesn't work

Hi @ematipico ,
I faced this issue in local as well. It is because of hash gets changed and snapshot test will fail.

To avoid that I have copied the file to be changed before test-case, and I'm restoring the original file after test-case. Please take a look at https://github.com/webpack/webpack-cli/pull/710/commits/1a6198687a17fb2532f696da0501210f5ef37fa0#diff-92fd40e106674a90ed774ea35c162bb6R17

ematipico:
Lot of work on this test! Amazing!!

hemal7735:
> Lot of work on this test! Amazing!!

Thanks @ematipico . I want to do this for all `watch` test-cases, once @evenstensberg  approves the approach.

test/binCases/watch/info-verbosity-off/info-verbosity-off.test.js
```diff
@@ -1,13 +1,41 @@
 "use strict";
 
-jest.setTimeout(10E6);
+jest.setTimeout(10e6);
```
evenstensberg: why change to lowercase?

test/binCases/watch/info-verbosity-off/info-verbosity-off.test.js
```diff
@@ -1,13 +1,41 @@
...
 /* eslint-disable node/no-unsupported-features  */
 /* eslint-disable node/no-unsupported-features/es-syntax  */
 
-const { runWatch } = require("../../../testUtils");
+const { extractSummary, extractHash, appendDataIfFileExists, runAndGetWatchProc } = require("../../../testUtils");
+const fs = require("fs");
```
evenstensberg: native requires before local ones

test/binCases/watch/info-verbosity-off/info-verbosity-off.test.js
```diff
@@ -1,13 +1,41 @@
...
-test("info-verbosity-off", async(done) => {
-	const result = await runWatch(__dirname, [
+const fileToChange = "index.js";
+const copyFile = "index_copy.js";
+const fileToChangePath = path.resolve(__dirname, fileToChange);
+const copyFilePath = path.resolve(__dirname, copyFile);
+
+// create copy of "index.js" => "index_copy.js"
+beforeEach(() => {
+	// fs.copyFileSync was added in Added in: v8.5.0
```
evenstensberg: we still support node v.6 so you will need to make a workaround for this

test/binCases/watch/info-verbosity-off/info-verbosity-off.test.js
```diff
@@ -22,12 +50,39 @@ test("info-verbosity-off", async(done) => {
...
-	expect(stdout).toMatchSnapshot();
-	done();
+
+	var outputCount = 0;
+	var hash1;
+
+	webpackProc.stdout.on("data", data => {
+		data = data.toString();
+
+		if (outputCount === 0) {
```
evenstensberg: if outputCount is 0 says me nothing, add comment to explain it's the first output count. You should extract this to a function explaining what it does in more detail

test/binCases/watch/info-verbosity-off/info-verbosity-off.test.js
```diff
@@ -22,12 +50,39 @@ test("info-verbosity-off", async(done) => {
...
+
+			expect(hash2.hash).not.toBe(hash1.hash);
+			webpackProc.kill();
+			done();
+		}
+	});
+
+	webpackProc.stderr.on("data", error => {
+		// fail test case if there is any error
+		expect(true).toBe(false);
```
evenstensberg: I don't like this. Is there a way to avoid writing redundant test cases and throw an error instead of using expect?

test/binCases/watch/info-verbosity-verbose/info-verbosity-verbose.test.js
```diff
@@ -1,12 +1,12 @@
 "use strict";
 
-jest.setTimeout(10E6);
+jest.setTimeout(10e6);
```
evenstensberg: `10E6`

test/testUtils.js
```diff
@@ -152,4 +162,23 @@ function extractHash(stdout) {
 	return hashInfo;
 }
 
-module.exports = { run, runWatch, extractHash };
+/*eslint valid-jsdoc: ["error", { "requireReturn": false }]*/
+/**
+ *
+ * @param {*} testCase - testCase directory
```
evenstensberg: Add types here please

test/binCases/watch/info-verbosity-off/info-verbosity-off.test.js
ematipico: Calling `done` with something will throw an error :)

test/binCases/watch/info-verbosity-off/info-verbosity-off.test.js
hemal7735: This is already a workaround. I did not use >=v8.5.0 feature `fs.copyFileSync` as we still support v6.

test/binCases/watch/info-verbosity-off/info-verbosity-off.test.js
hemal7735: Can you be a little bit more informative? Is this is a styling guide that we use? or is there a downside using it like this?

evenstensberg:
```
const fs = require('fs')
const {extractSummary} = ...
```

No, but it's standard convention in node

test/binCases/watch/info-verbosity-off/info-verbosity-off.test.js
evenstensberg: ok nice

test/binCases/watch/info-verbosity-off/info-verbosity-off.test.js
evenstensberg: Is there any way to avoid using a noop assertion for this?

hemal7735:
@webpack/cli-team can you help here understanding why Travis CI is failing?

ematipico:
There are a couple of tests that we have to fix that are not related to your issue

test/test-utils.test.js
```diff
@@ -101,3 +103,29 @@ Child ${config2Name}:
 		expect(hashInfo.config[1]).toEqual({ name: config2Name, hash: config2Hash });
 	});
 });
+
+describe("appendFile functionality", () => {
+	describe("positive test-cases", () => {
+		const junkFile = resolve(__dirname, "junkFile.js");
+
+		beforeEach(() => {
+			writeFileSync(junkFile, "");
```
evenstensberg: Here you're writing nothing an magically a junk comment appears if you see the analogy. Write something in `beforeEach` and then run `expect` after before each, then you need to make a change and check that it actually changes content as well

evenstensberg: Ok. Then you need to assert(expect) ∆data from the appending step (2) and number (4) and make sure that they hold.

hemal7735:
@ematipico @evenstensberg if you can review this again whenever possible?

test/binCases/watch/info-verbosity-verbose/info-verbosity-verbose.test.js
```diff
@@ -4,10 +4,36 @@ jest.setTimeout(10E6);
...
+const fileToChange = "index.js";
+const copyFile = "index_copy.js";
+const fileToChangePath = path.resolve(__dirname, fileToChange);
+const copyFilePath = path.resolve(__dirname, copyFile);
+
+// create copy of "index.js" => "index_copy.js"
+beforeEach(() => {
+	// fs.copyFileSync was added in Added in: v8.5.0
+	// We should refactor the below code once our minimal supported version is v8.5.0
+	fs.createReadStream(fileToChangePath).pipe(fs.createWriteStream(copyFilePath));
```
evenstensberg: you can abstract these to a util func

test/binCases/watch/info-verbosity-verbose/info-verbosity-verbose.test.js
```diff
@@ -22,12 +48,62 @@ test("info-verbosity-verbose", async(done) => {
...
+	// 2. webpack is watching the files...
+	// 3. Compilation  finished
+	// 4. Hash and other info
+	// 5. Compilation starting...
+	// 6. Compilation  finished
+	// 7. Hash and other info
+	var chunkNumber = 0;
+	var hash1, hash2;
+
+	webpackProc.stdout.on("data", data => {
```
evenstensberg: This is great, abstract this and use this for all watch tests

evenstensberg:
Good job so far

test/binCases/watch/info-verbosity-verbose/info-verbosity-verbose.test.js
hemal7735: Thank you @evenstensberg !
This pattern is used in all watch tests except `info-verbosity-off`. I will modify that to use this pattern.

test/binCases/watch/info-verbosity-verbose/info-verbosity-verbose.test.js
hemal7735: Sounds good!

test/binCases/watch/info-verbosity-verbose/info-verbosity-verbose.test.js
evenstensberg: with switch cases too? I didn't see :(

hemal7735: Yes yes 😉

evenstensberg: okay great, then abstract the common functions and you're good

hemal7735:
@evenstensberg can you review the changes?

hemal7735:
@evenstensberg even though it looks similar, and extracting it into common is the first thought that hit anyone's mind.
However, it contains `expect` statements which can differ when we evolve test-cases. Plus, having `expect` into some other file looks weird. This was the main reason I did not put into some other file.

I need var as I need to make changes to the variable.

evenstensberg:
Yes. That is fine. Add an argument that accept an array of expect assertions. You can make it 2Dimensional like 

`[{expect(..), afterEach]` or split into `[[expect, expect], [expect]]` where `i-th` row corresponds to the scope of assertion

hemal7735:
As I mentioned assertions are not independent of each other. They are sometimes involved with the previous step (for example, hash1 and hash2). Even if I abstract it, then it would not make a nice module, and the consumer would need to know how the underlying system works, which fails the point of abstraction.

I'm still not convinced that it will really work. It would make code look unreadable.
@ematipico I would like to your thoughts on this.

evenstensberg:
Ok

hemal7735:
> Mind rebasing?

yup, doing it 👍

evenstensberg:
great

hemal7735:
For some reason, CI is taking longer time. 
https://travis-ci.org/webpack/webpack-cli/builds/481769794

test/binCases/watch/info-verbosity-off/info-verbosity-off.test.js
```diff
@@ -3,11 +3,33 @@
...
 /* eslint-disable node/no-unsupported-features  */
 /* eslint-disable node/no-unsupported-features/es-syntax  */
+const fs = require("fs");
+const path = require("path");
+const { extractSummary, extractHash, appendDataIfFileExists, runAndGetWatchProc, copyFile } = require("../../../testUtils");
 
-const { runWatch } = require("../../../testUtils");
+const fileToChange = "index.js";
+const fileToChangePath = path.resolve(__dirname, fileToChange);
+var copyFilePath;
```
sendilkumarn: `let` ?

hemal7735:
@webpack/cli-team can someone check why Travis-CI test-phase is stuck while Azure is passed?

hemal7735:
It takes `23.955s` to run the tests in local.
Can anyone with Windows and Linux run the tests?
I'm using Mac.
I suspect this has something to do with folder permissions.

rishabh3112:
@hemal7735 Tests are passing on macOS as per `azure pipelines` too, failure seems to be platform specific as they passed on macOS only.
You can look for platform specific code changes in your code.
I shall look for linux in meanwhile.

rishabh3112:
@hemal7735 It took `113.973s` on my linux system and all passed.

hemal7735:
> @hemal7735 It took `113.973s` on my linux system and all passed.

@rishabh3112 thank you for running tests on linux machine. I guess you will also have Windows Machine with dual boot. If you can try on that as well?
I'm using node standard library `fs`. I'm not sure if that has something to do with platform-specific things.

rishabh3112:
No dual boot but `wsl`. I can't test in Windows as it will clutter my device environment with two nodes installations. You can use docker for this purpose though😉.

hemal7735:
@evenstensberg CI was failing for this PR. We should revert this PR.
@ematipico can you do the needful please?

rishabh3112:
Same with #732 , it was merged before work was completed.
Maybe I need to open a separate PR for rest work left there @evenstensberg ?

evenstensberg:
Ill fix

Issue #708, opened 2018-12-08
Don't scaffold "dist" property if user just chooses default.


TheLarkInn:
**What is the current behavior?**
`output: {path: path.resolve(__dirname, 'dist')}`

**Expected behavior**
If single entry and default output is chosen, do not even create it since webpack always defaults this. We want to stop creating config if we don't need to.

evenstensberg:
Having a look at it. Thanks for reporting!

EvsChen:
@evenstensberg Can I take a try on this?

evenstensberg:
Sure, go ahead

dhruvdutt:
@EvsChen Sure! Let us know if you run into any issues.

PR #716, opened 2018-12-22
misc: not scaffold 'dist' if the user chooses default

EvsChen:
As mentioned in #708 , we don't need to scaffold the "dist" property in output path if the entry is single and output path is not chosen by the user.

misc
N/A

**Summary**

No

**Other information**

### DIFF
--- a/packages/generators/init-generator.ts
+++ b/packages/generators/init-generator.ts
@@ -59,7 +59,6 @@ export default class InitGenerator extends Generator {
 		const self: this = this;
 		let regExpForStyles: string;
 		let ExtractUseProps: object[];
-		let outputPath: string = "dist";
 
 		process.stdout.write(
 			"\n" +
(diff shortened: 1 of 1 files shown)


evenstensberg:
Yes it will, but the default output path is `dist`. What @TheLarkInn is requesting is that we do not set the option at all in the generated configuration file 👍

EvsChen:
The `dist` will not appear in the output. If the user chooses default output and does not input anything, the `outputTypeAnswer.outputType.length` will be equal to 0 and the output will not be set. 

Did I miss some point here? I think the only place that sets the output option is in this `if` block.

evenstensberg:
If you set multiple entries and press enter you can see the default output option being set regardless. Idea is to remove that setting entirely

EvsChen:
As what mentioned in #708, I think @TheLarkInn only wants to remove it for the single entry case.

evenstensberg:
Webpack will set that property regardless, and it should not be in the generated configuration file if it is set to `dist`

PR #718, opened 2018-12-28
misc: reduce hop for creating an issue

hemal7735:
This is more of a **feasibility** kind of change for the benefit of Issue Raiser. Please feel free to close this PR if it adds a noise.

Currently, when we face any issue while running transformation we throw an error giving an URL
https://github.com/webpack/webpack-cli/issues

However, the bug reporter has to do one more hop to create an issue.

This PR changes that URL to  https://github.com/webpack/webpack-cli/issues/new?template=Bug_report.md

**Did you add tests for your changes?**  NA

**Does this PR introduce a breaking change?** NO

### DIFF
--- a/packages/utils/modify-config-helper.ts
+++ b/packages/utils/modify-config-helper.ts
@@ -123,7 +123,12 @@ export default function modifyHelperUtil(
 		);
 		return runTransform(transformConfig, action);
 	}).catch((err) => {
-		console.error(chalk.red("\nUnexpected Error, please file an issue to https://github.com/webpack/webpack-cli\n"));
+		console.error(
+			chalk.red(
+				// tslint:disable-next-line:max-line-length
+				"\nUnexpected Error, please file an issue to https://github.com/webpack/webpack-cli/issues/new?template=Bug_report.md \n",
+			),
+		);
 		console.error(err);
 	});
 }
--- a/packages/utils/modify-config-helper.ts
+++ b/packages/utils/modify-config-helper.ts
@@ -125,8 +125,10 @@ export default function modifyHelperUtil(
 	}).catch((err) => {
 		console.error(
 			chalk.red(
-				// tslint:disable-next-line:max-line-length
-				"\nUnexpected Error, please file an issue to https://github.com/webpack/webpack-cli/issues/new?template=Bug_report.md \n",
+				`
+Unexpected Error
+please file an issue here https://github.com/webpack/webpack-cli/issues/new?template=Bug_report.md
+				`,
 			),
 		);
 		console.error(err);


packages/utils/modify-config-helper.ts
```diff
@@ -123,7 +123,12 @@ export default function modifyHelperUtil(
 		);
 		return runTransform(transformConfig, action);
 	}).catch((err) => {
-		console.error(chalk.red("\nUnexpected Error, please file an issue to https://github.com/webpack/webpack-cli\n"));
+		console.error(
+			chalk.red(
+				// tslint:disable-next-line:max-line-length
+				"\nUnexpected Error, please file an issue to https://github.com/webpack/webpack-cli/issues/new?template=Bug_report.md \n",
```
rishabh3112: Use a template literal instead.
```suggestion
				`
				Unexpected Error
				please file an issue here https://github.com/webpack/webpack-cli/issues/new?template=Bug_report.md 
				`,
```

packages/utils/modify-config-helper.ts
```diff
@@ -123,7 +123,12 @@ export default function modifyHelperUtil(
 		);
 		return runTransform(transformConfig, action);
 	}).catch((err) => {
-		console.error(chalk.red("\nUnexpected Error, please file an issue to https://github.com/webpack/webpack-cli\n"));
+		console.error(
+			chalk.red(
+				// tslint:disable-next-line:max-line-length
```
rishabh3112: Remove this after using string literal
```suggestion

```

packages/utils/modify-config-helper.ts
rishabh3112: Also use `here` instead of  `to` as in above suggestion

hemal7735: Initially, I tried with template literal but it messes up the formatting. One can't just treat content inside template literal as it is and add a line break. It adds unwanted space, and if we remove that space then its alignment looks ugly with the rest of the code.

rishabh3112: Still I think it looks better that using `\n`.
I am Okay without template literals too if you like that way 🙂.
I have that unwanted spaces removed in suggestion if you like to consider.

packages/utils/modify-config-helper.ts
evenstensberg: don't do that

packages/utils/modify-config-helper.ts
hemal7735: Yup, I noticed that. Initially tried that, too, but it just does not align with the rest of the code. 😉

rishabh3112:
Bot gone nuts 😂

hemal7735:
> Who pushed commit using my account 🤔

It is a GitHub co-commit feature, where suggestions can be directly committed, and it will show both names. In this case yours **(suggested changes)**  and mine **(accepted changes)**

rishabh3112:
Well changed my comment accordingly.
I am on mobile right now which don't show co commiters 😅

Issue #724, opened 2019-01-03
`webpack.config.ts` can not be resolved any more after upgrading to 3.2.0


JounQin:
**Describe the bug**

As Title.

**What is the current behavior?**

```bash
(function (exports, require, module, __filename, __dirname) { import path from 'path';
                                                                     ^^^^

SyntaxError: Unexpected identifier
```

**To Reproduce**
use `webpack.config.ts` with `ts-node`

**Expected behavior**
build successfully

evenstensberg:
Checking it out.

evenstensberg:
can you make a repro repo for me to make it easier to debug?

matmannion:
Not specific to typescript, we have the same problem with webpack.config.babel.js. Downgrading webpack-cli to 3.1.2 fixes this

evenstensberg:
Gotcha. I think that we will need to add `interpret` again, we removed that dependency. Fixing this this week sorry for the inconvience

retorquere:
Looks like this was fixed in the 3.2.1 release.

ematipico:
@JounQin if you think the issue is fixed, feel free to close the issue!

Enase:
FYI: it's fixed for `webpack.config.babel.js` for me as well.

JounQin:
Thanks.

zenVentzi:
Still getting this error

```

import path from 'path';
       ^^^^

SyntaxError: Unexpected identifier
```


```
"devDependencies": {
    "@babel/core": "^7.3.4",
    "@babel/plugin-proposal-class-properties": "^7.3.4",
    "@babel/plugin-proposal-export-namespace-from": "^7.2.0",
    "@babel/plugin-proposal-object-rest-spread": "^7.3.4",
    "@babel/plugin-transform-runtime": "^7.3.4",
    "@babel/preset-env": "^7.3.4",
    "@babel/preset-typescript": "^7.3.3",
    "@types/node": "^11.10.4",
    "@types/webpack-bundle-analyzer": "^2.13.1",
    "babel-loader": "^8.0.5",
    "ts-node": "^8.0.2",
    "typescript": "^3.3.3333",
    "webpack": "^4.29.6",
    "webpack-bundle-analyzer": "^3.1.0",
    "webpack-cli": "^3.2.3"
  }
```

```
{
  "compilerOptions": {
[...]

PR #728, opened 2019-01-05
fix(bin): extension detection (#724)

lakatostamas:
bugfix

No, but I created test repositories: 
https://github.com/lakatostamas/webpack-babel
https://github.com/lakatostamas/webpack-ts

**Summary**
Issue: https://github.com/webpack/webpack-cli/issues/724
This commit introduced a bug in the default config file discovery: https://github.com/webpack/webpack-cli/pull/693/commits/b61220edb1bb7a61cf9ee7370e7aa752143e7684#diff-daabbe74e88640829a3ad9e9899efd8cR79
Before this commit, the extension of `webpack.config.ts` was `.ts` in the default configFile discovery, but after this commit, it became just `ts` (without `.`). `webpack.config.babel.js` extension is `js` now, but it should be `.babel.js`. With this pull request, I introduce a fix, which gets the name of the default config file and removes the `webpack.config` or `webpackfile` from it.
I don't know how to add tests when the extension of webpack config is not `js`. I haven't found any example, but if you tell me how to do it, I would happily add some tests to cover these cases.

No


zzh8829:
This should fix the problem of .babel.js discussed in https://github.com/webpack/webpack/issues/5960
Looks like the error was actually introduced in https://github.com/webpack/webpack-cli/commit/14ec0a5acf5fbe47bd04be0ec2712d6f355907eb#diff-daabbe74e88640829a3ad9e9899efd8cR90

ematipico:
We should add a bin test for that, if possible

ematipico:
I personally prefer install an additional devDependency and make sure that we don't break those kind of scenarios. After all this is a package used by lot of people :) what do you think?

lakatostamas:
Okay, that makes sense. I added `webpack.config.ts` test

PR #733, opened 2019-01-12
feat: add dotfolder lookup and update snapshot

evenstensberg:
Adds folder lookup for `.webpack`. I'm not so good with that optional regex prefix, is there anyone that knows? In defaultConfigFileRegexp

No

**Other information**
N/A

### DIFF
--- a/bin/convert-argv.js
+++ b/bin/convert-argv.js
@@ -72,15 +72,27 @@ module.exports = function(...args) {
 		const defaultConfigFileNames = ["webpack.config", "webpackfile"].join("|");
 		const webpackConfigFileRegExp = `(${defaultConfigFileNames})(${extensions.join("|")})`;
 		const pathToWebpackConfig = findup(webpackConfigFileRegExp);
-
 		if (pathToWebpackConfig) {
 			const resolvedPath = path.resolve(pathToWebpackConfig);
 			const actualConfigFileName = path.basename(resolvedPath);
 			const ext = actualConfigFileName.replace(new RegExp(defaultConfigFileNames), "");
 			configFiles.push({
 				path: resolvedPath,
-				ext,
+				ext
 			});
+		} else {
+			const pathToWebpackConfig = findup(webpackConfigFileRegExp, {
+				cwd: process.cwd() + "/.webpack/"
+			});
+			if (pathToWebpackConfig) {
+				const resolvedPath = path.resolve(pathToWebpackConfig);
+				const actualConfigFileName = path.basename(resolvedPath);
+				const ext = actualConfigFileName.replace(new RegExp(defaultConfigFileNames), "");
+				configFiles.push({
+					path: resolvedPath,
+					ext
+				});
+			}
 		}
 	}
 	if (configFiles.length > 0) {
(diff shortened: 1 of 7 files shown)


bin/convert-argv.js
```diff
@@ -72,15 +72,27 @@ module.exports = function(...args) {
 		const defaultConfigFileNames = ["webpack.config", "webpackfile"].join("|");
 		const webpackConfigFileRegExp = `(${defaultConfigFileNames})(${extensions.join("|")})`;
```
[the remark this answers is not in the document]
lakatostamas: This RegExp would be correct, but `findup-sync` uses `micromatch` as a matcher and it doesn't support non-capturing groups. It supports optional characters in the following way: `?(pattern)` or `(pattern|)`.  However `findup-sync` doesn't check files/dirs recursively on each level (it uses fs.readdirSync), so it returns only `.webpack` (not `.webpack/webpack.config.js`) thus I don't think it's possible to solve this task with `findup-sync` + one RegExp. 😞
patterns:
https://github.com/micromatch/extglob#extglob-cheatsheet
https://github.com/micromatch/micromatch#matching-features
findup-sync readDir:
https://github.com/gulpjs/findup-sync/blob/master/index.js#L61

bin/convert-argv.js
```diff
@@ -72,15 +72,27 @@ module.exports = function(...args) {
...
 			const resolvedPath = path.resolve(pathToWebpackConfig);
 			const actualConfigFileName = path.basename(resolvedPath);
 			const ext = actualConfigFileName.replace(new RegExp(defaultConfigFileNames), "");
 			configFiles.push({
 				path: resolvedPath,
-				ext,
+				ext
+			});
+		} else {
+			const pathToWebpackConfig = findup(webpackConfigFileRegExp, {
```
lakatostamas: If I understand correctly then using `findup` here is unnecessary. 
```
.
├── .webpack
│   └── webpack.config.js
└── packages
    └── foo
        └── .webpack  
            └── webpack.config.js
```
If you have this project structure and you run webpack in `./package/foo` it will resolve the config (`./packages/foo/.webpack/webpack.config.js`).
```
.
├── .webpack
│   └── webpack.config.js
└── packages
    └── foo
```
But if you don't have a webpack config inside foo, then the default config will be used, because it will perform the following checks.
1, `./packages/foo/.webpack` -- no match
2, `./packages/` -- no match
3, `.` -- no match, `findup-sync` will retrun `.webpack` (not it's content) and it won't match our current regex.

bin/convert-argv.js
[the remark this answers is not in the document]
evenstensberg: Gotcha, thanks!

test/binCases/help/help-output/__snapshots__/help-output.test.js.snap
```diff
@@ -1,7 +1,7 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`help-output 1`] = `
-"webpack-cli 3.1.2
+"webpack-cli 3.2.1
```
ematipico: I think we should remove the snapshot test.
Every time there's a release, it will break

evenstensberg: agree, but I think we'll have to revise this PR until the new cli is out, no way to look up dotfiles

Issue #741, opened 2019-01-24
Bug: "webpack-cli init" errors when trying to install @webpack-cli/init


Pomax:
**Describe the bug**

I tried to run `webpack-cli init` without having the separate init module installed, which makes webpack-cli inform me that there is a new module for that, and asks me whether it should install that. Saying "yes" leads to the following error:

```
i:\temp\mj>npx webpack-cli init
npx: installed 1 in 1.133s
Path must be a string. Received undefined
i:\temp\mj\node_modules\webpack-cli\bin\cli.js
The command moved into a separate package: @webpack-cli/init
Would you like to install init? (That will run npm install -g @webpack-cli/init) (yes/NO)yes
{ Error: Cannot find module 'i:\temp\mj\node_modules\@webpack-cli\init'
    at Function.Module._resolveFilename (module.js:547:15)
    at Function.Module._load (module.js:474:25)
    at Module.require (module.js:596:17)
    at require (i:\temp\mj\node_modules\v8-compile-cache\v8-compile-cache.js:159:20)
    at runCommand.then.result (i:\temp\mj\node_modules\webpack-cli\bin\prompt-command.js:94:15)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7) code: 'MODULE_NOT_FOUND' }
```

If I install `@webpack-cli/init` manually (and without `-g`, haven't tried with) then that error does not occur.

rishabh3112:
Thanks, @Pomax for reporting. It should be fixed soon.

ematipico:
PR merged, next release will contain the fix!

Pomax:
sweet, many thanks!

PR #742, opened 2019-01-25
fix(init): fix init prompt installation

rishabh3112:

bugfix
N/A
N/A
**Summary**
#741

No
**Other information**

### DIFF
--- a/bin/prompt-command.js
+++ b/bin/prompt-command.js
@@ -59,15 +59,6 @@ module.exports = function promptForInstallation(packages, ...args) {
 			options[0] = "add";
 		}
 
-		if (packages === "init") {
-			if (isYarn) {
-				options.splice(1, 1); // remove '-D'
-				options.splice(0, 0, "global");
-			} else {
-				options[1] = "-g";
-			}
-		}
-
 		const commandToBeRun = `${packageManager} ${options.join(" ")}`;
 
 		const question = `Would you like to install ${packages}? (That will run ${commandToBeRun}) (yes/NO)`;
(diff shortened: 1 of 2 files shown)


rishabh3112:
Hi @evenstensberg, `next` branch has refactored `bin` into `lib`, can you point me where to find installation prompt ?

evenstensberg:
You can navigate the lib folder, it's not that big, and you learn the new internals

rishabh3112:
@evenstensberg we don't do a global installation of `@webpack-cli/init` as per `next` branch. So `next` branch code doesn't encounter the error this PR supposed to solve. So there are two solutions:

1. revert global installation setup to local for `@webpack-cli/init` as in `next` branch ( I don't like this personally)
2. Get location to the global `@webpack-cli/init` package by getting `prefix` programatically using npm.

Which one I should implement?

evenstensberg:
Do as 1. Does it and ill review

rishabh3112:
Did that @evenstensberg

rishabh3112:
Any review here 😅?

bin/prompt-command.js
```diff
@@ -60,15 +59,6 @@ module.exports = function promptForInstallation(packages, ...args) {
 			options[0] = "add";
 		}
 
-		if (packages === "init") {
```
evenstensberg: why is this line removed?

rishabh3112: Doing local installation for `init` as decided, so this `if` is not required now in which we are replacing `-D` with global equivalents.

evenstensberg: I think we are allowing global installation as well

rishabh3112: Refer
https://github.com/webpack/webpack-cli/pull/742#issuecomment-460654658

Here you said for local setup.

Currently we are not allowing global installation in this scenario.

evenstensberg: Oh sorry. So we need global installation. This PR needs to support that

rishabh3112: Okay :+1:  will do it.

rishabh3112:
Done @evenstensberg

rishabh3112:
@evenstensberg NodeJS stable is failing on this PR.
Screenshot from 2019-03-09 15-39-18
On master too.
I have fixed commit messages here and is ready for review.

rishabh3112:
@evenstensberg After investigation error is certainly due to `jest-util` package having the error as per `Node v11.11.0` (stable). it's working fine at lower node versions.

pasting error for the reference.

```
TypeError: Cannot assign to read only property 'Symbol(Symbol.toStringTag)' of object '#<process>'
      at exports.default (node_modules/jest-util/build/create_process_object.js:15:34)
```

evenstensberg:
Any fix to this?

rishabh3112:
@evenstenberg This is under a fix at jest here.
https://github.com/facebook/create-react-app/issues/6591

Mean while we can either use Node v11.10 or don't test on stable while it's under fix.

rishabh3112:
Updating jest to v24.3.0 would fix this.

Issue #752, opened 2019-02-05
Enh: Opencollective post install script


evenstensberg:
**Is your feature request related to a problem? Please describe.**
As this is an open source project we need some sort of donation banner to indicate that we are looking for financial support. 

**Describe the solution you'd like**
A postinstall script in `package.json` that doesn't crash other platforms. I've created a bash script that should be written in JS instead of bash, being outputted when user installs webpack.

**Describe alternatives you've considered**
N/A

**Additional context**

https://github.com/evenstensberg/webpack-cli-scripts/blob/master/scripts/opencollective.sh

misterdev:
I think this can be closed :)

ematipico:
@misterdev 	A small advise: if you use magic comments in your PRs, you will be able to close automatically the issues once they are merged!

```
Closes #xxx
```

More here: https://help.github.com/en/articles/closing-issues-using-keywords

misterdev:
Thanks @ematipico! Yeah, I know but this time I forgot to include that somewhere :)

PR #755, opened 2019-02-06
enh(Add open collective prompt)

misterdev:
I'm adding a console message to ask the users for donations, following how it is implemented in [opencollective-cli](https://github.com/opencollective/opencollective-cli/blob/master/src/lib/print.js#L68)

OSX
screenshot 2019-02-06 at 16 15 01

Windows
postinstall

At the moment this is printed by the postinstall script. It can be improved by printing the banner sometimes during the cli usage as suggested in @evenstensberg [example](https://github.com/evenstensberg/webpack-cli-scripts/blob/master/scripts/opencollective.sh) 
```bash
LANG=C DAY=$(date +"%a")

if [ "$DAY" == "Mon" ];
then
    printDonationHeader
```
I think we can print it during cli usage with probability 1/50, I will work on this. 

No. Should I?

It's not relevant

**Summary**
Adds a "support us on open collective" closing #752 

No


package.json
```diff
@@ -1,165 +1,165 @@
 {
-  "name": "webpack-cli",
```
ematipico: Is the change here due to Spaces VS tabs?

bin/opencollective.js
```diff
@@ -0,0 +1,41 @@
...
+	str = chalk[color](str);
+
+	console.log(leftPadding, str);
+}
+
+console.log("");
+print(chalk.bold.blue("Thanks for using Webpack!"));
+print("Please consider donating to our open collective");
+print("to help us maintain this package.");
+console.log("");
```
ematipico: Just a nice to have: 
```js
console.log('\n\n')
```

Instead of double log?

bin/opencollective.js
```diff
@@ -0,0 +1,41 @@
...
+// Only show emoji on OSx (Windows shell doesn't like them that much ¯\_(ツ)_/¯ )
+function emoji(emoji) {
+	if (process.stdout.isTTY && process.platform === "darwin") {
+		return emoji;
+	} else {
+		return "";
+	}
+}
+
+function print(str, color) {
```
ematipico:
```suggestion
function print(str = '', color = 'dim') {
```

It's supported by node 6 :)

bin/opencollective.js
```diff
@@ -0,0 +1,41 @@
...
+	str = str || "";
+	color = color || "dim";
+	const terminalCols = process.platform === "win32" ? 80 : parseInt(child_process.execSync("tput cols").toString());
+	// eslint-disable-next-line no-control-regex
+	const strLength = str.replace(/\u001b\[[0-9]{2}m/g, "").length;
```
ematipico: could you store the regex into a variable and give to that variable a good name? So we know what's that regex for

bin/opencollective.js
```diff
@@ -0,0 +1,41 @@
+const child_process = require("child_process");
+const chalk = require("chalk");
+
+function padding(length) {
+	var padding = "";
```
ematipico: I think we are good to use `const`

bin/opencollective.js
misterdev: Makes sense :)

bin/opencollective.js
misterdev: Thanks, I replaced this function with `" ".repeat(length)` that's supported by node 4

misterdev:
Thanks @ematipico for your feedback, I've included those changes.

I've improved the formatting of the output, now it's prettier and well centered:

screenshot 2019-02-06 at 22 53 52

Also, this is showed once a week, I do this by saving the timestamp of the last print in a file called `./.lastocprint`, so every time we run webpack-cli it have to read from the file system. Is this a good solution?

I think this is complete, feedback are welcome 👍

misterdev:
@ematipico you can review this if you want

bin/opencollective.js
```diff
@@ -0,0 +1,44 @@
...
+
+function print(str = "", color = "dim") {
+	const terminalCols = process.platform === "win32" ? 80 : parseInt(child_process.execSync("tput cols").toString());
+	// eslint-disable-next-line no-control-regex
+	const ansiEscapeSeq = /\u001b\[[0-9]{1,2}m/g;
+	const strLength = str.replace(ansiEscapeSeq, "").length;
+	const leftPaddingLength = Math.floor((terminalCols - strLength) / 2);
+	const leftPadding = " ".repeat(leftPaddingLength);
+	str = chalk[color](str);
+	console.log(leftPadding, str);
```
evenstensberg: Hi, nice work. Can you add a day check to only output this on mondays?

misterdev: I had some problems with the azure CI but it's done now ;)

bin/opencollective.js
```diff
@@ -0,0 +1,46 @@
...
+	print(`Please consider donating to our ${chalk.bold.blue("Open Collective")}`);
+	print("to help us maintain this package.");
+	console.log("\n\n");
+	print(`${emoji("👉")} ${chalk.bold.yellow(" Donate:")} ${chalk.reset.underline.yellow("https://opencollective.com/webpack/donate")}`);
+	console.log("\n");
+}
+
+
+const now = new Date();
+if (now.getDay() === 1) {
```
evenstensberg: extract this to a variable, indicating which day it is. uppercase like this: `MONDAY`

bin/opencollective.js
```diff
@@ -0,0 +1,47 @@
...
+}
+
+
+const now = new Date();
+const MONDAY = 1;
+if (now.getDay() === MONDAY) {
+	const lastPrintFile = path.resolve(__dirname, "../.lastocprint");
+	fs.readFile(lastPrintFile, "utf8", (err, lastPrint = 0) => {
+		if (err && err.code !== "ENOENT") return;
+		if (now - lastPrint > 6 * 24 * 60 * 60 * 1000) {
```
evenstensberg: this needs to be in english as well

bin/opencollective.js
```diff
@@ -0,0 +1,47 @@
...
+	print(`${chalk.bold("Thanks for using")} ${chalk.bold.blue("Webpack!")}`);
+	print(`Please consider donating to our ${chalk.bold.blue("Open Collective")}`);
+	print("to help us maintain this package.");
+	console.log("\n\n");
+	print(`${emoji("👉")} ${chalk.bold.yellow(" Donate:")} ${chalk.reset.underline.yellow("https://opencollective.com/webpack/donate")}`);
+	console.log("\n");
+}
+
+
+const now = new Date();
```
evenstensberg: just printing on mondays regardless of diff is ok. Could you remove the file with the timestamp? It's unneeded and will bloat our code.

bin/opencollective.js
misterdev: ok!

bin/opencollective.js
misterdev: If we only check the day of the week, we will print every time the cli is executed on Monday.
I don't like the file too, other solutions are welcome

evenstensberg: https://nodejs.org/api/fs.html#fs_stats_atime

misterdev: I've updated the code using `atime`, thank you for the suggestion. Works on MacOS, I need to test it on Windows.
I'm busy with some exams, I'll keep working on this in the next days

misterdev: done

misterdev:
The prompt shows only on Monday if the last time it has been shown was more than 6 days earlier.
I've used `Stats.atime` as suggested by @evenstensberg to determine the last time the file (`opencollective.js`) has been executed. On windows seems that the atime is not updated when `opencollective.js` is executed, so I've added an instruction that sets it.

bin/cli.js
```diff
@@ -473,6 +473,22 @@ For more information, see https://webpack.js.org/api/cli/.`);
 					const statsString = stats.toString(outputOptions);
 					const delimiter = outputOptions.buildDelimiter ? `${outputOptions.buildDelimiter}\n` : "";
 					if (statsString) stdout.write(`${statsString}\n${delimiter}`);
+
+
+					const now = new Date();
```
evenstensberg: add comment here , maybe abstract to a local function. Needs more comments and explaination.

bin/cli.js
```diff
@@ -473,6 +473,22 @@ For more information, see https://webpack.js.org/api/cli/.`);
...
+					const MONDAY = 1;
+					const SIX_DAYS = 518400000;
+					const openCollectivePath = __dirname + "/opencollective.js";
+					if (now.getDay() === MONDAY) {
+						const { statSync, utimesSync } = require("fs");
+						const lastPrint = statSync(openCollectivePath).atime;
+						const lastPrintTS = new Date(lastPrint).getTime();
+						if (now.getTime() - lastPrintTS > SIX_DAYS) {
```
evenstensberg: abstract to expression or function that has the naming `installedLaterThanSixDays` or similar.

bin/cli.js
misterdev: I've added this comment on top:
```js
/**
* Show a hint to donate to our Opencollective
* once a week, only on Monday
*/
```

bin/cli.js
misterdev: I've abstracted to a variable:

```js
const timeSinceLastPrint = now.getTime() - lastPrintTS;
if (timeSinceLastPrint > SIX_DAYS) {
```

Thanks for your comments, let me know if I can make any other improvements 👍

bin/cli.js
```diff
@@ -473,6 +473,26 @@ For more information, see https://webpack.js.org/api/cli/.`);
...
+					const now = new Date();
+					if (now.getDay() === MONDAY) {
+						const { statSync, utimesSync } = require("fs");
+						const lastPrint = statSync(openCollectivePath).atime;
+						const lastPrintTS = new Date(lastPrint).getTime();
+						const timeSinceLastPrint = now.getTime() - lastPrintTS;
+						if (timeSinceLastPrint > SIX_DAYS) {
+							require(openCollectivePath);
+							// On windows we need to manually update the atime
+							utimesSync(openCollectivePath, now, now);
```
sbrunner: This line breaks the build on Linux when we install Webpack globally :-(

misterdev: :( Could you please open an issue with some details about the error?

sbrunner: => https://github.com/webpack/webpack-cli/issues/870 :-)

PR #613, opened 2018-09-28
tests(migration): typescript

dhruvdutt:
**What kind of change does this PR introduce?**
Tests: Refactoring to TypeScript

**Did you add tests for your changes?**
Yes

**If relevant, did you update the documentation?**
NA

**Summary**
Added `ts-test` branch on upstream. We will keep working on this branch to migrate all tests to TypeScript.

**Does this PR introduce a breaking change?**
No

**Other information**
cc @sendilkumarn @ematipico 

Packages that contains tests to port to typescript
- [x] add
- [x] generate-loader
- [x] generate-plugin
- [x] generators
- [X] info
- [x] init
- [x] make
- [X] migrate
- [x] remove
- [x] serve
- [x] update
- [x] utils
- [x] webpack-scaffold: #686 by @hemal7735

PR #765, opened 2019-02-19
tests(migration): including requested changes

misterdev:
Includes changes requested in #613 

**Summary**

Before committing I had to solve those issues:

- Removed [deprecated](https://conventional-changelog.github.io/commitlint/#/reference-rules?id=lang-deprecated) `lang: [0, "always", "eng"]` from commitlint configuration. It gave me errors preventing me from committing:
    >RangeError: Found invalid rule names: lang. Supported rule names are: body-case, body-empty

- Fixed a tslint error:
```
 ERROR: packages/migrate/index.ts:20:4 - Type literal has only a call signature — use 
`new(errors: string[]) => {
    message: string;
},` instead.
```

Those are the actual changes:
1. Adds an interface to `transformations` in `packages/migrate/migrate.ts`
2. Refactor every test in `migrate/**/__tests__/**.test.ts` 
```ts
//before
defineTest(join(__dirname, ".."), "resolve");

//after
const dirName: string = path.join(__dirname, "..");
defineTest(dirName, "resolve");
```
3. Adds a type notation in `packages/utils/__tests__/ast-utils.test.ts`

4. I've also refactored `packages/migrate/resolve/__tests__/__testfixtures__/resolve.input.js` following [this](https://github.com/webpack/webpack-cli/pull/613/files#r241777569) comment.
There's something I don't understand well, changing `const path = require("path")` with `const {resolve} = require("path")` gives me an error:
screenshot 2019-02-19 at 20 38 23

Why does this happens?

Feedbacks are welcome :)

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -12,7 +12,6 @@ module.exports = {
 		"body-leading-blank": [1, "always"],
 		"footer-leading-blank": [1, "always"],
 		"header-max-length": [2, "always", 80],
-		lang: [0, "always", "eng"],
 		"scope-case": [2, "always", "lowerCase"],
 		"scope-empty": [0, "never"],
 		"subject-case": [2, "never", ["sentence-case", "start-case", "pascal-case", "upper-case"]],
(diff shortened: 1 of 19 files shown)


ematipico:
If you inside a typescript file, I discorage the use of `require` if possible

misterdev:
Thanks for reviewing!
The only `require` I see are in `resolve.input.js` that's a js file and in `resolve.test.ts.snap` that I think is a jest snapshot

Issue #774, opened 2019-03-06
Docs: Publish docs on every push from gh-pages branch


anshumanv:
**Do you want to request a *feature* or report a *bug*?**
Docs feature maybe

**What is the current behavior?**
Docs are pushed with the CLI code.

**If the current behavior is a bug, please provide the steps to reproduce.**
NA

**What is the expected behavior?**
I think it makes sense to automate the docs publishing process so docs live separate from the CLI code and are updated on every push, we can simply run a travis hook for every push for this and then we can host them on gh-pages or elsewhere.

**If this is a feature request, what is motivation or use case for changing the behavior?**
Automated docs, cleaner code

**Please paste the results of `webpack-cli info` here, and mention other relevant information such as programming language.**
```
webpack-cli info

  System:
    OS: macOS 10.14.3
    CPU: (4) x64 Intel(R) Core(TM) i5-5250U CPU @ 1.60GHz
  Binaries:
    Node: 10.14.1 - /usr/local/bin/node
    Yarn: 1.12.3 - /usr/local/bin/yarn
    npm: 6.4.1 - /usr/local/bin/npm
  Browsers:
    Chrome: 72.0.3626.119
    Safari: 12.0.3
  npmPackages:
    webpack: 4.x.x => 4.29.6
    webpack-dev-server: ^3.1.10 => 3.2.1
```

Will do a quick PR in case we agree to proceed with this. 😄

anshumanv:
@evenstensberg using Netlify as the hosting service, is that alright?

anshumanv:
Here's a sample how it looks @evenstensberg 😉 

https://webpack-cli.netlify.com/

evenstensberg:
why netify over gh?

anshumanv:
@evenstensberg here's a comparative breakdown - https://www.netlify.com/github-pages-vs-netlify/

Also we don't need to push the docs anywhere too (gh-pages branch for instance) Netlify takes care of that. We can also get updated docs on each PR but we won't need that IMO.

evenstensberg:
I don't think we'll need netify

ematipico:
Is there a way to have github pages without pushing the new documentation every time?

anshumanv:
Yes, manually push them when you need to @ematipico 

[...]

PR #782, opened 2019-03-09
docs(all): removed hard-coded docs, moved md to /docs

anshumanv:
Removed all the hard-coded docs since we'll setup auto-deployment, moved some md docs to /docs
folder instead for a cleaner root dir.

ISSUES CLOSED: #774

Docs

Yes

Yes.

**Summary**
* Removed hard-coded docs, moved some markdowns to /docs instead
* Docs will be auto-deployed using Netlify.

No

**Other information**


anshumanv:
@evenstensberg I need to have push access to add the project on Netlify and enable docs generation on every push. I can do that or you can do it instead, works either way.

Also what domain do we take? `webpack-cli-docs.netlify.com` ? or Just `webpack-cli.netlify.com` 🤔

evenstensberg:
Could you make a fork and give us an example first?

anshumanv:
Hey @evenstensberg I already added my fork on netlify to test, here's the latest deploy link

https://app.netlify.com/sites/webpack-cli/deploys/5c8352334fb75f7efd11bacc

All we need to do it remove my fork and add this repo and we're done! 🎉

anshumanv:
> Is this same as this we already have?

Where is this hosted from? Generated docs that live in this repo? 🤔 

We can then opt to push the docs to a separate branch (say gh-pages) on every push so they're the latest imo. Everything remains the same then. What do you think? 😄 

> Instead we can add more content on the site.

Aren't they autogenerated? You mean hardcode?

rishabh3112:
> Where is this hosted from?

from `./docs/` at `master` using github pages.

> We can then opt to push the docs to a separate branch (say gh-pages) on every push so they're the latest imo. Everything remains the same then. What do you think? 😄

master would always be more updated  IMO.

> Aren't they autogenerated? You mean hardcode?

No I mean to add `bin` files too there

anshumanv:
> master would always be more updated IMO.

Yeah, travis will push the updated docs onto gh-pages branch on each commit done on master from where they will be hosted(gh-pages branch). I just find keeping the generated docs with the actual code irrelevant because everytime you generate them you need to edit the changes you mentioned, for the `/bin` we can just use a simple `.md` and keep it with the `INIT`, `SCAFFOLDING` and other md files I believe.

> They are auto-generated and updated at each release by even. But if we need this on each commit we can add this using husky.

Isn't it a lot of manual effort for him? Why need husky when travis can do it in the background (so we don't have to wait for commit everytime, even a simple character removal[or any quick change for instance] will build the docs while committing which won't be nice imo)? 😄 

Also for some reason other routes don't work https://webpack.github.io/webpack-cli/modules/_generate_loader_index_.html

while they work here - https://webpack-cli.netlify.com/modules/_generators_loader_generator_.html

cc @evenstensberg @dhruvdutt

rishabh3112:
> Why need husky when travis can do it in the background (so we don't have to wait for commit everytime, even a simple character removal[or any quick change for instance] will build the docs while committing which won't be nice imo)? 😄

Yeah, if it's possible using Travis then this would be alot better 😎

anshumanv:
Yes exactly, but I'm still in favour of Netlify considering - https://www.netlify.com/github-pages-vs-netlify/ 😎

ematipico:
Sorry if I chime in so suddenly. I think it's common sense to deploy the new docs when we release a new version. If we push at every commit/merge to master, we will publish things that are not published yet (and not yet available)

Surely I'm up to automate the task and to remove the docs from the repository. They shouldn't be tracked.

anshumanv:
Let's find a better solution, later! 😄

PR #785, opened 2019-03-10
fix(tests): update jest

evenstensberg:
Bugfix

Aye ish

N/A
**Summary**
Updates jest to make ci's pass. When updating Babel broke so we updated babel as well. This made the tests and snapshots take color ansi into stdin/stdout, so I've refactored some stuff to make tests assert for contains and not a regex, as it would fail when color codes comes in the way. 

es6 test is using our api for hooking up a config at runtime rather than just bundling a file. With no real test case.

No

**Other information**
:)


evenstensberg:
Also disables ts test in `/info`, @dhruvdutt will need to check it out

packages/utils/npm-packages-exists.test.js
```diff
@@ -15,15 +14,4 @@ describe("npmPackagesExists", () => {
 	test("throws a TypeError when an npm package name doesn't include the prefix", () => {
 		expect(() => npmPackagesExists(["my-webpack-scaffold"])).toThrowError(TypeError);
 	});
-
-	test("resolves packages when they are available on npm", done => {
```
evenstensberg: I've removed this, probably causes a delay and hangs the CI. Useless as the file is fragile when cross-compiling ts/js

package.json
```diff
@@ -122,13 +126,17 @@
...
   "devDependencies": {
+    "@babel/preset-env": "^7.3.4",
+    "@babel/register": "^7.0.0",
     "@commitlint/cli": "^7.2.1",
     "@commitlint/config-lerna-scopes": "^7.2.1",
-    "@commitlint/prompt-cli": "^7.2.1",
+    "@commitlint/prompt-cli": "^7.5.0",
     "@commitlint/travis-cli": "^7.2.1",
-    "@types/jest": "^23.3.9",
+    "@std/esm": "^0.26.0",
```
evenstensberg: devDep, so people can complain all they want

evenstensberg:
Green tests, golden. @webpack/cli-team I'm merging this as a temp fix so we don't get bugged with the automation complaining about the tests failing. Tests are likely to be fixed at the next branch when we're rolling out the new interface so we can discuss at tuesday if it's worth making an effort fixing the remaining tests.

anshumanv:
Neat ✨✅

PR #791, opened 2019-03-16
chore(types): add type information

sendilkumarn:

Adds some type related information and removed where it is obvious

N/A
N/A
**Summary**
leverage on type inference

r? @evenstensberg @ematipico  @dhruvdutt 

**Other information**


packages/generators/add-generator.ts
```diff
@@ -73,7 +72,7 @@ const traverseAndGetProperties = (arr: object[], prop: string): boolean => {
 const searchProps = (answers: object, input: string): Promise<string[]> => {
 	input = input || "";
 	return Promise.resolve(
-		PROPS.filter((prop: string): boolean =>
+		PROPS.filter((prop): boolean =>
```
misterdev: In other occurrences of `.filter` you are keeping the type of the parameter. I would keep here too

packages/generators/add-generator.ts
```diff
@@ -162,7 +161,7 @@ export default class AddGenerator extends Generator {
 							this.configuration.config.item = action;
 						});
 				}
-				const temp: string = action;
+				const temp = action;
```
misterdev: It doesn't look obvious to me. I would keep the type here

packages/init/init.ts
```diff
@@ -42,11 +42,11 @@ export default function runTransform(webpackProperties: IWebpackProperties, acti
 			.keys(webpackProperties)
 			.filter((p: string): boolean => p !== "configFile" && p !== "configPath");
 
-	const initActionNotDefined: boolean = (action && action !== "init") || false;
+	const initActionNotDefined = (action && action !== "init") || false;
 
 	webpackConfig.forEach((scaffoldPiece: string): Promise<void> => {
 		const config: IConfiguration = webpackProperties[scaffoldPiece];
-		const transformations: string[] = mapOptionsToTransform(config);
+		const transformations = mapOptionsToTransform(config);
```
misterdev: It doesn't look obvious to me what's the return type of `mapOptionsToTransform`

packages/generators/add-generator.ts
sendilkumarn: You are correct here. we need to have it here.

packages/init/init.ts
sendilkumarn: My inspiration for this and the above one is from the Typescript core repo. Again here the type is obvious if you know the method signature or the variable and we don't need to be verbose.

But well, I agree it is highly subjective...

ematipico: Actually there's no need to assign a type if the function already has a signature (and just one, where it returns always a string).

packages/generators/add-generator.ts
ematipico: `action` is already a string, so `temp` inherits its type

packages/utils/ast-utils.ts
```diff
@@ -230,7 +237,7 @@ function addOrUpdateConfigObject(
 
 	if (propertyExists) {
 		rootNode.properties
-			.filter((path: INode) => path.key.name === configProperty)
+			.filter((path: INode): boolean => path.key.name === configProperty)
```
ematipico: Is it needed the type here?

evenstensberg:
Needs a quick rebase @sendilkumarn

sendilkumarn:
Done rebasing, let us merge this once the builds are green 🎉

PR #797, opened 2019-03-20
fix(add): add handling of merge option

### LINKED ISSUE #795: [BUG]: @webpack-cli/add not working on topScope and merge
**Describe the bug**
When we want to add something to `topScope` raises following error
Screenshot from 2019-03-18 20-10-56

Currently this error raises as `add-generator` maps properties from `optionsSchema` of webpack. But this schema doesn't include properties that are part of `configuration` object we create in generator.

**To Reproduce**
1. `webpack add`
2. `give answer as topScope`

**Expected behavior**
Add items to topScope

**Screenshots**
uploaded above

```

  System:
    OS: Linux 4.15 Ubuntu 18.04.2 LTS (Bionic Beaver)
    CPU: (3) x64 Intel(R) Core(TM) i5-8250U CPU @ 1.60GHz
  Binaries:
    Node: 10.15.0 - ~/.nvm/versions/node/v10.15.0/bin/node
    Yarn: 1.13.0 - ~/.nvm/versions/node/v10.15.0/bin/yarn
    npm: 6.4.1 - ~/.nvm/versions/node/v10.15.0/bin/npm
  Browsers:
    Firefox: 65.0.1
  npmGlobalPackages:
    webpack-cli: 3.2.3
    webpack: 4.29.3
```
**Additional context**

rishabh3112:
BugFix #795 
Closed #795 

bugfix
N/A
N/A
**Summary**

**Other information**


packages/generators/add-generator.ts
```diff
@@ -173,6 +173,36 @@ export default class AddGenerator extends Generator {
...
+									} else {
+										return "Path doesn't corresponds to a javascript file";
+									}
+								}
+								return "Invalid path provided";
+							};
+							return this.prompt([
+								InputValidate(
+									"mergeFile",
+									"What is the location of webpack configuration file with which you want to merge current configuration?",
```
ematipico: Let's start doing something nice: could you move this long question inside a `constants` file?
Also, the same for the rest of the strings.

Ideally these strings could be exported from `webpack-cli/utils` so they can be reused across the board

packages/generators/add-generator.ts
```diff
@@ -94,7 +93,8 @@ export default class AddGenerator extends Generator {
 		config: {
 			configName?: string,
 			topScope?: string[],
-			item?: string;
+			item?: string,
+			merge?: object,
```
ematipico: `object` is way too generic. Is there a proper type for this `merge` option?

rishabh3112:
Hi @ematipico, I am pushing a sample refactoring ( will break lint as add-geneartor is importing non npm-published changes at `@webpack-cli/utils`.
If this refactoring seems fine to you, I will continue refactoring this and all generators as time permits in seperate PRs. :+1:

packages/generators/add-generator.ts
rishabh3112: Added corresponding type @ematipico :+1:

rishabh3112:
@ematipico Have a review when free 😅.

rishabh3112:
CLA bot is showing that I have not signed 🤔. Some issue here @ematipico

dhruvdutt:
@rishabh3112 It shows signed for me. All checks success.

packages/utils/generators/add/questions/index.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export const topScopeQuestion = Input(
+	"topScope",
+	"What do you want to add to topScope?",
+);
+
+const mergeFileQuestionFunction = () => {
+ const question = "What is the location of webpack configuration with which you want to merge current configuration?";
+ const validator = (path: string) => {
+		const resolvedPath = resolve(process.env.PWD, path);
```
evenstensberg: `process.cwd()`

packages/utils/generators/add/questions/index.ts
```diff
@@ -0,0 +1,63 @@
...
+		if (existsSync(resolvedPath)) {
+			if (/\.js$/.test(path)) {
+				if (typeof require(resolvedPath) !== "object") {
+					return "Given file doesn't export an Object";
+				}
+				return true;
+			} else {
```
evenstensberg: premature if-else

packages/generators/add-generator.ts
```diff
@@ -162,16 +103,24 @@ export default class AddGenerator extends Generator {
...
 								this.configuration.config.topScope.push(topScopeAnswer.topScope);
 								done();
 							});
 						}
+						if (action === "merge") {
+							return this.prompt([mergeFileQuestion])
+							.then((mergeFileAnswer: {
+								mergeFile: string;
+							}) => {
+								const resolvedPath = resolve(process.env.PWD, mergeFileAnswer.mergeFile);
```
evenstensberg: `process.cwd()`

packages/generators/add-generator.ts
```diff
@@ -93,7 +47,8 @@ export default class AddGenerator extends Generator {
 		config: {
 			configName?: string,
 			topScope?: string[],
-			item?: string;
+			item?: string,
+			merge?: IWebpackOptions,
 			webpackOptions?: IWebpackOptions,
```
evenstensberg: merge is in webpackOptions, feel free to make it a config prop but then the ast checking for merge needs to be removed. This is nice if we want to spit out a pure webpack config in webpackOptions

rishabh3112: @evenstensberg I also like to emit a pure config in final configuration built,
But with `merge` prop in `config` (like in current scenario of this PR), final configuration is still having merge prop in `module.exports`. Is that has to be changed too?

evenstensberg: No merge in module exports in my point being. This is ok but ast needs to be checked for code involving this

rishabh3112: The config has `merge` as prop only and is not in `webpackOptions` as you said [here](https://github.com/webpack/webpack-cli/blob/master/packages/utils/modify-config-helper.ts#discussion_r268748106)!
and `runTransform()` also consider `merge` as prop in config prior to my changes too.
https://github.com/webpack/webpack-cli/blob/b96857f182481623db2c1fc799fdab59dab81107/packages/utils/modify-config-helper.ts#L11-L19

So, IMO problem is solved as standalone AST parsing checking of merge is insignificant as it may be using functions outside the scope of object, thus parsing of final file makes more sense.

packages/utils/package.json
```diff
@@ -31,6 +31,7 @@
     "@types/node": "^10.12.9",
     "@types/p-each-series": "^1.0.0",
     "@types/prettier": "^1.15.0",
+    "@webpack-cli/webpack-scaffold": "^0.1.5",
```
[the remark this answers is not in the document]
rishabh3112: Emanuele suggested moving questions to `@webpack-cli/utils`. And questions require `@webpack-cli/webpack-scaffold`, For methods like: `Confirm`, `List`, etc. So, that's why I have put that in dependencies

packages/utils/scaffold.ts
```diff
@@ -48,7 +48,7 @@ export default function runTransform(transformConfig: ITransformConfig, action:
 			transformations.push("topScope");
 		}
 
-		if (config.merge) {
+		if (config.merge && transformations.indexOf("merge") === -1) {
```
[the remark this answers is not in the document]
rishabh3112: Here we don't want to add merge to transformations if it is already there. And, `indexOf("merge")` method returns `-1` when it is not present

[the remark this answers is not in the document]
rishabh3112: `indexOf("merge") >= 0` : Is exactly opposite of what this condition should do. It is checking if the merge is there in transformations. Will result in readdition of `merge` which is not expected
`~!indexOf("merge")` will also not work as anything other than `0` is true. thus wrong condition.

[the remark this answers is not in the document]
evenstensberg: Ok thanks for clearing out your intention. You could do `~!ndexOf("merge")`, which is a bitwise NOT to flip the bits. (Which is true for this case ;) )

packages/utils/package.json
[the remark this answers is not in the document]
evenstensberg: Nice, okay. I don't think user questions should be moved to `utils`, only src that touches functionality. If your functions using inquirer could be used in a general basis, don't put this in utils, but rather as a utils folder within the generator itself. This is to provide a seperation of concern / abstraction

packages/utils/scaffold.ts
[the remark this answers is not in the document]
rishabh3112: For this to work, order should be reversed IMO. i.e.
`!(~indexOf("merge"))`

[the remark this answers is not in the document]
evenstensberg: Yes. Anyway, I think it's fine, it's a nitpick

packages/utils/package.json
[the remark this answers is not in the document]
rishabh3112: Okay I will do that way. Thanks :)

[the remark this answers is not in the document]
evenstensberg: I don't see it being used, so you could remove it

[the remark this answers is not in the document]
rishabh3112: Well currently is in use at `packages/utils/generators/add/questions/index.ts`
nevertheless, I removing all of the related code from `@webpack-cli/utils` and refactoring generator questions under `@webpack-cli/generators`. Hope that will be fine :)

rishabh3112:
Done @evenstensberg.

packages/generators/add-generator.ts
```diff
@@ -151,13 +104,29 @@ export default class AddGenerator extends Generator {
...
 									done();
 								}
 							);
 						}
+						if (action === "merge") {
+							return this.prompt([mergeFileQuestion])
+							.then((mergeFileAnswer: {
+								mergeFile: string;
+							}) => {
+								const resolvedPath = resolve(process.cwd(), mergeFileAnswer.mergeFile);
```
evenstensberg: You'll need to parse this yourself. Extract this logic into a function `parseMergeConfig`, where you read the config, the `webpackOptions` is the module export of the config itself, you already have the filename from `basename(resolvedPath)` . I don't think that merge actually works like this. Merge takes a config name and attaches the name to a function. 

What you will need to do, is to make merge take in either `<Array>[name, path]` and add a require on top if the require is not in topScope (which I think is the best solution) or just add the name (bad DX).

So what you'll need to do, is to have this in the object of the generator:

```js
{
merge: ['webpackBaseConfig', 'path/to/foo'],
config {...}
}
```

and make the function parseMerge ( https://github.com/webpack/webpack-cli/blob/master/packages/utils/ast-utils.ts#L608-L640 ) accept an array that will convert the first idx to the function parameter, the second one to a topScope property. You could use the existing API to make this happen more easily, and that is to make a require statement based on the indices and add that to topScope before running the transform. This will leave you with the same API as it is, but providing support for `merge` in add.

packages/generators/utils/add/index.ts
```diff
@@ -0,0 +1,64 @@
+import PROP_TYPES from "@webpack-cli/utils/prop-types";
```
evenstensberg: Other improvements is nice, cause CLI in add throws if dev-server is not a dependency, so you can keep the rest of the changes

packages/generators/add-generator.ts
```diff
@@ -85,6 +42,12 @@ export default class AddGenerator extends Generator {
 			configName?: string;
 			topScope?: string[];
 			item?: string;
+			merge?: {
```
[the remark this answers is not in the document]
rishabh3112: Sure!
I had a problem with this one 😅, as the merge strictly can't have merge property, it will act as a bad interface. Can you suggest a workaround?

evenstensberg:
Made a repro for you @rishabh3112 to help you get where I'm going with this:

https://gist.github.com/evenstensberg/87978d3e21bcc325f5fcef67f372fd68

packages/generators/add-generator.ts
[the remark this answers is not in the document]
evenstensberg: I think maybe we should discuss how we should handle merge (see the comment on add generator) and then resume this if necessary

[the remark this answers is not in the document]
rishabh3112: Never mind, I just read the next comment, not a problem now :)

packages/generators/add-generator.ts
rishabh3112: Nice solution.
Or we can just take up the path to the config file from user. Internally, add `webpack-merge` as dependency for the user and use it to scaffold a config that will use `webpack-merge`.
WDYT?

packages/generators/utils/add/index.ts
rishabh3112: I didn't get what you want to say here?

packages/generators/add-generator.ts
rishabh3112: Done @evenstensberg

packages/generators/add-generator.ts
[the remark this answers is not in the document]
rishabh3112: Solved now, Thanks!

packages/generators/utils/add/index.ts
evenstensberg: Things that has been changed in logic not related to the transformation w/AST is fine

evenstensberg: Things that has been changed in logic not related to the transformation w/AST is fine

packages/utils/scaffold.ts
```diff
@@ -11,6 +11,29 @@ import astTransform from "./recursive-parser";
 import runPrettier from "./run-prettier";
 import { Node } from "./types/NodePath";
 
+
+
+function mergeHandler(config: Config, transformations: string[]): [Config, string[]]{
+	if(!config["topScope"])
```
evenstensberg: Assuming you already declared the array in the constructor, you won't need to check if it's existent, you only need to push to the array

packages/utils/scaffold.ts
```diff
@@ -45,16 +68,16 @@ export default function runTransform(transformConfig: TransformConfig, action: s
...
+			let transformations = mapOptionsToTransform(config);
 
 			if (config.topScope && transformations.indexOf("topScope") === -1) {
 				transformations.push("topScope");
 			}
 
-			if (config.merge) {
-				transformations.push("merge");
+			if (config.merge && transformations.indexOf("merge") === -1) {
+				[config, transformations] = mergeHandler(config, transformations);
```
evenstensberg: This is logic relevant to the generator providing the code, not the ast transform, this would make the abstraction bad. The generator is responsible to provide:

1. an array of indices with relevant data
2. The topScope property with the path

In the `handleMerge` function, you'll need to check if the part of the require is present in the most outerscope and append the second index if not present. 

Sorry for giving you an indication of what the easiest choice would be, it was a stepping stone towards a final solution

packages/utils/scaffold.ts
```diff
@@ -45,16 +68,16 @@ export default function runTransform(transformConfig: TransformConfig, action: s
 
 	webpackConfig.forEach(
 		(scaffoldPiece: string): Promise<void> => {
-			const config: Config = transformConfig[scaffoldPiece];
+			let config: Config = transformConfig[scaffoldPiece];
```
evenstensberg: Config should be immutable/const here as we don't want to modify a presupplied config

packages/utils/scaffold.ts
```diff
@@ -11,6 +11,29 @@ import astTransform from "./recursive-parser";
 import runPrettier from "./run-prettier";
 import { Node } from "./types/NodePath";
 
+
+
+function mergeHandler(config: Config, transformations: string[]): [Config, string[]]{
```
evenstensberg: I'm not a fan of this function as it isn't descriptive enough and it passes the transformation object in an indrection we don't want the AST to use. The transformation and it's data should have a 1:1 mapping

packages/utils/modify-config-helper.ts
```diff
@@ -14,7 +14,7 @@ export interface Config extends Object {
 	};
 	topScope?: string[];
 	configName?: string;
-	merge: object;
+	merge: string|string[];
```
evenstensberg: nice

packages/generators/utils/add/index.ts
evenstensberg: It's all good is what I'm saying. We need to look at the merge logic first in the ast part and then we can look at this, but it looks ok

packages/generators/add-generator.ts
evenstensberg: Nice. Have a look at the comment further down

packages/utils/scaffold.ts
evenstensberg: To give you a hint: We're pushing the transformation, which means that it's probably the latest transform to be run, so what you'll need to do, is this logic, but directly as an ast transform. The data gets passed down by the each astTransform func

rishabh3112: Thanks for review! As far as I could understand now, I need to update `parseMerge` function directly in `ast-utills` 
to just use the 1st index of the array 
And push to the topScope here itself. Right?

packages/utils/scaffold.ts
rishabh3112: Added to avoid errors if runTransform function is directly used, like I do in webpack UI

packages/utils/scaffold.ts
rishabh3112: I was updating config later using `mergeHandler`, that's why this change. See the comment above for more info on my view on this now.

packages/utils/scaffold.ts
rishabh3112: Remove this function

packages/utils/scaffold.ts
rishabh3112: Updated

packages/utils/scaffold.ts
rishabh3112: Will found some other way around for UI 👍 . Removed from here.

rishabh3112:
Done @evenstensberg

packages/generators/add-generator.ts
```diff
@@ -151,13 +99,25 @@ export default class AddGenerator extends Generator {
...
 						}
+						if (action === "merge") {
+							return this.prompt(mergeFileQuestion)
+							.then((mergeFileAnswer: {
+								mergeFile: string,
+								mergeConfigName: string
+							}) => {
+								const resolvedPath = resolve(process.cwd(), mergeFileAnswer.mergeFile);
+								// eslint-disable-next-line
+								this.configuration.config["merge"] = [mergeFileAnswer.mergeConfigName, resolvedPath];
```
evenstensberg: `action` here instead of hard-coded value

packages/utils/recursive-parser.ts
```diff
@@ -15,7 +15,7 @@ export default function recursiveTransform(
 		console.error("Error in parsing top scope, Array required");
 		return false;
 	} else if (key === "merge") {
-		if (typeof value === "string") {
+		if (typeof value[0] === "string" && Array.isArray(value)) {
```
evenstensberg: both indices are strings here, start by checking array, then that both indices are strings

packages/utils/scaffold.ts
```diff
@@ -53,10 +54,17 @@ export default function runTransform(transformConfig: TransformConfig, action: s
 				transformations.push("topScope");
 			}
 
-			if (config.merge) {
+			if (config.merge && transformations.indexOf("merge") === -1) {
```
evenstensberg: This is a double up if, you can nest into a single one

packages/utils/scaffold.ts
```diff
@@ -53,10 +54,17 @@ export default function runTransform(transformConfig: TransformConfig, action: s
...
 				transformations.push("merge");
 			}
 
+			if (config.merge) {
```
evenstensberg: nest this ^

packages/utils/ast-utils.ts
```diff
@@ -605,7 +605,7 @@ function parseTopScope(j: JSCodeshift, ast: Node, value: string[], action: strin
  */
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-function parseMerge(j: JSCodeshift, ast: Node, value: string, action: string): boolean | Node {
+function parseMerge(j: JSCodeshift, ast: Node, value: string[], action: string): boolean | Node {
 	function createMergeProperty(p: Node): boolean {
 		// FIXME Use j.callExp()
```
evenstensberg: Split into two variables, give them meaningful names. First one is supposed to go here, and then you'll need to check if the require of this is present in the outer body of the AST. If it's present, you don't need to do anything, if it's not existent you will need to add it. 

Remove the topScope push and manually add that here. You could reuse the logic used in topScope to a certain extent. Remember to abstract this code so it looks neat

packages/utils/ast-utils.ts
```diff
@@ -626,14 +626,35 @@ function parseMerge(j: JSCodeshift, ast: Node, value: string, action: string): b
...
 		};
 		(p.value as Node).body[bodyLength - 1] = newVal;
 		return false; // TODO: debug later
 	}
+
+	function addMergeImports(ast: Node, mergeImports: string[]) {
+		ast.find(j.Program).filter((program: Node): boolean => {
+			mergeImports.forEach(
+				(imp: string): void => {
+					if ((program.value as Node).body.indexOf(imp) === -1){
```
rishabh3112: Currently checking whether in the current iteration these two imports are added or not. Similar to `parseTopScope` which also not check this.

I want to make a better checking if a particular package is already imported or not by iterating over all `CallExpression`s and checking values of `argument` where `callee` is `require`. This same function could also be used for `parseTopScope` for better checking when imports are pushed to the topScope.
Should I do that in this PR itself on open a new one for this? asking as scope of that change is larger that that of this PR.

rishabh3112:
Rebased, look at the above review.

packages/utils/ast-utils.ts
evenstensberg: Take it in this PR, but could you rephrase your naming conventions such that it makes sense

packages/utils/scaffold.ts
evenstensberg: aye

packages/utils/ast-utils.ts
rishabh3112: Like?

evenstensberg: What is an `imp` ? what does it do?

packages/utils/ast-utils.ts
```diff
@@ -626,14 +626,35 @@ function parseMerge(j: JSCodeshift, ast: Node, value: string, action: string): b
...
+	}
+
 	if (value) {
-		return ast.find(j.Program).filter((p: Node): boolean => createMergeProperty(p));
+		const [configIdentifier, configPath] = value;
+		const mergeImports = [
+			`const ${configIdentifier} = require('${configPath}')`,
+			`const merge = require('webpack-merge')`
+		];
+		ast = addMergeImports(ast, mergeImports);
```
evenstensberg: Use the jscodeshift api for this together with `replace`

packages/utils/ast-utils.ts
rishabh3112: Short for `import`, import is a reserved keyword and can't use as a variable name.

It is one of the imports in `mergeImports` array.

evenstensberg: exactly, just getting you on the right track here from a different perspective

rishabh3112: Okay 😂, any other naming you felt bad ?

rishabh3112:
Done @evenstensberg

rishabh3112:
@evenstensberg can I know if anything left to be done on this? Also, I suggest you merge this first before 
#843 as that way I would be able to fix conflicts for obvious file `add-generator.ts` at #843.

evenstensberg:
Going through it now

packages/utils/ast-utils.ts
```diff
@@ -626,14 +637,27 @@ function parseMerge(j: JSCodeshift, ast: Node, value: string, action: string): b
...
 		(p.value as Node).body[bodyLength - 1] = newVal;
 		return false; // TODO: debug later
 	}
+
+	function addMergeImports(configIdentifier: string, configPath: string) {
+		ast.find(j.Program).forEach(p => {
+			if (!isImportPresent(j, ast, 'webpack-merge')) {
+				(p.value as Node).body.splice(-1, 0, `const merge = require('webpack-merge')`);
+			}
+			(p.value as Node).body.splice(-1, 0, `const ${configIdentifier} = require('${configPath}')`);
```
evenstensberg: you need to check if this import is present too, look for the configIdentifier in the body. So a new `if` here: `if importPresent(configIdentifier)`

packages/utils/ast-utils.ts
```diff
@@ -626,14 +637,27 @@ function parseMerge(j: JSCodeshift, ast: Node, value: string, action: string): b
...
+		})
+	}
+
 	if (value) {
-		return ast.find(j.Program).filter((p: Node): boolean => createMergeProperty(p));
+		const [configIdentifier, configPath] = value;
```
evenstensberg: Gotta check these types if they're present

packages/utils/recursive-parser.ts
```diff
@@ -15,7 +15,7 @@ export default function recursiveTransform(
 		console.error("Error in parsing top scope, Array required");
 		return false;
 	} else if (key === "merge") {
-		if (typeof value === "string") {
+		if (Array.isArray(value) && typeof value[0] === "string" && typeof value[1] === "string") {
```
evenstensberg: We can skip these comparrisons until later on in the import functions you've added. Checking for array is sufficient here, comparing the destructed values later

packages/utils/ast-utils.ts
rishabh3112: Throwing errors now in case of type mismatch.

evenstensberg: awesome sauce

rishabh3112:
Done @evenstensberg

evenstensberg:
Mind rebasing @rishabh3112 ? I'll have a look, just got caught up in other issues

evenstensberg:
Hmu once that's done and I'l do a review to get this merged

rishabh3112:
Done @evenstensberg

packages/generators/add-generator.ts
```diff
@@ -1,57 +1,25 @@
...
- *
- * @returns {Promise} Returns promise which resolves to filtered props
- *
- */
-const searchProps = (answers: object, input: string): Promise<string[]> => {
-	input = input || "";
-	return Promise.resolve(PROPS.filter((prop: string): boolean => prop.toLowerCase().includes(input.toLowerCase())));
-};
+import entryQuestions from "./utils/entry";
+import * as AutoComplete from "inquirer-autocomplete-prompt";
```
evenstensberg: Move package installed deps as the first imports

packages/generators/add-generator.ts
```diff
@@ -136,13 +94,23 @@ export default class AddGenerator extends Generator {
...
 									this.configuration.config.topScope.push(topScopeAnswer.topScope);
 									done();
 								}
 							);
 						}
+						if (action === "merge") {
+							return this.prompt(mergeFileQuestion).then(
+								(mergeFileAnswer: { mergeFile: string; mergeConfigName: string }): void => {
+									const resolvedPath = resolve(process.cwd(), mergeFileAnswer.mergeFile);
+									// eslint-disable-next-line
```
evenstensberg: Don't disable this lint rule

packages/generators/add-generator.ts
```diff
@@ -1,57 +1,25 @@
...
-import * as path from "path";
+import {
+	actionTypeQuestion,
+	entryTypeQuestion,
+	manualOrListInput,
+	mergeFileQuestion,
+	topScopeQuestion
+} from "./utils/add/questions";
+
+import { traverseAndGetProperties, webpackDevServerSchema, webpackSchema } from "./utils/add";
```
evenstensberg: You don't need to abstract the schemas

packages/generators/utils/add/questions/index.ts
```diff
@@ -0,0 +1,63 @@
+import {
```
evenstensberg: rename to questions.js to avoid parent index and this index.js being confusing

packages/generators/utils/add/questions/index.ts
```diff
@@ -0,0 +1,63 @@
...
+
+const mergeFileQuestionsFunction = (): Question[] => {
+ const mergePathQuestion = "What is the location of webpack configuration with which you want to merge current configuration?";
+ const mergePathValidator = (path: string): boolean|string => {
+		const resolvedPath = resolve(process.cwd(), path);
+		if (existsSync(resolvedPath)) {
+			if (/\.js$/.test(path)) {
+				return true;
+			}
+			return "Path doesn't corresponds to a javascript file";
```
evenstensberg: doesn't correspond

packages/generators/utils/add/questions/index.ts
```diff
@@ -0,0 +1,63 @@
+import {
+	AutoComplete,
```
evenstensberg: Could you rephrase your questions to the ones that are on master right now?

packages/utils/ast-utils.ts
```diff
@@ -1,6 +1,24 @@
 import { JSCodeshift, Node, valueType } from "./types/NodePath";
 import * as validateIdentifier from "./validate-identifier";
 
+function isImportPresent(j: JSCodeshift, ast: Node, path: string): boolean {
+	if (typeof path !== "string") {
+		throw new Error(`path parameter should be string, recieved ${typeof path}`);
+	}
+	let isPresent = false;
```
evenstensberg: `mergeImportExists`

packages/utils/ast-utils.ts
```diff
@@ -626,14 +644,37 @@ function parseMerge(j: JSCodeshift, ast: Node, value: string, action: string): b
...
 		};
+
 		(p.value as Node).body[bodyLength - 1] = newVal;
 		return false; // TODO: debug later
 	}
+
+	function addMergeImports(configIdentifier: string, configPath: string): void {
+		if (typeof configIdentifier !== "string" || typeof configPath !== "string") {
+			throw new Error(
+				`Both parameters should be string. recieved ${typeof configIdentifier}, ${typeof configPath}`
```
evenstensberg: strings

packages/utils/ast-utils.ts
rishabh3112: It is a general utility function, can be used for `topScope` too while adding new packages. Are you okay with "importExists"?

packages/generators/utils/add/questions/index.ts
rishabh3112: Questions are the same as in the master

rishabh3112:
Have applied suggestions @evenstensberg and left some comments for you :)

packages/generators/add-generator.ts
```diff
@@ -1,57 +1,26 @@
...
-import * as autoComplete from "inquirer-autocomplete-prompt";
-import * as path from "path";
-
 import npmExists from "@webpack-cli/utils/npm-exists";
 import { getPackageManager } from "@webpack-cli/utils/package-manager";
-import PROP_TYPES from "@webpack-cli/utils/prop-types";
-import { AutoComplete, Confirm, Input, List } from "@webpack-cli/webpack-scaffold";
+import { Input, List } from "@webpack-cli/webpack-scaffold";
+// eslint-disable-next-line
+export const webpackDevServerSchema = require("webpack-dev-server/lib/options.json");
```
evenstensberg: with ts you can import without require

packages/generators/add-generator.ts
```diff
@@ -1,57 +1,26 @@
...
- *
- */
-const searchProps = (answers: object, input: string): Promise<string[]> => {
-	input = input || "";
-	return Promise.resolve(PROPS.filter((prop: string): boolean => prop.toLowerCase().includes(input.toLowerCase())));
-};
+import entryQuestions from "./utils/entry";
+import { generatePluginName } from "./utils/plugins";
+// eslint-disable-next-line
+export const webpackSchema = require("../optionsSchema.json");
```
evenstensberg: same here

packages/generators/add-generator.ts
rishabh3112: Done @evenstensberg

packages/generators/add-generator.ts
```diff
@@ -1,57 +1,24 @@
...
-
 import npmExists from "@webpack-cli/utils/npm-exists";
 import { getPackageManager } from "@webpack-cli/utils/package-manager";
-import PROP_TYPES from "@webpack-cli/utils/prop-types";
-import { AutoComplete, Confirm, Input, List } from "@webpack-cli/webpack-scaffold";
+import { Input, List } from "@webpack-cli/webpack-scaffold";
+import webpackDevServerSchema from "webpack-dev-server/lib/options.json";
+import * as AutoComplete from "inquirer-autocomplete-prompt";
+import path, { resolve } from "path";
+import glob from "glob-all";
```
evenstensberg: glob and path on top of import, then the autocomplete. Add a space after that and the cli imports after that

```
const fs ...
const path ....
const autocomplete

const cli..
const cli2..
```

packages/generators/utils/add/questions.ts
```diff
@@ -0,0 +1,46 @@
+import { AutoComplete, Confirm, Input, InputValidate } from "@webpack-cli/webpack-scaffold";
+import { existsSync } from "fs";
```
evenstensberg: same import order as mentioned in prev file

rishabh3112:
Have moved local imports (example `../utils`)  at last.

packages/generators/add-generator.ts
```diff
@@ -1,57 +1,25 @@
...
+import * as AutoComplete from "inquirer-autocomplete-prompt";
 import * as Generator from "yeoman-generator";
-
-import * as glob from "glob-all";
-import * as autoComplete from "inquirer-autocomplete-prompt";
-import * as path from "path";
+import webpackDevServerSchema from "webpack-dev-server/lib/options.json";
```
evenstensberg: group this with the ones below

packages/generators/utils/add/questions.ts
```diff
@@ -0,0 +1,48 @@
...
+import { resolve } from "path";
+import { Question } from "inquirer";
+
+import { AutoComplete, Confirm, Input, InputValidate } from "@webpack-cli/webpack-scaffold";
+
```
evenstensberg: - whitespace, group these

rishabh3112:
No problem, Thanks🎉🎉

Issue #709, opened 2018-12-08
[Feature Request]: Let's rework some of the main scaffolding layout


TheLarkInn:
**Describe the solution you'd like**
Let's rework how development and production configs are laid out.

- [x] Always no matter what, config is generated in root of project
- [ ] Config is generated _IF_ defaults are not used
- [x] Config is always named `webpack.config.js` 
- [ ] Development and Production configs are composed or combined using `webpack-merge` not Object.assign
- [ ] Development and Production configs use names webpack.development.js and webpack.production.js (even if shorter is more convenient it matches with mode names).
- [ ] Scripts generated should use `--env.mode development` `--env.mode production` to determine whether or not production or development modes are used. 
- [ ] any config used is a function that returns a config.
- [ ] any config that is needed for both production and development goes into the 	`webpack.config.js` file, or the common config.
- [ ] For multipage application we do not use webpack-dev-server in dev, and instead use --watch in scripts.
- [ ] For SPA (or single entry) we consider using webpack-dev-server as default instead of --watch
- [ ] Dev script is called `dev` and production script is called `prod`.
- [ ] A script called `webpack` is generated so users can run `npm run webpack`. This will be used inside of `dev` and `prod` scripts for composition sakes.

![screen shot 2018-12-08 at 2 44 01 pm](https://user-images.githubusercontent.com/3408176/49691465-c09a3d80-faf7-11e8-8931-2482fa70609e.png)

Hopefully all of this is clear and I added them in checkbox format. These ideas I've created based on positive feedback I've received from my workshops and talks. 

Since all this code is JavaScript i would understand that there are many opinions on how configs are formatted but I would like to go this direction because I believe it is just easier to maintain this way. 

If you look at my repo: https://github.com/TheLarkInn/webpack-workshop-2018/tree/feature/19-sm-css-code-split 

This shows you a fully fleshed out example with many features, integrations, etc and hopefully we can follow this style (even presets which I can document in a separate issue).

EvsChen:
Are we going to track these feats in this issue? I would like to handle some of them.

evenstensberg:
Probably. If there's no PR's open, take one down @EvsChen , thanks!

EvsChen:
I guess I'll handle the config naming first.

evenstensberg:
I've already submitted a PR for that. Other than that, up for grabs

anshumanv:
Working on 1 - Generate webpack config in the root of the project.

😄

[...]

PR #801, opened 2019-03-22
fix(scaffold): config file is always generated at the project root

anshumanv:
config file is generated at the project root

bugfix

yes

NA

**Summary**
Determine project root and create the file there.

No

**Other information**
Ref - #709

### DIFF
--- a/packages/utils/find-root.ts
+++ b/packages/utils/find-root.ts
@@ -1,0 +1,8 @@
+import * as findup from "findup-sync";
+import * as path from "path";
+
+export function findProjectRoot(curDir: string): string {
+  const rootFilePath = findup(`package.json`);
+  const projectRoot = path.dirname(rootFilePath);
+  return projectRoot;
+}
--- a/packages/utils/find-root.ts
+++ b/packages/utils/find-root.ts
@@ -1,7 +1,7 @@
 import * as findup from "findup-sync";
 import * as path from "path";
 
-export function findProjectRoot(curDir: string): string {
+export function findProjectRoot(): string {
   const rootFilePath = findup(`package.json`);
   const projectRoot = path.dirname(rootFilePath);
   return projectRoot;
--- a/packages/utils/scaffold.ts
+++ b/packages/utils/scaffold.ts
@@ -76,7 +76,7 @@ export default function runTransform(transformConfig: ITransformConfig, action: 
 					configurationName = "webpack." + config.configName + ".js";
 				}
 
-				const projectRoot = findProjectRoot(process.cwd());
+				const projectRoot = findProjectRoot();
 				const outputPath: string = initActionNotDefined
 				? transformConfig.configPath
 				: path.join(projectRoot || process.cwd(), configurationName);
--- a/packages/utils/scaffold.ts
+++ b/packages/utils/scaffold.ts
@@ -2,7 +2,7 @@ import chalk from "chalk";
 import * as j from "jscodeshift";
 import pEachSeries = require("p-each-series");
 import * as path from "path";
-import pkgDir from "pkg-dir";
+import { findProjectRoot } from "./find-root";
 
 import { IError } from "../init/types";
 import { IConfig, ITransformConfig } from "./modify-config-helper";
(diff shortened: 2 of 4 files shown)


anshumanv:
Looking into CI

packages/utils/scaffold.ts
```diff
@@ -74,14 +76,16 @@ export default function runTransform(transformConfig: ITransformConfig, action:
 					configurationName = "webpack." + config.configName + ".js";
 				}
 
-				const outputPath: string = initActionNotDefined
-					? transformConfig.configPath
-					: path.join(process.cwd(), configurationName);
-				const source: string = ast.toSource({
-					quote: "single",
+				pkgDir(process.cwd()).then((projectRoot) => {
```
rishabh3112: Using `process.env.PWD` instead of `process.cwd()` in deleted code will also fix the issue IMO.

rishabh3112:
Also, commit `package-lock.json` when you install or update packages :)

packages/utils/scaffold.ts
anshumanv: You mean fix the CI or the issue? 🤔

anshumanv:
> Also, commit `package-lock.json` when you install or update packages :)

Sure thing, on it. 👍

packages/utils/scaffold.ts
rishabh3112: issue

anshumanv: Ah I just tested and it prints the current dir for me and not the project root dir🤔

rishabh3112: What is your current dir? This `process.env.PWD` remains constant throughout a process with where it started. And I am assuming that your running `webpack-cli init` or any other of scaffolding packages in project root directory.

anshumanv: for instance - 

Project root is `test`

cur dir is `test/a/b`

I run `webpack-cli init`

`webpack.config.js` is generated at the root rather than `/a/b` path

cc @evenstensberg is this what we want?

anshumanv: Makes sense, will make changes 👍

anshumanv:
@ematipico this approach looks amazing ✨ 

Thanks for suggesting this, looks way more maintainable and cleaner. 👍

packages/utils/find-root.ts
```diff
@@ -0,0 +1,8 @@
+import * as findup from "findup-sync";
```
ematipico: We are almost there! So, the `findup-sync` package is in the root project but is not installed in the utils project.

So you have two alternatives:
- (the quick one) is to add the `findup-sync` package also here in the utils project
- (the better one) remove the `findup-sync` package from the root project, add it in the utils project and **expose a utility function to fix the usage of package in the root project**.

Choose the solution that you prefer. I'd prefer the second but it can tackled in a second PR

evenstensberg: Could we go for no.2?

anshumanv: Yes, added the package in utils for now, doing it in a new PR asap.

@evenstensberg @ematipico

packages/utils/find-root.ts
```diff
@@ -0,0 +1,8 @@
...
+import * as path from "path";
+
+export function findProjectRoot(curDir: string): string {
```
ematipico: `curDir` is not used, you can remove it

packages/utils/scaffold.ts
```diff
@@ -74,14 +76,15 @@ export default function runTransform(transformConfig: ITransformConfig, action:
 					configurationName = "webpack." + config.configName + ".js";
 				}
 
+				const projectRoot = findProjectRoot(process.cwd());
```
ematipico: No need to pass `process.cwd()`

packages/utils/find-root.ts
anshumanv: Fixed, I thought in future I can make this extendible, but removing it for now.

packages/utils/scaffold.ts
anshumanv: Removed. 👍

anshumanv:
@ematipico fixed the issues, can we go ahead with this once builds are green? Will start working on the follow-up PR soon. ✨

dhruvdutt:
Looks good to go. Nice work. 💯

packages/utils/find-root.ts
anshumanv: Hi @ematipico 

I created util in `packages/utils` but then what's the workflow to import it in the root project?

anshumanv: One way could be to import it from `@webpack-cli/utils/fileName` but for it we'll need to publish it right?

Issue #800, opened 2019-03-22
Enhancement: Move the file to utils and other minor refactors


anshumanv:
**Do you want to request a *feature* or report a *bug*?**
refactor: move opencollective to utils and use constants from file.

**What is the current behavior?**
It exists with the `cli.js` file and has local constants.

**If the current behavior is a bug, please provide the steps to reproduce.**
NA

**What is the expected behavior?**
Since it provides utility functions, it should be moved to utils.

**If this is a feature request, what is motivation or use case for changing the behavior?**
refactor

**Please paste the results of `webpack-cli info` here, and mention other relevant information such as programming language.**

```
webpack-cli info

  System:
    OS: macOS 10.14.3
    CPU: (4) x64 Intel(R) Core(TM) i5-5250U CPU @ 1.60GHz
  Binaries:
    Node: 10.14.1 - /usr/local/bin/node
    Yarn: 1.12.3 - /usr/local/bin/yarn
    npm: 6.4.1 - /usr/local/bin/npm
  Browsers:
    Chrome: 72.0.3626.119
    Safari: 12.0.3
  npmPackages:
    webpack: 4.x.x => 4.29.6
    webpack-dev-server: ^3.1.10 => 3.2.1
```

misterdev:
Hi @anshumanv, seems a good idea to move it into utils. 
I think I would keep the local constants where they are since they have a name just to make the code cleaner and they are not meant to be used anywhere else.

If you still think it could be better to refactor the constants let's go ahead 👍

anshumanv:
> I think I would keep the local constants where they are since they have a name just to make the code cleaner and they are not meant to be used anywhere else.

Makes sense, I'll open up a PR to move the file initially and move constants as per feedback. 😄

ematipico:
I'm not convinced. That file is not an utility as it's in one place for a specific thing. If it was an utility you'd think there is another usage somewhere in the code but there isn't.

anshumanv:
@ematipico can you suggest something else then?

ematipico:
Nothing to change really, I'm happy with the code like this. Unless you want try to convince me :)

evenstensberg:
I think current approach is good enough for us. Thanks for reporting & providing a fix @anshumanv <3

PR #804, opened 2019-03-26
chore(opencollective): move opencollective to utils

anshumanv:
move opencollective to utils

ISSUES CLOSED: #800

minor refactor

yes

NA

**Summary**
Simply fixed the file's path

No

**Other information**
I need suggestions whether to move the constants used in opencollective function to be moved to `constants.js` since they're just used once so does it make sense to move them as well?


anshumanv:
What do you think of the current approach? Can it be improved? How about making a closure and keeping a variable whether it is to be shown?

fokusferit:
@anshumanv what problem was exactly solved? I do get that refactoring. Was there anything reported? Just curious.

anshumanv:
Yeah @fokusferit there's an issue here where we had some discussion - https://github.com/webpack/webpack-cli/issues/800

anshumanv:
Sure thing, can you suggest an approach to refactor or any other change?

evenstensberg:
See issue #800 for explaination. Altough you are the best 💙

Issue #805, opened 2019-03-27
Enhancement: remove findup-sync from root package and move to utils


anshumanv:
**Do you want to request a *feature* or report a *bug*?**
refactor

**What is the current behavior?**
findup-sync is present in project root, it should be moved to utils package.
**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**
findup-sync is present in both project root and utils package.
**If this is a feature request, what is motivation or use case for changing the behavior?**
https://github.com/webpack/webpack-cli/pull/801#discussion_r268735064

Sending a quick fix.

PR #806, opened 2019-03-27
chore(cli): remove findup-sync from root move to utils package && format ts files

anshumanv:
remove findup-sync from package dir and move to utils

ISSUES CLOSED: #805

refactor mostly

yes

NA

**Summary**
removed findup-sync package from root, ported to utils

Yes

**Other information**


bin/utils/convert-argv.js
```diff
@@ -6,7 +6,7 @@ const prepareOptions = require("./prepareOptions");
 const webpackConfigurationSchema = require("../config/webpackConfigurationSchema.json");
 const validateSchema = require("webpack").validateSchema;
 const WebpackOptionsValidationError = require("webpack").WebpackOptionsValidationError;
-const findup = require("findup-sync");
+const { webpackConfigPath } = require("../../packages/utils/webpack-config-path");
```
anshumanv: This should be imported from `@webpack-cli/utils/webpack-config-path` but won't it need for the package to be published? Tell me when I should change it.

anshumanv: @ematipico @dhruvdutt is this the right direction? 🤔

ematipico: The utils folder is a totally different project (it has its own package.json). Because of this it must be treated as a third party library. So you have to import your utility like this:

```js
const { myUtility } = require('@webpack-cli/utils');
```

So you have to make sure that utils project exports correctly the function

packages/utils/webpack-config-path.ts
```diff
@@ -0,0 +1,7 @@
+import * as findup from "findup-sync";
```
ematipico: Is there a reason why you created a specific file just for this function?

anshumanv: Not really, what do you suggest where we put this function?

anshumanv: I think it makes sense to rename the file `find-root` to `find-path` and we can merge this file into it and export all the path related function from there? 🤔

evenstensberg: We can put it in `is-local-path.ts` and rename it to `path-utils.ts`

anshumanv: Alright, on it. 👍

anshumanv:
Can we possibly disable Travis comments as everyone can just follow the stack trace on Travis page? Comments seem to create a lot of noise. 😅

ematipico:
Sometimes people don't know how get to Travis logs and also from mobile it's not easy to read it. Here it helps and you also get a notification.

anshumanv:
@ematipico makes sense, can you help me figure out why tests are failing? Probably I'm missing something simple. 🤔

bin/utils/convert-argv.js
```diff
@@ -6,7 +6,7 @@ const prepareOptions = require("./prepareOptions");
 const webpackConfigurationSchema = require("../config/webpackConfigurationSchema.json");
 const validateSchema = require("webpack").validateSchema;
 const WebpackOptionsValidationError = require("webpack").WebpackOptionsValidationError;
-const { webpackConfigPath } = require("@webpack-cli/utils/path-utils");
+const { webpackConfigPath } = require("../../packages/utils/path-utils");
```
evenstensberg: what should we do about this? 🤔

anshumanv: Precisely, wondering :(

ematipico: This is the correct approach 

```js
const { webpackConfigPath } = require("@webpack-cli/utils/path-utils")
```

Does this create some problems?

anshumanv: Yes, this is what I wrote when a large number of tests failed right before the last commit. @ematipico

ematipico: Before pushing the code, did you test it locally? Debug it locally might give more insights.

Probably when the code gets compiled by typescript, the function doesn't get exported correctly

ematipico: Look!! `@webpack-cli/utils` has to become a dependency of the main package :) That's the missing piece of the puzzle. We can't use a package if we don't add it to the `package.json` and we don't install it

anshumanv: Hey @ematipico thanks for your findings, I did `npm i @webpack-cli/utils` in the project root to add it to deps, but I'm still seeing the same errors, wonder what we're missing 🤔

anshumanv: Ah my bad, need to link the local version, on it

anshumanv:
@ematipico still no help, it installs the npm version and not the local version, can you suggest something?

misterdev:
Have you tried `npm link` inside utils and then `npm link @webpack-cli/utils` where you want to use it?
It should also be added as a dependency :)

anshumanv:
Yeah works that way, locally. How do we make it work here?

bin/utils/convert-argv.js
ematipico: In theory you don't need to do link but you have to run Lerna bootstrap. That's the correct way

ematipico:
You have to run Lerna bootstrap in order to link correctly the packages

anshumanv:
So the key is about listing `@webpack-cli/utils` as a local depency which installs from file, but that would mean every inter package dependency needs to be converted to local dep which will be installed from the local package and not from npm, should we go for it @ematipico

ematipico:
The real issue is that during the CI we don't call the command `npm run bootstrap` to link the packages before running the tests. We don't need to any sort of link.

anshumanv:
If I understand correctly it seems we are @ematipico  - https://github.com/webpack/webpack-cli/blob/b96857f182481623db2c1fc799fdab59dab81107/azure-pipelines-template.yml#L24

🤔

ematipico:
That's weird 😮

anshumanv:
There seems to be some issue with linking, It should work once we publish the utils package though with the latest code.

anshumanv:
@evenstensberg @ematipico after searching for days I couldn't find a solution that links local deps to be installed after symlink, you either need to use a package path but then it isn't suggested by lerna.

I've introduced the required functions in `path-utils` which can be used in the project root after the next release (since then they'll be exposed by utils package), can we go ahead with this meanwhile if you don't have any input?

bin/utils/convert-argv.js
anshumanv: No help 🆘

ematipico:
Sounds good!

bin/utils/convert-argv.js
```diff
@@ -7,6 +7,7 @@ const webpackConfigurationSchema = require("../config/webpackConfigurationSchema
 const validateSchema = require("webpack").validateSchema;
 const WebpackOptionsValidationError = require("webpack").WebpackOptionsValidationError;
 const findup = require("findup-sync");
+// const { webpackConfigPath } = require("@webpack-cli/utils/path-utils");
```
dhruvdutt: Can we clean this now?

anshumanv: Cleaned up 👍

packages/utils/__tests__/is-local-path.test.ts
```diff
@@ -1,7 +1,7 @@
 "use strict";
 
 import * as path from "path";
-import isLocalPath from "../is-local-path";
+import { isLocalPath } from "../path-utils";
```
evenstensberg: absolute import by package here, to avoid breaking stuff

anshumanv: Hey @evenstensberg if you mean importing by the package name like this - 
image this doesn't work clearly even after symlink (same issue as before) and if you intend to do something like `import x from 'utils/path-utils` then that would need changes to the tsconfig file to add entries like `baseUrl` and `paths` and stuff which I don't think needs to be done here since tests seem to work fine both locally and on CI.

Or maybe you can suggest something? :)

evenstensberg: I think you'll need to change how the CI bootstraps in that case. You should get it up running locally first.

anshumanv: @evenstensberg I investigated on this, everything works if I do 
```sh
cd packages/utils
npm link
npm link @webpack-cli/utils
```
which basically creates a symlink of the utils package which should originally be done by lerna.
There seems to be some issue with lerna https://github.com/webpack/webpack-cli/pull/806#issuecomment-480594467 and thus I'm still figuring this out since I started working on this. Can we go ahead with this meanwhile?

anshumanv:
@evenstensberg PTAL ✨ 

https://github.com/webpack/webpack-cli/pull/806#discussion_r277530633

anshumanv:
@evenstensberg it's recommended to use relative imports for tests - https://github.com/lerna/lerna/issues/2052#issuecomment-492443271

packages/utils/__tests__/is-local-path.test.ts
evocateur: Literally _all_ of the sibling tests in this module use relative imports of the file under test. Really unclear how using absolute package imports will help "avoid breaking stuff."

anshumanv: cc @evenstensberg 

Thanks @evocateur 🙏

packages/utils/path-utils.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+/**
+ * Find the root directory path of a project.
+ *
+ * @returns {String} Absolute path of the project root.
+ */
+
+export function findProjectRoot(): string {
+	const rootFilePath = findup(`package.json`);
```
evenstensberg: no need to use string literals here

packages/utils/path-utils.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+/**
+ * Get absolute path of a webpack config in a project.
+ *
+ * @param {String[]} str - array of extensions to look for.
+ * @returns {String} Absolute path of the config.
+ */
+
+export function webpackConfigPath(extensions: string[]): string {
```
evenstensberg: I don't see this being used anywhere in the PR, could you find where it is used and replace it with the new location and if not used, remove it?

Issue #572, opened 2018-08-24
bug: require(...).default is not a function


huixisheng:
**Describe the bug**

`package.json` with content 

```
...
  "devDependencies": {
    "@webpack-cli/generate-plugin": "0.0.8",
    "@webpack-cli/init": "^0.1.0",
    "@webpack-cli/migrate": "^0.1.0",
    "webpack": "^4.17.1",
    "webpack-cli": "^3.1.0"
  },
...
```

**What is the current behavior?**

```
TypeError: require(...).default is not a function
    at promptForInstallation (/Users/huixisheng/x/lab/webpack/webpack-plugin-demo/node_modules/webpack-cli/bin/prompt-command.js:108:30)
    at /Users/huixisheng/x/lab/webpack/webpack-plugin-demo/node_modules/webpack-cli/bin/cli.js:45:37
    at Object.<anonymous> (/Users/huixisheng/x/lab/webpack/webpack-plugin-demo/node_modules/webpack-cli/bin/cli.js:530:3)
    at Module._compile (module.js:652:30)
    at Object.Module._extensions..js (module.js:663:10)
    at Module.load (module.js:565:32)
    at tryModuleLoad (module.js:505:12)
    at Function.Module._load (module.js:497:3)
    at Module.require (module.js:596:17)
    at require (internal/module.js:11:18)
```

**To Reproduce**

1. npx webpack-cli generate-plugin
2. vi remote default
```
	} else {
		require(pathForCmd)(...args); // eslint-disable-line
	}
```

**Expected behavior**

only  my-webpack-plugin create.

**Screenshots**

![image](https://user-images.githubusercontent.com/1518967/44596785-19232480-a800-11e8-8867-a71c99a09b8c.png)

```
  System:
    OS: macOS High Sierra 10.13.6
    CPU: x64 Intel(R) Core(TM) i5-5257U CPU @ 2.70GHz
  Binaries:
    Node: 8.11.1 - /usr/local/bin/node
    Yarn: 1.7.0 - /usr/local/bin/yarn
    npm: 6.2.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 68.0.3440.106
    Firefox: 54.0
    Safari: 11.1.2
  npmPackages:
    webpack: ^4.17.1 => 4.17.1 
    webpack-cli: ^3.1.0 => 3.1.0 
    webpack-defaults: ^2.3.0 => 2.3.0 
  npmGlobalPackages:
    webpack-cli: 3.1.0
[...]

PR #589, opened 2018-09-15
Fix: fix prompt-installation #572

duzy:
the strange things in this fix is about packages -- if it's not indicating the sub-command or a list of commands?

**What kind of change does this PR introduce?**

**Did you add tests for your changes?**

**If relevant, did you update the documentation?**

**Summary**

**Does this PR introduce a breaking change?**

**Other information**
Fixes #572

PR #815, opened 2019-04-01
fix(bin, serve): force default package export, add serve default

Loonride:
ISSUES CLOSED: #572

This PR introduces a bugfix for the serve package to export a default function, and a more useful error message when a package does not export a default function.

No, I could not seem to find any tests relating to the packages, eg. serve, init. If you would like me to write tests I can, I might just need some guidance.

No

**Summary**

I ran into this issue: https://github.com/webpack/webpack-cli/pull/589 while trying `webpack-cli serve`. The reason for the issue is that `serve` does not export a default function. There was a PR earlier about this that never got finished https://github.com/webpack/webpack-cli/pull/589. `serve` still has problems, but I think those should be addressed in a separate PR.

The changes in serve are not breaking, but the changes in bin are breaking for serve if you do not consider the changes I made. I'm not exactly sure how the npm publishing process works for these multiple packages. If everything is published at the same time, there is no problem, and if serve is published before the bin, there is no problem. In any case, serve seems to still have other problems with the current webpack-cli version outside of this.

**Other information**

### DIFF
--- a/bin/utils/prompt-command.js
+++ b/bin/utils/prompt-command.js
@@ -37,6 +37,15 @@ const npmGlobalRoot = () => {
 	});
 };
 
+const runWhenInstalled = (packages, pathForCmd, ...args) => {
+	const package = require(pathForCmd);
+	const func = package.default;
+	if (typeof func !== 'function') {
+		throw new Error(`@webpack-cli/${packages} failed to export a default function`);
+	}
+	return func(...args);
+}
+
 module.exports = function promptForInstallation(packages, ...args) {
 	const nameOfPackage = "@webpack-cli/" + packages;
 	let packageIsInstalled = false;
@@ -113,10 +122,7 @@ module.exports = function promptForInstallation(packages, ...args) {
 							}
 
 							pathForCmd = path.resolve(process.cwd(), "node_modules", "@webpack-cli", packages);
-							if (packages === "serve") {
-								return require(pathForCmd).default.serve();
-							}
-							return require(pathForCmd).default(...args); //eslint-disable-line
+							return runWhenInstalled(packages, pathForCmd, ...args);
 						})
 						.catch(error => {
 							console.error(error);
@@ -132,6 +138,6 @@ module.exports = function promptForInstallation(packages, ...args) {
 			}
 		});
 	} else {
-		require(pathForCmd).default(...args); // eslint-disable-line
+		return runWhenInstalled(packages, pathForCmd, ...args);
 	}
 };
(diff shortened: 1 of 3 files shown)


bin/utils/prompt-command.js
```diff
@@ -113,10 +122,7 @@ module.exports = function promptForInstallation(packages, ...args) {
 							}
 
 							pathForCmd = path.resolve(process.cwd(), "node_modules", "@webpack-cli", packages);
-							if (packages === "serve") {
-								return require(pathForCmd).default.serve();
```
evenstensberg: you cannot remove `.serve()`

packages/serve/index.ts
```diff
@@ -170,7 +170,7 @@ function serve() {
 	}
 }
 
-export = {
+export {
```
evenstensberg: we're using these, so needs to be grouped export

evenstensberg: refactor opportunity here

bin/utils/prompt-command.js
Loonride: The idea is that `.serve()` should just be the default function exported by the `serve` package, the way all the other packages export a default function. `.default.serve()` as it is in this file is not even correct, as `serve` currently does not have a default export. It's required and used correctly in the example: https://github.com/webpack/webpack-cli/blob/master/packages/serve/README.md#webpack-cli-serve

evenstensberg: sounds good, is there any other place the exported modules are used?

packages/serve/index.ts
Loonride: What do you mean? There are no breaking changes here, its simply adding the default export onto the object of existing exported functions.

evenstensberg: we actually don't need these exports. Safe to remove

bin/utils/prompt-command.js
Loonride: I don't see any other places they are used. `prompt-command.js` should be the only place since it should make sure it is installed before trying to use a module.

packages/serve/index.ts
evenstensberg: do this and I'll do a lgtm

Loonride: We should keep `serve` in case someone uses this method: https://github.com/webpack/webpack-cli/tree/master/packages/serve

evenstensberg: You can do a default export of serve only, adjust the docs to use default export:

` const serve = require('rick-n-morty/serve'=`

Loonride: Oh wait then it would be `const serve = require("@webpack-cli/serve").default;`, seems kind of ugly compared to `.serve`.

evenstensberg: Maybe you could do the same as in `init` ?

https://github.com/webpack/webpack-cli/tree/master/packages/init#node

evenstensberg:
@webpack/cli-team PTAL, please do a local test against symlinked packages

PR #816, opened 2019-04-02
[WIP] misc: adds the description in each file

pranshuchittora:
It's hard to understand what each file in the code does. By adding a bit of description at the top in each file can be quite helpful.
```
Name : <FILE NAME>
Description : <Info about the core functionality>
```

Improves DX

### DIFF
--- a/packages/generators/init-generator.ts
+++ b/packages/generators/init-generator.ts
@@ -1,3 +1,8 @@
+/*
+Name 			: init-generators
+Description 	: Generates new webapck configuration as per user requirements.
+*/
+
 import chalk from "chalk";
 import * as logSymbols from "log-symbols";
 import Generator = require("yeoman-generator");


pranshuchittora:
As of now, I have added the description to only one file. 
Let's discuss its pros & cons. :fire:

packages/generators/init-generator.ts
```diff
@@ -1,3 +1,8 @@
+/*
+Name 			: init-generators
+Description 	: Generates new webapck configuration as per user requirements.
```
evenstensberg: spelling

ematipico:
@pranshuchittora you could mark this PR as Draft PR, if you think this is not ready. If you can't mark it as "Draft", could you add the "[WIP]" inside the title?

packages/generators/init-generator.ts
```diff
@@ -1,3 +1,8 @@
+/*
```
rishabh3112:
```suggestion
/**
```

evenstensberg:
Hi. As long as we have descriptive code and JSDocs (instead of comments on top), I think this is redundant. Thanks for putting in the time!

PR #817, opened 2019-04-04
misc(init-generator) cli improvements & refactor

### LINKED ISSUE #788: Scaffold: Better defaults for webpack-cli init questions
`webpack-cli init` from my point of view has 2 main purposes, making the setup faster and making it easier expecially for beginners. For these reasons I think the default answers for `webpack-cli init` should be modified to allow creating the simpler possible setup.

1.  `? Will your application have multiple bundles? (Y/n)`
Having multiple bundles is an "advanced" and less common configuration compared to having just one entry point.
I think the default should be `(y/N)`

2. `? Will you use one of the below CSS solutions?`
Same here, the default option is `SASS` where a better default would be `No`:
    ```
    > SASS
      LESS
      CSS
      PostCSS
      No
    ```
    Also, I use `init` mainly to scaffold node projects where I don't use CSS.

**Additional context**
I think this is a very very small modification that can slightly improve the user experience.

I can work on this if it is something desirable :)

### LINKED ISSUE #814: Enhancement: Init generator improvements
The init generator executed while running `webpack-cli init` creates a webpack configuration file and installs the dependencies, I think it could be improved:
- There are some unused imports and comments in the generated configuration file
- The keys in the configuration object are in a random order, IMO this should be an example of good practices for beginners
- The entry point provided by the user is never set in the final configuration file
- We check many times a variable [isProd](https://github.com/webpack/webpack-cli/blob/master/packages/generators/init-generator.ts#L29) that is initialized to `false` and never updated
- The code could be refactored, `topScope` and `webpackOptions` are initialized sparsely
- When multientry configuration is chosen we suggest naming the entries "app" and "vendor", this seems to be [discouraged](https://webpack.js.org/concepts/entry-points/#separate-app-and-vendor-entries) in v4

- [x] generate some boilerplate too (eg. generate the entry point in `src/index.js`)
- [ ] cleaner code
- [x] cleaner configuration with ordered keys 

I think this would benefit the programmatic usage too.

I'm already working on this, please let me know any opinion and if you think there's something else that can be improved :)

misterdev:
Fixes some problems and adds new features

Yes, for `webpack-scaffold`

**Summary**
Fixes: #884
Closes #814 
Closes #788

From #814:  
- [x] Init generator refactor
- [x] Improves command line defaults adding defaultChoice to Input and InputValidate
- [x] Improves generated webpack configuration (functionality & formatting)
- [x] Generate the entry file/files, `package.json` with a build script and a README
- [x] Update documentation (webpack-scaffold & init)
- [x] Add a prompt to tell the user what to do when the scaffolding completes
- [x] Add the possibility to choose between vanilla, es6 or typescript
- [x] Fixes: 
    > { parser: "babylon" } is deprecated; we now treat it as { parser: "babel" }
- [x] Updates init-generator to async/await (#836)

**CLI output when selecting defaults:**

- Before
before

- After
after

**CLI output when multientry is chosen:**
> Naming the entries "app" and "vendor" seems to be [discouraged](https://webpack.js.org/concepts/entry-points/#separate-app-and-vendor-entries) in v4

- Before
beforemult

- After
aftermult

No


INIT.md
```diff
@@ -1,6 +1,6 @@
 # webpack-cli init
 
-`webpack-cli init` is used to initialize `webpack` projects quickly by scaffolding configuration and installing modules required for the project as per user preferences.
+`webpack-cli init` is used to initialize `webpack` projects quickly by scaffolding configuration, generating some boilerplate and installing modules required for the project as per user preferences.
```
evenstensberg: B specific, better to tell them that we create a config and make a runnable config, not boilerplate

packages/generators/init-generator.ts
```diff
@@ -40,16 +43,63 @@ export default class InitGenerator extends Generator {
 		super(args, opts);
 		this.isProd = false;
 		this.usingDefaults = false,
+
 		this.dependencies = [
 			"webpack",
 			"webpack-cli",
-			"terser-webpack-plugin",
```
evenstensberg: why is terser removed

packages/generators/init-generator.ts
```diff
@@ -40,16 +43,63 @@ export default class InitGenerator extends Generator {
...
+			"webpack-dev-server",
```
evenstensberg: use dev server if config is devmode

packages/generators/init-generator.ts
```diff
@@ -40,16 +43,63 @@ export default class InitGenerator extends Generator {
...
 		];
+		if (this.isProd) {
+			this.dependencies.push("terser-webpack-plugin");
+		}
+
 		this.configuration = {
 			config: {
 				topScope: [],
-				webpackOptions: {},
+				// tslint:disable: object-literal-sort-keys
```
evenstensberg: don't disable rules

packages/generators/init-generator.ts
```diff
@@ -40,16 +43,63 @@ export default class InitGenerator extends Generator {
...
+				webpackOptions: {
+					mode: this.isProd ? "'production'" : "'development'",
```
evenstensberg: let's try to set as much of a minimal config as possible

packages/generators/init-generator.ts
```diff
@@ -40,16 +43,63 @@ export default class InitGenerator extends Generator {
...
+
+		if (this.isProd) {
+			this.configuration.config.topScope.push(
+				tooltip.terser(),
+				"const TerserPlugin = require('terser-webpack-plugin');",
+				"\n",
+			);
+		}
+
+		this.configuration.config.webpackOptions.optimization = {
```
evenstensberg: I'm in favor for using `splitChunks: all`. Works good as a initial default, at least in dev mode

packages/generators/templates/package.json.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
```
evenstensberg: This is great. Let's try to omit the flags and bake them into the config

packages/generators/utils/entry.ts
```diff
@@ -25,8 +25,9 @@ export default function entry(self: IEntry, answer: {
 			.prompt([
 				InputValidate(
 					"multipleEntries",
-					"Type the names you want for your modules (entry files), separated by comma [example: app,vendor]",
```
evenstensberg: don't remove examples

packages/generators/utils/entry.ts
```diff
@@ -87,16 +89,17 @@ export default function entry(self: IEntry, answer: {
 	} else {
 		result = self
 			.prompt([
-				InputValidate(
+				Input(
```
evenstensberg: Have you tested on empty strings when removing this?

packages/generators/utils/entry.ts
misterdev: I've removed them in favor of default values:

multixd

packages/generators/utils/entry.ts
misterdev: The InputValidate was used as an Input since we didn't provide it a validator function.
I've added default values to both so now when the answer is an empty string it defaults the entry to "src/index"

packages/generators/init-generator.ts
misterdev: I'm trying to make the generated webpack configuration "standard", where for standard I mean with keys ordered like [this](https://webpack.js.org/configuration#options):

```js
module.exports = {
  mode: "production", // "production" | "development" | "none"
  entry: "./app/entry", // string | object | array
  output: {
    /* ... */
  },
  module: {
    /* ... */
  }
```

The other solution I see is to keep the keys ordered is to set them this way: 

```js
webpackOptions.mode = ...
webpackOptions.entry = ...
webpackOption.output = ...
```

ematipico: It makes sense to me, that's the order I've been using for ages

packages/generators/init-generator.ts
ematipico: Only in dev mode is fine as long as we use the in memory output (that's what dev server does). But not in production, we can't assume people have the correct version of `HtmlWebpackPlugin` with the correct configuration

packages/generators/init-generator.ts
ematipico: It gets added later only in production. Makes sense

misterdev: I agree with @ematipico.

I'm confused because in the [current](https://github.com/webpack/webpack-cli/blob/master/packages/generators/init-generator.ts#L430) code we remove terser in production and we include it only in devmode (also, we never use the plugin in the configuration) 

```typescript
if (this.isProd) {
	this.dependencies = this.dependencies.filter(
		(p: string): boolean => p !== "terser-webpack-plugin",
	);
} else {
	this.configuration.config.topScope.push(
		tooltip.terser(),
		"const TerserPlugin = require('terser-webpack-plugin');",
		"\n",
	);
}
```

Is terser only in production the way to go?

ematipico: Yes! It doesn't make much sense having it in production

packages/generators/init-generator.ts
misterdev: Thanks! So, I'm going to keep it in production and using "all" in dev mode, let me know if it should be done differently

packages/generators/init-generator.ts
misterdev: I'm keeping this if it's ok

misterdev:
Ok, I'm happy about the work so far, I would appreciate to also receive feedback on further works;

I see `isProd` and `useDefaults` in the code that are initialized to false and never updated. How are they meant to be used? 

Would be desirable to:
1. Have the possibility to choose between vanilla, es6 or typescript?
2. Take the name of the project as an input from the user and generate the also the project folder?
3. Have the possibility to choose the target (node, browser, electron, ...)?

ematipico:
I would really love to have typescript in this enhancement of the `init-generator`!

packages/generators/utils/style.ts
```diff
@@ -0,0 +1,172 @@
...
+				);
+			} else {
+				ExtractUseProps.push(
+					{
+						loader: "'style-loader'",
+					},
+					{
+						loader: "'css-loader'",
+						options: {
+							importLoaders: 1,
```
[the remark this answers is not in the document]
misterdev: Seems like this is a way to handle the dependencies of a CSS file (explained [here](https://github.com/webpack-contrib/css-loader/issues/228))

packages/generators/utils/entry.ts
```diff
@@ -64,8 +63,9 @@ export default function entry(self: IEntry, answer: {
 					self.prompt([
 						InputValidate(
 							`${entryProp}`,
-							`What is the location of "${entryProp}"? [example: ./src/${entryProp}]`,
+							`What is the location of "${entryProp}"?`,
 							validate,
+							`./src/${entryProp}`,
```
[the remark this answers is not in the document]
misterdev: Yeah, that's how you described:

bug

This prompt allows the user to customize the entry point of each bundle suggesting as value `./src/bundle_name`

Btw. I found a bug looking into this, thanks ;)

packages/generators/init-generator.ts
```diff
@@ -38,20 +41,98 @@ export default class InitGenerator extends Generator {
...
+			);
+		}
+
+		this.configuration.config.webpackOptions.plugins.push(
+			"new webpack.ProgressPlugin()",
+		);
+
+		if (this.isProd) {
+			this.configuration.config.webpackOptions.plugins.push(
+				"new TerserPlugin()",
```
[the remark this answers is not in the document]
misterdev: Nice, thank you!

ematipico:
Do you want to try to move promises to `async`/`await`?

misterdev:
Yes, what do you think about that?

packages/generators/utils/style.ts
```diff
@@ -0,0 +1,172 @@
...
+		case "SASS":
+			self.dependencies.push(
+				"sass-loader",
+				"node-sass",
+				"style-loader",
+				"css-loader",
+			);
+			regExpForStyles = `${new RegExp(/\.(scss|css)$/)}`;
+			if (self.isProd) {
+				ExtractUseProps.push(
```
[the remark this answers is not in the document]
misterdev: I just found out that we currently use it in the init-generator

ematipico:
It's a good opportunity doing it now as you're working on revamping the whole generator :)

packages/generators/utils/style.ts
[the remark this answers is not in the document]
ematipico: I must have missed it 😅 thanks!

packages/generators/templates/index.js
```diff
@@ -0,0 +1,5 @@
+/**
+ * Application entry point
+ */
+
+console.log("Hello World from <%= name %>");
```
rishabh3112:
```suggestion
// Start writing code here
```
We can clearly guide the user here instead of a `console.log`

packages/generators/templates/index.js
```diff
@@ -0,0 +1,5 @@
+/**
+ * Application entry point
```
rishabh3112: IMO this comment is not required.
Instead, we can have a README.md at user's root to guide users about each file. This is also followed by other scaffolding tools.

packages/generators/init-generator.ts
```diff
@@ -59,6 +59,7 @@ export default class InitGenerator extends Generator {
 			config: {
 				configName: this.isProd ? "prod" : "config",
 				topScope: [],
+				// TODO migrate tslint
```
ematipico: This is now done, so you can rebase from master. I think that comment has been deleted there.

packages/generators/templates/tsconfig.json.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+	compilerOptions: {
+		noImplicitAny: true,
+		module: "es6",
+		target: "es5",
+		jsx: "react",
```
ematipico: I'd remove it. We don't know what the user will use for their project. 
We could set `allowSyntheticDefaultImports` to `true`

packages/generators/utils/language.ts
```diff
@@ -4,9 +4,10 @@ export enum LangType {
 }
 
 interface IModuleRule extends Object {
```
ematipico: Yeah I think you should rebase from `master`. The interfaces name as default rule can't start with `I`

packages/generators/utils/language.ts
```diff
@@ -36,23 +40,60 @@ export function getBabelPlugin(): IModuleRule {
...
+	};
+}
+
+export function getTypescriptLoader(): IModuleRule {
+	return {
+		// TODO migrate tslint
+		// tslint:disable: object-literal-sort-keys
+		test: "/\.tsx?$/",
+		loader: "'ts-loader'",
+		include: ["path.resolve(__dirname, 'src')"],
```
ematipico: `'src'` is hardcoded and it shouldn't. If the user chose a different folder, we should set that one.

packages/generators/utils/language.ts
```diff
@@ -36,23 +40,60 @@ export function getBabelPlugin(): IModuleRule {
...
+			};
+
+			// Update the entry files extensions to .ts
+			const jsEntryOption = self.configuration.config.webpackOptions.entry;
+			const jsExtension = new RegExp("\.js(?!.*\.js)");
+			let tsEntryOption = {};
+			if (typeof jsEntryOption === "string") {
+				tsEntryOption = jsEntryOption.replace(jsExtension, ".ts");
+			} else if (typeof jsEntryOption === "object") {
+				Object.keys(jsEntryOption).map((entry) => {
```
ematipico: In this case `forEach` is a better fit. Using a `map` without assigning it to a variable could lead to memory leaks.

packages/utils/run-prettier.ts
```diff
@@ -19,7 +19,7 @@ export default function runPrettier(outputPath: string, source: string, cb?: Fun
 		try {
 			prettySource = prettier.format(source, {
 				filepath: outputPath,
-				parser: "babel",
+				parser: "babylon",
```
ematipico: Here they use `babel`: https://prettier.io/docs/en/api.html#prettierformatsource-options
Does it depend by the version that we have installed?

packages/generators/templates/index.js
misterdev: You are right!

packages/generators/templates/index.js
misterdev: I'd keep this since the purpose here is to scaffold a basic project so that users can have something working right after executing init. Another idea would be to generate a simple website, but I prefer to be less opinionated. Any suggestion on improving the generated boilerplate is welcome

packages/generators/init-generator.ts
misterdev: Yeah, I left that comment as a reminder for myself :) I'm waiting to complete my work on the generator, just to avoid rebasing multiple times

packages/generators/utils/language.ts
misterdev: Good point, I didn't notice, thanks!

packages/utils/run-prettier.ts
misterdev: You are right, `babylon` is deprecated. I'm still trying to figure out what is happening here, most of the time `babel` works but sometimes It gives me an error. I may have a dirty workspace in one of the computers I use... It'll be fixed before merging ;)

misterdev:
Took a while but is finally complete

If you want to take a look again:
@ematipico @evenstensberg @rishabh3112

misterdev:
This is rebased and ready for review :) @ematipico

ematipico:
What do you think about updating the PR with the new types (yeoman) you just changed? That's why I merged that one first :D

misterdev:
I think that's a good idea :) Rebased successfully @ematipico 🎉

packages/generators/templates/index.js
evenstensberg: agreeing with @misterdev

misterdev:
Would be great! @eventstensberg :)

packages/generators/utils/language.ts
```diff
@@ -0,0 +1,128 @@
...
+ * Returns an module.rule object for the typescript loader
+ * @param {string[]} includeFolders An array of folders to include
+ * @returns {ModuleRule} A configuration containing the ts-loader
+ */
+export function getTypescriptLoader(includeFolders: string[]): ModuleRule {
+	const include = includeFolders.map((folder: string): string =>
+		`path.resolve(__dirname, '${folder}')`
+	);
+	return {
+		test: "/\.tsx?$/",
```
ematipico:
```suggestion
		test: "/\.(ts|tsx)?$/",
```

packages/generators/utils/language.ts
```diff
@@ -0,0 +1,128 @@
...
+	exclude?: string[];
+	loader: string;
+	options?: {
+		plugins: string[];
+		presets: Preset[][];
+	};
+	test: string;
+}
+
+type Preset = string | object;
```
ematipico: I would use the types coming from `webpack`: https://github.com/webpack/webpack/blob/master/declarations/WebpackOptions.d.ts#L82

misterdev:
Thanks for reviewing @ematipico!
I included your suggestions, I've used the WebpackOptions types defined here: https://github.com/webpack/webpack-cli/blob/master/packages/generators/types/index.ts

ematipico:
Which is not ideal IMHO. Those were done when webpack didn't have any declarations. Now that webpack is typescript friendly, I believe we should consume them as they are heavily used in their codebase.

We could do it in a second moment, it's not crucial. I'm happy with the PR! Amazing work here!!

packages/generators/add-generator.ts
```diff
@@ -140,7 +140,7 @@ export default class AddGenerator extends Generator {
 							.then(
 								(entryTypeAnswer: { entryType: boolean }): Promise<void | {}> => {
 									// Ask different questions for entry points
-									return entryQuestions(self, entryTypeAnswer);
+									return entryQuestions(self, entryTypeAnswer.entryType);
```
evenstensberg: I think that the entrytype here is already module destructed, no?

packages/webpack-scaffold/README.md
```diff
@@ -163,38 +169,38 @@ CheckList("entry", "what kind of entry do you want?", ["Array", "Function"]);
...
-	}
-};
-InputValidate("entry", "what is your entry point?", validation);
+const InputValidate = require('@webpack-cli/webpack-scaffold').InputValidate;
+
+const validation = (value) => {
+    if(value.length > 4) {
+        return true;
+    } else {
+        return 'Wow, that was short!'
```
evenstensberg: Give the user actions to improve explicitly instead

packages/generators/utils/entry.ts
```diff
@@ -18,19 +18,18 @@ interface CustomGenerator extends Generator {
...
 	let result: Promise<void | {}>;
-	if (answer.entryType) {
+	if (multiEntries) {
 		result = self
 			.prompt([
 				InputValidate(
 					"multipleEntries",
-					"Type the names you want for your modules (entry files), separated by comma [example: app,vendor]",
-					validate
+					"How do you want to name your bundles? (separated by comma)",
```
evenstensberg: What instead of how

packages/generators/utils/entry.ts
```diff
@@ -18,19 +18,18 @@ interface CustomGenerator extends Generator {
...
+					validate,
+					"pageOne, pageTwo"
```
evenstensberg: I'd CC @addyosmani here (sorry) for a help on how to name the default name of multiple bundles in webpack. 

Config will look like this: 

```js
{
pageOne: './foo/bar',
pageTwo: './baz'
}
```

What naming should be instead of `PageOne` and `PageTwo` to give the user an indication of their entry points? I'm suggesting that we have `pageEntryOne` instead to explicitly tell the user that this is an entry in terms of bundles

packages/generators/utils/entry.ts
evenstensberg: Example here is probably a good idea as this is the confusing part of webpack

packages/generators/utils/language.ts
```diff
@@ -0,0 +1,117 @@
...
+			tsEntryOption[entry] = replaceExt(jsEntryOption[entry], newExt);
+		});
+	}
+	self.configuration.config.webpackOptions.entry = tsEntryOption;
+}
+
+const getFolder = (path: string): string =>
+	path.replace("'./", "").split("/").slice(0, -1).join("/");
+
+function getEntryFolders(self): string[] {
```
evenstensberg: This splits an entry property to an object if its an object and to a string if it's a string? If so, I think maybe we should name this to `getEntryProperties`

packages/generators/utils/language.ts
```diff
@@ -0,0 +1,117 @@
...
+	);
+	return {
+		test: "/\.(ts|tsx)?$/",
+		loader: "'ts-loader'",
+		include,
+		exclude: ["/node_modules/"],
+	};
+}
+
+export default function language(self, langType: string): void {
```
evenstensberg: Language is probably a bit vague for this namespace. Perhaps something that indicates that it's the extention support we're actually providing: `extentionSupport` or similar.

packages/generators/utils/plugins.ts
```diff
@@ -1,11 +0,0 @@
-/**
```
evenstensberg: Are we using this file somewhere?

packages/generators/utils/entry.ts
misterdev: I've removed the example from the prompt in favor of default answers. It now looks like this:

bundles

packages/generators/utils/language.ts
misterdev: It returns an array of strings with the paths of all the folders containing an entry point

For example:
```ts
entry: string = "./this/is/sparta.js"
// getEntryFolders returns ["this/is"]

entry: object = {
  pageOne: './this/is/pageone.js'
  pageTwo: './this/pageTwo.js'
} 
// getEntryFolders returns ["this/is", "this"]
```

I use this inside the babel/typescript loaders configuration to include the folders containing an entry point :)

packages/generators/utils/plugins.ts
misterdev: No, we used it inside the `init-generator` ([link](https://github.com/webpack/webpack-cli/blob/master/packages/generators/init-generator.ts#L12)) calling it `getDefaultPlugins`

I've just moved this code into the default production configuration ([here](https://github.com/misterdev/webpack-cli/blob/misc/improve-init-generator/packages/generators/init-generator.ts#L101))

packages/generators/utils/language.ts
misterdev: Makes sense, would `languageSupport` or `langExtSupport` be ok? `extensionSupport` seems too generic for me, style extension, for example, are handled in another file called [style.ts](https://github.com/misterdev/webpack-cli/blob/misc/improve-init-generator/packages/generators/utils/style.ts) (that I guess can be renamed too)

packages/generators/add-generator.ts
misterdev: Not really, `entryTypeAnswer: { entryType: boolean }`  is a type annotation :)

packages/generators/utils/language.ts
evenstensberg: Ok thanks for informing

packages/generators/utils/language.ts
evenstensberg: langaugeSupport is nice

packages/generators/utils/plugins.ts
evenstensberg: Okay. Could you abstract that part into a func? Looks messy

packages/webpack-scaffold/README.md
```diff
@@ -186,15 +192,15 @@ const validation = value => {
 	if (value.length > 4) {
 		return true;
 	} else {
-		return "Wow, that was short!";
+		return "The entry point must be longer than 4, try again";
```
evenstensberg:
```suggestion
		return "Your answer must be longer than 4 characters, try again";
```

packages/webpack-scaffold/__tests__/__snapshots__/index.test.ts.snap
```diff
@@ -2,6 +2,17 @@
 
 exports[`utils Inquirer should make an Input object with validation 1`] = `
 Object {
+  "default": undefined,
```
evenstensberg: Don't snapshot undefined values, hard to verify if there's actually an undefined bug somewhere

packages/webpack-scaffold/__tests__/index.test.ts
```diff
@@ -76,6 +76,15 @@ describe("utils", () => {
 		});
 		it("should make an Input object", () => {
 			expect(utils.Input("plugins", "what is your plugin?")).toEqual({
+				default: undefined,
```
evenstensberg: ref reply ^

packages/generators/utils/language.ts
misterdev: Renamed to `languageSupport` and `styleSupport`

packages/generators/utils/plugins.ts
misterdev: I've moved it to a new file [`generators/utils/webpackConfig`](https://github.com/misterdev/webpack-cli/blob/misc/improve-init-generator/packages/generators/utils/webpackConfig.ts) . I chose this name so that it can be possibly used for other configuration utilities (eg. defaultConfiguration)

packages/webpack-scaffold/__tests__/__snapshots__/index.test.ts.snap
misterdev: Done, I had to change the code ([link](https://github.com/webpack/webpack-cli/pull/817/commits/fb25bd2cb1d72def40e14ba431fe368cd6df6653))

packages/webpack-scaffold/index.ts
```diff
@@ -76,21 +76,30 @@ export function CheckList(name: string, message: string, choices: string[]): Gen
...
+		defaultChoice?: string,
+	): Generator.Question {
+
+	const input: Generator.Question = {
 		message,
 		name,
 		type: "input",
 		validate: cb
 	};
+	if (defaultChoice !== undefined) input.default = defaultChoice;
```
evenstensberg: Could you do `!defaultCoice` here?

misterdev: Sure! Done, thanks!

packages/generators/utils/entry.ts
misterdev: Those names are taken from the [documentation](https://github.com/misterdev/webpack/blob/output-reporter/lib/ReporterPlugin.js) where there are 2 examples, one with `app, adminApp` and one (more generic) with `pageOne, pageTwo, pageThree`.

Since multiple bundles are a more "advanced" option that you have select before using, I expect the majority of people to be aware of what they are doing. Also, someone looking for information may end up in the documentation and find a consistent example.

IMO if we want to move on we can keep those and then change them if we find something better. What do you think @evenstensberg? :)

packages/webpack-scaffold/README.md
```diff
@@ -186,15 +192,15 @@ const validation = value => {
 	if (value.length > 4) {
 		return true;
 	} else {
-		return "Wow, that was short!";
+		return "Your answer must be longer than 4, try again";
```
rishabh3112:
```suggestion
		return "Your answer must be longer than 4 characters, try again";
```

packages/utils/run-prettier.ts
```diff
@@ -19,18 +19,17 @@ export default function runPrettier(outputPath: string, source: string, cb?: Fun
 		try {
 			prettySource = prettier.format(source, {
 				filepath: outputPath,
-				parser: "babylon",
+				parser: "babel",
```
rishabh3112: `jscodeshift` makes transforms using `babylon` parser currently. This may give inconsistent results.

ematipico: Why would it? Here we just format a string using prettier (with babel, because babylon is deprecated), I don't any `jscodeshift` thing that could cause troubles

rishabh3112: I wasn't sure @ematipico wrote this review because, prettier will be formating the source string that also includes the transforms made using babylon parser and as babel is quiet new as compared to babylon and may not match the some of the old syntactical structures of commonJS ( if there's any ).

But still I may be wrong here :)

packages/webpack-scaffold/README.md
misterdev: Thanks! :)

packages/utils/run-prettier.ts
misterdev: Looking at `jscodeshift`'s [README](https://www.npmjs.com/package/jscodeshift) seems that it uses `babel` as default parser

```
 --parser=babel|babylon|flow|ts|tsx  the parser to use for parsing the source files
                                          (default: babel)
```

Searching in `webpack-cli` source code I can't find any `babylon` occurence, so I guess we are not overriding that default value.
What do you think @rishabh3112? :)

rishabh3112: We are using an older version of jscodeshift (0.5.x) which uses babylon for parsing.

rishabh3112: 0.5.1 to be precise , refer utils package:)

ematipico: Fair point @rishabh3112 	
I wouldn't too much though, babel and babylon are the same parser at the end. I am optimistic and I'd say it won't break things 😄

rishabh3112: Ya sure, it was tinkling me at back of my mind 😂

evenstensberg: Need to change it to babylon as tsc is complaining

packages/generators/utils/entry.ts
evenstensberg: aye let's keep it like this

packages/generators/utils/webpackConfig.ts
```diff
@@ -0,0 +1,32 @@
...
+                cacheGroups: {
+                    vendors: {
+                        priority: -10,
+                        test: "/[\\\\/]node_modules[\\\\/]/",
+                    },
+                },
+                chunks: "'async'",
+                minChunks: 1,
+                minSize: 30000,
+                // for production name is recommended to be off
```
evenstensberg: remove this

misterdev: Done ;)

misterdev:
PS: @evenstensberg changing prettier parser to `"babylon"` will make appear a message every time we run `init` `{ parser: "babylon" } is deprecated; we now treat it as { parser: "babel" }`

evenstensberg:
@misterdev Could you make it so that tsc doesn't complain while having parser: babel?

misterdev:
I don't have errors, but I remember I had that problem some times ago.
Sorry for asking,  have you executed `yarn bootstrap`? It should install a new version of prettier and @types/prettier

evenstensberg:
I didn't, had some old lockfiles, was fixed once I ran that. Thanks!

PR #826, opened 2019-04-11
chore(refactor): refactor scaffolding and AST utils

rishabh3112:

Refactoring for removing console printing from core functions so that while UI is running there is no noise in UI.
No
No
**Summary**
WIP, the effort to reuse the codebase of webpack-cli for UI

No (most probably)
**Other information**
N/A


rishabh3112:
Are these changes acceptable? Also, Can I override `this.prompt` function to use functions from socket class?

This way would be able to use the exact same yeoman generator without any duplicate changes :)

If so, after merging of this PR, would work on getting the above approach done.

Pinging for having a bit of feedback,
@evenstensberg @ematipico @dhruvdutt

packages/utils/modify-config-helper.ts
```diff
@@ -87,59 +95,95 @@ export default function modifyHelperUtil(
...
+					chalk.red("\nCould not find a yeoman configuration file.\n"),
+				);
 				console.error(
 					chalk.red(
-						`
-Unexpected Error
-please file an issue here https://github.com/webpack/webpack-cli/issues/new?template=Bug_report.md
-				`
-					)
+						// tslint:disable-next-line: max-line-length
```
misterdev: This can be removed since we don't use tslint anymore

rishabh3112:
> What do you mean when you talk about the socket's method to replace `this.prompt`? :)

I am working for a new approach for supporting generators on UI. 
By default Yeoman class prompts to `process.stdout`. By overriding the prompt method, when using UI, I can make the same class prompt over sockets.

packages/utils/modify-config-helper.ts
rishabh3112:
```suggestion

```

rishabh3112: Did a fun thing here 🙈

ematipico:
I guess what's left is a rebase! 🎆

rishabh3112:
Done @ematipico

packages/utils/modify-config-helper.ts
```diff
@@ -87,59 +96,94 @@ export default function modifyHelperUtil(
...
-					console.error(chalk.red("\nCould not find a yeoman configuration file.\n"));
-					console.error(
-						chalk.red(
-							"\nPlease make sure to use 'this.config.set('configuration', this.configuration);' at the end of the generator.\n"
-						)
-					);
-					Error.stackTraceLimit = 0;
-					process.exitCode = -1;
+	if (!usingCLI) {
+		/**
```
evenstensberg: no need to add comment here, you will need to disable prompting and fully generate a scaffold file by running the transformation on the config object

rishabh3112: Not exactly, I meant to use `prompting` but instead of directly using `this.prompt` which internally uses inquirer questions, I meant to overload `this.prompt` with prompting on the socket.

Rest of the work, like running transformations, generate scaffold, etc. is already implemented for CLI and would be reused.

Is this approach is OK for you?

rishabh3112:
Now any generator in `UI` mode will prompt over a TCP connection :confetti_ball: .
I studied yeoman codebase for this and finally get it working smoothly.

rishabh3112:
I think this approch serves the purpose as no code is rewritten here.
@ematipico have a review when free :+1:

rishabh3112:
As UI is now a separate project, this PR is not needed anymore. Thanks for everyone who gave their reviews to this PR ❤

Issue #830, opened 2019-04-16
[cli]: Installation prompt should look for package from project root.


anshumanv:
**Describe the bug**

Installation prompt is presently searching for packages that are executed like `serve, init` etc in the `cwd` which is not ideal for cases when user is in some other dir inside root and wishes to execute the package so the lookup for whether a package is installed or not is done in the `cwd` and not in the project-root

**What is the current behavior?**
Package lookup is done in present working dir.

**To Reproduce**
`in a webpack project dir, @webpack-cli/serve is already installed`
`
cd some-dir && webpack-cli serve
`

This will throw a prompt to install the package but it is already installed 😟 

**Expected behavior**
Package lookup from project root or the nearest package location.

**Screenshots**
NA

**Please paste the results of `webpack-cli info` here, and mention other relevant information**

**Additional context**
Add any other context about the problem here like linking to an similar issue you might think is the cause.

anshumanv:
@ematipico @evenstensberg think this fix is needed?

evenstensberg:
No, we need to symlink

anshumanv:
> No, we need to symlink

Can you help me understand what changes are needed or none at all?

evenstensberg:
You will first need to look for the package in global and then local

anshumanv:
Okay! Updating the PR to look for global -> project root -> cwd

In this order

anshumanv:
@evenstensberg you sure we need to change the workflow? Right now it's looking for cwd first and then goes to global to check if a package is installed, I've proposed to check the project root -> cwd -> global which makes sense IMO, what do you think?

For the serve package it is presently initiated in cwd, but I think the project root is more ideal location to start the webpack-dev-server.

ematipico:
Changing the flow means breaking change... So major release.

anshumanv:
Yeah, I guess for now we can just make a minor improvement of starting off the server from the project root then go to the cwd, just a minor change since I think it makes more sense to start the project from it's root. @ematipico 
No changes to the core flow as per the current PR.

Issue #829, opened 2019-04-16
[serve]: serving from anywhere in the project is not functional


anshumanv:
**Describe the bug**
Presently `webpack-cli serve` requires you to be in the project root to function properly but user can be anywhere in the project and till it's a webpack project serving should work from anywhere in the project it is executed.

**What is the current behavior?**
Only works from  project root,

**To Reproduce**
Run `webpack-cli serve` from a dir inside project root.

**Expected behavior**
It should work from anywhere in the project.

**Screenshots**
NA

**Please paste the results of `webpack-cli info` here, and mention other relevant information**

**Additional context**
Add any other context about the problem here like linking to an similar issue you might think is the cause.

ematipico:
Is it really a bug or it's an enhancement/feature?

anshumanv:
> Is it really a bug or it's an enhancement/feature?

Should be a bug since the user needs to be able to run serve from any place in the project, should I send a PR? @ematipico

ematipico:
That's the thing: where is it written this requirement? Honest question, I wasn't aware of this behaviour

anshumanv:
> where is it written this requirement? 

I don't think this is stated as a requirement anywhere, but IMO this should be how it should work where user be inside any dir of the project and can serve the project.

What do you think? @ematipico 🤔

ematipico:
I think that this is not a bug because there's not a requirement and few people have complained about the current behaviour. Loads of people run their scripts from project root, less people from a sub directory. It's a nice to have!

So I'd say it's a feature because we are extending a functionality that we already have.

anshumanv:
> So I'd say it's a feature because we are extending a functionality that we already have.

Yeah works, https://github.com/webpack/webpack-cli/pull/831 aims to solve this PTAL 🙏 😄

PR #831, opened 2019-04-16
fix(serve): use project root while serving a project

anshumanv:
use project root to serve any project

ISSUES CLOSED: #829
ISSUES CLOSED: #830

bugfix

Yes

NA

**Summary**
Package lookup from project root, start webpack-dev-server from the project root.

No

**Other information**


anshumanv:
Will remove junit.xml if need be.

anshumanv:
> also introduces another dependency

It's already present and will be imported from utils once it's fixed.

Can you suggest any changes or if this should be closed? 🤔

bin/utils/prompt-command.js
```diff
@@ -53,7 +56,15 @@ module.exports = function promptForInstallation(packages, ...args) {
 	try {
 		const path = require("path");
 		const fs = require("fs");
-		pathForCmd = path.resolve(process.cwd(), "node_modules", "@webpack-cli", packages);
+		const findup = require("findup-sync");
```
evenstensberg: You cannot use the findProj root function for this. This should exactly match cwd as we cannot traverse up directories to search for a project. We need to assume that the user is in the current folder. Otherwise, people might install webpack-cli at /Desktop instead of /Desktop/myproj or something.

packages/serve/index.ts
```diff
@@ -82,6 +83,9 @@ export default function serve(): Promise<void | Function> {
 			);
 			process.exit(1);
 		}
+		// Switch the dir to project root before executing webpack-dev-server
+		process.chdir(findProjectRoot());
```
evenstensberg: No, we are not manipulating the cwd through the CLI, this is undefined user behaviour and up to the user themselves. You will need to provide the serve command context property (rootpath), which will be sent to webpack.

evenstensberg:
Hi. Thanks for the PR. I think this change is not needed for us right now, as the benefit of having it such that the current solution is better. Thanks for thanking the effort to put this together!

PR #834, opened 2019-04-18
Remove tslint in favour of eslint

ematipico:
Feature, better a chore.

At the moment we use two linters to lint the projects

N/A

Using eslint to lint all the files.

**Main changes that there are in this PR:**
- created a super generic `.eslintrc.js` file. Which we already have and I removed all the the rules regarding `eslint-plugin-node`.
- inside the `cli` folder I created a new eslint config file that inherits from the main one and it adds all the node rules that have been removed from the main config file
- inside each package I crated an eslint file that takes care of using the proper typescript parser plus and plugin. I also added the prettier configuration so it won't collide with formatting
- tests should be always inside a folder so we can have a specific eslint config file for those: by default typescript-eslint wants for every function to return the type (which is good), but when we define the tests, we always have `describe("", () => {})` which is annoying. So better disable the rule for test files.


pranshuchittora:
IMO both ESLint & TSLint have some voids. Pls read [this](https://medium.com/palantir/tslint-in-2019-1a144c2317a9)

ematipico:
Yeah I know but it's difficult to maintain two linters with two different way of configuring things. Now that eslint supports TypeScript and core members of TypeScript are helping `@typescript-eslint` I'd say it's a good thing to move to just one to help and make maintainability easier.

anshumanv:
> In order to avoid bifurcating the linting tool space for TypeScript, we therefore plan to deprecate TSLint and focus our efforts instead on improving ESLint’s TypeScript support. 

Positive change IMO. ✨

ematipico:
TypeScript interfaces are a bit messed up it's a full of duplicates. 
After this PR we should think about doing the following things: 
- remove our types about JSCodeshift and use `@types/jscodeshift`
- remove out yeoman types and use `@types/yeoman-generator`

.eslintrc.js
```diff
@@ -1,110 +1,62 @@
 module.exports = {
-	"root": true,
-	"plugins": ["node"],
-	"extends": ["eslint:recommended", "plugin:node/recommended"],
-	"env": {
-		"node": true,
-		"es6": true,
-		"jest": true
+	// "root": true,
```
evenstensberg: remove unneeded comments?

packages/generate-loader/index.ts
```diff
@@ -1,8 +1,6 @@
 import LoaderGenerator from "@webpack-cli/generators/loader-generator";
 import * as yeoman from "yeoman-environment";
 
-import { IYeoman } from "./types/Yeoman";
```
evenstensberg: Why is type removed?

.eslintrc.js
ematipico: Will do

ematipico:
Yes changes to the functions were entirely to the types, most of them it was to apply the cast. I don't why it was building honestly, there were a number of typescript issues. Anyway, they are all fixed

packages/utils/package-manager.ts
```diff
@@ -109,3 +99,12 @@ export function getPathToGlobalPackages(): string {
 
 	return require("global-modules");
 }
+
+export function spawnChild(pkg: string): SpawnSyncReturns<Buffer> {
```
anshumanv: Docstrings for this are left above.

packages/utils/validate-identifier.ts
```diff
@@ -79,17 +71,13 @@ function isKeyword(code: string): boolean {
 
 /* tslint:disable: max-line-length */
```
anshumanv: Can be removed.

packages/utils/validate-identifier.ts
```diff
@@ -79,17 +71,13 @@ function isKeyword(code: string): boolean {
...
 /* tslint:enable: max-line-length */
```
anshumanv: Remove

.gitignore
```diff
@@ -37,3 +37,4 @@ yarn.lock
 
 # source maps of docs
 docs/**/*.map
+junit.xml
```
anshumanv: thanks

PR #837, opened 2019-04-20
[init]: Do not generate config when defaults are used

anshumanv:
scaffolding layout rework

ye

NA

**Summary**
Just checked the prompts that user uses while init generator, if they match defaults do skip config generation and display a message.

No

**Other information**
This is just a starting point, this would probably need more changes so please drop in your feedback about the approach.

@evenstensberg @ematipico @dhruvdutt


packages/generators/utils/checkDefaults.ts
```diff
@@ -0,0 +1,23 @@
...
+ * Utility function to check if the user config is same as default.
+ *
+ * @param	{Object[]} initConfig 	- config generated by the user
+ * @returns {Boolean} - Whether the default condif is used or not
+ */
+
+export default function checkDefaults(initConfig: Object[]) {
+  // This is the default prompts selections generated on picking default options.
+  const defaultConfig = {
+		babelConfirm: true,
```
anshumanv: Do we need to check all prompts or just the entry and output would do? 🤔

anshumanv: @evenstensberg @ematipico @dhruvdutt is this the correct approach?

ematipico: All of them. If the user would decide to change **only** the styling config, we would need to generate the config file, even though `output` and `entry` are untouched. 

On the other hand, the check can be done at the very end. No need to check it every time, correct?

anshumanv: > All of them. If the user would decide to change only the styling config, we would need to generate the config file, even though output and entry are untouched.

Yeah, works this way rn 👍 

> On the other hand, the check can be done at the very end. No need to check it every time, correct?

yeah it is checking at the end, I'm just storing the user's pick and comparing it at the end before the project scaffolding starts. Did you face any issue while doing a test run? 🤔

ematipico: Oh I see it now! Sorry, I overlooked the overall logic! Looks good to me

packages/generators/init-generator.ts
```diff
@@ -376,6 +385,17 @@ export default class InitGenerator extends Generator {
 							this.configuration.config.webpackOptions.module.rules.push(moduleRulesObj);
 						}
 					}
+
+					// Check if the options selected by the user are defaults
+				if (checkDefault(initConfig)) {
```
evenstensberg: nice

packages/generators/init-generator.ts
```diff
@@ -81,15 +83,19 @@ export default class InitGenerator extends Generator {
 			"\n"
 		);
 
+		const initConfig = []
```
evenstensberg: You already have an object that you have access to already in two ways hint hint

anshumanv: So do I pass in the entire object, it contains a lot of other data which is a lot level nested? So I thought to pass in just the options user picked makes it a lot easier to implement the compare function and seems work pretty well(one liner) looks cleaner. WDYT? :) @evenstensberg

evenstensberg: this.configuration is better than alloc memory for all that

anshumanv: Okay then! Changing in a sec ✨ 😄

package-lock.json
```diff
@@ -7543,8 +7543,7 @@
         "ansi-regex": {
           "version": "2.1.1",
```
misterdev: Do we need to update package-lock?

packages/generators/utils/checkDefaults.ts
```diff
@@ -0,0 +1,23 @@
...
+ * Utility function to check if the user config is same as default.
+ *
+ * @param	{Object[]} initConfig 	- config generated by the user.
+ * @returns {Boolean} - Whether the default config is used or not.
+ */
+
+export default function checkDefaults(initConfig: Record<string, any>): boolean {
+	// This is the default prompts selections generated on picking default options.
+	const defaultConfig = {
+		babelConfirm: true,
```
misterdev: If you agree, I think it would be nice to have those as constants so that we can use them as default values while prompting

package-lock.json
anshumanv: It seemed to be out of sync.

packages/generators/utils/checkDefaults.ts
anshumanv: Yeah would be much cleaner that way, how about a `constants` file in the utils folder? From where we can export all constants? 🤔

misterdev: Sound good

packages/generators/init-generator.ts
anshumanv: It turns out that not all prompt answers are present in the config and therefore some memory is needed anyways, for now I've shifted the config object inside `this.configuration` and passing the config object from there to check, if we don't want this way we can store the missing things in this.configuration and then check it, but this approach looks more maintainable to me. WDYT about this?

packages/generators/init-generator.ts
```diff
@@ -81,15 +86,22 @@ export default class InitGenerator extends Generator {
...
+
 		return this.prompt([Confirm("entryType", "Will your application have multiple bundles?", false)])
 			.then(
 				(entryTypeAnswer: { entryType: boolean }): Promise<{}> => {
+					this.configuration.config.initConfig.push(entryTypeAnswer);
+					initConfig.push(entryTypeAnswer);
```
misterdev: Should we remove the `initConfig.push` if we are saving the answers to `this.configuration.config.initConfig`? :)

packages/generators/init-generator.ts
misterdev: My 2 cents, I agree with @anshumanv and I'd put the answers into `this.configuration.answers`.
Iside `this.configurations.config` we have `topScope` and `webpackConfig` that directly influence the generated webpack config file

packages/generators/init-generator.ts
anshumanv: Correct! Thanks, just pushed. :)

Yeah `this.configuration.answers` sounded better/ 👍

packages/generators/init-generator.ts
```diff
@@ -158,6 +168,7 @@ export default class InitGenerator extends Generator {
 			)
 			.then(
 				(stylingTypeAnswer: { stylingType: string }): void => {
+					this.configuration.answers.push(stylingTypeAnswer);
```
evenstensberg: No need to push to create more objects on heap. Check directly against `this.configuration`

evenstensberg:
This is great. Could you make it like explained in https://github.com/webpack/webpack-cli/issues/902 ? Thanks!

anshumanv:
@evenstensberg can you please go through - https://github.com/webpack/webpack-cli/pull/837#discussion_r279616749

Since not all are stored, I don't think having an array (of 5 elements) of chosen config outside of the `this.configuration` hurts much, how do I tackle the issues of fields which are not tracked by `this.configuration` :)

evenstensberg:
Doesn't hurt, but hurts the abstraction we're providing. You're able to use the config object to determine prod or dev, I want that to be used rather than making a clone. The answers are present dynamically at the object at runtime

evenstensberg:
@anshumanv could you work on this after the prettier PR? This is desperately needed

packages/generators/init-generator.ts
```diff
@@ -84,12 +89,15 @@ export default class InitGenerator extends Generator {
 		return this.prompt([Confirm("entryType", "Will your application have multiple bundles?", false)])
 			.then(
 				(entryTypeAnswer: { entryType: boolean }): Promise<{}> => {
```
anshumanv: this result var is not stored in `this.configuration`, so I think I'm missing something in the way you're suggesting, how do I check this against `this.configuration`?

evenstensberg: `checkDefaults(this.config......entry)` after those are set

evenstensberg:
Could you rebase?

anshumanv:
Definitely, thanks for the reminder, picking this now. 👍

packages/generators/init-generator.ts
anshumanv: I think I get it, thanks man, sounding neat, trying this out.

packages/generators/init-generator.ts
anshumanv: Cool, on it

evenstensberg:
I'll pick this up

evenstensberg:
#937 closing in favor for that, it's a lot of work that needs to be adjusted against different targets like the next branch

anshumanv:
Thanks for picking @evenstensberg, sorry for the delay ✨

Issue #840, opened 2019-04-20
[DOCUMENTATION]: Migration of docs to webpack.js.org


pranshuchittora:
Docs are something which is one of the most important references for any developer. What's worse than outdated docs :cry: .

Like these
- [Scaffolding](https://github.com/webpack/webpack-cli/blob/master/SCAFFOLDING.md) 
- [Migrate](https://github.com/webpack/webpack-cli/blob/master/MIGRATE.md)
and many more.

### Solution
The idea is to migrate/shift all the documentation to the webpack.js.org to make it the only source of truth.
Keep only those docs which are required for the development of the cli project.

**Do you want to request a *feature* or report a *bug*?**
Better codebase

**What is the current behaviour?**
Docs are spread all over the project.
 
**What is the expected behaviour?**
Migrate docs to webpack.js.org, for better maintainability.

**Pls discuss :+1:**

misterdev:
There's already a discussion on this: [#2894](https://github.com/webpack/webpack.js.org/pull/2894) :)

evilebottnawi:
bump

evilebottnawi:
We should do some improvement after refactor

alexander-akait:
`Migrate` was deprecated and removed, updating `Scaffolding` will be after https://github.com/webpack/webpack-cli/issues/2430

PR #841, opened 2019-04-20
chore(docs): removes scaffolding docs from the root

pranshuchittora:
As per #840
Removes the scaffolding docs from the root of the project folder. As it's already been maintained at webpack.js.org. [Here](https://webpack.js.org/guides/scaffolding) 

chore


anshumanv:
One thing that worries me 😟 
image

ematipico:
Something more worrisome is that in the results I can't see the other page, the one that is webpack.js.org

PR #853, opened 2019-04-22
feat(scaffold): create a webpack script on running a scaffold

anshumanv:
create a webpack script on running a scaffold

feat

yes

NA

**Summary**
* On creating a scaffold, add a webpack script to it

No

**Other information**
Not yet, do lmk if this is the right direction.
Ref - #709

### DIFF
--- a/packages/utils/add-script.ts
+++ b/packages/utils/add-script.ts
@@ -9,15 +9,16 @@ import * as fs from "fs";
  * @returns {void}
  */
 
+const PACKAGE_JSON_FILE = "package.json";
+
 export default function addScript(key: string, value: string): void {
   const localDir = process.cwd();
-  // eslint-disable-next-line
-  const packageJSON = require(path.resolve(localDir, "package.json"));
-  if (!packageJSON["scripts"]) {
-    packageJSON["scripts"] = {};
+  const packageJSON = JSON.parse(fs.readFileSync(path.resolve(localDir, PACKAGE_JSON_FILE), "utf8"));
+  if (!packageJSON.scripts) {
+    packageJSON.scripts = {};
   }
-  packageJSON["scripts"][key] = value;
-  fs.writeFile('package.json', JSON.stringify(packageJSON, null, 2), function(err): void {
+  packageJSON.scripts[key] = value;
+  fs.writeFile(PACKAGE_JSON_FILE, JSON.stringify(packageJSON, null, 2), function(err): void {
 		if (err) console.error(err);
   });
 }
--- a/packages/utils/add-script.ts
+++ b/packages/utils/add-script.ts
@@ -2,16 +2,16 @@ import { json } from "mrm-core";
 import * as path from "path";
 
 /**
- * Utility function to add a script to a package.json file given a key and a value
+ * Utility function to add a script in a package.json file given a key and a value
  *
  * @param {String} key - key of the script to be added in package.json
  * @param {String} value - command to be executed on running that script
- * @returns {Boolean} whether the string could be a path to a local file or directory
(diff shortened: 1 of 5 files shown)


packages/utils/scaffold.ts
```diff
@@ -98,6 +99,8 @@ export default function runTransform(transformConfig: TransformConfig, action: s
 				);
 		}
 	);
+	// Add a webpack script to package.json
+	addScript("webpack", "webpack")
```
evenstensberg: This script is dependant on which config you have generated. Sometimes it's prod, then you'll need `--config` flag, otherwise not

packages/utils/add-script.ts
```diff
@@ -0,0 +1,20 @@
+import { json } from "mrm-core";
```
evenstensberg: I don't think we need a dependency to do a JSON write

packages/utils/scaffold.ts
anshumanv: Refer - https://github.com/webpack/webpack-cli/issues/709

`webpack` script is to be anyhow generated, depending on the config, `prod` and `dev` scripts are generated.

packages/utils/add-script.ts
anshumanv: Removed 👍

packages/utils/add-script.ts
```diff
@@ -0,0 +1,23 @@
...
+ * Utility function to add a script in a package.json file given a key and a value
+ *
+ * @param {String} key - key of the script to be added in package.json
+ * @param {String} value - command to be executed on running that script
+ * @returns {void}
+ */
+
+export default function addScript(key: string, value: string): void {
+  const localDir = process.cwd();
+  // eslint-disable-next-line
```
anshumanv: ESLint suggests to use import in the next line but it doesn't work for some reason.

ematipico: Because the `require` it's used, usually, to resolve modules. Here you want to read a file, so you should use `fs.readFile` instead

packages/utils/add-script.ts
```diff
@@ -0,0 +1,23 @@
...
+  const packageJSON = require(path.resolve(localDir, "package.json"));
```
ematipico: `"package.json"` can be moved inside a constant

packages/utils/add-script.ts
```diff
@@ -0,0 +1,23 @@
...
+  if (!packageJSON["scripts"]) {
```
ematipico:
```suggestion
  if (!packageJSON.scripts) {
```

ematipico: If you accept this suggestion, please change also the other parts

packages/utils/add-script.ts
anshumanv: 😮 ofcourseeee

packages/utils/add-script.ts
anshumanv: Sounds great! Updated.

packages/utils/add-script.ts
anshumanv: Updated elsewhere 👍

anshumanv:
@evenstensberg @ematipico PTAL ✨

misterdev:
Hi @anshumanv, sorry for joining this now but I've just noticed that this feature is in the scope of #817. Some differences:

- #817 also generates the package.json, if it will get merged this code will become unused
- in #817 the script is called "build" instead of "webpack"

anshumanv:
Sound good @misterdev, looks like a better approach there. Happy to let you take over, closing! 😄

PR #858, opened 2019-04-25
chore(deps): lockdown versions

dhruvdutt:

**Summary**

**Other information**
Working on upgrading all the deps.


package.json
```diff
@@ -116,63 +116,63 @@
...
+    "eslint-config-prettier": "4.1.0",
+    "eslint-plugin-node": "8.0.0",
+    "esm": "3.2.14",
+    "execa": "1.0.0",
+    "husky": "1.1.4",
+    "jest": "24.3.1",
+    "jest-cli": "24.3.1",
+    "jest-junit": "6.3.0",
+    "lerna": "3.13.2",
     "lint-staged": "7.x.x",
```
ematipico: Also this one should be pinned with a specific version number

ematipico:
It is indeed, the fact that other people don't do it with their libraries doesn't mean that we can't.

PR #862, opened 2019-04-26
fix(cli): throws err when no args passed

### LINKED ISSUE #861: 🐞[cli]: throws err when no args passed.
**Describe the bug**

CLI throws an error when no args passed. Occurs only if `webpack` is not installed.

Screenshot from 2019-04-26 11-20-44

**What is the current behaviour?**
Throws an error.
**To Reproduce**
Steps to reproduce the behaviour:

1. Create a new folder.
2. `npm init -y`
3. `npm i webpack-cli`
4. ` ./node_modules/webpack-cli/bin/cli.js `

**Expected behavior**
Must provide some useful information in order to resolve the issue, instead of showing err logs.

  System:
    OS: Linux 4.18 Ubuntu 18.10 (Cosmic Cuttlefish)
    CPU: (8) x64 Intel(R) Core(TM) i5-8250U CPU @ 1.60GHz
  Binaries:
    Node: 10.12.0 - ~/.nvm/versions/node/v10.12.0/bin/node
    Yarn: 1.12.3 - /usr/bin/yarn
    npm: 6.5.0 - ~/.nvm/versions/node/v10.12.0/bin/npm
  Browsers:
    Chrome: 73.0.3683.103
    Firefox: 66.0.3
  npmPackages:
    webpack: ^4.30.0 => 4.30.0 
    webpack-cli: ^3.3.1 => 3.3.1 
  npmGlobalPackages:
    webpack-cli: 3.3.1
    webpack: 4.27.1

pranshuchittora:
Closes #861 
** Behaviour after fix **
Screenshot from 2019-04-26 11-50-10

bugfix

Nopes


bin/cli.js
```diff
@@ -71,7 +71,12 @@ For more information, see https://webpack.js.org/api/cli/.`);
 			options = require("./utils/convert-argv")(argv);
 		} catch (err) {
 			if (err.name !== "ValidationError") {
-				throw err;
+				console.error("\n\u001b[1m\u001b[31mWebpack not found, please install webpack");
```
rishabh3112:
```suggestion
				console.error("\n\u001b[1m\u001b[31mwebpack not found, please install webpack");
```
as per branding guidelines of webpack

bin/cli.js
```diff
@@ -71,7 +71,12 @@ For more information, see https://webpack.js.org/api/cli/.`);
...
+				console.error(
+					"\u001b[1m\u001b[31mAlternatively, run 'webpack(-cli) --help' for usage info.\u001b[39m\u001b[22m\n"
```
rishabh3112: Merge two `console.error` here using concatenation or template literals

bin/cli.js
ematipico: The message is misleading. What if you have actually installed webpack?

pranshuchittora: Thanks @rishabh3112. Will make the following changes.
Can you please suggest some better message for this error?

bin/cli.js
pranshuchittora: Will do that :smile:

bin/cli.js
pranshuchittora: So the behaviour is something like this.
_In both cases, webpack-cli has been installed._

1. **When you haven't installed webpack & no args passed to the CLI.**

The output is an error which is currently not been handled. :arrow_down: 

Screenshot from 2019-04-26 18-03-46

2. **When webpack is installed, but no args passed to the CLI.**

After installing webpack, if you run webpack-cli with no args. It shows this message :arrow_down: 

Screenshot from 2019-04-26 18-04-18

#### Pls, acknowledge if you can able to achieve similar behaviour.

rishabh3112: Sure you can,
```suggestion
				console.error("\n\u001b[1m\u001b[31mwebpack not found, please install webpack using\n\tnpm install --save-dev webpack");
```
P.S.
I feel that not having `ValidationError` doesn't strictly implies `webpack` is not installed.
You should check the error's corresponding `err.name` for better handling of this error.

pranshuchittora: @rishabh3112 For this scenario the `err.code` is `MODULE_NOT_FOUND`
IMO I should define a new scope(condition) for handling this particular err.

What do you think about this approach?

Thanks

rishabh3112: Yeah, you got my point now :), something like below would work IMO:
```suggestion
			if (err.name === "MODULE_NOT_FOUND" && err.message.indexOf("webpack") !== -1) {
      				console.error("\n\u001b[1m\u001b[31m\n\u001b[1m\u001b[31mwebpack not found, please install webpack using\n\tnpm install --save-dev webpack");
			}
```

pranshuchittora:
Progress so far:
- Asks the user to install webpack if webpack is not installed.

Screenshot from 2019-04-27 01-20-49

bin/cli.js
```diff
@@ -70,6 +70,22 @@ For more information, see https://webpack.js.org/api/cli/.`);
 		try {
 			options = require("./utils/convert-argv")(argv);
 		} catch (err) {
+			// When webpack is not installed and no args passed to the CLI
+			if (err.code === "MODULE_NOT_FOUND") {
+				let errorMessage =
+					"\n\u001b[31mwebpack not found, please install webpack using\n\t\u001b[33mnpm install --save-dev webpack\n";
+
+				if (process.env.npm_execpath !== undefined) {
+					if (process.env.npm_execpath.indexOf("yarn") !== -1) {
```
EugeneHlushko: I think includes is the preferred way nowadays e.g.
```suggestion
					if (process.env.npm_execpath.includes("yarn")) {
```

bin/cli.js
```diff
@@ -70,6 +70,22 @@ For more information, see https://webpack.js.org/api/cli/.`);
 		try {
 			options = require("./utils/convert-argv")(argv);
 		} catch (err) {
+			// When webpack is not installed and no args passed to the CLI
+			if (err.code === "MODULE_NOT_FOUND") {
+				let errorMessage =
+					"\n\u001b[31mwebpack not found, please install webpack using\n\t\u001b[33mnpm install --save-dev webpack\n";
+
+				if (process.env.npm_execpath !== undefined) {
```
EugeneHlushko: Not really making it more readably by having another if nested, maybe just use `&&` and second condition on newline?

bin/cli.js
anikethsaha: @pranshuchittora the last line error message in the second image about the `./src` , Is it genuine?
I mean does your config file really have `./src` mentioned in the resolve? Cause I am getting that error even though I didn't mention that path in my config file. PTAL

bin/cli.js
pranshuchittora: @EugeneHlushko  pls review the new changes.

EugeneHlushko:
Yes thats what i meant, thank you for implementing 👍

bin/cli.js
```diff
@@ -70,6 +70,20 @@ For more information, see https://webpack.js.org/api/cli/.`);
...
+				let errorMessage =
+					"\n\u001b[31mwebpack not found, please install webpack using\n\t\u001b[33mnpm install --save-dev webpack\n";
+
+				if (process.env.npm_execpath !== undefined && process.env.npm_execpath.includes("yarn")) {
+					errorMessage =
+						"\n\u001b[31mwebpack not found, please install webpack using\n\t\u001b[33myarn add webpack --dev\n";
+				}
+
+				console.error(errorMessage);
+				return;
```
rishabh3112: Set `process.exitCode = 1`

pranshuchittora:
Need help: Reason for failing CI

ematipico:
Please rebase, we removed node 6 from the CI

bin/cli.js
evenstensberg: I think the process is already exiting in this case, `return` is redundant too @pranshuchittora.

evenstensberg: Could you remove the return here and doing what we're suggesting @pranshuchittora ? After that I can do a manual test of this an verify, then we can approve and merge it

pranshuchittora: If I remove the `return`, then the following scope will throw an error, which might freak out a user.

Here is a screenshot which shows what will happen if the func won't return.
Screenshot from 2019-05-30 20-15-41

pranshuchittora: IMO we must return from the func as soon as the above condition satisfies.

evenstensberg: So to clear things out here. Assign the process exit code as @rishabh3112 points out, shorten the stack trace by https://developer.mozilla.org/en-US/docs/Web/JavaScript/Microsoft_Extensions/Error.stackTraceLimit and make more use of coloring to make the suggestion to install webpack more visible. All good

pranshuchittora: Done :+1: 

Screenshot from 2019-05-30 22-26-58

bin/cli.js
```diff
@@ -70,6 +70,22 @@ For more information, see https://webpack.js.org/api/cli/.`);
...
+
+				if (process.env.npm_execpath !== undefined && process.env.npm_execpath.includes("yarn")) {
+					errorMessage =
+						"\n\u001b[31mwebpack not found, \u001b[33mplease install webpack using\n\t\u001b[32myarn add webpack --dev\n";
+				}
+
+				console.error(errorMessage);
+				Error.stackTraceLimit = 1;
+				process.exitCode = 1;
+				return;
```
evenstensberg: you need to remove the return

bin/cli.js
```diff
@@ -70,6 +70,22 @@ For more information, see https://webpack.js.org/api/cli/.`);
 		try {
 			options = require("./utils/convert-argv")(argv);
 		} catch (err) {
+			// When webpack is not installed and no args passed to the CLI
```
evenstensberg: I think these lines speak for themselves

bin/cli.js
pranshuchittora: Done

bin/cli.js
pranshuchittora: ??

evenstensberg: Remove the comment, code is clear enough

bin/cli.js
```diff
@@ -70,6 +70,20 @@ For more information, see https://webpack.js.org/api/cli/.`);
 		try {
 			options = require("./utils/convert-argv")(argv);
 		} catch (err) {
+			if (err.code === "MODULE_NOT_FOUND") {
+				let errorMessage =
+					"\n\u001b[31mwebpack not found, \u001b[33mplease install webpack using\n\t\u001b[32mnpm install --save-dev webpack\n";
```
evenstensberg: `webpack not installed, consider installing it using npm install --save-dev webpack`

evenstensberg: Indentation seems weird here, you don't need different color schemes for the `not found` part, the `please install ` part and the installation text. Two colors is sufficient, one for the error stack and one for the npm save

pranshuchittora: Done

pranshuchittora:
Some clarification regarding the return statement.
- If the func won't return, then a frightening err message will logs out.

Screenshot from 2019-06-01 21-46-29

pranshuchittora:
@evenstensberg pls review the final changes.
#ReadyToMerge

evenstensberg:
Tests are failing, could you revise? Having a look now at how it's looking

bin/cli.js
```diff
@@ -70,6 +70,21 @@ For more information, see https://webpack.js.org/api/cli/.`);
...
+
+				if (process.env.npm_execpath !== undefined && process.env.npm_execpath.includes("yarn")) {
+					errorMessage =
+						"\n\u001b[31mwebpack not installed, consider installing it using \n\u001b[32myarn add webpack --dev\n";
+				}
+
+				console.error(errorMessage);
+				Error.stackTraceLimit = 1;
+				process.exitCode = 1;
+				return;
```
evenstensberg: I guess this is fine, as long as we're doing it in the scope of only `MODULE_NOT_FOUND`

PR #869, opened 2019-04-28
chore(types): move to @types/yeoman-generator

### LINKED ISSUE #865: [Chore]: move types to @types/yeoman-generator
At the moment we use internal interfaces for our generator classes.

We should move to [`@types/yeoman-generator`](https://www.npmjs.com/package/@types/yeoman-generator)

misterdev:
Removes custom Yeoman types in favour of `@types/yeoman-generator`

**Summary**
Closes #865 

It compiles and seems to be working, I've marked with  `// TODO` where there's some work left to do

Any feedback is appreciated :)


packages/generators/add-generator.ts
```diff
@@ -98,17 +98,19 @@ export default class AddGenerator extends Generator {
 				webpackOptions: {}
 			}
 		};
-		const { registerPrompt } = this.env.adapter.promptModule;
-		registerPrompt("autocomplete", autoComplete);
+		// TODO adapter doesn't exists
+		// const { registerPrompt } = this.env.adapter.promptModule;
+		// registerPrompt("autocomplete", autoComplete);
```
misterdev: Can anyone explain me what is this? :)

ematipico: @dhruvdutt might know something

misterdev: Some context, the type of env in [DefinitelyTyped ](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/yeoman-generator/index.d.ts#L89) is:

```
    env: {
        error(...e: Error[]): void
    };
```

misterdev: The adapter is the part that actually handler the user interaction ([doc](https://yeoman.io/authoring/integrating-yeoman.html)).
`registerPrompt` is used to define a custom Inquirer prompt ([source](https://github.com/SBoudrias/Inquirer.js/blob/master/packages/inquirer/lib/inquirer.js#L46)).

I'm going to open a PR to include those in DefinitelyTyped, I just need to understand how to write tests :)

packages/generators/addon-generator.ts
```diff
@@ -27,21 +25,20 @@ import { YeoGenerator } from "../generate-loader/types/Yeoman";
...
+const addonGenerator = (
+	prompts: Generator.Questions,
 	templateDir: string,
 	copyFiles: string[],
 	copyTemplateFiles: string[],
 	templateFn: Function
-): YeoGenerator {
-	return class AddOnGenerator extends Generator {
-		public props: InquirerScaffoldObject;
+): typeof Generator => class AddonGenerator extends Generator {
```
misterdev: Not sure about `typeof Generator`, just `Generator` doesn't work

Issue #884, opened 2019-05-16
[BUG]: 🐞init command doesnt change the entryOptions value when passed 


anikethsaha:
**Describe the bug**

**What is the current behavior?**
in `$ npx webpack-cli init` command, In the second question, it is asking for the entry point and the default is `dist` but if we write the entrypoint in the answer then also it is keeping `dist` only in the `webpack.config.js` which is generated

**To Reproduce**

1. RUN :  ` $ npx webpack-cli init  `
2.  In 1st question , select `n` and then in the second question *Which module will be the first to enter the application? [default: ./src/index]* give any path for entry like `./index.js ` or any `path or filename`
3. in the webpack.config.js file generated, it won't write any `entry` property as it is taking the `dist` which is the default one

**Expected behavior**

It should have an `entry` property with the value as the answer passed for 2nd question in a path resolved method

**Fix**
line causing the bug:
in **[this line](https://github.com/webpack/webpack-cli/blob/f0031fdd037a5ad3d5dca97b23aaac96c9edd4fb/packages/generators/init-generator.ts#L93)**
The first if statement will run always even though it is getting **`  '[file-or-pathname]'  `**  from **[here](https://github.com/webpack/webpack-cli/blob/f0031fdd037a5ad3d5dca97b23aaac96c9edd4fb/packages/generators/utils/entry.ts#L114)** as it will always return **` '' `**  and its length will be greater than 0 and type is always `string`

To **Fix** this : 
in **[ line](https://github.com/webpack/webpack-cli/blob/f0031fdd037a5ad3d5dca97b23aaac96c9edd4fb/packages/generators/init-generator.ts#L93)** change replacing the existing with this one work fine. 
```bash
		.then(
				(entryOptions: object | string): Promise<{}> => {
					if (typeof entryOptions === "string" && entryOptions.length > 0) {
						
						if (entryOptions !== "\"\"" &&  entryOptions !== "\'\'") {
							this.configuration.config.webpackOptions.entry = entryOptions;
						}
					}
			
					return this.prompt([
						Input("outputType", "In which folder do you want to store your generated bundles? (dist):")
					]);
				}
			)
```

I have tested this and it's working fine in all types of answers for that.

**Please paste the results of `webpack-cli info` here, and mention other relevant information**
```bash

  System:
    OS: Windows 10
    CPU: (4) x64 Intel(R) Core(TM) i5-6200U CPU @ 2.30GHz
  Binaries:
[...]

PR #883, opened 2019-05-12
fix(generator): using configFile in configPath to get the config file name

anikethsaha:
Passing {configFile:configFile} to the env.run(generator) as second
argument in packages/utils/modify-config-helper.ts and using this property in generator class to get the
configPath in generator class constructor in remove-generator.ts

**What kind of change does this PR introduce?**
bugfix

**Did you add tests for your changes?**
No

**If relevant, did you update the documentation?**
Not required

**Summary**
Issue link : https://github.com/webpack/webpack-cli/issues/882
In order to avoid the `typeError` error while running the remove command when passing the config file name other than `webpack.config.js` 
Added a `configFile` property while running the generator from `modify-config-helper.ts` :
```bash
	env.registerStub(generator, generatorName);

	env.run(generatorName,{
		configFile
	}).then()
```
In order to pass the user's config file name to the generator class and replace the hardcoded `webpack.config.js`  to this 
```bash
	const {configFile} = opts
		let configPath = path.resolve(process.cwd(), configFile);
		const webpackConfigExists = fs.existsSync(configPath);
```

**Does this PR introduce a breaking change?**

No

**Other information**
No need of this statement I think 
```bash

		if (!webpackConfigExists) {
			configPath = null;
			// end the generator stating webpack config not found or to specify the config
		}
```

Cause the  below statement in `modify-config-helper.ts` will take care of that
```bash
	if (webpackConfigExists) {
			process.stdout.write(
				"\n" +
					logSymbols.success +
					chalk.green(" SUCCESS ") +
					"Found config " +
					chalk.cyan(configFile + "\n") +
					"\n"
			);

		} else {
			process.stdout.write(
				"\n" +
					logSymbols.error +
					chalk.red(" ERROR ") +
					chalk.cyan(configFile) +
					" not found. Please specify a valid path to your webpack config like \n " +
					chalk.white("$ ") +
					chalk.cyan(`webpack-cli ${action} webpack.dev.js`) +
					"\n"
			);
			return;
		}
```

PR #886, opened 2019-05-21
Fix(init): init entry point fix when path is mentioned

anikethsaha:

bugfix

No

Not Required

**Summary**
fix https://github.com/webpack/webpack-cli/issues/884
Refactored the `init-genereator.ts` to change the `entry` options in the webpack config file to the one answered if the answer is not empty

NA

**Other information**
The `regex ` can be improved to more efficient with something like this
```bash
if(!(/(\'\'|\"\")/g.test(entryOptions))){
...
}
```


anikethsaha:
@evenstensberg we are getting `entryOptions` from that entry file. They are not the same I guess.

**PS** : I changed the `step to reproduce` in the issue https://github.com/webpack/webpack-cli/issues/884 as the earlier steps were not correct.

misterdev:
Hi @anikethsaha, thanks for the PR :)

#817 is already addressing some of the issues in the `init-generator` and fixes #884 too

anikethsaha:
@misterdev that one looks good. Are you sure about the particular entry property issue https://github.com/webpack/webpack-cli/issues/884 is addressing?. I need to test that but if you say that one is fine then we can go with that :)
Also, your review is pending for https://github.com/webpack/webpack-cli/pull/883 this too, PTAL 👍

misterdev:
Yeah, it fixes the problem you reported about the custom entry point not set in the configuration :)
I would keep #884 open anyway until it is fixed

anikethsaha:
@misterdev let's keep that then 👍

PR #843, opened 2019-04-21
chore: fix and refactor add generator

rishabh3112:
**What kind of change does this PR introduce?**

refactoring

**Did you add tests for your changes?**
No

**If relevant, did you update the documentation?**
N/A

**Summary**
Updating `prompting()` to use async/await

**Does this PR introduce a breaking change?**

Maybe

**Other information**
#836

PR #893, opened 2019-05-24
chore(add-generator): changed the naming of the plugin in config file

anikethsaha:
refactoring/bugfix

Yes

NA

**Summary**
Changed the naming of the current naming of the webpack plugin in the `add-generator` to a better and standard way of naming. The earlier one was creating error name when passed something like `extract-text-webpack-plugin`, it would give the plugin in config file like this

**Bug**

```bash
	plugins: new Extract() - textPlugin,

```

**With this refractor**
It will output the config file plugin with this
```bash
plugins: new ExtractTextWebpackPlugin(),

```

No

**Other information**
NA

### DIFF
--- a/packages/generators/add-generator.ts
+++ b/packages/generators/add-generator.ts
@@ -11,26 +11,11 @@ import { AutoComplete, Confirm, Input, List } from "@webpack-cli/webpack-scaffol
 
 import { SchemaProperties, WebpackOptions } from "./types";
 import entryQuestions from "./utils/entry";
-
+import { generatePluginName } from "./utils/plugins";
 import webpackDevServerSchema from "webpack-dev-server/lib/options.json";
 import webpackSchema from "./utils/optionsSchema.json";
 const PROPS: string[] = Array.from(PROP_TYPES.keys());
 
-/**
- *
- * Replaces the string with a substring at the given index
- * https://gist.github.com/efenacigiray/9367920
- *
- * @param	{String} str - string to be modified
- * @param	{Number} index - index to replace from
- * @param	{String} replace - string to replace starting from index
- *
- * @returns	{String} string - The newly mutated string
- *
- */
-function replaceAt(str: string, index: number, replace: string): string {
-	return str.substring(0, index) + replace + str.substring(index + 1);
-}
 
 /**
  *
@@ -395,11 +380,7 @@ export default class AddGenerator extends Generator {
 								(p: boolean): void => {
 									if (p) {
 										this.dependencies.push(answerToAction.actionAnswer);
-										let myPluginNameArray = answerToAction.actionAnswer.split("-")
-										for (let i = 0; i < myPluginNameArray.length; i++) {
-											myPluginNameArray[i] = replaceAt(myPluginNameArray[i], 0, myPluginNameArray[i].charAt(0).toUpperCase());
-										}
-										const pluginName = myPluginNameArray.join("")
+										const pluginName = generatePluginName(answerToAction.actionAnswer)
 										this.configuration.config.topScope.push(
 											`const ${pluginName} = require("${answerToAction.actionAnswer}")`
 										);
--- a/packages/generators/add-generator.ts
+++ b/packages/generators/add-generator.ts
@@ -12,8 +12,8 @@ import { AutoComplete, Confirm, Input, List } from "@webpack-cli/webpack-scaffol
 import { SchemaProperties, WebpackOptions } from "./types";
 import entryQuestions from "./utils/entry";
 import { generatePluginName } from "./utils/plugins";
-import webpackDevServerSchema from "webpack-dev-server/lib/options.json";
-import webpackSchema from "./utils/optionsSchema.json";
+import * as webpackDevServerSchema from "webpack-dev-server/lib/options.json";
+import * as webpackSchema from "./utils/optionsSchema.json";
 const PROPS: string[] = Array.from(PROP_TYPES.keys());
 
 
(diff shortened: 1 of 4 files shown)


packages/generators/add-generator.ts
```diff
@@ -395,15 +395,11 @@ export default class AddGenerator extends Generator {
...
-										const normalizePluginName = answerToAction.actionAnswer.replace(
-											"-webpack-plugin",
-											"Plugin"
-										);
-										const pluginName = replaceAt(
-											normalizePluginName,
-											0,
-											normalizePluginName.charAt(0).toUpperCase()
-										);
+										let myPluginNameArray = answerToAction.actionAnswer.split("-")
```
ematipico: Could you extract this new logic into a separate function and test it against few different plugins? There are the native ones too `webpack.DefinePlugin`.

rishabh3112:
Thanks @anikethsaha,
Some Bug fixes and refactoring of add generator is underway at #843. As add generator is written in async/await now, this change may need a rework, else will cause merge conflict.

anikethsaha:
@rishabh3112 true!
I guess I can wait for that PR to merge , Then I will rebase this and resolve the conflict manually if this is fine!...what say?

anikethsaha:
@ematipico yeah I can do that.
Actually I tested with those , it will simply change the first most char of that name. 
If you want I can add it in `utils` then can add tests for that. Is it fine?

packages/generators/add-generator.ts
evenstensberg: +1, this could quickly turn into a mess

evenstensberg: ^ What I mean by that is that like @ematipico is saying, is that it might be hard to verify that this works unless we test it thoroughly. A better thing to do, would be to abstract the logic so we can run unit tests on it and make sure that it is working as expected.

anikethsaha: Okay, So Should I add it in `utils` and write tests for it. If it's fine? @evenstensberg @ematipico

evenstensberg: Probably nice idea, or to add it within the generator itself if it's only used here

anikethsaha: @evenstensberg  Yeah....I was talking about the `utils` folder inside the `generators` package. Will add it there with test then

evenstensberg: That's a great location. Go for it

anikethsaha: okay 👍

anikethsaha:
cc @ematipico @evenstensberg , I made those changes. PTAL

packages/generators/utils/plugins.ts
```diff
@@ -43,8 +43,11 @@ export const replaceAt = (str: string, index: number, replace: string) : string
 export const generatePluginName = (rawPluginName: string): string => {
 	let myPluginNameArray : string[];
 	myPluginNameArray = rawPluginName.split("-");
-	for (let i = 0; i < myPluginNameArray.length; i++) {
-		myPluginNameArray[i] = replaceAt(myPluginNameArray[i], 0, myPluginNameArray[i].charAt(0).toUpperCase());
+	if( myPluginNameArray.length <= 1 ){}
```
evenstensberg: is the indenting here correct?

anikethsaha: there was no linting error in my editor...okay I will check again

evenstensberg: `if( myPluginNameArray.length <= 1 ){}`
```suggestion
	if( myPluginNameArray.length <= 1 ){}
```

anikethsaha: okay...I will change it .

anikethsaha: done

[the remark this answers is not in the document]
anikethsaha: like this  ? 
```bash
	for (let i = 0;( i < myPluginNameArray.length && myPluginNameArray.length > 1) ; i++) {
			myPluginNameArray[i] = replaceAt(myPluginNameArray[i], 0, myPluginNameArray[i].charAt(0).toUpperCase());
		}

```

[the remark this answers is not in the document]
evenstensberg: yeah that works but without `(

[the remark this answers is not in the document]
anikethsaha: yeah sure...okay i will update with these....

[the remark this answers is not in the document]
anikethsaha: done

packages/generators/utils/plugins.ts
```diff
@@ -6,6 +7,44 @@
...
+ *
+ * @returns	{String} string - the webpack standard plugin name
+ *
+ */
+
+
+export const generatePluginName = (rawPluginName: string): string => {
+	let myPluginNameArray : string[];
+	myPluginNameArray = rawPluginName.split("-");
+	for (let i = 0; i < myPluginNameArray.length && myPluginNameArray.length > 1 ; i++) {
```
evenstensberg: Great. So the next step now is to extract the length of the array into a variable so that GC will be better. The v8 compiler engine by Google recommends this because the optimizations at parse time will be improved, so there's no reason not to change this. 

You simply declare `const pluginArrLength = myPlufinName.length` and use that to compare

https://stackoverflow.com/questions/6973942/is-optimizing-javascript-for-loops-really-necessary

anikethsaha: done @evenstensberg.

PR #909, opened 2019-06-03
chore(utils): separated interfaces from the core module

pranshuchittora:
Separated the interfaces to a separate file to make the code cleaner

refactoring

nopes

no

Nopes


packages/utils/types/Config.ts
```diff
@@ -0,0 +1,15 @@
+export interface Config extends Object {
+	item?: {
+		name: string;
+	};
+	topScope?: string[];
+	configName?: string;
+	merge: object;
```
rishabh3112:
```suggestion
	merge: string|string[];
```
We have now changed the `merge` functionality at #797, and this is the type we need now.

Issue #906, opened 2019-06-02
[BUG]: 🐞 webpack-cli add wont lookup for webpack's native plugin in plugins options


anikethsaha:
**Describe the bug**
When using the `add` command and then selecting `plugins` to add and giving webpack's native plugin like `webpack.DefinePlugin` , it will throw error with saying no plugin exists.

**What is the current behavior?**
`webpack-cli add` wont add webpack's native plugin
The error will throw
```bash
? What property do you want to add to? plugins
? What do you want to add to plugins? webpack.DefinePlugin
webpack.DefinePlugin doesn't exist on NPM or is built in webpack, please check for any misspellings.
```

**To Reproduce**
1. `npx webpack-cli add`
2.  in plugins options
3. Give any webpack's native plugin like `webpack.DefinePlugin`
4. It will throw  
> webpack.DefinePlugin doesn't exist on NPM or is built in webpack, please check for any misspellings.

**Fix**
We can have a curated list of all the native plugins and then we can look up to that list and the `npmExist` method before adding the plugin

**Expected behavior**
Should add even webpack's native plugin

**Please paste the results of `webpack-cli info` here, and mention other relevant information**
```
  System:
    OS: Windows 10
    CPU: (4) x64 Intel(R) Core(TM) i5-6200U CPU @ 2.30GHz
  Binaries:
    Yarn: 1.15.2 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 6.9.0 - C:\Program Files\nodejs\npm.CMD
  npmPackages:
    webpack-cli: ^3.3.2 => 3.3.2

```

evenstensberg:
feel free to open a pr

anikethsaha:
@evenstensberg  sure but I couldn't find much of native plugins other than the mention. Do you have any list or can you mention where I can find them?
I checked at the docs and at webpack repo too. Please mention if I missed any!

evenstensberg:
We're searching through the available plugins that are hooked up on the webpack object. You could dig and debug yourself to figure out?

anikethsaha:
okay I will try it out at best......

anikethsaha:
there are few in here https://github.com/webpack/webpack/tree/master/lib 
atleast we can start with them ...what say ? @evenstensberg

evenstensberg:
That’s where we are finding them from, the logic is already present

anikethsaha:
👍 okay

anikethsaha:
[...]

PR #910, opened 2019-06-04
fix(generator): fixed the support of native plugins in add command

anikethsaha:
Refactored the code to fix the error of supporting the native plugin in
add-generator. removed the toLowerCase() method, added the logic for
more general plugin name from the input

bugfix/refactoring

NA
NA
**Summary**
 fix https://github.com/webpack/webpack-cli/issues/906
fixed support for adding  native plugin in add-generator

No

**Other information**
NA

### DIFF
--- a/packages/generators/add-generator.ts
+++ b/packages/generators/add-generator.ts
@@ -291,6 +291,13 @@ export default class AddGenerator extends Generator {
 					 * find the names of each natively plugin and check if it matches
 					 */
 					if (action === "plugins") {
+						let answeredPluginName = answerToAction.actionAnswer
+						let isPrefixPresent = /webpack./.test(answeredPluginName);
+						if( isPrefixPresent ){
+							answeredPluginName = answeredPluginName.replace("webpack.","").trim()
+						}else{
+							answeredPluginName = answeredPluginName.trim()
+						}
 						const pluginExist: string = glob
 							.sync(["node_modules/webpack/lib/*Plugin.js", "node_modules/webpack/lib/**/*Plugin.js"])
 							.map(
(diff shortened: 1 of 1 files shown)


packages/generators/add-generator.ts
```diff
@@ -300,7 +308,7 @@ export default class AddGenerator extends Generator {
 										.pop()
 										.replace(".js", "")
 							)
-							.find((p: string): boolean => p.toLowerCase().indexOf(answerToAction.actionAnswer) >= 0);
```
evenstensberg: Don't remove the lowercase here, it's needed because if the user just types we will install the wrong plugin

anikethsaha: I found that this `toLowerCase` was one of the main issues, 
it will convert the `DefinePlugin` to `defineplugin` and then in `glob.sync` method..this plugin won't be addressed as in the directory its name as `DefinePlugin` only.

evenstensberg: it will check indexof, so if you start typing `define` it will convert to `DefinePlugin` since we're not assigning the lowercase. Important not to strict typecheck here bc of user experience. Ideally should have substring matching instead

anikethsaha: I didnt get that.
suppose we have this array 
```
let x = ["DefinePlugin", "ProgessPlugin"]
```
Now here are the output with/without `toLowerCase`
```bash
x.map(p => p.toLowerCase()).indexOf("DefinePlugin")
// -1
x.map(p => p).indexOf("DefinePlugin")
// 1
x.map(p => p.toLowerCase()).indexOf("defineplugin")
// 1
x.map(p => p).indexOf("defineplugin")
// -1
```
I guess we should replace it with regex , what say @evenstensberg

Issue #908, opened 2019-06-03
Feature: allow related field in generator questions


rishabh3112:
**Is your feature request related to a problem? Please describe.**
For UI, we want to show the user that to which property in webpack config is this question-related to. As the logic is implemented in generators itself, we can't fill this property from outside. 

**Describe the solution you'd like**
Questions from the generator to have a `related` field

**Describe alternatives you've considered**
No alternatives IMO.

**Additional context**
Will implement once decided

rishabh3112:
UI label can be added to the issue for visibility.

anikethsaha:
Something like property name along with the description?

rishabh3112:
Yes up to some extent, I am mainly looking to have a general opinion and thus standardize this field across the webpack-scaffold ecosystem. As CLI should be consistent with the behaviour of UI and vice-versa, thus CLI should also provide this with the question in an appropriate way if this has to there.
So, implementation after having the final consensus.

anikethsaha:
LGTM. 
ok, so what I understood *(please correct me If I am wrong)* is that a `related` field in `webpack-scaffold` 's type of question.
but we need to make sure it doesn't break the types. yeoman-generator's prompt method take the question of type `inquirer.Question` so I think If adding an extra field (`related`) in the webpack-scaffold's question objects may break it.
Or do you have a plan for a different approach?

evenstensberg:
This might be hard as we're asking things such as "are you gonna use SASS", which in this case wouldn't be equivalent to `module.rules`. I don't think we would need to provide this mapping, but a mapping is needed, which is the `.yo-rc` file

rishabh3112:
Thanks!! I was looking for this so as to provide a insight on what this question do thus scaffolding wouldn't be a complete black box for user.
Can you explain a bit more on `.yo-rc` file and it's usecase here ?

evenstensberg:
The yo file is a reflection of a webpack configuration, so there's a direct mapping there

rishabh3112:
Okay, you meant to give a overview to user at the end of scaffolding?

evenstensberg:
> For UI, we want to show the user that to which property in webpack config is this question-related to. As the logic is implemented in generators itself, we can't fill this property from outside.

We can, because we already have a config to use `.yo-rc`. So we have a mapping here

rishabh3112:
I am asking for providing the details during prompting, because `.yo-rc` is generated at the end of execution of generator.
[...]

PR #932, opened 2019-06-05
[Not to be merged] explain usage of related field

rishabh3112:
feature proposal
N/A
N/A
**Summary**
Explained in comments in init-generator

N/A

**Other information**
Issue #908

### DIFF
--- a/packages/generators/init-generator.ts
+++ b/packages/generators/init-generator.ts
@@ -142,11 +142,18 @@ export default class InitGenerator extends Generator {
 		}
 
 		const { langType } = await this.prompt([
+			// At this point of time we don't know that it will be adding a plugin
+			// And `this.configuration.config` would only have changes corressponding to previous question
+			// So we can't point user what this question will be doing
 			List("langType", "Will you use one of the below JS solutions?", [LangType.ES6, LangType.Typescript, "No"])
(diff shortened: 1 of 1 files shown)


rishabh3112:
cc @evenstensberg

evenstensberg:
Submit a draft PR next time

packages/generators/init-generator.ts
```diff
@@ -142,11 +142,18 @@ export default class InitGenerator extends Generator {
 		}
 
 		const { langType } = await this.prompt([
+			// At this point of time we don't know that it will be adding a plugin
```
evenstensberg: We don't, but we can write code on client to handle this. Other than that, I think that making a scaffold starter in specific for UI would do good

rishabh3112: Okay makes sense, than what about starters available at npm ? Their's generator can't be edited by us for that info.

evenstensberg: No, but if they fill inn say for instance `typescript`, then the author would add the loaders etc needed in the background and since we're reflecting the webpack config object about to be generated at UI, this is fine

rishabh3112: Okay, then maybe we can chill this feature untill we have final designs and the migration of features is completed.

rishabh3112:
Thanks!

Issue #870, opened 2019-04-29
Regression in version 3.3.1


sbrunner:
**Describe the bug**

Regression on vwersion 3.3.1

**What is the current behavior?**

I get the following error:
```
     + 1 hidden asset
    Entrypoint undefined = desktop.html
    [0] ../html-webpack-plugin/lib/loader.js!/src/geoportal/testgeomapfish_geoportal/static-ngeo/js/apps/desktop.html.ejs 15.7 KiB {0} [built]
    [2] ../webpack/buildin/global.js 177 bytes {0} [built]
    [3] ../webpack/buildin/module.js 511 bytes {0} [built]
    [4] /src/geoportal/testgeomapfish_geoportal/static-ngeo/js/apps/image/favicon.ico 64 bytes {0} [built]
    [5] /src/geoportal/testgeomapfish_geoportal/static-ngeo/js/apps/image/logo.png 61 bytes {0} [built]
    [6] /src/geoportal/testgeomapfish_geoportal/static-ngeo/js/apps/image/background-layer-button.png 80 bytes {0} [built]
        + 1 hidden module
Child html-webpack-plugin for "iframe_api.html":
     3 assets
    Entrypoint undefined = iframe_api.html
    [0] ../html-webpack-plugin/lib/loader.js!/src/geoportal/testgeomapfish_geoportal/static-ngeo/js/apps/iframe_api.html.ejs 4.62 KiB {0} [built]
    [2] ../webpack/buildin/global.js 177 bytes {0} [built]
    [3] ../webpack/buildin/module.js 511 bytes {0} [built]
    [4] /src/geoportal/testgeomapfish_geoportal/static-ngeo/js/apps/image/favicon.ico 64 bytes {0} [built]
    [5] /src/geoportal/testgeomapfish_geoportal/static-ngeo/js/apps/image/background-layer-button.png 80 bytes {0} [built]
        + 1 hidden module
Child html-webpack-plugin for "mobile.html":
     3 assets
    Entrypoint undefined = mobile.html
    [0] ../html-webpack-plugin/lib/loader.js!/src/geoportal/testgeomapfish_geoportal/static-ngeo/js/apps/mobile.html.ejs 9.76 KiB {0} [built]
    [2] ../webpack/buildin/global.js 177 bytes {0} [built]
    [3] ../webpack/buildin/module.js 511 bytes {0} [built]
[...]

PR #935, opened 2019-06-06
fix: improve checking file permission

pastak:

improvement checking permission on open collective prompt EPERM error.

**Summary**

This error is related on https://github.com/webpack/webpack-cli/issues/870 and fixed https://github.com/webpack/webpack-cli/pull/876 but It looks not enough.

For example, It is happen EPERM when node_modules is owned by `root` and webpack-cli is executed by user excepted `root`.

It should check file and process owners are same because updating utime requires it.


pastak:
Thanks for your maintenance and quick response.

PR #942, opened 2019-06-07
feat: better defaults (0CJS) with config generation

evenstensberg:
Better default settings for init generator

N/A
N/A
**Summary**
Sets better defaults if user is in production and if the user is using defaults.

N/A

**Other information**

After this is done, we can use the default config generated from `webpack init --auto` and use that as the default config for a webpack instance. After doing so, we can remove the `isUsingDefaults` logic from the generator (or most of it) and put it as a non-config requirement in the core cli.

## TODO:

- Prefetching
- other things that might come up
- More aggressive chunksplitting strategies on specialized builds


evenstensberg:
We can make it offline support + PWA here as well. Will implement

PR #948, opened 2019-06-10
chore: added await in order to resolve the pending promise

anikethsaha:
bugfix/refactoring

NA

NA

**Summary**
Added `await` in order to resolve the pending promise which was resulting in prompting error for 3rd question in init command

NA

**Other information**
NA

### DIFF
--- a/packages/generators/init-generator.ts
+++ b/packages/generators/init-generator.ts
@@ -118,7 +118,7 @@ export default class InitGenerator extends Generator {
 			this.configuration.config.webpackOptions.entry = entryOption;
 		}
 
-		let { outputDir } = Input(
+		let { outputDir } = await Input(
 			self,
 			"outputDir",
 			"In which folder do you want to store your generated bundles?",


packages/generators/init-generator.ts
```diff
@@ -133,6 +133,8 @@ export default class InitGenerator extends Generator {
 				chunkFilename: "'[name].[chunkhash].js'",
 				filename: "'[name].[chunkhash].js'"
 			};
+		} else {
```
evenstensberg: remove this

anikethsaha: but what about the output directory then ?....if it is specified by the user then shouldn't it be added ?

evenstensberg: If not specified we are setting a default one in webpack-cli no need to add it to config

anikethsaha: What about when its specified?

evenstensberg: I think we're setting that prop further up the Lines.

anikethsaha: I guess it was not adding any, ok will test this particular one more time and will get back here

anikethsaha: @evenstensberg I checked with when specified the output directory, `output.path` not reflecting in the config file, its simply giving the chunk file support.

evenstensberg: I'm working on these defaults next so leave them to me, need to check in with some folks and speak to various of people to set the correct defaults

anikethsaha: sure then!!

anikethsaha:
yeah, messed up the rebasing. trying to fix it. 👍

packages/generators/init-generator.ts
anikethsaha: done

PR #954, opened 2019-06-11
chore: v3.3.4

evenstensberg:
CI Test Run for v.3.3.4


ematipico:
It's `v3.3.4`

evenstensberg:
Gotcha, will tag it as that

ematipico:
Also, from the change log, looks like there are new features. Should it be `3.4.0`?

evenstensberg:
No, those are for monorepo pkgs, hard to maintain a changelog for each of those, maybe we should introduce a new flag in the changelog generator called `mono:`, wdyt?

PR #955, opened 2019-06-11
docs(readme): adds deprecated warning and commands

pranshuchittora:
adds deprecated commands warning

docs


README.md
```diff
@@ -57,6 +57,12 @@ Supporting developers is an important task for webpack CLI. Thus, webpack CLI pr
 -   [`webpack-cli generate-loader`](./packages/generate-loader/README.md#webpack-cli-generate-loader) - Initiate new loader project.
 -   [`webpack-cli serve`](./packages/serve/README.md#webpack-cli-serve) - Use webpack with a development server that provides live reloading.
 
+> Deprecated commands as of v3.3.3
```
sendilkumarn: Would it make sense to explicitly say `Deprecated commands since v3.3.4` here

anikethsaha: I guess the current is fine. but it would be good if we say which version still supports those command. What you think ?

rishabh3112: Deprication refers that these commands will be removed subsequent versions. But as these commands have been removed already w/o a deprication notice, we must write something like `Removed Commands` IMO

evenstensberg: Depreciated is ok

evenstensberg: Might do a range instead <= 3.3.3

anikethsaha: Yeah....this one can be done 👍

sendilkumarn: @anikethsaha Yeah, the intention is to make it easier for people to catch the change. Most of them (including me) will just skim the readme at first. It will be really catchy if we have specified the version in which the change came into effect.

I like @evenstensberg suggestion too.

evenstensberg: Yes, let's use 'since'

ematipico: If they have been removed, they are removed, not deprecated.

evenstensberg: We decided on `since` right?

evenstensberg:
Could you address changes here and rebase?

ematipico:
Are they removed or deprecated?

evenstensberg:
Removed

PR #957, opened 2019-06-11
chore: changed the package manager condition

### LINKED ISSUE #953: [Enhancement]: Refactoring the prompt-command.js package manager check
**Is your refactoring related to a problem? Please describe.**
Currently, the package manager check is done using this condition
```javascript 
const isYarn = fs.existsSync(path.resolve(process.cwd(), "yarn.lock"));
// then
if (isYarn){ ... }
```

it can be replaced with this 
```javascript
process.env.npm_execpath !== undefined && process.env.npm_execpath.includes("yarn")
```

**Why**
it will work fine(better) when this package is installed globally and a user is in the fresh project folder with no lock file
its been used in `cli.js` so I guess it can be implemented here too

```javascript
process.env.npm_execpath !== undefined && process.env.npm_execpath.includes("yarn")
```

The current one

**Additional context**
NA

anikethsaha:
refactoring

NA

NA

**Summary**
closes #953 
In the issue, I mentioned that 
```javascript
fs.existsSync(path.resolve(process.cwd(), "yarn.lock")) 

// to changed with 

process.env.npm_execpath !== undefined && process.env.npm_execpath.includes("yarn")

// in the prompt-command.js
```
but I did the opposite as that `process.env.npm_execpath` has most chance to result in `undefined`

NA

**Other information**
NA

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -80,7 +80,10 @@ For more information, see https://webpack.js.org/api/cli/.`);
 					instructions = `Install webpack to start bundling: \u001b[32m\n  $ npm install --save-dev ${moduleName}\n`;
 					const path = require("path");
 					const fs = require("fs");
-					if (fs.existsSync(path.resolve(process.cwd(), "yarn.lock"))) {
+					const isYarn =
+						fs.existsSync(path.resolve(process.cwd(), "yarn.lock")) ||
+						(process.env.npm_execpath !== undefined && process.env.npm_execpath.includes("yarn"));
+					if (isYarn) {
 						instructions = `Install webpack to start bundling: \u001b[32m\n $ yarn add ${moduleName} --dev\n`;
 					}
 					Error.stackTraceLimit = 1;
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -4,7 +4,8 @@
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
-
+const path = require("path");
+const fs = require("fs");
 const { NON_COMPILATION_ARGS } = require("./utils/constants");
 
 (function() {
@@ -78,8 +79,6 @@ For more information, see https://webpack.js.org/api/cli/.`);
 				if (moduleName === "webpack") {
 					errorMessage = `\n${moduleName} not installed`;
 					instructions = `Install webpack to start bundling: \u001b[32m\n  $ npm install --save-dev ${moduleName}\n`;
-					const path = require("path");
-					const fs = require("fs");
 					const isYarn =
 						fs.existsSync(path.resolve(process.cwd(), "yarn.lock")) ||
 						(process.env.npm_execpath !== undefined && process.env.npm_execpath.includes("yarn"));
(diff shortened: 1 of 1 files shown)


bin/cli.js
```diff
@@ -78,8 +78,9 @@ For more information, see https://webpack.js.org/api/cli/.`);
 				if (moduleName === "webpack") {
 					errorMessage = `\n${moduleName} not installed`;
 					instructions = `Install webpack to start bundling: \u001b[32m\n  $ npm install --save-dev ${moduleName}\n`;
-
-					if (process.env.npm_execpath !== undefined && process.env.npm_execpath.includes("yarn")) {
+					const path = require("path");
+					const fs = require("fs");
+					if (fs.existsSync(path.resolve(process.cwd(), "yarn.lock"))) {
```
ematipico: You can't use `existsSync` because it's not available on node `6.0.0` (check the history dropdown => https://nodejs.org/api/fs.html#fs_fs_existssync_path)

anikethsaha:
:laughing: :laughing:

bin/cli.js
evenstensberg: Have you tested this against yarn plug n play?

anikethsaha: No !!! , never used yarn pnp. will do it then

evenstensberg: Nice

sendilkumarn: This is not 100% fool-proof either. I think the better is use both the checks

anikethsaha: yeah, we can add both too !

anikethsaha:
I couldn't able to test with yarn pnp, so if any1 would like to test against it would be great. Though implied both, may not require that test now I guess. 
cc @evenstensberg

bin/cli.js
evenstensberg: Yeah let's go for that

anikethsaha: done already in the last commit  !!! 👍

evenstensberg:
Run docker

anikethsaha:
Can't do with docker, I am on my windows and I tried earlier too to add docker and it failed. It would be great if anyone can do that test about yarn pnp

anikethsaha:
These two should be add here too https://github.com/webpack/webpack-cli/blob/08edf5766afc3201689eeaf56cfaaad567a22633/bin/utils/prompt-command.js#L72 .

bin/cli.js
```diff
@@ -78,8 +78,12 @@ For more information, see https://webpack.js.org/api/cli/.`);
 				if (moduleName === "webpack") {
 					errorMessage = `\n${moduleName} not installed`;
 					instructions = `Install webpack to start bundling: \u001b[32m\n  $ npm install --save-dev ${moduleName}\n`;
-
-					if (process.env.npm_execpath !== undefined && process.env.npm_execpath.includes("yarn")) {
+					const path = require("path");
```
evenstensberg: put them on top level for now. Bad for runtime perf but good for readability

anikethsaha: done 👍

anikethsaha: What about this? https://github.com/webpack/webpack-cli/pull/957#issuecomment-501775706

anikethsaha:
Closing this as the interface has been changed in the new release

PR #958, opened 2019-06-11
chore(packages): lock dependencies versions

dhruvdutt:

**Summary**

No

**Other information**


dhruvdutt:
Yes, both are working fine.

Screenshot 2019-06-12 at 6 31 51 PM

dhruvdutt:
@evenstensberg Let me know if anything else is needed. Let's get this merged. 🚀

milesj:
Why was this change allowed? Now all downstream consumers are going to have duplicates of all their modules since these are fixed versions.

ematipico:
Mainly because of this event: https://snyk.io/blog/malicious-code-found-in-npm-package-event-stream/

We believe all libs should lock their versions of the dependencies.

evenstensberg:
Bazel and Angular is doing the same thing, this is ok

dhruvdutt:
`webpack`, `webpack-cli` are anyways going to be `devDependencies` for most applications so it shouldn't matter anyway.

milesj:
@ematipico That's nice and all but that can easily be avoided without having to lock all dependencies.

You're now in the scenario where this package will *never* receive any bug fixes without the package dependencies being manually updated. So unless you're going to do a dependency audit on a weekly basis, this does more harm than good.

ematipico:
> @ematipico That's nice and all but that can easily be avoided without having to lock all dependencies.

How?

milesj:
@evenstensberg Unless it's part of a build process, most of their packages in the GitHub source use `^` versions. Same goes for React, Vue, Babel, and more.

@ematipico I could say the same thing about pinned versions... How do you know the version you pinned to is safe and secure? What if it's revealed a month later to also be problematic? This argument goes both ways but is also a personal opinion. So I suppose this discussion is over.

ematipico:
React and Vue don't have direct dependencies.

Anyway feel free to terminate the discussion, although I asked you what's your alternative solution!

The event-stream issue was caused because a person took advantage of unpinned dependencies (of dependencies, etc.) and people lost money! What's the less evil?

milesj:
@ematipico 

- Not use NPM packages for such minor functionality.
- Rely only on NPM packages from more credible owners (like orgs, foundations, etc).
- Pin untrusted packages, not all of them.

I'm hoping with the new GitHub registry, a lot of these problems will go away as the artifacts now sit alongside the source code and are verified and signed before publishing. Bad actors will always be thing but being entirely proactive isn't a perfect solution.

PR #966, opened 2019-06-19
fix: removed process.getuid method as its not supported in winos

### LINKED ISSUE #962: Webpack crushes when tries to print a message about donating in Windows 8.1

Webpack-cli crushes:
```
.\node_modules\webpack-cli\bin\cli.js:352

if (!e && fileOwnerId === process.getuid()) utimesSync(openCollectivePath, now, now);

TypeError: process.getuid is not a function
```

**To Reproduce**
1. run `npx webpack --config webpack.config.js --mode development` on Monday

**Expected behavior**
It should not to stop a webpack process

**Versions**
webpack: 4.33.0
webpack-cli: 3.3.3
nodejs: 12.4.0
OS: Windows 8.1

anikethsaha:
bugfix

NA

NA

**Summary**
fixes #962 

No

**Other information**
So here is the thing
As far as I looked, windows seem to give `users sid` for a user and I guess there is no support for that to check with any file/folders as they don't keep that info with it 
*(maybe ! please comment if I am wrong about it)* 
So there is no way to check them with the file. the ` sid ` looks weird too.
```bash
H:\testing>WMIC useraccount get sid
SID
S-1-5-XX-352825XXXX-XXXXXXXXX-XXXXXXXXX-XXX
S-1-5-XX-352825XXXX-XXXXXXXXX-XXXXXXXXX-XXX
S-1-5-XX-352825XXXX-XXXXXXXXX-XXXXXXXXX-XXX
S-1-5-XX-352825XXXX-XXXXXXXXX-XXXXXXXXX-XXX
S-1-5-XX-352825XXXX-XXXXXXXXX-XXXXXXXXX-XXX
S-1-5-XX-352825XXXX-XXXXXXXXX-XXXXXXXXX-XXX
```
Also in windows as far as I know **`statSync("anyfile-or-folder").uid`**  will always be **0**
So I think removing that check seems fare or can add try-catch, at least it will be there for OSX users 
```javascript
try {
    if(fileOwnerId === process.getuid()){
        // actions code here , setting the utime
    }
} catch (error) {
    // setting the utime -- here
}
```
or no need for try-catch cause any way the timestamp needs to update in that file.

what say @evenstensberg @ematipico

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -352,7 +352,13 @@ For more information, see https://webpack.js.org/api/cli/.`);
 							// On windows we need to manually update the atime
 							// Updating utime requires process owner is as same as file owner
 							access(openCollectivePath, constants.W_OK, e => {
-								if (!e) utimesSync(openCollectivePath, now, now);
+								if (process.platform === "darwin" || process.platform === "linux") {
+									const fileOwnerId = stat.uid;
+									if (!e && fileOwnerId === process.getuid())
+										utimesSync(openCollectivePath, now, now);
+								} else {
+									if (!e) utimesSync(openCollectivePath, now, now);
+								}
 							});
 						}
 					}
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -349,16 +349,10 @@ For more information, see https://webpack.js.org/api/cli/.`);
 						const timeSinceLastPrint = now.getTime() - lastPrintTS;
 						if (timeSinceLastPrint > SIX_DAYS) {
 							require(openCollectivePath);
-							// On windows we need to manually update the atime
-							// Updating utime requires process owner is as same as file owner
 							access(openCollectivePath, constants.W_OK, e => {
-								if (process.platform === "darwin" || process.platform === "linux") {
-									const fileOwnerId = stat.uid;
-									if (!e && fileOwnerId === process.getuid())
-										utimesSync(openCollectivePath, now, now);
-								} else {
+								try {
 									if (!e) utimesSync(openCollectivePath, now, now);
-								}
+								} catch (err) {}
 							});
 						}
 					}
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -345,7 +345,6 @@ For more information, see https://webpack.js.org/api/cli/.`);
 						const { access, constants, statSync, utimesSync } = require("fs");
 						const stat = statSync(openCollectivePath);
 						const lastPrint = stat.atime;
-						const fileOwnerId = stat.uid;
 						const lastPrintTS = new Date(lastPrint).getTime();
 						const timeSinceLastPrint = now.getTime() - lastPrintTS;
 						if (timeSinceLastPrint > SIX_DAYS) {
@@ -353,7 +352,7 @@ For more information, see https://webpack.js.org/api/cli/.`);
 							// On windows we need to manually update the atime
 							// Updating utime requires process owner is as same as file owner
 							access(openCollectivePath, constants.W_OK, e => {
-								if (!e && fileOwnerId === process.getuid()) utimesSync(openCollectivePath, now, now);
+								if (!e) utimesSync(openCollectivePath, now, now);
 							});
 						}
 					}


rseeberg:
This pullrequest should not be approved.
While it does fix the Windows monday issue by removing the fileowner check,
it reintroduces the original issue that @pastak fixed for linux & mac
in https://github.com/webpack/webpack-cli/commit/de413515ed0bba4e72e24e758f6701a5c0ec0886

The main problem is that if you try to update a file that you are not the owner of,
you get an error.

I don't think there's any great way of getting the fileowner (or process owner) in node on windows.
You could probably wire something up in powershell, but spawning a child_process in order to update a timestamp to control a donation message.. ludicrous!

As a user of your software, i would prefer you just print the message on every build.
Then you can remove all of this logic. IT might even speed up our builds ever so slightly, not having to do IO look ups..

anikethsaha:
> I would prefer you just print the message on every build.

This is actually very disturbing for some users as many others may be using webpack-cli as their dependency and if they install that package this message will pop up every time. The Current approach is the workaround to save that. 
as far as https://github.com/webpack/webpack-cli/commit/de413515ed0bba4e72e24e758f6701a5c0ec0886 is concerned I will refactor the current PR to check for the file-owner when its mac or Linux and to skip it when its windows.

rseeberg:
@anikethsaha
Looking better now :+1:

Still, for windows users, you can get the EPERM issue that @pastak originally attempted to fix.
This happens because you try to update the utime on a file you are not the owner of.
Since you have no way of checking if it will work or not, I don't see any way around wrapping the
`if (!e) utimesSync(openCollectivePath, now, now);`
in a try catch
`if(!e) try { utimesSync(openCollectivePath, now, now); } catch() { /*utime update failed due to permission issues */ }`

anikethsaha:
yeah true! thats why I dropped the idea for try-catch , I guess for windows we cant get any check for uid for file cause I guess windows doesnt keep any info on its file or folder I think

anikethsaha:
windows check in catch ?

rseeberg:
> Let's do the try catch thing described in the issue

Do keep in mind though, that this will mean people with permission issues, will see the donation message every time they build, until they fix their permission issues. Personally i don't mind, but..

> > I would prefer you just print the message on every build.
> 
> This is actually very disturbing for some users as many others may be using webpack-cli as their dependency and if they install that package this message will pop up every time. The Current approach is the workaround to save that.
> 
> as far as [de41351](https://github.com/webpack/webpack-cli/commit/de413515ed0bba4e72e24e758f6701a5c0ec0886) is concerned I will refactor the current PR to check for the file-owner when its mac or Linux and to skip it when its windows.

This could be mitigated by:

- not showing the message
- showing a message about why you are seeing it (due to permission issues)
- by showing it "randomly"

`const showMessage = (Math.random() * 1000) > 500; `

In the end, like most things, it boils down to picking the less of two (three?) evils 🙂

rseeberg:
> windows check in catch ?
> 
> ```js
> try {
>     if (!e && fileOwnerId === process.getuid())
>         utimesSync(openCollectivePath, now, now);
> } catch (error) {
>     if (!e)
>         utimesSync(openCollectivePath, now, now);
> 
> }
> ```

```js
if (!e) {
	if ((process.platform === "darwin" || process.platform === "linux") && stat.uid === process.getuid()) {
		utimesSync(openCollectivePath, now, now);
	} else {
		try {
			utimesSync(openCollectivePath, now, now);
		} catch(error) {
			/*
			 * Unable to update utime.
			 */
		}
	}
}
```

anikethsaha:
why `utimesSync(openCollectivePath, now, now);` inside try....that stmt gonna run anyway

rseeberg:
> why `utimesSync(openCollectivePath, now, now);` inside try....that stmt gonna run anyway
> I guess the condition should be in it
> this alone
> 
> ```js
> fileOwnerId === process.getuid()
> ```

Nope, that is the issue that @pastak was fixing.
If the current process is not the owner of the file, you will *NOT* be allowed to update it.
> It is happen EPERM when node_modules is owned by `root` and webpack-cli is executed by user excepted `root`.

anikethsaha:
but the code you mentioned
 ```js 
// ----- this one
if (!e) {
	if ((process.platform === "darwin" || process.platform === "linux") && stat.uid === process.getuid()) {
		utimesSync(openCollectivePath, now, now);
	} else {
		try {
			utimesSync(openCollectivePath, now, now);
		} catch(error) {
			/*
			 * Unable to update utime.
			 */
		}
	}
}
```
will update the time in no matter what with the process owner !

rseeberg:
🤷‍♂ I'm not too much into node etc.
All I can say is that I have had plenty of eperm issues, 
and @pastak believes they are caused by this.
Which I have no reason to doubt.
But it's more than possible you are correct.

anikethsaha:
I got your issue and really thankful for you to point out that. It seems I am a bit confused about what to catch in it. 
@evenstensberg any help !. 
Adding `fileOwnerId === process.getuid()` in try will not be effective as it will go to catch and if we put update time method in the catch statement for windows then it will run in mac also when the permission won't be satisfying 
I think we should leave the try-catch

evenstensberg:
> Adding fileOwnerId === process.getuid() in try will not be effective as it will go to catch and if we put update time method in the catch statement for windows then it will run in mac also when the permission won't be satisfying

It will, it will be a gracefully done exit if perm is an issue.

> And so if a user is not having permission, then this code should simply throw a nice message, it should not update the time. that's the flow I guess it should have.

Let's just try to check if the api is accessible and silent exit the script if not please

anikethsaha:
> Let's just try to check if the api is accessible and silent exit the script if not please

what about windows time update then? it will exit for windows then

evenstensberg:
This is a valid fix, but we've removed this from master now. Submitted a commit to show you how we would do it. Might add later

evenstensberg:
Thanks for taking the time!

Issue #746, opened 2019-01-31
Enh: `--config-register` not resolved relative to `process.pwd()`


Swatinem:
**Describe the bug**

the `--config-register / -r` flag differs from the `-r` flag of `node` which is confusing. In particular relative paths are not resolved relative to `process.pwd()`

**What is the current behavior?**

    $ webpack -r ./tools/register.js --config packages/xxx/tools/webpack.config

That fails with

> Error: Cannot find module './tools/register.js'

**To Reproduce**

* **fails**: `webpack -r ./tools/register.js --config packages/xxx/tools/webpack.config`
* **succeeds**: `webpack -r $(pwd)/tools/register.js --config packages/xxx/tools/webpack.config`
* **succeeds**: `node -r ./tools/register.js node_modules/.bin/webpack --config packages/xxx/tools/webpack.config`

**Expected behavior**

The behavior of `-r` should match the behavior of `node`, in particular that relative paths are resolved relative to `process.pwd()`.

evenstensberg:
Oh didn't know, do you mind submitting a PR for this?

Swatinem:
> do you mind submitting a PR for this?

Can do, give me some time :-)

jamesgeorge007:
@evenstensberg I would gladly take this up :+1: It would be great if you could give me a better picture of the files to be worked with as I'm not that aware of the code-base.

evenstensberg:
Great, hmu if you have any questions! @Swatinem since this hasn't had activity since January, I'll let @jamesgeorge007 do this

jamesgeorge007:
@evenstensberg Can you point me to the files to work with :thinking:

Swatinem:
Go for it. I completely forgot about this…

evenstensberg:
Here you go: https://github.com/webpack/webpack-cli/blob/master/bin/utils/convert-argv.js#L106-L120

jamesgeorge007:
@evenstensberg It seems that this is the required workaround :-
```js
return require(path.resolve(process.cwd(), configPath));
```
This is what I infer from the context. Shall I propose a PR with the respective changes :thinking:

evenstensberg:
Do that!

stephtr:
Is this really fixed? Shouldn't `path.resolve` be added to [line 111](https://github.com/webpack/webpack-cli/pull/903/files#diff-f28fd412eca1697c5b34e761c9bbd562R111)?

evenstensberg:
[...]

PR #971, opened 2019-06-23
fix: improve local configRegister file resolution

stephtr:
bugfix

no

~**If relevant, did you update the documentation?**~

**Summary**
Fixes [#746](https://github.com/webpack/webpack-cli/issues/746#issuecomment-504755804) by first checking, if the dependency is a local file or a package. `module.paths` doesn't work for requiring local paths.

I couldn't think of any.

**Other information**

### DIFF
--- a/lib/groups/config.js
+++ b/lib/groups/config.js
@@ -46,14 +46,17 @@ class ConfigGroup extends GroupHelper {
 
         this.configOptions = (() => {
             if (register && register.length) {
-                module.paths.unshift(resolve(process.cwd(), 'node_modules'), process.cwd());
+                module.paths.unshift(resolve(process.cwd(), 'node_modules'));
                 register.forEach(dep => {
-                    require(dep);
+                    const isRelative = ["./", "../", ".\\", "..\\"].some(start => dep.startsWith(start));
+                    if (isRelative) {
+                        require(resolve(process.cwd(), dep));
+                    } else {
+                        require(dep);
+                    }
                 });
-                return require(configPath);
-            } else {
-                return require(configPath);
             }
+            return require(configPath);
         })();
     }
 


bin/utils/convert-argv.js
```diff
@@ -107,14 +107,17 @@ module.exports = function(...args) {
 		const requireConfig = function requireConfig(configPath) {
 			let options = (function WEBPACK_OPTIONS() {
 				if (argv.configRegister && argv.configRegister.length) {
-					module.paths.unshift(path.resolve(process.cwd(), "node_modules"), process.cwd());
+					module.paths.unshift(path.resolve(process.cwd(), "node_modules"));
 					argv.configRegister.forEach(dep => {
-						require(dep);
+						const isRelative = dep.startsWith("./") || dep.startsWith("../") || dep.startsWith(".\\") || dep.startsWith("..\\");
```
jamesgeorge007: This seems to be redundant at least IMO :thinking: 

```js
const isRelative = ["./", "../", ".\", "..\\"].indexOf(dep.charAt(0));
```

bin/utils/convert-argv.js
```diff
@@ -107,14 +107,17 @@ module.exports = function(...args) {
...
+						if (isRelative) {
```
jamesgeorge007: This seems to be redundant at least IMO :thinking: 

```js
const isRelative = ["./", "../", ".\", "..\\"].indexOf(dep.charAt(0)) !== -1;
```

jamesgeorge007: cc @evenstensberg

stephtr: `dep.charAt(0)` would just return '.'. And this also (more or less) the way how it's done [within node](https://github.com/nodejs/node/blob/master/lib/internal/modules/cjs/loader.js#L578).

jamesgeorge007: Yeah :sweat_smile: seems like there is no perfect way to have a similar sort of implementation to match those characters.

stephtr: Would you prefer `const isRelative = ["./", "../", ".\\", "..\\"].some(start => dep.startsWith(start));`? (even though it makes the code a bit harder to read)

jamesgeorge007: It would become more complicated as evident from the recent edit above.

jamesgeorge007: cc @evenstensberg

jamesgeorge007: > Would you prefer `const isRelative = ["./", "../", ".\\", "..\\"].some(start => dep.startsWith(start));`? (even though it makes the code a bit harder to read)

Much concise IMO :clap:

bin/utils/convert-argv.js
```diff
@@ -109,11 +109,7 @@ module.exports = function(...args) {
 				if (argv.configRegister && argv.configRegister.length) {
 					module.paths.unshift(path.resolve(process.cwd(), "node_modules"));
 					argv.configRegister.forEach(dep => {
-						const isRelative =
-							dep.startsWith("./") ||
-							dep.startsWith("../") ||
-							dep.startsWith(".\\") ||
-							dep.startsWith("..\\");
+						const isRelative = ["./", "../", ".\\", "..\\"].some(start => dep.startsWith(start));
```
anikethsaha: nice refactoring ! 👍

stephtr:
The CLA actually is signed. @jsf-clabot switches just more or less randomly between signed/not signed.

stephtr:
@evenstensberg Done. CI is currently failing for the whole `next` branch.

evenstensberg:
Thank you <3

PR #984, opened 2019-07-06
tests(next): added tests for path-utils

anikethsaha:

Tests

Yes

NA

**Summary**
Added tests for `path-utils` 

No

**Other information**
NA


anikethsaha:
cc @evenstensberg

packages/utils/__tests__/path-utils.test.ts
```diff
@@ -0,0 +1,13 @@
+import { join } from "path";
+import { isLocalPath, findProjectRoot } from "../path-utils";
+
+describe("path-utils tests", () => {
+  it("isLocalPath Testing : __dirname", () => {
+    const answer = isLocalPath(join(__dirname, "../../"));
```
ematipico: You should test this against different formats:
- ./path
- path/path
- ..\path (win)

And also you should test the error cases

anikethsaha:
I have submitted a PR https://github.com/webpack/webpack-cli/pull/999 regarding the `isLocalPath` method which seems to be buggy. tests in this commit https://github.com/webpack/webpack-cli/pull/984/commits/b9c4260c40d50a3ae68a4a9a4b7f02554a3d162d will be passed if https://github.com/webpack/webpack-cli/pull/999 is considered cause currently for non-valid local path the `isLocalPath` will return `true`

@ematipico @evenstensberg PTAL

evenstensberg:
We cannot merge anything before the tests are passing, so we need to fix those first

PR #987, opened 2019-07-07
Docs(next): changed init docs , added auto flag for the node 

anikethsaha:

refactoring

NA

NA

**Summary**
as in the init package, its using `args.slice(3)` in order to detect whether any custom scaffold or `--auto` flag is passed or not, in the init docs it was 
`init([null,null,..myPackages])`

**there are two conflicts with this**

- in the `init/index.ts`, in `initializeInquirer` the arguments are spread so its actually converting it into array , so if there is `init([ ])` used to call the API, then its actually spreading it and making array of array `[ [ ] ]`, and it will create an error, as there won't be any **3rd** index so it will be skipped

- there are 3 arguments mentioned in the docs which are passing to the method, so as its `slice(3)` again the scaffold will be skipped

Also 
I added docs for running the `init` using node for `--auto` flag too

No

**Other information**
NA

### DIFF
--- a/packages/init/README.md
+++ b/packages/init/README.md
@@ -25,7 +25,7 @@ const init = require("@webpack-cli/init").default;
 init();
 
 // we're slicing node.process, ...myPacakges is a webpack-scaffold name/path
-init(null, null, ...myPacakges);
+init(null, null, null, ...myPacakges);
 
 // for auto flag to use the default configs
 init(null, null, null, "--auto")
--- a/packages/init/README.md
+++ b/packages/init/README.md
@@ -25,7 +25,10 @@ const init = require("@webpack-cli/init").default;
 init();
 
 // we're slicing node.process, ...myPacakges is a webpack-scaffold name/path
-init([null, null, ...myPacakges]);
+init(null, null, ...myPacakges);
+
+// for auto flag to use the default configs
+init(null, null, null, "--auto")
 ```
 
 ### CLI (via `webpack-cli`)


anikethsaha:
cc @evenstensberg

packages/init/README.md
```diff
@@ -25,7 +25,10 @@ const init = require("@webpack-cli/init").default;
 init();
 
 // we're slicing node.process, ...myPacakges is a webpack-scaffold name/path
-init([null, null, ...myPacakges]);
```
evenstensberg: We actually need to support both, so this change needs to be adjusted accordingly

anikethsaha: support both ?
cli and node ?

evenstensberg: No we're supporting both auto flag and multiple pkgs

anikethsaha:
```js
options = [...myPackages,"--auto"]
init(null, null, null, options);
```
like this can we add in the docs ?

evenstensberg: Let's keep the array convention, and stick with either having a spread plus the auto flag or just the packages.

anikethsaha: > stick with either having a spread plus the auto flag or just the packages.

thinking of keeping with the auto flag. 
 
> Let's keep the array convention, 

but then in the init method , the method argument will be an array of array. of slicing of 3 may not work

evenstensberg: Then you can tweak the api

PR #988, opened 2019-07-08
BREAKING CHANGE: Webpack-CLI v4 Foundation

evenstensberg:
Merges the work on v4 to master branch.

No
No
**Summary**
This PR will make it easier for outside collaborators to help out developing the new cli version. It's a lot of work left and some help on getting this ready would be great. After this has been merged, we can close outdated issues and triage new ones related to the master branch that has been updated.

Even though you could argue that keeping your LTS version on master intact, I do not agree in this instance, because the next branch is a bit mysterious right now, as well as there's no real contributions from multiple collaborators. This will make it easier to change that. 

The core work which introduces the kind of structure we want is there. We need to adjust now to make this production ready, which means that we'll need to simplify some logic in the code, write some comments in the code and make the src easier to work with.

Not yet 

**Other information**
N/A


evenstensberg:
Squash merge fyi, I'm not good at commit messages

evenstensberg:
After discussion with @ematipico and the rest of the team, we're gonna make the next default branch instead. Makes it easier for contributors to help develop and we do not use that much time reviewing PRs on the master branch (which takes away time from the new version)

PR #1004, opened 2019-07-20
[WIP] Add ReporterPlugin

misterdev:
This is related to the work on the [ReporterPlugin](https://github.com/misterdev/webpack-reporter-plugin), this PR is a placeholder for the work required to integrate it into webpack-cli as the default way of printing the output.

### DIFF
--- a/lib/utils/compiler.js
+++ b/lib/utils/compiler.js
@@ -88,34 +88,36 @@ async function invokeWatchInstance(compiler, lastHash, options, outputOptions, w
 }
 function compilerCallback(compiler, err, stats, lastHash, options, outputOptions) {
     let statsErrors = [];
-    const stdout = options.silent
-        ? {
-              write: () => {},
-		  } // eslint-disable-line
-        : process.stdout;
+    // TODO handle this in default reporter
+    // const stdout = options.silent
+    //     ? {
+    //           write: () => {},
+	// 	  } // eslint-disable-line
+    //     : process.stdout;
     if (!outputOptions.watch || err) {
         // Do not keep cache anymore
         compiler.purgeInputFileSystem();
     }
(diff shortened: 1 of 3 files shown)


misterdev:
I have two considerations on this:

1. Using the ReporterPlugin as CLI's default way of making output is not necessary, but if we don't do that, we'll have to do something in order to avoid duplicated output when the plugin is used by a project (e.g. both the cli and the reporter plugin will print stats), and I think that will require a change in the CLI's code

2. This plugin requires the [`outputOptions`](https://github.com/webpack/webpack/blob/master/lib/Compilation.js#L435) object used by webpack to customize the output appearence (e.g. `{ color: true }`). This object is created by webpack if we set, for example, the "stats" property in the configuration file, otherwise, it is created by webpack-cli (see [here](https://github.com/webpack/webpack-cli/blob/next/lib/webpack-cli.js#L88)). In order to make this plugin work, it needs to access this object, and the only way I know is that it receives it from the cli (like [here](https://github.com/misterdev/webpack-cli/blob/next-webpack-reporter-plugin/lib/utils/compiler.js#L159))
    ```js
    new ReporterPlugin(reporterPluginOptions).apply(compiler, outputOptions);
    ```

    If we won't do this the plugin won't work correctly (e.g. will remove colors from stats)


PS: The output option [`colors`](https://github.com/webpack/webpack-cli/blob/master/bin/cli.js#L169) seems to have been ranemed to [`color`](https://github.com/webpack/webpack-cli/blob/next/lib/utils/compiler.js#L128), is it the intended behaviour?

evenstensberg:
@misterdev that's on the next branch? Yes, it is. Also, re 1) 

This should extend the existing options, not override it unless supplied

Issue #900, opened 2019-05-28
Migrate webpack-dev-server CLI to webpack-cli serve


Loonride:
**Is your feature request related to a problem? Please describe.**
The Dev Server's CLI should live in `webpack-cli`. This will make project configuration easier for users, and will force minimization of configuration changes that the CLI makes, creating uniformity between the Dev Server's CLI and API.

**Describe the solution you'd like**
- `webpack-cli` serve should call the Dev Server API, not the CLI
- Move logic from [`bin/webpack-dev-server`](https://github.com/webpack/webpack-dev-server/blob/master/bin/webpack-dev-server.js) to `serve` package
- Minimalize any configuration changes that the CLI makes - it should simply normalize the CLI arguments into Dev Server options, then pass it into the API, and do no further config.
- Dev Server CLI should tell the user to switch to `webpack-cli serve`
- CLI should eventually be removed from the Dev Server in a major version

**Describe alternatives you've considered**
N/A

**Additional context**

EDIT:

- Also, tests need to be added here for `webpack-cli serve`

evilebottnawi:
Bump

evilebottnawi:
Fixed, but with a lot of bugs :disappointed:

PR #1011, opened 2019-07-28
[NEXT]: webpack-cli serve refactor draft

Loonride:
/cc @evilebottnawi @hiroppy 

Initial draft for https://github.com/webpack/webpack-cli/issues/900

This makes `webpack-cli serve` use the new `command-line-args` argument parsing system, and it makes this command call the `webpack-dev-server` API, rather than the CLI.

No

No

**Summary**

The goal of this change is ultimately to make `webpack-cli serve` work directly with the `webpack-dev-server` API. I have placed this API call in `startDevServer`, with this file mostly copied from [`bin/webpack-dev-server.js`](https://github.com/webpack/webpack-dev-server/blob/master/bin/webpack-dev-server.js). Ideally, we can minimalize what this helper does.

This draft is still very incomplete, what still needs to be done/considered:

- Add all `webpack-dev-server` flags
- Make the system use the correct webpack config file
- Is it best that this works mostly independently from the new `webpack-cli` system?
- make `false` work as a CLI flag definition, like with yargs where you could do `--inline=false` for bools

Reorganization:

- the dev server flags I think should live in the `webpack-dev-server` repository, but it is easier to draft them here.

Edit:

- We will remove the `--color` CLI flag in favor of simply retrieving the `colors` option from the compiler options.
- The `--info` and `--quiet` flags will also be removed

TBD

**Other information**


packages/serve/startDevServer.js
```diff
@@ -0,0 +1,93 @@
+const fs = require('fs');
```
hiroppy: Why do you use `js`?

Loonride: @hiroppy Just easier for me to draft it faster. Can change it later once plan is more solidified.

packages/serve/startDevServer.js
```diff
@@ -0,0 +1,93 @@
...
+                    if (err) {
+                        throw err;
+                    }
+                });
+            })
+            .catch((err) => {
+                throw err;
+            });
+    }
+}
```
evilebottnawi: /cc @Loonride we need discussion about this, keep this stuff here will not allow us to change logic and fix very quickly in the future, we need create api on webpack-dev-server side, maybe idea with new api method is not bad

Loonride: @evilebottnawi I agree, I simply copied and pasted it because I knew I would probably be changing it on the `webpack-dev-server` side. An alternative to a new API method is introducing some minor breaking changes to the `listen` method, as I started doing here: https://github.com/webpack/webpack-dev-server/pull/2061, so that the `listen` method works without any additional configuration for the CLI. I think either approach could work.

evilebottnawi: @Loonride yes, let's finish this in `webpack-dev-server` and continue development here :+1:

lib/commands/external.js
```diff
@@ -68,9 +68,9 @@ class ExternalCommand {
             pkgLoc = await ExternalCommand.promptInstallation(scopeName, name);
         }
         // Serve needs to be checked for
-        if (name === 'serve') {
-            return pkgLoc ? require(pkgLoc).serve(args) : null;
-        }
+        // if (name === 'serve') {
```
evenstensberg: Instead of uncommenting, just remove it after wip

Loonride:
I've made it so that `webpack-cli serve` now uses the compiler that the `webpack-cli` command would create.

Loonride:
A few thoughts:

- should `serve` package use async/await?
- at what point should the Promise resolve that is returned by the `serve` package? Currently, it resolves once the compiler is retrieved, but does not wait for anything else. Should it wait until both the server is serving, and the first compilation has completed?

packages/serve/startDevServer.ts
```diff
@@ -0,0 +1,23 @@
+import * as Server from "webpack-dev-server/lib/Server";
```
Loonride: /cc @evilebottnawi I made some progress here. In here I put basically the ideal version of what we want, so minimal options handling before starting the server. 

I think `flags.js` should live in `webpack-dev-server` in the future, rather than here, because then we can easily make updates without changing `webpack-cli`, but we can put that in `webpack-dev-server` `next` later. Do you agree?

Loonride:
/cc @evilebottnawi @hiroppy Thoughts? I know I still may need to make changes here based on dev server changes, but I think that, optimally, there should be very minimal interaction with dev server options as I have done here.

Where should tests be? Should tests related to running the basic `webpack-cli serve` command be in `webpack-cli`, and commands related to specific CLI args be in `webpack-dev-server`?

Where should the list of dev server CLI flags be? I think `webpack-dev-server`, so that they can easily be maintained.

packages/serve/flags.js
```diff
@@ -0,0 +1,222 @@
...
+        {
+            name: 'allowed-hosts',
+            type: String,
+            describe:
+                'A list of hosts that are allowed to access the dev server, separated by spaces',
+            group: CONNECTION_GROUP,
+            multiple: true,
+        },
+    ],
+};
```
evilebottnawi: /cc @Loonride we should avoid this file

evilebottnawi: It will be changed in next version

Loonride:
> Do you mind rebasing such that you only have the changes to the `serve` command and related?

@evenstensberg I did rebase. But I don't think Files Changed is any different. Did you see a problem with something?

lib/bootstrap.js
```diff
@@ -46,6 +55,6 @@ async function runCLI(cli, commandIsUsed) {
 // eslint-disable-next-line space-before-function-paren
 (async () => {
```
hulkish: Not a real reason to use `async`/`await`, here.

Since you're not processing multiple promises, or catching an async error - you're not benefiting from `async`/`await`. Instead, I suggest:
```js
const commandIsUsed = isCommandUsed(commands);
const cli = new WebpackCLI();
runCLI(cli, commandIsUsed);
```

Which then lets you get rid of the IIFE `(async () => {})()`.

lib/utils/compiler.js
```diff
@@ -112,7 +112,11 @@ function compilerCallback(compiler, err, stats, lastHash, options, outputOptions
     }
 }
 
-module.exports = async function webpackInstance(opts, shouldUseMem) {
+async function getCompiler(opts) {
```
hulkish: This function doesn't need to be `async`, it can just return the unhandled promise. Same goes for the subsequent usage of `await`.

lib/utils/compiler.js
```diff
@@ -197,3 +202,8 @@ module.exports = async function webpackInstance(opts, shouldUseMem) {
         return await invokeCompilerInstance(compiler, lastHash, options, outputOptions);
     }
 };
+
+module.exports = {
```
hulkish: No need for new object. Instead, I suggest:

```js
exports.getCompiler = getCompiler;
exports.webpackInstance = webpackInstance;
```

packages/serve/flags.js
```diff
@@ -0,0 +1,222 @@
...
+            name: 'useLocalIp',
+            type: Boolean,
+            describe: 'Open default browser with local IP',
+        },
+        {
+            name: 'open-page',
+            type: String,
+            describe: 'Open default browser with the specified page',
+        },
+        // instead of this option, we should consider retrieving the 'color' option
```
hulkish: I recommend moving this dialog to your PR description - or as a comment.

packages/serve/startDevServer.ts
hulkish: The tests for options should live inside `webpack-cli` - since it owns the corresponding API for said options. 

`webpack-dev-server` should have tests that validates integration & compatibility with `webpack-cli` api. For ex, this would apply to your changes for `webpack-cli/lib/webpack-cli.js` where you have added the `processArgs` fn, etc.

lib/bootstrap.js
```diff
@@ -1,11 +1,20 @@
...
-const normalizeFlags = (args, cmd) => args.slice(2).filter(arg => arg.indexOf('--') < 0 && arg !== cmd.name && arg !== cmd.alias);
+const isArgCommandName = (arg, cmd) => arg === cmd.name || arg === cmd.alias;
+const normalizeFlags = (args, cmd) => {
+    const slicedArgs = args.slice(2);
+    if (cmd.name === 'serve') {
+        return slicedArgs.filter(arg => !isArgCommandName(arg, cmd));
+    }
+    else {
+        return slicedArgs.filter(arg => arg.indexOf('--') < 0 && !isArgCommandName(arg, cmd));
+    }
```
jamesgeorge007:
```suggestion
return slicedArgs.filter(arg => arg.includes('--') && !isArgCommandName(arg, cmd));
    }
```

lib/bootstrap.js
```diff
@@ -1,11 +1,20 @@
...
 const isFlagPresent = (args, flag) => args.find(arg => [flag, `--${flag}`].includes(arg));
-const normalizeFlags = (args, cmd) => args.slice(2).filter(arg => arg.indexOf('--') < 0 && arg !== cmd.name && arg !== cmd.alias);
+const isArgCommandName = (arg, cmd) => arg === cmd.name || arg === cmd.alias;
+const normalizeFlags = (args, cmd) => {
+    const slicedArgs = args.slice(2);
+    if (cmd.name === 'serve') {
+        return slicedArgs.filter(arg => !isArgCommandName(arg, cmd));
+    }
+    else {
+        return slicedArgs.filter(arg => arg.indexOf('--') < 0 && !isArgCommandName(arg, cmd));
```
jamesgeorge007:
```suggestion
        return slicedArgs.filter(arg => arg.includes('--') && !isArgCommandName(arg, cmd));
    }
```

jamesgeorge007:
```suggestion
        return slicedArgs.filter(arg => arg.includes('--') && !isArgCommandName(arg, cmd));
```

jamesgeorge007:
```suggestion
        return slicedArgs.filter(arg => !arg.includes('--') && !isArgCommandName(arg, cmd));
```

lib/utils/compiler.js
Loonride: Are you sure the subsequent `await` is not needed? Is there something coming in webpack 5 where it might return a promise when `webpack()` is called? I assume that's why the `await` was there originally.

evenstensberg:
I'm ready to get this merged to the next branch if you rebase @Loonride

Loonride:
> I'm ready to get this merged to the next branch if you rebase @Loonride

@evenstensberg I rebased. I've moved the dev server flags into the `lib` directory temporarily only to pass CI (It is complaining about it being a .js file). Eventually, we will move this file out of `webpack-cli` and into `webpack-dev-server`, but it is convenient to keep it here for now when making CLI changes.

lib/bootstrap.js
```diff
@@ -1,11 +1,20 @@
...
+const WebpackCLI = require('./webpack-cli');
 const { core, commands } = require('./utils/cli-flags');
 const cmdArgs = require('command-line-args');
 
 require('./utils/process-log');
 
 const isFlagPresent = (args, flag) => args.find(arg => [flag, `--${flag}`].includes(arg));
-const stripDashedFlags = (args, cmd) => args.slice(2).filter(arg => ~arg.indexOf('--') && arg !== cmd.name && arg !== cmd.alias);
+const isArgCommandName = (arg, cmd) => arg === cmd.name || arg === cmd.alias;
+const stripDashedFlags = (args, cmd) => args.filter(arg => ~arg.indexOf('--') && !isArgCommandName(arg, cmd));
```
jamesgeorge007:
```suggestion
const stripDashedFlags = (args, cmd) => args.filter(arg => arg.includes('--') && !isArgCommandName(arg, cmd));
```

evenstensberg:
@Loonride that might be something with your tsconfig file, try tweaking around with it and running `npm run build` and `npm run bootstrap`

lib/bootstrap.js
Loonride: Actually I think the logic got messed up and I just followed along with it :slightly_frowning_face:

It should be `!arg.includes('--')`, because we want to get rid of all the arguments with `--` in this helper, correct?

packages/serve/argsToCamelCase.ts
```diff
@@ -0,0 +1,31 @@
...
+ * @returns {String} - new camel case string
+ */
+function dashesToCamelCase(str): string {
+    return str.replace(/-([a-z])/g, (g): string => g[1].toUpperCase());
+}
+
+/**
+ *
+ * Converts CLI args to camel case from dash-separated words
+ *
```
evenstensberg: could you make this filename hyphen based instead?

lib/bootstrap.js
```diff
@@ -1,11 +1,20 @@
...
 const cmdArgs = require('command-line-args');
 
 require('./utils/process-log');
 
 const isFlagPresent = (args, flag) => args.find(arg => [flag, `--${flag}`].includes(arg));
-const stripDashedFlags = (args, cmd) => args.slice(2).filter(arg => ~arg.indexOf('--') && arg !== cmd.name && arg !== cmd.alias);
+const isArgCommandName = (arg, cmd) => arg === cmd.name || arg === cmd.alias;
+const stripDashedFlags = (args, cmd) => args.filter(arg => !arg.includes('--') && !isArgCommandName(arg, cmd));
+const normalizeFlags = (args, cmd) => {
+    const slicedArgs = args.slice(2);
```
evenstensberg: you can return early and remove the else here

Loonride:
> Small comments and then this is ready, thank you so much @Loonride

I'm still not sure about the failing CI @evenstensberg

evenstensberg:
Don't worry about it, is my fault, will deal with that

lib/utils/dev-server-flags.js
```diff
@@ -0,0 +1,193 @@
...
+        {
+            name: 'allowed-hosts',
+            type: String,
+            describe:
+                'A list of hosts that are allowed to access the dev server, separated by spaces',
+            group: CONNECTION_GROUP,
+            multiple: true,
+        },
+    ],
+};
```
evilebottnawi: /cc @Loonride it is very very very bad, next PR will change a lot of options, we don't need this do in webpack-cli

packages/serve/startDevServer.ts
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    const server = new Server(compiler, options);
+    server.listen(socket || port, host, (err): void => {
+        if (err) {
+            throw err;
+        }
+        if (typeof onListening === 'function') {
+            onListening();
+        }
```
evilebottnawi: We should solve this on `webpack-dev-server` side, not here

packages/serve/startDevServer.ts
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * @param {Object} compiler - a webpack compiler
+ * @param {Object} options - devServer options
+ * @param {Function} onListening - optional callback for when the server starts listening
+ * 
+ * @returns {Void} 
+ */
+export default function startDevServer(compiler, options, onListening): void {
+    const firstWpOpt = compiler.compilers
+        ? compiler.compilers[0].options
```
evilebottnawi: Why only `0`? What about multi-mode compilation

evenstensberg:
@Loonride could you address these in a new PR?

packages/serve/startDevServer.ts
Loonride: > We should solve this on `webpack-dev-server` side, not here

This `onListening` is different, it is just a callback as one of the function parameters. It can be removed if needed.

packages/serve/startDevServer.ts
Loonride: @evilebottnawi In the past on `webpack-dev-server` we only looked at the first compiler in a multi-compiler array to see if it has the `devServer` object. Do you want to change this behavior?

PR #1014, opened 2019-07-30
tests: add no-args help test

evenstensberg:
Adds test to running `webpack help` with no args

Kinda ⛺️ 
N/A
**Summary**
Adds test for empty help flag

No

**Other information**


evenstensberg:
I think this is fine for splitting up work, we can add tests gradually. Size of the PR doesn't really matter if we continue to add more tests over time? That's my approach

Issue #1023, opened 2019-08-14
TypeError: Cannot read property 'length' of undefined


wind4gis:
**Describe the bug**

A 💯 way to do this is to provide your configuration via a GitHub gist and/or to use markdown when describing which commands you used when the error occurred! Be precise and clear in your description of the bug.

**What is the current behavior?**
when I run `webpack --mode development(or production)`, the webpack-cli shows `TypeError: Cannot read property 'length' of undefined` on the line `node_modules\webpack-cli\bin\cli.js:93`. my computer is win10

This is where you elaborate on the current behavior of the cli, how it behaves normally differing from the previous version/build you did. 

**To Reproduce**

1.just create a folder on win10
2.npm i -S webpack@next
3.npm i -S webpack-cli@latest
4.write "dev": "webpack --mode development" in package.json
5.npm run dev

**Expected behavior**

**Screenshots**
![image](https://user-images.githubusercontent.com/3963291/63011085-2ecea580-beba-11e9-881e-349e1d1e202f.png)

**Please paste the results of `webpack-cli info` here, and mention other relevant information**
![image](https://user-images.githubusercontent.com/3963291/63011134-44dc6600-beba-11e9-998e-f02944031b85.png)

**Additional context**
Add any other context about the problem here like linking to an similar issue you might think is the cause.

evenstensberg:
Hi, thanks for submitting an issue. It's a bug, I've done a patch here:

https://github.com/webpack/webpack-cli/pull/1024

Which hopefully will land as a patch soon

wind4gis:
@evenstensberg  It seems that there is still this bug

evenstensberg:
Could you try out the #1024 branch and check if it fixes the bug?

ryanspice:
I ran into this issue and pulled your branch it seems to be working now.

ryanspice:
@evenstensberg not sure if this is to be expected but an empty webpack config object
![image](https://user-images.githubusercontent.com/3946233/63227173-dd703000-c1b1-11e9-8c3d-57cf794b9b0d.png) brings up the same error.

ryanspice:
actually it might be something else, not quite sure, but trying to plug in my existing config I have seen this error pop up.

ryanspice:
yeah im not sure why but when i tried to run it with a config now it was always giving me this error 

[...]

PR #1024, opened 2019-08-14
fix: resolve opts when no-config & fix vulns

evenstensberg:
Closes https://github.com/webpack/webpack-cli/issues/1023
Yes
N/A
**Summary**
Dupes out of resolving options if there are none to throw an error

No
**Other information**
N/A

### DIFF
--- a/bin/utils/convert-argv.js
+++ b/bin/utils/convert-argv.js
@@ -128,7 +128,7 @@ module.exports = function(...args) {
 	}
 
 	if (!configFileLoaded) {
-		return processConfiguredOptions({});
+		return processConfiguredOptions(null);
 	} else if (options.length === 1) {
 		return processConfiguredOptions(options[0]);
 	} else {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -142,7 +142,7 @@
     "babel-preset-env": "1.7.0",
     "babel-preset-jest": "24.6.0",
     "codecov": "3.5.0",
-    "commitizen": "3.1.1",
+    "commitizen": "^4.0.3",
```
ematipico:
```suggestion
    "commitizen": "4.0.3",
```

package.json
```diff
@@ -131,7 +131,7 @@
   "devDependencies": {
     "@babel/preset-env": "7.4.5",
     "@babel/register": "7.4.4",
-    "@commitlint/cli": "8.0.0",
+    "@commitlint/cli": "^8.1.0",
```
ematipico:
```suggestion
    "@commitlint/cli": "8.1.0",
```

bin/utils/convert-argv.js
```diff
@@ -128,28 +128,32 @@ module.exports = function(...args) {
...
-			process.exit(-1); // eslint-disable-line
+		if (options) {
+			const webpackConfigurationValidationErrors = validateSchema(webpackConfigurationSchema, options);
+			if (webpackConfigurationValidationErrors.length) {
+				const error = new WebpackOptionsValidationError(webpackConfigurationValidationErrors);
+				console.error(error.message, `\nReceived: ${typeof options} : ${JSON.stringify(options, null, 2)}`);
+				process.exit(-1); // eslint-disable-line
+			}
+		} else {
+			options = {};
```
ematipico: Nothing wrong with this code, but I usually avoid mutations like this one.

bin/utils/convert-argv.js
```diff
@@ -128,28 +128,32 @@ module.exports = function(...args) {
...
 		}
 
 		// process Promise
-		if (typeof options.then === "function") {
+		if (options && typeof options.then === "function") {
```
ematipico: Now that you have the if/else, `options` will never be undefined right? So this additional check is not needed

evenstensberg: checks for `options.then` object prop so will throw. We're setting props on the object later these checks, I think we "assumed" we either found or did not find the props before validating args that aren't protected by 0CJS (entry and output)

bin/utils/convert-argv.js
evenstensberg: Yea, agreed. I think we would have needed a larger refactor to process args without validating options, as explained a bit further down

bin/utils/convert-argv.js
ematipico: I'm not talking about .then, I'm talking about `options &&`. It's redundant. You already assigned it before with `options = {}`. 

So or you keep this check and remove the assignment or keep the assignment and remove the check.

evenstensberg: Thanks for pointing it out, changed now!

PR #1026, opened 2019-08-14
tests: add tests for version, improve conventions

evenstensberg:
Tests and a fix in type for modifyconfighelper

You bet!

N/A
**Summary**
Adds tests for version flag. Keep in mind that help goes before version in precedence.

N/A
**Other information**
N/A


lib/bootstrap.js
```diff
@@ -5,6 +5,7 @@ const cmdArgs = require('command-line-args');
 require('./utils/process-log');
 
 const isHelpFlagPresent = args => args.find(arg => ['help', '--help'].indexOf(arg) !== -1);
```
evenstensberg: we could generalize this func

lib/bootstrap.js
```diff
@@ -5,6 +5,7 @@ const cmdArgs = require('command-line-args');
...
+const isVersionFlagPresent =  args => args.find(arg => ['version', '--version'].indexOf(arg) !== -1);
```
ematipico: You could use `includes` instead of `indexOf`

lib/bootstrap.js
ematipico: You could use `includes` instead of `indexOf`

lib/webpack-cli.js
```diff
@@ -111,6 +111,13 @@ class webpackCLI extends GroupHelper {
         const HelpGroup = require('./groups/help');
         return new HelpGroup().outputHelp();
     }
+    async runVersion() {
```
ematipico: Why `async` if it doesn't have `await`?

test/binCases/version/mixed/precedence-command.test.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
```
ematipico: In these tests you are mixing commands and the test doesn't help to understand what's the expected outcome. Using the snapshot is not enough. I'd suggest to add to those test more assertion about what's correct what is not

lib/webpack-cli.js
evenstensberg: Was following the async/await convention for each method, will remove

test/binCases/version/mixed/precedence-command.test.js
evenstensberg: I was planning to use snapshots for version and help flags, because they are static, will not change much. In hindsight, it might be beneficiary to assert correct versions, yes. 

The convention I've used to diff these two are the naming, but they need a more consistent and clear separation, indeed. 

Do you have any suggestion for the naming?

evenstensberg:
PTAL <3

PR #1035, opened 2019-08-17
chore: give feedback when no config

rishabh3112:
feature

No

No

**Summary**
Give feedback to the user if the config is not found

No

**Other information**
NA


evenstensberg:
I don't think we'll have this in now, but might change later once we're stable

PR #1038, opened 2019-08-18
feat: use JS Webpack Plugin

evenstensberg:
Feature
yes
N/A
**Summary**
Adds a 0CJS wrapper when no-config is found

No

**Other information**
N/A


evenstensberg:
Not necessarily the problem we’re trying to solve

rishabh3112:
Better defaults are always better

PR #1050, opened 2019-08-19
tests: add tests for entry defaults, try to set better defaults

evenstensberg:
Adds tests for entries, test if `./src/index.js` is present and set that default if it exists.

Yes bby
N/A
**Summary**
Adds tests, improves code

N/A
**Other information**

Keep in mind 1 of the cases fails, because we cannot override mixed arguments yet, I'll try to get that to work. For instance:

`webpack-cli --entry index.js entry index2.js` will fail because we have duplicate keys. Same goes for the opposite scenario or if we have `webpack-cli --entry index.js --entry index2.js`

in case of using `--` twice, we can convert those two into an array, which will be the best case, but for mixed arguments I think we need to decide to override the first key and set a notation that mixing conventions is not encouraged.

### DIFF
--- a/build/changelog-generator/index.js
+++ b/build/changelog-generator/index.js
@@ -5,6 +5,7 @@
  *  Based on: https://github.com/GoogleChrome/lighthouse/blob/master/build/changelog-generator/
  */
 'use strict';
+/* eslint-disable @typescript-eslint/explicit-function-return-type */
 const readFileSync = require('fs').readFileSync;
 const resolve = require('path').resolve;
 const mainTemplate = readFileSync(resolve(__dirname, 'templates/template.hbs')).toString();
--- a/lib/bootstrap.js
+++ b/lib/bootstrap.js
@@ -5,7 +5,7 @@ const cmdArgs = require('command-line-args');
 require('./utils/process-log');
 
 const isFlagPresent = (args, flag) => args.find(arg => [flag, `--${flag}`].includes(arg));
-const normalizeFlags = (args, cmd) => args.slice(2).filter(arg => arg.indexOf('--') < 0 && arg !== cmd.name && arg !== cmd.alias);
+const stripDashedFlags = (args, cmd) => args.slice(2).filter(arg => ~arg.indexOf('--') && arg !== cmd.name && arg !== cmd.alias);
 
 const isCommandUsed = commands =>
     commands.find(cmd => {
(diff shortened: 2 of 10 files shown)


lib/bootstrap.js
```diff
@@ -27,18 +27,24 @@ async function runCLI(cli, commandIsUsed) {
...
-        args = cmdArgs(core, { stopAtFirstUnknown: false });
         try {
+        args = cmdArgs(core, { stopAtFirstUnknown: false });
             const result = await cli.run(args, core);
             if (!result) {
                 return;
             }
         } catch (err) {
+
+            // TODO: normalzie and retry args when keys are same
```
jamesgeorge007:
```suggestion
            // TODO: normalize and retry args when keys are same
```

build/changelog-generator/index.js
```diff
@@ -5,6 +5,7 @@
  *  Based on: https://github.com/GoogleChrome/lighthouse/blob/master/build/changelog-generator/
  */
 'use strict';
+/* eslint-disable @typescript-eslint/explicit-function-return-type */
```
ematipico: Can be removed

lib/bootstrap.js
```diff
@@ -5,7 +5,7 @@ const cmdArgs = require('command-line-args');
 require('./utils/process-log');
 
 const isFlagPresent = (args, flag) => args.find(arg => [flag, `--${flag}`].includes(arg));
-const normalizeFlags = (args, cmd) => args.slice(2).filter(arg => arg.indexOf('--') < 0 && arg !== cmd.name && arg !== cmd.alias);
+const stripDashedFlags = (args, cmd) => args.slice(2).filter(arg => ~arg.indexOf('--') && arg !== cmd.name && arg !== cmd.alias);
```
ematipico: You could use `includes` instead of `indexOf`

lib/webpack-cli.js
```diff
@@ -32,9 +33,21 @@ class webpackCLI extends GroupHelper {
 
     checkDefaults(groupResult) {
         const { options } = groupResult;
-        if (!options.entry) {
-            this.shouldUseMem = true;
+        if(options.entry && options.entry === './index.js') {
```
ematipico: Sometimes people don't set the `.js`. What should we do in this case?

build/changelog-generator/index.js
evenstensberg: Linter threw here so I don't think so

lib/webpack-cli.js
evenstensberg: We gotta check for `index` too, but we will treat it as a directory, so if the user supplies `index`, we will look for the index directory and a filename under that dir. If the filename isn't present, we will set `index/bundle.js`

build/changelog-generator/index.js
ematipico: Ok it might be an issue of the linter (we could upgrade later). This is not a ts file after all

lib/webpack-cli.js
ematipico: If it's a directory, it should be treated as `index/index.js`

PR #1052, opened 2019-08-19
Support `export default` from configuration files.

izaakschroeder:

Feature.

Nope 😢 Tested it locally? Not sure how to add tests or what tests one would like to see.

Not necessary hopefully, since I imagine people want ES modules as config files to just work™.

**Summary**

This changes adds a `require` wrapper to the `config` group that allows it to detect and return the default export of ES modules. This is useful when your configuration is a TS(X) file or something that is being transpiled by babel.

Nope!

**Other information**

If one could possibly publish `webpack-cli@4.0.0-alpha-6` with these changes that would be 👌

### DIFF
--- a/lib/groups/config.js
+++ b/lib/groups/config.js
@@ -41,6 +41,14 @@ class ConfigGroup extends GroupHelper {
         };
     }
 
+    require(path) {
+        const result = require(path);
+        if (result && result.__esModule && result.default) {
+            return result.default;
+        }
+        return result;
+    }
+
     requireConfig(configPath) {
         const { register } = this.args;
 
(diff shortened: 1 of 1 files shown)


izaakschroeder:
Could this version possibly be published @evenstensberg? 😄

ematipico:
@izaakschroeder the PR was done against `next` branch, which is v4 and it's not ready to be published. If you would like to have this fixed released, you should make a PR against `master`

evenstensberg:
We could push a new alpha sure, but not yet

izaakschroeder:
@ematipico @evenstensberg that is correct, I'd love a new alpha version 😄 The short version of a long story is I made a plugin based on `webpack@5`'s `ConcatedModule` to emit an ESM library like rollup does (no runtime, no header, ESM exports, etc), and as a consequence of using that, the old `webpack-cli` does not work with `webpack@5`. This version does, but the current published version is broken. The above patch makes everything work for me 🎉  Open to other suggestions!

izaakschroeder:
Any thoughts @ematipico / @evenstensberg? 😄

PR #1059, opened 2019-08-23
chore(GroupHelper): minor tweak

jamesgeorge007:
refactoring

Nope

N/A

**Summary**
Beautify code.

Nope

**Other information**
N/A

### DIFF
--- a/lib/utils/group-helper.js
+++ b/lib/utils/group-helper.js
@@ -57,8 +57,7 @@ class GroupHelper {
     }
     mapArgToBoolean(name, optionName) {
         ifArg(name, function(bool) {
-            if (bool === true) options[optionName || name] = true;
-            else if (bool === false) options[optionName || name] = false;
+            options[optionName || name] = bool;
         });
     }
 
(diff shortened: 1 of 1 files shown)


lib/utils/group-helper.js
```diff
@@ -57,8 +57,7 @@ class GroupHelper {
     }
     mapArgToBoolean(name, optionName) {
         ifArg(name, function(bool) {
-            if (bool === true) options[optionName || name] = true;
-            else if (bool === false) options[optionName || name] = false;
+            options[optionName || name] = bool;
```
ematipico: What if `bool` is not a real boolean?

jamesgeorge007: The check assumes `bool` to be either `true` or `false` right?

jamesgeorge007: Oh yeah, making sure none other than `true` or `false` comes up right?

ematipico: Yes exactly. Your change changes the functionality of the code.

PR #1062, opened 2019-08-25
chore: changed the outDir to entry test

anikethsaha:

refactoring/bugfix

NA

NA

**Summary**
the current matching method, the test is failing in windows as the `Output Director` is as follows
```
\\Path-to\\webpack-cli\\test\\entry\\bin
``` 
and this is been matched with `entry/bin` which is failing .
So if we refactored the current to just check with the last two directory, we can make it to pass

No

**Other information**
Also disabled the `@typescript-eslint/explicit-function-return-type` for `test-utils` file as its causing to failed while committing.

### DIFF
--- a/test/entry/entry-single-arg.test.js
+++ b/test/entry/entry-single-arg.test.js
@@ -10,7 +10,12 @@ describe('single entry flag', () => {
         const { stdout, stderr } = run(__dirname);
         const summary = extractSummary(stdout);
         const outputDir = 'entry/bin';
-        expect(summary['Output Directory']).toContain(outputDir);
+        const outDirectoryFromCompiler = summary['Output Directory'];
+        const outDirToMatch = outDirectoryFromCompiler
+            .split('\\')
+            .slice(outDirectoryFromCompiler.split('\\').length - 2, outDirectoryFromCompiler.split('\\').length)
+            .join('/');
+        expect(outDirToMatch).toContain(outputDir);
         expect(stderr).toContain('Entry module not found');
         stat(resolve(__dirname, './bin'), (err, stats) => {
             expect(err).toBe(null);
--- a/test/entry/entry-single-arg.test.js
+++ b/test/entry/entry-single-arg.test.js
@@ -1,7 +1,7 @@
 'use strict';
 
 const { stat } = require('fs');
-const { resolve } = require('path');
+const { resolve, sep } = require('path');
 
 const { run, extractSummary } = require('../utils/test-utils');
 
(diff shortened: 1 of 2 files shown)


anikethsaha:
Actually, the current test is failing in windows, as the `Output Directory` is having path using `\\`  and also it is giving the whole relative path from the drive, not just the last basename, so I did this.
Also please check I used the node default separator for splitting in the latest commit

anikethsaha:
cc @evenstensberg @ematipico

test/entry/entry-single-arg.test.js
```diff
@@ -10,7 +10,9 @@ describe('single entry flag', () => {
         const { stdout, stderr } = run(__dirname);
         const summary = extractSummary(stdout);
         const outputDir = 'entry/bin';
-        expect(summary['Output Directory']).toContain(outputDir);
+        const outDirectoryFromCompiler = summary['Output Directory'].split(sep);
+        const outDirToMatch = outDirectoryFromCompiler.slice(outDirectoryFromCompiler.length - 2, outDirectoryFromCompiler.length).join('/');
```
ematipico: `.join('/')` is not correct. Windows don't have `/` in their directory paths

test/entry/entry-single-arg.test.js
```diff
@@ -10,7 +10,9 @@ describe('single entry flag', () => {
...
+        expect(outDirToMatch).toContain(outputDir);
         expect(stderr).toContain('Entry module not found');
         stat(resolve(__dirname, './bin'), (err, stats) => {
```
ematipico: `./bin` is not correct. It should be just `bin`

test/entry/entry-single-arg.test.js
anikethsaha: I just used `'/'` to make it in the format of the `outputDir` i.e `entry/bin` .

test/entry/entry-single-arg.test.js
```diff
@@ -10,7 +10,12 @@ describe('single entry flag', () => {
         const { stdout, stderr } = run(__dirname);
         const summary = extractSummary(stdout);
         const outputDir = 'entry/bin';
-        expect(summary['Output Directory']).toContain(outputDir);
+        const outDirectoryFromCompiler = summary['Output Directory'];
```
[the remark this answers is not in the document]
anikethsaha: I used the `node` method, `sep`

test/entry/entry-single-arg.test.js
anikethsaha: done

test/entry/entry-single-arg.test.js
ematipico: Yes, I know. Webpack under the hoods takes care of it. And when it's parsed, it's transformed with the path convention used by the operative system.

The use `path` utility to deal with paths.

Issue #1012, opened 2019-07-28
[TESTS]: Write tests for the arguments we support


evenstensberg:
**Is your feature request related to a problem? Please describe.**
In the next branch, we're replacing old test in favor for new ones. We can still use logic from the old test suite, but we need to write new assertions. Those include:

- [x] create
- [x] migrate
- [x] loader
- [x] plugin
- [x] info
- [x] negated flags
- [x] entry
- [x] dotfolder lookup
- [x] output
- [x] watch
- [x] merge
- [x] global
- [x] sourcemap
- [x] help
- [x] json
- [x] dev
- [x] prod
- [x] version
- [x] config

 **Describe the solution you'd like**

As listed in the previous section, these are the arguments we need to test. In addition, we need a replacement for the `mode` property, by reading the node env instead or by letting users pass the environment as a param to the webpack config, maybe both. I prefer that we read the node environment, because this means that we'll opt out of having to support a flag for setting mode, rather users can use the environment to set webpack mode. 

Furthermore, we need to in specific, test the way we combine the cli arguments with a config specified, in such fashion that we know that we're not creating any bugs.

I'd prefer that CLI arguments outrule configuration, but I'm open for discussion on that part. 

**Describe alternatives you've considered**
N/A

**Additional context**
Please comment in this issue for questions instead of reaching out to my personal accounts. This way it's public and everyone can chime in on what's happening.

evenstensberg:
FYI after this issue is closed, the args that (might) cause trouble are fixed, we can tag a beta

rishabh3112:
> Furthermore, we need to in specific, test the way we combine the cli arguments with a config specified, in such fashion that we know that we're not creating any bugs.
>
> I'd prefer that CLI arguments outrule configuration, but I'm open for discussion on that part.

I think this makes sense (i.e. CLI arguments should have a higher priority in case of conflict with config).
Are there some tests I can start working on @evenstensberg?

evenstensberg:
Just pick a flag to test

rishabh3112:
@evenstensberg I started on `--version` flag. I am afraid that I can't understand the requirements of this use case:

`webpack-cli --version` (in the empty project)
Instead of printing the version of the webpack CLI, it complies with errors.

Just for info, I understand this behaviour of command as per the current codebase, but is this the expected behaviour?

anshumanv:
Writing tests for info ℹ️

anikethsaha:
Trying tests for `config`

[...]

PR #1063, opened 2019-08-25
tests: added tests for config

anikethsaha:
Test

Yes

No

**Summary**
Added tests for config https://github.com/webpack/webpack-cli/issues/1012

No

**Other information**
More configs tests (advance) are coming soon. Just want to check if this way is fine by the team


test/config/basic/basicConfig.test.js
```diff
@@ -0,0 +1,23 @@
...
+const { resolve } = require('path');
+const { run, extractSummary } = require('../../utils/test-utils');
+
+describe('config flag test : Basic config file', () => {
+    it('Should not throw any module not found', done => {
+        const { stdout } = run(__dirname, ['-c', resolve(__dirname, 'webpack.config.js')]);
+        const summary = extractSummary(stdout);
+        const outputDir = 'basic/bin';
+        const outDirectoryFromCompiler = summary['Output Directory'];
+        const outDirToMatch = outDirectoryFromCompiler
```
ematipico: This is a test done in Windows environment. This will fail in Unix environment. I'd suggest to test it in other way

test/config/empty/emptyConfig.test.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+const { stat } = require('fs');
+const { resolve } = require('path');
+const { run, extractSummary } = require('../../utils/test-utils');
+
+describe('config flag test : Empty config file', () => {
+    it('compile but throw missing entry module error', done => {
+        const { stdout, stderr } = run(__dirname, ['-c', resolve(__dirname, 'webpack.config.js')]);
+        const summary = extractSummary(stdout);
+        const outputDir = 'empty/bin';
```
ematipico: Also this test has to be reviewed

anikethsaha: done

test/config/basic/basicConfig.test.js
anikethsaha: done

test/config/basic/basicConfig.test.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
+const { stat } = require('fs');
+const { resolve, sep } = require('path');
+const { run, extractSummary } = require('../../utils/test-utils');
+
+describe('config flag test : Basic config file', () => {
```
ematipico: Each test (`it`) should not have dependencies from each other. Instead here, the second and the second have a dependency from the first one.

I'd suggest to use the `run` function for each `it`.

anikethsaha: Actually here, I am trying to use the same instance of the run method to test each part of its result.
they will use that same run instance in a single test suite. 
I think a single test suite should contain a single instance of the run method.
Is it fine or should I change it then !?  
@ematipico

test/config/basic/bin/bundle.js
```diff
@@ -0,0 +1,100 @@
+/******/ (function(modules) { // webpackBootstrap
```
evenstensberg: Should not be committed

anikethsaha: okay I will remove it !

test/config/basic/basicConfig.test.js
ematipico: Or you wrap everything in one single `it` and there you do your assertions or you split in different `it` but you don't share same execution instance. You're using a mix which could case issue

test/config/basic/bin/bundle.js
anikethsaha: @evenstensberg should I add `test/**/bin/` in `.gitignore` ?

test/config/basic/basicConfig.test.js
anikethsaha: ohkay I will add everything in a single `it` .

test/config/basic/bin/bundle.js
ematipico: Add a `.gitignore` inside your `config` folder. It's better, so we don't include files by accident.

anikethsaha: done

PR #1065, opened 2019-08-26
fix: support both webpack versions

evenstensberg:
https://github.com/webpack/webpack-cli/issues/1023
Yes
N/A
**Summary**

Tries to resolve validation for both versions.

no

**Other information**

@sokra could you verify? Tested locally, removed the `if/else` but you might be aware of anymore pitfalls.


bin/utils/convert-argv.js
```diff
@@ -137,11 +136,20 @@ module.exports = function(...args) {
 
 	function processConfiguredOptions(options) {
 		if (options) {
-			const webpackConfigurationValidationErrors = validateSchema(webpackConfigurationSchema, options);
-			if (webpackConfigurationValidationErrors.length) {
-				const error = new WebpackOptionsValidationError(webpackConfigurationValidationErrors);
-				console.error(error.message, `\nReceived: ${typeof options} : ${JSON.stringify(options, null, 2)}`);
-				process.exit(-1); // eslint-disable-line
+			let error;
```
ematipico: Could you make a function out of it and export it? So we can test it and make sure 100% that it works.

evenstensberg: yup will do

evenstensberg:
Thanks for reviewing @ematipico . Singing this of and releasing 💙

PR #1074, opened 2019-09-01
fix: fixed the entry path error to run for all format and test path splitting

anikethsaha:
bugfix

Yes

Not required

**Summary**
**The bug**
with `entry` flag or without any flag and just passing the entry file, if we give input like these
- `path/to/entry.js`
- `path/to/entry` (having `entry.js` as the entry file )
will fail as in [this line](https://github.com/webpack/webpack-cli/blob/next/lib/groups/basic.js#L47) it will check with reference to `index.js` and so on.

**Step to reproduce**
- run command in the webpack-cli cloned - `node cli ./test/config/basic/a` or  `node cli ./test/config/basic/a.js`
- see the error

**Fix**
I fixed this in the `resolveFilePath` method of `group-helper` file, here if the 2nd parameter is not passed then it will check for the `filename` string and checking if the last part of the string is present as a file. if not the last part is folder and use `index.js`.
If present, then add `.js` extension if it not provided else simply use it.

**Test**
Tested this against inputs like
-  path/entry.js 
-  path/entry 
-  path
-  path/index.js 
-  path/index 

No

**Other information**
NA


lib/utils/group-helper.js
```diff
@@ -115,6 +115,29 @@ class GroupHelper {
...
+                        join(
+                            fileNameArray.slice(0, fileNameArray.length - 1).join('/'),
+                            defaultValue.includes('.js') ? defaultValue : defaultValue + ".js"
+                        )
+                    )
+                );
+                defaultValue = isLastNameAFile ? defaultValue : 'index.js';
+                defaultValue = defaultValue.includes('.js') ? defaultValue : defaultValue + ".js"
+                filename = (defaultValue == 'index.js' && (fileNameArray[fileNameArray.length - 1] != 'index' && fileNameArray[fileNameArray.length - 1] != 'index.js')) ? fileNameArray.join('/') : fileNameArray.slice(0, fileNameArray.length - 1).join('/')
+            }
```
jamesgeorge007: All these chained conditionals makes the code block a bit hard to read :thinking:

anikethsaha: yeah kind of, do you have any better refactoring in mind!? would love to get some help then !

anikethsaha:
> Do you mind not having the tests in lib, but rather in test?

@evenstensberg wherein `lib`? like inside `lib/test/`

lib/utils/group-helper.js
ematipico: It's really hard to understand what you try to achieve in the new code. Why don't you make a separate function and try to use the old fashion if/else?

anikethsaha: like to extract everything inside this if statement to a separate function? or just this line Line number 132 ?

ematipico: All of it. Having a function with a proper name would help to understand what we try to achieve.

anikethsaha: done

anikethsaha:
@ematipico @evenstensberg Please check if this one is fine!
@jamesgeorge007 hopes this one is much cleaner!

anikethsaha:
> Could you have every test with the same folder structure within the tests folder? Thanks!

@evenstensberg  I didnt get this !. You want all the tests to be in the `tests` folder in the root right? Currently its there only but inside `lib-utils`  so do you want that folder removed?

evenstensberg:
Tests under the folder `tests`, that's it

anikethsaha:
in `tests/group-helper/` here ? Cause here sample folder and `index.js` is also present and then simply keeping it in `tests/` may create conflicts in the future for new tests

anikethsaha:
@evenstensberg please check if this is fine?

test/group-helper.test.js
```diff
@@ -0,0 +1,52 @@
+const { parse, sep } = require('path');
```
evenstensberg: make this under `test/lib/my/path/as/it/is/in/lib`

anikethsaha: @evenstensberg I kept in `test/lib/utils/` in the latest commit. Please check if this is fine !

anikethsaha:
cc @evenstensberg

lib/utils/group-helper.js
```diff
@@ -1,6 +1,22 @@
 const { resolve, join } = require('path');
 const { existsSync } = require('fs');
 
+const parseAndValidatedFilename = pathToResolve => {
```
jamesgeorge007:
```suggestion
const parseAndValidateFilename = pathToResolve => {
```

lib/utils/group-helper.js
```diff
@@ -115,6 +131,17 @@ class GroupHelper {
         if (filename && Array.isArray(filename)) {
             return filename.map(fp => this.resolveFilePath(fp, defaultValue)).filter(e => e);
         }
+        if (filename) {
+            if (!defaultValue) {
+                [filename, defaultValue] = parseAndValidatedFilename(filename);
```
jamesgeorge007:
```suggestion
                [filename, defaultValue] = parseAndValidateFilename(filename);
```

lib/utils/group-helper.js
```diff
@@ -1,6 +1,22 @@
...
 const { existsSync } = require('fs');
 
+const parseAndValidatedFilename = pathToResolve => {
+    const pathToArray = pathToResolve.split('/');
+    let pathBeforeFileName;
+    let fileName = pathToArray[pathToArray.length - 1];
+    const isLastNameAFile = existsSync(resolve(join(pathToArray.slice(0, pathToArray.length - 1).join('/'), fileName.includes('.js') ? fileName : fileName + '.js')));
+    fileName = isLastNameAFile ? fileName : 'index.js';
+    fileName = fileName.includes('.js') ? fileName : fileName + '.js';
+    const isPathNotHavingIndexFile = pathToArray[pathToArray.length - 1] != 'index' && pathToArray[pathToArray.length - 1] != 'index.js';
```
jamesgeorge007:
```suggestion
    const isPathNotHavingIndexFile = !['index', 'index.js'].includes(pathToArray.slice(-1).pop())
```

lib/utils/group-helper.js
```diff
@@ -1,6 +1,22 @@
 const { resolve, join } = require('path');
 const { existsSync } = require('fs');
 
+const parseAndValidatedFilename = pathToResolve => {
+    const pathToArray = pathToResolve.split('/');
+    let pathBeforeFileName;
+    let fileName = pathToArray[pathToArray.length - 1];
+    const isLastNameAFile = existsSync(resolve(join(pathToArray.slice(0, pathToArray.length - 1).join('/'), fileName.includes('.js') ? fileName : fileName + '.js')));
+    fileName = isLastNameAFile ? fileName : 'index.js';
+    fileName = fileName.includes('.js') ? fileName : fileName + '.js';
```
jamesgeorge007: This could be turned into a helper method as it is being reused.

lib/utils/group-helper.js
```diff
@@ -1,6 +1,22 @@
 const { resolve, join } = require('path');
 const { existsSync } = require('fs');
 
+const parseAndValidatedFilename = pathToResolve => {
+    const pathToArray = pathToResolve.split('/');
+    let pathBeforeFileName;
+    let fileName = pathToArray[pathToArray.length - 1];
```
jamesgeorge007:
```suggestion
    let fileName = pathToArray.slice(-1).pop();
```
This seems to be much concise.

lib/utils/group-helper.js
jamesgeorge007:
```js
    const isPathNotHavingIndexFile = ['index', 'index.js'].every(file => pathToArray.slice(-1).pop() !== file)
```
What do you say about this version?

lib/utils/group-helper.js
anikethsaha: May be! I guess we can export it from here itself. what do you think! @evenstensberg what do you think of adding it to the separate utils file or exporting from here itself?

lib/utils/group-helper.js
anikethsaha: > `   const isPathNotHavingIndexFile = ['index', 'index.js'].every(file => pathToArray.slice(-1).pop() !== file)`

it looks good but I guess as far as time factor is concerned, this is more functionally heavy than the current one as there are too many calls and we can do that even just jumping to those indexes.
What do you think?

lib/utils/group-helper.js
anikethsaha: a somewhat similar opinion like this https://github.com/webpack/webpack-cli/pull/1074#discussion_r327230213

anikethsaha: Anyway, the functional approach is good way to refactor. If you want then I guess I will change it! 😃

anikethsaha:
cc @evenstensberg @ematipico Fixed CI errors 👍

anikethsaha:
cc @evenstensberg

lib/utils/group-helper.js
evenstensberg: I think the proposed solution is cleaner

test/entry/defaults-index/entry-multi-args.test.js
```diff
@@ -10,8 +10,9 @@ describe('single entry flag index present', () => {
         const { stdout, stderr } = run(__dirname);
         const summary = extractSummary(stdout);
         const outputDir = 'entry/defaults-index/bin';
-
-        expect(summary['Output Directory']).toContain(outputDir);
+        const outDirectoryFromCompiler = summary['Output Directory'].split(sep);
+        const outDirToMatch = outDirectoryFromCompiler.slice(outDirectoryFromCompiler.length - 3, outDirectoryFromCompiler.length).join('/');
```
evenstensberg: need a util for these

lib/utils/group-helper.js
anikethsaha: Sure ! will change it soon

test/entry/defaults-index/entry-multi-args.test.js
anikethsaha: test utils ? like extracting these lines ?

lib/utils/group-helper.js
anikethsaha: done

test/entry/defaults-index/entry-multi-args.test.js
evenstensberg: yup, make them a utility function in `test-utils`, use that function instead for all occurences

test/lib/utils/group-helper.test.js
```diff
@@ -0,0 +1,52 @@
+const { parse, sep } = require('path');
+const GroupHelper = require('../../../lib/utils/group-helper');
+
+describe('group-helper tests', () => {
+    describe('resolveFilePath', () => {
+        it('should return correct path : sample/entry.js', done => {
+            const GroupHelperOptions = [{ entry: 'test/lib/utils/sample/entry.js' }, { progress: 'bar' }, { sourcemap: 'eval' }, { dev: false }, { prod: true }];
+            const GroupHelperObj = new GroupHelper(GroupHelperOptions);
+            const entryPath = GroupHelperObj.resolveFilePath(GroupHelperOptions[0].entry);
+            const { dir, base } = parse(entryPath);
```
evenstensberg: Use `afterEach` to extract the dir and base, if possible

test/lib/utils/group-helper.test.js
```diff
@@ -0,0 +1,52 @@
...
+            const entryOutput = dir.split(sep).concat(base);
+            expect(entryOutput.join('/').includes(GroupHelperOptions[0].entry)).toBe(true);
```
evenstensberg: I think this functionality also would need a better abstraction

test/utils/test-utils.js
```diff
@@ -178,4 +178,18 @@ function copyFile(testCase, file) {
...
-module.exports = { run, runWatch, runAndGetWatchProc, extractSummary, appendDataIfFileExists, copyFile, copyFileAsync, appendDataToMultipleIfFilesExists };
+/**
+ * It splits the arr passed as an argument using node.js's seperator for supporting cross-windows
+ * matching for the paths, and slice the elements from the idx passed and
+ * joins them to create a string which can be matched against the output directory
+ * @param {Array} arr - the array element, in this case the outputDir array from summary
+ * @param {Number} idx - index from the end to start slicing
+ * @returns {string} - output directory to test against
+ */
+function outDirExtractor(arr, idx) {
```
evenstensberg: Is there a way to find the idx from the array? I think so, and it would make the abstraction better

anikethsaha: Actually idx is kind of depenedent on the `outdir` where it needs to match against. 
Like here https://github.com/webpack/webpack-cli/pull/1074/files#diff-b82e1bc49af5272a0d78bc41925fb337R10 , idx needs to be 3 as there are 3 sub paths it's checking, so either that string need to be passed and then checking the len after slitting it with `\`,  so either way we need 2nd argument to pass  I am ready to change if you want that string to be passed instead to index

test/lib/utils/group-helper.test.js
anikethsaha: I will try to refactore it.

test/utils/test-utils.js
anikethsaha: cc @evenstensberg

test/lib/utils/group-helper.test.js
anikethsaha: `afterEach` of jest ? 
I didn't get how it will work! You want to use this line
```js
const { dir, base } = parse(entryPath);
```
in `afterEach` ?
Then how'd I take those values inside the test cases?

evenstensberg: Declare them as global variables and reassign them, so we can reduce duplicate code 👍

test/lib/utils/group-helper.test.js
evenstensberg: Awesome!

test/utils/test-utils.js
evenstensberg: If you iterate the array and try to concat those? Maybe even check if one those three paths are the correct one, if not present, fail the test?

anikethsaha: > If you iterate the array and try to concat those?

I didn't get this !?

the `arr` array will be joined to form a string which will then test against the `outDir` ! so we need both right?

Did you mean removing the `outDir` and just testing the `arr` (which is the output came from the summary) by checking the file system directories? if so, wouldn't it be a lot of work which can be reduced by doing manually?

evenstensberg: You'll always have a knowing of these output directories, so it's not always needed to provide an array.

anikethsaha: @evenstensberg I didn't refactore this as I am still not clear about the approach. 
Should I remove this function and also those assertions where this is being used.

test/lib/utils/group-helper.test.js
```diff
@@ -0,0 +1,57 @@
...
+ * @param {string} entryPathFromOutput - entry path came from the output of the group helper
+ * @param {string} EntryFromGroupHelperInput - entry path passed as an input to the group helper
+ * @returns {boolean} - returns whether entry paths are matched or not
+ */
+const isEntryOuputMatchedInput = (entryPathFromOutput, EntryFromGroupHelperInput) => {
+    let dir = parse(entryPathFromOutput).dir;
+    let base = parse(entryPathFromOutput).base;
+    let entryOutput = dir.concat(`/${base}`).split(sep);
+    return entryOutput.join('/').includes(EntryFromGroupHelperInput);
+};
```
anikethsaha: I didnt place this function to test-utils because I think its quite specific to `group-helper` only. 
If you want this to be in utils, please let me know I will move it

evenstensberg:
Hi, have you tested the issue reporting the bug against the current next branch? I do not get the same error that you are experiencing on the branch right now

anikethsaha:
Yeah I have checked this with next branch. It's coming as bug in my machine

evenstensberg:
As far as I've understood, if you entry is a file, but you omit the file extention, it will fail? What will fail, the test or the functionality? If the functionality fails, that means we'll need to check if that path is either a directory or a file.

If it's a file, we can set that to the new entry point, just adding the extention, which I believe we are already doing.

If it's a directory, we will need to set a default file inside the interpreted directory `a/b` that will turn into `a/b/bundle.js`. I believe that is the current functionality?

anikethsaha:
It will throw an error while compiling like this 

image

evenstensberg:
Could you make a reproduction in a seperate gh repo?

anikethsaha:
you can try with having a folder name `a` and having `index.js` inside it and then giving `--entry a` should leave this error

evenstensberg:
Could you make the repository? It would be easier to follow

anikethsaha:
ok will try to re-create it

ematipico:
@anikethsaha  what's the situation of this PR? I am trying to catch up

ematipico:
@anikethsaha  I need to close this PR unfortunately. It's been in stale for too long and things changed. If you wish to resume your work, feel free to open a new PR!

anikethsaha:
@ematipico sorry for this much delayed!
I actually forgot the context of this PR. SO if anyone wants to take this, feel free to work on this !

snitin315:
@anikethsaha I would like to work on this. Thanks!

PR #1090, opened 2019-10-06
[Test]: added tests for output flag

anikethsaha:

Tests

yup

Not required

**Summary**

#1012 

Added tests for two cases
- not having `webpack.config.js` and using `--output` flag to output the file.
- having `output` property in `webpack.config.js` and then using the `--output` file to overwrite that 

nop

**Other information**
NA

Update
---
I added tests for the following cases
- `--output <directory-name>`
- `--output <Empty>`
- In `overwrite-and-chunk` folder there are three tests
  - simple use of output flag `--output ./bin/a.bundle.js` to overwrite the output file mentioned in the `webpack.config.js` which is `bundle.js`
  - there is a `webpack.single.config.js` file present which is having multiple entries and output is in the single bundle file. But this is been overwrite by the `--output [name].bundle.js` flag.
  - There is one more config file present named `webpack.chunk.config.js` with the same entries as above but having output as chunks. In the CLI flag using a single bundle file for output which will thow error as ` Multiple chunks emit assets to the same filename`.


test/output/basic/output.test.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+const { stat } = require('fs');
+const { resolve, sep } = require('path');
+const { run, extractSummary } = require('../../utils/test-utils');
+
+describe('Simple Output flag ', () => {
+    it('should create a file passed in the output file', done => {
+        const { stdout } = run(__dirname, ['--entry', './a.js', '--output', './bin/bundle.js']);
```
evenstensberg: You'll need to pass something else here, because output defaults to this path. Try setting an unique name and directory.

anikethsaha: I guess this simple one is not needed. 
I for the second one, the flag path should override the path mentioned in the webpack.config,js , so I tested that.

evenstensberg: Look at the tests that use `binary/something` instead

evenstensberg: Nice work so far, looks promising! 💙

anikethsaha: > Try setting an unique name and directory.

Like these
- `--output bin`
- `--output <empty>`
- `--output [name].bundle.js`

??

evenstensberg: yes those would be great! Nice!

anikethsaha: okay will add them 👍

anikethsaha: For `--output [name].bundle.js` , it should give `a.bundle.js` for entry of `a.js` right ? or `main.bundle.js` which is I am getting ?

evenstensberg: I'd test those 3 options you have, then we'll work on that in another PR

anikethsaha: I updated the tests as follows
- `--output <directory-name>`
- `--output <Empty>`
- In `overwrite-and-chunk` folder there are three tests
  - simple use of output flag `--output ./bin/a.bundle.js` to overwrite the output file mentioned in the `webpack.config.js` which is `bundle.js`
  - there is a `webpack.single.config.js` file present which is having multiple entries and output is in the single bundle file. But this is been overwrite by the `--output [name].bundle.js` flag.
  - There is one more config file present named `webpack.chunk.config.js` with the same entries as above but having output as chunks. In the CLI flag using a single bundle file for output which will thow error as ` Multiple chunks emit assets to the same filename`.

anikethsaha: @evenstensberg is there any way to check for a substring in jest matchers. Cause [`here`](https://github.com/webpack/webpack-cli/pull/1090/files#diff-e0d8aa47f72006fc56bedfe5c82494f3R41), Its a two-line error log, And the hexagon(webpack) symbol is not getting printed for window's terminal. It would be enough to check for the substring of that message. 
If there isn't any matchers from jest (Which I checked and couldn't found any), then I will custom comparing condition.

anikethsaha: cc @evenstensberg

anikethsaha:
cc @evenstensberg  https://github.com/webpack/webpack-cli/pull/1090#discussion_r331811746

anikethsaha:
cc @evenstensberg

PR #1098, opened 2019-10-15
feat: help for commands

### LINKED ISSUE #1053: [ENHANCEMENT]: Add help for core commands
We can have help commands for individual core commands for ex `webpack-cli create --help`

Check if the help flag is present along with the core command and then print relative command.

NA

**Additional context**
Would love suggestions on this and will implement once agreed.

rishabh3112:
feature

Pending

No

**Summary**
Closes #1053 

No

**Other information**
Current status:
image


rishabh3112:
cc @evenstensberg @ematipico 
If output is satisfying, will write up tests for it.

lib/webpack-cli.js
```diff
@@ -162,8 +162,14 @@ class WebpackCLI extends GroupHelper {
         return await require('./commands/external').run(command.name, ...args);
     }
 
-    runHelp() {
+    runHelp(args) {
         const HelpGroup = require('./groups/help');
+        const Commands = ['create', 'loader', 'plugin', 'migrate', 'info', 'serve'];
```
evenstensberg: We need a better way of extracting these, maybe we can do the same for flags?

lib/groups/help.js
```diff
@@ -20,6 +20,16 @@ class HelpGroup {
         process.stdout.write(`\nwebpack ${webpack.version}\n`);
     }
 
+    outputCommand(command) {
```
evenstensberg: I think that we can intertwine this with the `outputHelp` method, so we avoid duplicate functionality?

lib/webpack-cli.js
rishabh3112: Would extracting from `cli-flag.js` be fine?

lib/groups/help.js
rishabh3112: Ok!
Just one concern functionalites of both differs alot (one focusing on options and flags, others on descriptions and examples ) and thus duplication shouldn't be a concern imo.
Thanks.

lib/webpack-cli.js
rishabh3112: Did that in latest commits.

lib/groups/help.js
evenstensberg: I think that it should be one, maybe abstracted inside the `outputHelp` method itself when you validate in `bootstrap.js`. It's more clean as an abstraction rather than adding more duplicate code

rishabh3112: Okay, did that in latest commit.

lib/groups/help.js
```diff
@@ -8,8 +8,17 @@ class HelpGroup {
     l() {
         const c = chalk.cyan;
     }
-    outputHelp() {
-        process.stdout.write(this.run().outputOptions.help);
+    outputHelp(command=null) {
+        if (command != null) {
```
evenstensberg: `!==`

lib/utils/commands.js
```diff
@@ -0,0 +1,10 @@
+const arguments = require('./cli-flags');
```
evenstensberg: This is nice, but instead of mutating, maybe `Object.keys` the name and map/filter the names instead?

lib/webpack-cli.js
```diff
@@ -162,9 +162,14 @@ class WebpackCLI extends GroupHelper {
         return await require('./commands/external').run(command.name, ...args);
     }
 
-    runHelp() {
+    runHelp(args) {
         const HelpGroup = require('./groups/help');
-        return new HelpGroup().outputHelp();
+        const Commands = require('./utils/commands').names;
```
evenstensberg: You can do `Object.keys` here as well?

lib/groups/help.js
```diff
@@ -8,8 +8,17 @@ class HelpGroup {
...
+            const commands = require('../utils/commands').info;
+            const options = commands.find(cmd => (cmd.name == command || cmd.alias == command));
```
evenstensberg: `===`

lib/utils/commands.js
rishabh3112: I couldn't follow how `Object.keys` would be helpful here as `cli-flags.js` contains all commands in an array instead of a map.

rishabh3112:
Okay, would do changes as in your comments by tommorow hopefully.

evenstensberg:
No worries

lib/utils/commands.js
rishabh3112: @evenstensberg please let me know how to do it.

evenstensberg: Object.keys iteration instead of using array.forEach to create those values in another array. Just create one directly

rishabh3112: Thanks for guidance, Learnt new tricks and about mutability from this task.
Did that in latest commit.

rishabh3112:
Commands and their alias specified in `cli-flags.js`.

evenstensberg:
I'm not being able to reproduce this PR, what command are you running to output the help?

rishabh3112:
Current state of PR
image
( Ran from root )

evenstensberg:
Nice, thanks for clearing out! Could you add descriptions to the flags as well and show a more detailed description on those? If you have multiple flags with help, output entire help option, if there's one flag, output a single help flag, like in the description

evenstensberg:
And also, really nice, thanks for elaborating, it looks amazing!

rishabh3112:
I think flag based helps aren't needed as they are just extra liabilities. Instead I suggest to have flags for each command in a single place.
> And also, really nice, thanks for elaborating, it looks amazing!

Thanks, no need to mention.

rishabh3112:
Now it looks like:
image

evenstensberg:
> I think flag based helps aren't needed as they are just extra liabilities.  

why?

rishabh3112:
As usage of flag is self explanatory from the help ouput of a command.

evenstensberg:
It is, but maybe people don't know what the flag is supposed to do and need more information about what it does without opening the docs page

rishabh3112:
That is why description of each flag is shown at the command it self and docs has nothing extra as far as commands of `webpack cli` are concerned (which are mono repo repositories), which is the focus of this PR.

evenstensberg:
Could you write this for flags as well? I'm not concerned about the commands, but the flags in addition

rishabh3112:
Yes sure, will open a separate PR as webpack supports a lot of flags and this PR isn't focussing on that.

evenstensberg:
Could you open a PR against this branch?

evenstensberg:
@ematipico what do you think?

ematipico:
Different PR might be better IMHO

rishabh3112:
@evenstensberg and @ematipico, let me know what's left in this PR.

PR #1100, opened 2019-10-20
tests(serve): add serve tests

Loonride:
Tests

Yes

N/A

**Summary**

/cc @evilebottnawi 

This adds basic tests for the `serve` command, confirming that it runs and applies the provided CLI flags.

There is a slightly hacky method I used to simulate having the package `@webpack-cli/serve` in your node_modules folder. Better ideas for how to do this are welcome.

No

**Other information**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@
 
 # Node dependencies
 node_modules
+!test/config/**/node_modules
 
 # npm-debug log
 npm-debug.*
--- a/packages/serve/package.json
+++ b/packages/serve/package.json
@@ -23,14 +23,6 @@
     "@types/node": "12.7.2",
     "typescript": "3.5.3"
   },
-  "peerDependencies": {
-    "webpack-dev-server": ">=4.0.0"
-  },
-  "peerDependenciesMeta": {
-    "webpack-dev-server": {
-      "optional": true
-    }
-  },
   "scripts": {
     "build": "tsc",
     "watch": "npm run build && tsc -w"
--- a/packages/serve/package.json
+++ b/packages/serve/package.json
@@ -14,14 +14,14 @@
     "@webpack-cli/utils": "0.2.2",
     "chalk": "2.4.2",
     "cross-spawn": "6.0.5",
-    "inquirer": "6.5.1"
+    "inquirer": "6.5.1",
+    "webpack-dev-server": "^3.9.0"
   },
   "devDependencies": {
     "@types/cross-spawn": "6.0.0",
     "@types/inquirer": "6.5.0",
     "@types/node": "12.7.2",
-    "typescript": "3.5.3",
-    "webpack-dev-server": "^3.8.2"
+    "typescript": "3.5.3"
   },
   "peerDependencies": {
     "webpack-dev-server": ">=4.0.0"
--- a/packages/serve/package.json
+++ b/packages/serve/package.json
@@ -23,6 +23,14 @@
     "@types/node": "12.7.2",
     "typescript": "3.5.3"
   },
+  "peerDependencies": {
+    "webpack-dev-server": ">=3.0.0"
+  },
+  "peerDependenciesMeta": {
+    "webpack-dev-server": {
+      "optional": false
+    }
+  },
   "scripts": {
     "build": "tsc",
     "watch": "npm run build && tsc -w"
--- a/packages/serve/package.json
+++ b/packages/serve/package.json
@@ -14,14 +14,14 @@
     "@webpack-cli/utils": "0.2.2",
     "chalk": "2.4.2",
     "cross-spawn": "6.0.5",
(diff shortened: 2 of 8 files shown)


test/serve/serve-basic.test.js
```diff
@@ -0,0 +1,33 @@
...
+const { runWatch } = require('../utils/test-utils');
+
+jest.setTimeout(20000);
+
+const runServe = args => {
+    return runWatch(path.resolve(__dirname, '../config/serve-basic'), ['serve'].concat(args), false, 'main');
+};
+
+describe('basic serve usage', () => {
+    // eslint-disable-next-line space-before-function-paren
```
rishabh3112: Consider ignoring this rule for the whole file.

packages/serve/package.json
```diff
@@ -14,14 +14,14 @@
...
-    "webpack-dev-server": "^3.8.1"
+    "inquirer": "6.5.1"
   },
   "devDependencies": {
     "@types/cross-spawn": "6.0.0",
     "@types/inquirer": "6.5.0",
     "@types/node": "12.7.2",
-    "typescript": "3.5.3"
+    "typescript": "3.5.3",
+    "webpack-dev-server": "^3.8.2"
```
evilebottnawi: I think it should be in `dependencies` because this package can't be work without `webpack-dev-server`

packages/serve/package.json
```diff
@@ -14,14 +14,14 @@
...
   },
   "peerDependencies": {
     "webpack-dev-server": ">=4.0.0"
```
evilebottnawi: No need it here, apparently we missed it somewhere. 

We should put it:
```
"peerDependenciesMeta": {
    "@webpack-cli/serve": {
      "optional": true
    }
  },
```

in webpack-cli main package

packages/serve/package.json
Loonride: @evilebottnawi I thought we wanted it as a peer dependency like I did here:

https://github.com/webpack/webpack-cli/blob/d7799ade420cc31cfe29d1ca1e8cd58368f6a08a/packages/serve/package.json#L26-L33

Then we also need it as a dev dependency just so that we can test it.

packages/serve/package.json
Loonride: @evilebottnawi Can we do this in a separate PR? And should it be done with all webpack-cli packages? (e.g. `@webpack-cli/...`)

packages/serve/package.json
evilebottnawi: @webpack-cli/serve require `webpack-dev-server` to use so we should put it in `dependencies`

packages/serve/package.json
evilebottnawi: > Can we do this in a separate PR?

Yes, just don't forget

> And should it be done with all webpack-cli packages? 

Maybe, need look in other packages

evenstensberg:
@evilebottnawi mind guiding this PR?

evilebottnawi:
@evenstensberg need fix the problem above, feel free to review too

packages/serve/package.json
evenstensberg: You can have the seperate PRs towards this one @Loonride ?

Loonride: > You can have the seperate PRs towards this one @Loonride ?

Sounds good.

evenstensberg: Let us know when you've put one up

.gitignore
```diff
@@ -11,6 +11,7 @@
 
 # Node dependencies
 node_modules
+!test/config/**/node_modules
```
evenstensberg: what is this for?

packages/serve/package.json
```diff
@@ -15,22 +15,14 @@
...
-    "webpack-dev-server": "^3.8.1"
+    "webpack-dev-server": "^3.9.0"
   },
   "devDependencies": {
     "@types/cross-spawn": "6.0.0",
     "@types/inquirer": "6.5.0",
     "@types/node": "12.7.2",
     "typescript": "3.5.3"
   },
-  "peerDependencies": {
```
evenstensberg: Could you set this to `>= 3` and optional to false? After that, we are good! 

CC @evilebottnawi

packages/serve/package.json
```diff
@@ -23,6 +23,14 @@
...
     "typescript": "3.5.3"
   },
+  "peerDependencies": {
+    "webpack-dev-server": ">=3.0.0"
+  },
+  "peerDependenciesMeta": {
+    "webpack-dev-server": {
+      "optional": false
+    }
+  },
```
evilebottnawi: I think we can simplify this and just remove, because something in `peerDependencies` always mean non optional :smile:

Loonride: I guess we can make adjustments in another PR.

PR #1101, opened 2019-10-22
[Minor] code refactors 

sendilkumarn:

Minor code refactoring.

> Wow the PR number is binary format of `13`, just in case if you are curious 😉 

NA

NA

**Summary**

Few things will provide us minor performance wins and simply to get back in pace with the code base.

NA

**Other information**

### DIFF
--- a/lib/utils/compiler.js
+++ b/lib/utils/compiler.js
@@ -175,8 +175,8 @@ function compilerCallback(compiler, err, stats, lastHash, options, outputOptions
     }
 }
 
-function getCompiler(opts) {
-    return webpack(opts.options);
+function getCompiler(options) {
+    return webpack(options);
 }
 
 async function webpackInstance(opts) {
@@ -186,13 +186,12 @@ async function webpackInstance(opts) {
         throw new Error(processingErrors);
     }
 
-    const compiler = getCompiler(opts);
+    const compiler = getCompiler(options);
     let lastHash = null;
 
-    const ProgressPlugin = webpack.ProgressPlugin;
-    const progressPluginExists = options.plugins ? options.plugins.find(p => p instanceof ProgressPlugin) : false;
-    if (progressPluginExists) {
-        options.plugins = options.plugins.filter(e => e !== progressPluginExists);
+    const { ProgressPlugin } = webpack;
+    if (options.plugins) {
+        options.plugins = options.plugins.filter(e => e instanceof ProgressPlugin);
     }
 
     if (outputOptions.interactive) {
--- a/lib/utils/group-helper.js
+++ b/lib/utils/group-helper.js
@@ -20,7 +20,7 @@ class GroupHelper {
         });
     }
     mergeRecursive(acc, curr) {
-        for (var p in curr) {
+        for (const p in curr) {
             try {
                 // Property in destination object set; update its value.
                 if (curr[p].constructor == Object) {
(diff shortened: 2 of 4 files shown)


lib/utils/zero-config.js
```diff
@@ -1,40 +1,38 @@
...
         return 'development';
-    } else if (!optionsDevFlag && !optionsProdFlag && mode) {
+    } else if (mode) {
         return mode;
     }
     return 'production';
 }
 
-module.exports = function setDefaultConfigBasedOnEnvironment(webpackObject) {
+function getConfigurations(options, outputOptions) {
```
evenstensberg: I think the old filename is more clear to what it is doing

lib/utils/compiler.js
```diff
@@ -186,13 +186,12 @@ async function webpackInstance(opts) {
         throw new Error(processingErrors);
     }
 
-    const compiler = getCompiler(opts);
+    const compiler = getCompiler(options);
     let lastHash = null;
 
-    const ProgressPlugin = webpack.ProgressPlugin;
```
evenstensberg: Reason for doing so is mutation on several runs, I think it's fine, but if a regression comes up, this is the reason. We have to find an exact match, not instance match

lib/utils/group-helper.js
```diff
@@ -40,7 +40,7 @@ class GroupHelper {
         if (!arr) {
             return;
         }
-        var result = {};
+        const result = {};
```
evenstensberg: result is mutated

lib/utils/zero-config.js
sendilkumarn: The old name is still there, this is just stubbed out function.

evenstensberg: Noticed, thx for ensuring!

lib/utils/group-helper.js
sendilkumarn: This prevents the re-assignment and I think we will need this.

ematipico: We can add the eslint rule `prefer-const` to be 100% sure (if it's not there)

evenstensberg: Hmm, I'd say adding a property to a declared object is mutation but I'm probably wrong. Small thing though, nothing blocking

ematipico: Well this is JavaScript, we don't have immutable structures. In JavaScript what it counts it's just the reassignment and if the object is not reassigned, `const` is the way to go.
But you're right. But in JS world that distinction doesn't exist :D

lib/utils/zero-config.js
```diff
@@ -1,40 +1,38 @@
 function getEnvFromOptionsAndMode(mode, optionsObject) {
-    const optionsDevFlag = optionsObject.dev;
-    const optionsProdFlag = optionsObject.prod;
+    const { dev, prod } = optionsObject;
 
     if (process.env.NODE_ENV && process.env.NODE_ENV === 'production' || process.env.NODE_ENV === 'development') {
```
jamesgeorge007:
```js
if (process.env.NODE_ENV && ['production', 'development'].includes(process.env.NODE_ENV)) {
```
Maybe :thinking:

sendilkumarn: I assume, includes will be slower here. (eventhough it is very negligible but still)

ematipico: create a new array just for a comparison? Not needed I think

Issue #1095, opened 2019-10-11
[Feature Request]: --no-cache flag for webpack@5


Andarist:
**Is your feature request related to a problem? Please describe.**
I often (well, from time to time) find myself editing node_modules by hand to debug a particular problem with some dependency. And webpack@5 with its persistent caching might make this process rather difficult - because it would require changing a config.

Actually you already have `--cache=false` here, but I'm not sure if this right now can disable persistent caching and Tobias has kinda redirected me to here with this feature request. If this is already handled - then sorry for reporting and feel free to close this.

**Describe the solution you'd like**
a special CLI flag that would disable persistent caching for a current build

**Describe alternatives you've considered**
n/a

**Additional context**
https://github.com/webpack/webpack/issues/9802#issuecomment-541089703

anikethsaha:
This is actually good if it's not present already! It will be too handy for the user to use it from CLI
I don't think right now the persistent caching flag is there. Is it @evenstensberg?
But as far as the webpack@5 is concerned, I guess we can wait for the major release as that is currently in beta, and then it can be added in the CLI. What do you think?

evenstensberg:
We can add the cache flag for the new cli

Andarist:
Would be great to at least prepare a PR for this before webpack@5 hits stable, but please do whatever you see most appropriate here

evenstensberg:
Can do both

evenstensberg:
Fixed https://github.com/webpack/webpack-cli/releases/tag/v3.3.10

PR #1102, opened 2019-10-27
feat: add new no-cache flag and patch sec dep

evenstensberg:
Fixes https://github.com/webpack/webpack-cli/issues/1095

N/A
No, this needs to be done
**Summary**
Adds `no-cache` flag, (we do not check version here for setting the property)

No
**Other information**

- Might need to add validation against webpack version?
- Monorepo test are failing, might worth removing the entire mono repo at master branch? (Is that good for versioning against next? I think it's bad)

### DIFF
--- a/bin/config/config-yargs.js
+++ b/bin/config/config-yargs.js
@@ -117,6 +117,11 @@ module.exports = function(yargs) {
 				group: BASIC_GROUP,
 				requiresArg: true
 			},
+			"no-cache": {
+				type: "boolean",
+				describe: "Disables cached builds",
+				group: BASIC_GROUP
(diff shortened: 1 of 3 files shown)


evenstensberg:
That would be great, but we need to write docs, migration guide and make sure it is stable

anikethsaha:
> Cli v4 will support only webpack v5

webpack v5 is in beta, isn't it ?
Why removing support for v4? I guess many projects will still be using webpack.v4 for quite a few time in the future. cause I still see projects using v2.
Will it be like webpack-cli v3 will be for webpack v4 and less ?

PR #1103, opened 2019-10-27
tests: configuration lookup

evenstensberg:
Bugfix and tests

Hehe :)
No
**Summary**
Adds tests for lookup

nah fam

**Other information**
N/A


lib/groups/config.js
```diff
@@ -93,20 +92,31 @@ class ConfigGroup extends GroupHelper {
...
-        
+
         let configOptions = [];
         if (configFiles) {
             // TODO: support mjs etc..
             const resolvedConfigurationFiles = this.requireConfig(configFiles.path);
             configOptions = resolvedConfigurationFiles;
+            if (configOptions && configFiles.path.includes('.webpack')) {
+                const currentPath = configFiles.path;
+                const parentContext = path.dirname(currentPath).split(sep).slice(0, -1).join('/');
```
ematipico: With this operation, what would be the `parentContext` value? Could you give some context so I can understand?

evenstensberg: parent context would be the parent directory that we will have the base of webpack from, if not the context prop on a configuration has been set. Since we have a `.webpack` folder, we assume that the user has their code in parent

PR #1105, opened 2019-11-03
chore: minor code refactoring

jamesgeorge007:
refactoring

Nope

N/A

**Summary**
Make use of object destructuring assignment approach and a couple of other minor tweaks.

Nope

**Other information**
N/A

### DIFF
--- a/lib/utils/commands.js
+++ b/lib/utils/commands.js
@@ -1,12 +1,11 @@
-const arguments = require('./cli-flags');
+const {commands} = require('./cli-flags');
 
-const info = arguments.commands;
-const names = Object.keys(info).map( i => {
-    if (info[i].alias) {
-        return [info[i].name, info[i].alias];
+const names = Object.keys(commands).map( i => {
+    if (commands[i].alias) {
+        return [commands[i].name, commands[i].alias];
     }
-    return [info[i].name];
+    return [commands[i].name];
 }).reduce((arr, val) => arr.concat(val), []);
 
 module.exports.names = names;
-module.exports.info = arguments.commands;
+module.exports.info = commands;
--- a/lib/utils/commands.js
+++ b/lib/utils/commands.js
@@ -1,10 +1,10 @@
 const { commands } = require('./cli-flags');
 
-const names = Object.keys(commands).map( i => {
-    if (commands[i].alias) {
-        return [commands[i].name, commands[i].alias];
+const names = Object.keys(commands).map( key => {
+    if (commands[key].alias) {
+        return [commands[key].name, commands[key].alias];
     }
-    return [commands[i].name];
+    return [commands[key].name];
 }).reduce((arr, val) => arr.concat(val), []);
 
 module.exports.names = names;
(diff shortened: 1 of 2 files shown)


lib/utils/commands.js
```diff
@@ -1,12 +1,11 @@
-const arguments = require('./cli-flags');
+const { commands } = require('./cli-flags');
 
-const info = arguments.commands;
-const names = Object.keys(info).map( i => {
-    if (info[i].alias) {
-        return [info[i].name, info[i].alias];
+const names = Object.keys(commands).map( i => {
```
ematipico: `i` can be called `key`

lib/webpack-cli.js
```diff
@@ -11,6 +11,11 @@ class WebpackCLI extends GroupHelper {
         this.groupMap = new Map();
         this.groups = [];
         this.processingErrors = [];
+        this.commandNames = {
+            'create': 'init',
```
ematipico: You can remove the apostrophe from the key, not needed

lib/webpack-cli.js
```diff
@@ -11,6 +11,11 @@ class WebpackCLI extends GroupHelper {
         this.groupMap = new Map();
         this.groups = [];
         this.processingErrors = [];
+        this.commandNames = {
```
ematipico: Main this could be called `defaultCommands`. What do you think?

lib/webpack-cli.js
```diff
@@ -152,19 +157,12 @@ class WebpackCLI extends GroupHelper {
...
-        } else if (command.name === 'plugin') {
-            command.name = 'generate-plugin';
-        }
-        return await require('./commands/external').run(command.name, ...args);
+        return await require('./commands/external').run(this.commandNames[command.name], ...args);
     }
 
     runHelp(args) {
         const HelpGroup = require('./groups/help');
-        const names = require('./utils/commands').names;
```
ematipico: What's the reason of this change?

jamesgeorge007: Object destructuring assignment makes the respective code snippet much concise.

lib/webpack-cli.js
jamesgeorge007: yeah, makes sense :+1:

lib/utils/commands.js
```diff
@@ -1,12 +1,11 @@
-const arguments = require('./cli-flags');
+const { commands } = require('./cli-flags');
 
-const info = arguments.commands;
-const names = Object.keys(info).map( i => {
-    if (info[i].alias) {
-        return [info[i].name, info[i].alias];
+const names = Object.keys(commands).map( key => {
+    if (commands[key].alias) {
+        return [commands[key].name, commands[key].alias];
```
evenstensberg: I think you could use destruct syntax here?

lib/webpack-cli.js
```diff
@@ -11,6 +11,11 @@ class WebpackCLI extends GroupHelper {
         this.groupMap = new Map();
         this.groups = [];
         this.processingErrors = [];
+        this.defaultCommands = {
```
evenstensberg: Nice. You can move this outside the class

ematipico:
Thank you! 
FYI, this is not a feature (`feat`). That's why I changed to a simple chore

PR #1108, opened 2019-11-04
Feature/add typings to cli

ematipico:

This PR adds JSDocs to the JavaScript files of the CLI to make sure they have proper types.

We already use TypeScript and we could use it to make sure that everything is well maintained by checking the JSDoc. The main webpack repo already uses it, we ca do the same.

I will need some help to identify the correct types.
I also found some bugs.

I removed `process.cliLogger` because I think it's anti-pattern to tarnish the main `process` object. I created a simple module and updated all the places where we use `process.cliLogger`.

No

Not needed. It creates documentation by itself

**Summary**
The motivation is basically to maintain a good quality of the code while other people can check and the code and understand the types of everything.

Also, it will prevent problems and bugs.

Possibly yes, that's why I created the PR on the `next` branch.

**Other information**


ematipico:
Closing because of unwanted changes

PR #1109, opened 2019-11-04
[WIP] feat(types): added types to the cli

ematipico:

This PR adds JSDocs to the JavaScript files of the CLI to make sure they have proper types.

We already use TypeScript and we could use it to make sure that everything is well maintained by checking the JSDoc. The main webpack repo already uses it, we ca do the same.

I will need some help to identify the correct types.
I also found some bugs.

I removed `process.cliLogger` because I think it's anti-pattern to tarnish the main `process` object. I created a simple module and updated all the places where we use `process.cliLogger`.

No

Not needed. It creates documentation by itself

**Summary**
The motivation is basically to maintain a good quality of the code while other people can check the code and understand the types of everything.

Also, it will prevent problems and bugs during the build.

Possibly yes, that's why I created the PR on the `next` branch.

**Other information**


lib/utils/commands.js
```diff
@@ -1,12 +1,23 @@
...
-    if (info[i].alias) {
-        return [info[i].name, info[i].alias];
-    }
-    return [info[i].name];
-}).reduce((arr, val) => arr.concat(val), []);
-
-module.exports.names = names;
-module.exports.info = arguments.commands;
+
+const names = Object.keys(info)
```
ematipico: I wasn't able to understand what's the result of this `names`. What would be the final result so I can add it to the types? @evenstensberg

.eslintrc.js
```diff
@@ -7,7 +7,7 @@ module.exports = {
     },
     parser: '@typescript-eslint/parser',
     root: true,
-    plugins: ['node'],
+    plugins: ['node', 'prettier'],
```
ematipico: My IDE was complaining it couldn't find `eslint-plugin-prettier` (we are using it inside `extends`). I added it to the plugins and here.

ematipico:
I noticed a lot of inconsistencies about the usage of double quotes and single quotes.

As we are using `prettier`, it all defaults to single quote (that's why the lot of changes) but I'd need to understand if we want to use single quotes or not. So I can apply the proper change.

lib/groups/config.js
```diff
@@ -1,142 +1,136 @@
...
-	}
+    constructor(options) {
+        super(options);
+        this.extensions = ['.mjs', '.js', '.json', '.babel.js', '.ts'];
+        this.defaultConfigFiles = this.getDefaultConfigFiles();
+        this.configFiles = null;
+        this.configOptions = [];
+    }
+    getDefaultConfigFiles() {
+        const { mode } = this.args;
```
ematipico: @evenstensberg This fails because args doesn't have a proper interface.
Could you help and tell us what should we have inside `args`?

cli.js
```diff
@@ -1,42 +1,37 @@
...
 
-require("v8-compile-cache");
-
-const importLocal = require("import-local");
+const importLocal = require('import-local');
 
 // Prefer the local installation of webpack-cli
 if (importLocal(__filename)) {
-	return;
+    // @ts-ignore
```
ematipico: The compiler gives error because a `return` is not allowed outside a function. For now it's ignored but shouldn't we use `process.exit`?

lib/commands/external.js
```diff
@@ -1,83 +1,107 @@
...
+     * @param {any[]} args
+     * @returns {Promise<null|*>}
+     */
+    static async run(name, ...args) {
+        let pkgLoc = ExternalCommand.validateEnv(name);
+        const scopeName = '@webpack-cli/' + name;
+        if (!pkgLoc) {
+            pkgLoc = await ExternalCommand.promptInstallation(scopeName, name);
+        }
+        if (typeof pkgLoc !== 'boolean') {
```
ematipico: `pkgLog` returns a string or a boolean. That's why I had to add this change

lib/commands/internal.js
```diff
@@ -1,16 +1,35 @@
...
+     */
+    constructor(command, ...args) {
+        /**
+         *
+         * @type {String}
+         */
+        this.command = command;
+        /**
+         *
+         * @type {any[]}
```
ematipico: @evenstensberg here we need to understand what's inside `args`.
Also, we should change the name because it's misleading. Maybe `cliArguments`?

lib/tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
```
ematipico: I added this new file. This will be used to check the js files inside the `lib` folder. It won't emit any javascript files

ematipico:
Another issue: tabs vs spaces. I've seen files with two different stylings. Should we force just one styling? I'd say tabs.

.eslintrc.js
evenstensberg: Gotcha, thanks for the PR. I was hesitating to add linting to the `lib` folder until we were at a stable point, but I think it's time

cli.js
evenstensberg: Depends on what we want to do, import local is ensuring that we're not in a global - local scope or vice versa

https://github.com/sindresorhus/import-local#readme

lib/commands/internal.js
evenstensberg: thought the same thing, you can rename it, invoked from `bootstrap.js` and it's missleading indeed

lib/commands/internal.js
```diff
@@ -1,16 +1,35 @@
...
+         */
+        this.args = args;
+    }
+    interactive() {}
+    make() {}
```
evenstensberg: you can remove the `make` and `interactive` methods, `make` is not going to be supported, `interactive` is inside `compiler.js` as an `outputOptions` flag

lib/groups/config.js
evenstensberg: this.args is a mimic of a partial webpack Config object and `outputOptions`

```js
this.args = {
webpackOptions: {},
outputOptions: {}
}
```

lib/tsconfig.json
evenstensberg: great, thanks for the heads up

lib/groups/output.js
```diff
@@ -5,26 +5,27 @@ class OutputGroup extends GroupHelper {
...
-			} else {
-				newArgs.output.filename = "bundle.js";
-			}
-			if(args.output.path) {
-				newArgs.output['path'] = ArgsValidator.resolveFileDirectory(args.output.path, "dist")
-			}
-		}
-		this.opts.options = this.mergeRecursive(this.opts.options, newArgs);
-	}
+	// TODO: delete or fix ArgsValidator
```
evenstensberg: What's wrong with argsvalidator

lib/utils/commands.js
evenstensberg: array of strings, check #1098

lib/groups/config.js
ematipico: So why there's a `mode` in there?

lib/groups/output.js
ematipico: It doesn't exist inside the file :)

evenstensberg: Okay, will have a look after you rebase

PR #1130, opened 2019-12-01
chore: improved the project setup process

pranshuchittora:
**Summary**
I recently formated my laptop, while setting up the project `webpack-cli` project I found the initial setup cycle of the codebase quite weird.

**Current behaviour**
1. Clone
2. npm i (to install the node_modules)
3. npm run bootstrap (this is quite weird, `npm run clean:all && npm install && lerna bootstrap`)

**Another approach**
What if we run `lerna` directly, but it won't work as lerna is not yet installed.

**Solution**
Install node_modules in the root so that lerna gets installed, then we can run `lerna` to install the deps recursively.

In short, we are installing node_modules and then deleting them.

No

kinda feature/build 

NA

No, I have to.

**Other information**

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -67,12 +67,12 @@ In case you are suggesting a new feature, we will match your idea with our curre
 
     ```bash
     #npm
-    npm install
+    npm run setup
     npm link
     npm link webpack-cli
     ---------------------------
     #yarn
-    yarn
+    yarn run setup
     yarn link
     yarn link webpack-cli
     ```
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
   ],
   "scripts": {
     "bootstrap": "npm run clean:all && npm install && lerna bootstrap",
+    "setup": "npm install && lerna bootstrap",
     "build": "tsc",
     "changelog": "conventional-changelog --config ./build/changelog-generator/index.js --infile CHANGELOG.md --same-file",
     "clean:all": "rimraf node_modules packages/*/{node_modules}",


package.json
```diff
@@ -28,6 +28,7 @@
   ],
   "scripts": {
     "bootstrap": "npm run clean:all && npm install && lerna bootstrap",
+    "setup": "npm install && lerna bootstrap",
```
rishabh3112: `bootstrap` command does that.

.github/CONTRIBUTING.md
```diff
@@ -67,12 +67,12 @@ In case you are suggesting a new feature, we will match your idea with our curre
...
     ```bash
     #npm
-    npm install
+    npm run setup
     npm link
     npm link webpack-cli
     ---------------------------
     #yarn
-    yarn
+    yarn run setup
```
rishabh3112: I think strictly enforcing new contributors to install dependencies for side packages isn't a good idea as PR may not be related to the sub packages and just adds extra friction.
If user wants to, they should follow the steps at "Bootstrap all the submodules before building for the first time" heading below.

pranshuchittora:
> I don't think we need a setup step

Generally `bootstrap` is the single entry point for end to end project setup.
https://docs.firefox-dev.tools/getting-started/build.html#building-and-running-locally

What about these changes
-> Bootstrap => Install Modules for the entire project. (One-step setup)
-> clean:all => Clean the modules just like the current behaviour of `bootstrap`. (Already there)

ematipico:
I think we are fine like this for now. Please next time open an issue and let's discuss that there, so won't spend time on a PR.

PR #1131, opened 2019-12-03
chore(utils): added tests and minor changes

pranshuchittora:
- Added tests for path-utils
- Minor formatiing for the package-manager func

tests + refacto

Yes

No

No

**Other information**

### DIFF
--- a/packages/package-utils/src/packageUtils.ts
+++ b/packages/package-utils/src/packageUtils.ts
@@ -1,8 +1,8 @@
 import fs from 'fs';
 import path from 'path';
-import { sync } from 'execa';
+import execa from 'execa';
 import spawn from 'cross-spawn';
-import  chalk = require('chalk');
+import chalk = require('chalk');
 import { prompt } from 'enquirer';
 import { runCommand } from './processUtils';
 
@@ -17,18 +17,12 @@ import { runCommand } from './processUtils';
 type PackageName = 'npm' | 'yarn';
 
 export function getPackageManager(): PackageName {
-    const hasLocalYarn = fs.existsSync(path.resolve(process.cwd(), 'yarn.lock'));
     try {
-        if (hasLocalYarn) {
-            return 'yarn';
-        } else if (sync('yarn', [' --version'], { stdio: 'ignore' }).stderr) {
-            return 'yarn';
-        } else {
-            return 'npm';
-        }
+        execa.sync('yarn', ['-v']);
     } catch (e) {
         return 'npm';
     }
+    return 'yarn';
 }
 
 /**
--- a/packages/utils/__tests__/path-utils.test.ts
+++ b/packages/utils/__tests__/path-utils.test.ts
@@ -1,0 +1,12 @@
+import * as pathUtils from '../src/path-utils';
+
+describe('local path test', () => {
+    it('should check the local path', () => {
+        const pathStatus = pathUtils.isLocalPath('./');
+        expect(pathStatus).toBe(true);
+    });
+    it('should check the wrong local path', () => {
+        const pathStatus = pathUtils.isLocalPath('xyz/');
+        expect(pathStatus).toBe(false);
+    });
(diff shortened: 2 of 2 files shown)


packages/utils/package-manager.ts
```diff
@@ -13,17 +13,18 @@ import * as path from "path";
...
-		return "npm";
-	}
+    const pwd = path.resolve(process.cwd());
+    const hasLocalNPM = fs.existsSync(path.resolve(pwd + "/package-lock.json"));
+    const hasLocalYarn = fs.existsSync(path.resolve(pwd + "/yarn.lock"));
+    if (hasLocalNPM) {
+        return "npm";
+    } else if (hasLocalYarn) {
+        return "yarn";
+    } else if (spawn.sync("yarn", [" --version"], { stdio: "ignore" }).error) {
```
rishabh3112: Here we are returning "npm" in else block too, maybe this if can be removed?

pranshuchittora: I thought of that too, here else works like a default case. What about removing the check for package-lock.json, as default is npm, so why check for it.

rishabh3112: Yes, I think what you're suggesting would be better, just checking for `yarn` if yarn is not there then return `npm`.

pranshuchittora: Have a look at the recent changes. Any idea about the failing tests.

packages/utils/package-manager.ts
```diff
@@ -13,17 +13,16 @@ import * as path from "path";
...
-		return "npm";
-	} else {
-		return "npm";
-	}
+    const pwd = path.resolve(process.cwd());
+    const hasLocalYarn = fs.existsSync(path.resolve(pwd + '/yarn.lock'));
+
+    if (hasLocalYarn) {
+        return 'yarn';
+    } else if (spawn.sync('yarn', [' --version'], { stdio: 'ignore' }).error) {
```
evenstensberg: Could you do a early return instead and do a if/return statement?

pranshuchittora: I didn't get it, could you pls elaborate.

ematipico: It means
```js
if () {
   return
}
return;
```
No need of `else` statement.
Also, here you forgot to kill the process.

ematipico: Also, could you use `execa` instead?

packages/utils/package-manager.ts
```diff
@@ -13,17 +13,17 @@ import * as path from "path";
...
-	} else if (hasLocalYarn) {
-		return "yarn";
-	} else if (spawn.sync("yarn", [" --version"], { stdio: "ignore" }).error) {
-		return "npm";
-	} else {
-		return "npm";
-	}
+    const pwd = path.resolve(process.cwd());
+    const hasLocalNPM = fs.existsSync(path.resolve(pwd, '/package-lock.json'));
+    const hasLocalYarn = fs.existsSync(path.resolve(pwd + '/yarn.lock'));
```
ematipico: Let's use `execa` to spawn a process. Also, let's not check the lock file because a first time installation would fail.

Use execa to run a synchronous command: yarn -v

If the result is an error, it means the user doesn't have yarn installed

packages/package-utils/src/packageUtils.ts
```diff
@@ -17,18 +17,12 @@ import { runCommand } from './processUtils';
 type PackageName = 'npm' | 'yarn';
 
 export function getPackageManager(): PackageName {
-    const hasLocalYarn = fs.existsSync(path.resolve(process.cwd(), 'yarn.lock'));
```
ematipico: I'd prefer to keep the logic as it is: it's better to check a local file than running a sub process

ematipico:
Going to close this PR. Code changed meanwhile

PR #1122, opened 2019-11-15
chore: Minor code refactor adhering to ES6 semantics

jamesgeorge007:
**What kind of change does this PR introduce?**
refactoring

**Did you add tests for your changes?**
Nope

**If relevant, did you update the documentation?**
N/A

**Summary**
Use object destructuring assignment wherever possible.

**Does this PR introduce a breaking change?**
Nope

**Other information**
N/A

PR #1134, opened 2019-12-06
refactor(module): module exporting

pranshuchittora:
**Summary**
`module.exports.a = a` can be written as `module.exports = {a:a}` or `module.exports = {a}`
If we have multiple exports from a single file then writing `module.exports.a = a` could be tedious task.

Refactoring
No
No

No

**Other information**
NaN

### DIFF
--- a/lib/utils/commands.js
+++ b/lib/utils/commands.js
@@ -9,5 +9,7 @@ const names = commands
     })
     .reduce((arr, val) => arr.concat(val), []);
 
-module.exports.names = names;
-module.exports.info = commands;
+module.exports = {
+    names,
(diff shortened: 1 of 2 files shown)


lib/utils/commands.js
```diff
@@ -9,5 +9,7 @@ const names = commands
     })
     .reduce((arr, val) => arr.concat(val), []);
 
-module.exports.names = names;
-module.exports.info = commands;
+module.exports = {
+    names,
+    info: commands,
+};
```
jamesgeorge007: #1122 takes account of this one.

Issue #1138, opened 2019-12-07
[ref]: use enum instead of direct string comparison


anshumanv:
**Do you want to request a *feature* or report a *bug*?**
We're using string comparison in bootstrap.js (will take a look at other files as well later), which can lead to bugs due to possible typos in future, use enums to stick to standard errs.

**What is the current behavior?**
Need refactor

**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**
Define enums and use switch to pick the appropriate err.

**If this is a feature request, what is motivation or use case for changing the behavior?**
more maintainable code

**Please paste the results of `webpack-cli info` here, and mention other relevant information such as programming language.**

ematipico:
Closing the issue as it doesn't bring enough value to the code

PR #1139, opened 2019-12-07
chore(bootstrap.js): refactor to use enums and switch in errs

anshumanv:
refactor to use enums and switch to execute correct error code

ISSUES CLOSED: #1138

refactor

NA

NA

**Summary**
Use enums rather than strict switch comparison to execute correct err code as per the switch.

No

**Other information**

### DIFF
--- a/lib/bootstrap.js
+++ b/lib/bootstrap.js
@@ -73,37 +73,46 @@ async function runCLI(cli, commandIsUsed) {
                 return;
             }
         } catch (err) {
-            if (err.name === 'UNKNOWN_VALUE') {
-                process.cliLogger.error(`Parse Error (unknown argument): ${err.value}`);
-                return;
-            } else if (err.name === 'ALREADY_SET') {
-                const argsMap = {};
-                const keysToDelete = [];
-                process.argv.forEach((arg, idx) => {
-                    const oldMapValue = argsMap[arg];
-                    argsMap[arg] = {
-                        value: process.argv[idx],
-                        pos: idx,
-                    };
-                    // Swap idx of overriden value
-                    if (oldMapValue) {
-                        argsMap[arg].pos = oldMapValue.pos;
-                        keysToDelete.push(idx + 1);
-                    }
-                });
-                // Filter out the value for the overriden key
-                const newArgKeys = Object.keys(argsMap).filter(arg => !keysToDelete.includes(argsMap[arg].pos));
-                // eslint-disable-next-line require-atomic-updates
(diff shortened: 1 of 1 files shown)


lib/bootstrap.js
```diff
@@ -73,37 +73,46 @@ async function runCLI(cli, commandIsUsed) {
...
-                // Filter out the value for the overriden key
-                const newArgKeys = Object.keys(argsMap).filter(arg => !keysToDelete.includes(argsMap[arg].pos));
-                // eslint-disable-next-line require-atomic-updates
-                process.argv = newArgKeys;
-                args = cmdArgs(core, { stopAtFirstUnknown: false, partial: true });
+            const errs = {
+                UNKNOWN_VALUE: 'UNKNOWN_VALUE',
+                ALREADY_SET: 'ALREADY_SET',
+            };
+            switch (err.name) {
```
rishabh3112: @anshumanv I think we should stick to if statements for comparisons as we can't add complex conditions if we needed in future on switch.

anshumanv: Makes sense, but this seems more readable and if we need complex conditions we can use conditionals inside the switch that we're working with.

rishabh3112: @anshumanv It is my personal choice but lets consider a scenario like below having if
```js
if (err.name == errs.UNKNOWN_VALUE and err.name == errs.ALREADY_SET) {/** code **/}
```
is more readable to me as compared to
```js
switch(err.name) {
  case errs.UNKNOWN_VALUE:
     if (err.name == errs.ALREADY_SET) {/** code **/}
     break;
}
```

anshumanv: Ah, yes. Seems more readable the way you suggested to me as well. I believe err.name will be a single value and from your example, it depends and say there are 3 params in future, it can lead up to 9-if-else combinations at max. But this is highly data-oriented and if err.name is a single value then at all times then we're cool with this arrangement as well. Let's see though, I'm okay with both approaches and as per the majority. 👍

rishabh3112: Oops got wrong with the example, but you may got the gist of what I was trying to say :)

ematipico:
I don't think this PR is bringing enough value to the code, here's why - I think:
- `enum` don't belong to JavaScript, so it's misleading to call a simple constant like that
- the strings used inside the `if` statements are used only once, so it's OK for now to keep them inline inside the if
- using a `switch` statement is the same as using the `if`

PR #1140, opened 2019-12-07
chore(lib): refactored lib utils and groups

pranshuchittora:
     Refactoring
    No
    No
**Summary**
- Object destructing to prevent deep level object refs (quite confusing and tedious to write)
- repeated else if to switch

No

**Other information**

### DIFF
--- a/lib/groups/AdvancedGroup.js
+++ b/lib/groups/AdvancedGroup.js
@@ -43,34 +43,39 @@ class AdvancedGroup extends GroupHelper {
     //     }
     // }
     resolveOptions() {
-        const { args } = this;
+        const {
+            args,
+            opts: { options },
+        } = this;
+
         if (args.hot) {
             const { HotModuleReplacementPlugin } = require('webpack');
             const hotModuleVal = new HotModuleReplacementPlugin();
-            if (this.opts.options && this.opts.options.plugins) {
-                this.opts.options.plugins.unshift(hotModuleVal);
+            if (options && options.plugins) {
+                options.plugins.unshift(hotModuleVal);
             } else {
-                this.opts.options.plugins = [hotModuleVal];
+                options.plugins = [hotModuleVal];
             }
         }
         if (args.prefetch) {
             const { PrefetchPlugin } = require('webpack');
             const prefetchVal = new PrefetchPlugin(null, args.prefetch);
-            if (this.opts.options && this.opts.options.plugins) {
-                this.opts.options.plugins.unshift(prefetchVal);
+            if (options && options.plugins) {
+                options.plugins.unshift(prefetchVal);
             } else {
-                this.opts.options.plugins = [prefetchVal];
+                // Currently the Plugin function is not functional  -> https://github.com/webpack/webpack-cli/pull/1140#discussion_r376761359
+                // options.plugins = [prefetchVal];
             }
         }
         if (args.plugin) {
-            if (this.opts.options && this.opts.options.plugins) {
-                this.opts.options.plugins.unshift(this.loadPlugin(args.plugin));
+            if (options && options.plugins) {
+                options.plugins.unshift(this.loadPlugin(args.plugin));
             } else {
-                this.opts.options.plugins = [this.loadPlugin(args.plugin)];
+                options.plugins = [this.loadPlugin(args.plugin)];
             }
         }
         if (args.target) {
-            this.opts.options.target = args.target;
+            options.target = args.target;
         }
 
         if (args.global) {
@@ -119,10 +124,10 @@ class AdvancedGroup extends GroupHelper {
 
             const { ProvidePlugin } = require('webpack');
             const globalVal = new ProvidePlugin(providePluginObject);
-            if (this.opts.options && this.opts.options.plugins) {
-                this.opts.options.plugins.unshift(globalVal);
+            if (options && options.plugins) {
+                options.plugins.unshift(globalVal);
             } else {
-                this.opts.options.plugins = [globalVal];
+                options.plugins = [globalVal];
             }
         }
     }
--- a/lib/utils/interactive.js
+++ b/lib/utils/interactive.js
@@ -124,12 +124,19 @@ module.exports = async function(config, outputOptions, processingMessageBuffer) 
         if (isExitCtrl(key)) {
             console.clear();
             process.exit();
-        } else if (key.name === 'down') {
-            process.stdout.write(ansiEscapes.cursorNextLine);
-        } else if (key.name === 'up') {
-            process.stdout.write(ansiEscapes.cursorPrevLine);
-        } else if (key.name === 'return') {
-            // TODO: get line and do stuff
+        }
+        switch (key.name) {
+            case 'down':
+                process.stdout.write(ansiEscapes.cursorNextLine);
+                break;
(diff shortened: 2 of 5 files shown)


lib/utils/interactive.js
```diff
@@ -124,12 +124,19 @@ module.exports = async function(config, outputOptions, processingMessageBuffer)
...
+                process.stdout.write(ansiEscapes.cursorNextLine);
+                break;
+            case 'up':
+                process.stdout.write(ansiEscapes.cursorPrevLine);
+                break;
+            case 'return':
+                // TODO: get line and do stuff
+                break;
+            default:
+                break;
```
rishabh3112: It's my personal opinion, I think `if else` provides more flexibility for future need as compared to `switch` and are comparatively more readable too. Consider if you like.

lib/groups/basic.js
```diff
@@ -8,24 +8,32 @@ class BasicGroup extends GroupHelper {
...
+                case 'sourcemap':
+                    this.opts.options.devtool = args[arg] || 'eval';
+                    outputOptions.devtool = args[arg];
+                    break;
+
+                case 'entry':
+                    this.opts.options[arg] = this.resolveFilePath(args[arg], 'index.js');
+                    break;
+                default:
+                    break;
```
rishabh3112: Same here

lib/utils/interactive.js
pranshuchittora: I completely agree that `if-else` are easier to read.

> As it turns out, the switch statement is faster in most cases when compared to if-else, but significantly faster only when the number of conditions is large. The primary difference in performance between the two is that the incremental cost of an additional condition is larger for if-else than it is for switch.

https://www.oreilly.com/library/view/high-performance-javascript/9781449382308/ch04.html#:~:targetText=As%20it%20turns%20out%2C%20the,than%20it%20is%20for%20switch%20.

rishabh3112: @pranshuchittora a valid point. My personal opinions ahead:
That excerpt reminded me of my debate with 1st semester C language teacher why one would use switch over if else and this was the major point for me to get satisfied. Still, If a particular block of code is to be executed only one, I would prefer `if-else` only as the gain is not that much.

pranshuchittora: :smile: That's a good old custom in the engineering of having debates with the profs. Switch case won't match for all conditions if `break` is there.

rishabh3112: Ofcourse, I meant to say this:
execution of whole switch if we writing switch in a for loop, thats the adequate performance gain.
But if the conditions are meant checked only for a single time, then we need not have a switch.

pranshuchittora: Uhm.. But I think a switch is an optimal & feasible choice.

lib/utils/compiler.js
```diff
@@ -41,19 +41,22 @@ function setUpHookForCompiler(compiler, outputOptions, options) {
     });
 
     if (outputOptions.infoVerbosity === 'verbose') {
+        const resolveCompilationName = compilation => {
```
jamesgeorge007: Why not make use of the inline approach?

```js
        const resolveCompilationName = compilation => 
          compilation.name ? compilation.name : '';
```

pranshuchittora: :thinking:  IMO this is much more inferable.

jamesgeorge007: Just personal preference :smile:

lib/utils/compiler.js
```diff
@@ -41,19 +41,22 @@ function setUpHookForCompiler(compiler, outputOptions, options) {
...
+            return compilation.name ? compilation.name : '';
+        };
         if (outputOptions.watch) {
             compiler.hooks.watchRun.tap('WebpackInfo', compilation => {
-                const compilationName = compilation.name ? compilation.name : '';
+                const compilationName = resolveCompilationName(compilation);
```
ematipico: You could actually remove the need of a variable:
```js
process.cliLogger.info('Compilation ' + resolveCompilationName(compilation) + ' starting…'); 
```

lib/groups/advanced.js
```diff
@@ -43,33 +43,34 @@ class AdvancedGroup extends GroupHelper {
     }
     resolveOptions() {
         const { args } = this;
+        const Options = this.opts.options;
```
ematipico: Lower case `options`

lib/utils/zero-config.js
```diff
@@ -1,8 +1,8 @@
 function getEnvFromOptionsAndMode(mode, optionsObject) {
     const { dev, prod } = optionsObject;
-
-    if ((process.env.NODE_ENV && process.env.NODE_ENV === 'production') || process.env.NODE_ENV === 'development') {
-        return process.env.NODE_ENV;
+    const NODE_ENV = process.env.NODE_ENV;
+    if ((NODE_ENV && NODE_ENV === 'production') || NODE_ENV === 'development') {
```
rishabh3112:
```suggestion
    if (NODE_ENV && (NODE_ENV === 'production' || NODE_ENV === 'development')) {
```

lib/groups/AdvancedGroup.js
```diff
@@ -44,33 +44,34 @@ class AdvancedGroup extends GroupHelper {
...
-            if (this.opts.options && this.opts.options.plugins) {
-                this.opts.options.plugins.unshift(prefetchVal);
+            if (options && options.plugins) {
+                options.plugins.unshift(prefetchVal);
             } else {
-                this.opts.options.plugins = [prefetchVal];
+                options.plugins = [prefetchVal];
             }
         }
         if (args.plugin) {
```
ematipico: Please comment this check. Plugin function is commented at the moment.

lib/groups/AdvancedGroup.js
```diff
@@ -44,33 +44,34 @@ class AdvancedGroup extends GroupHelper {
     // }
     resolveOptions() {
         const { args } = this;
+        const options = this.opts.options;
```
jamesgeorge007:
```suggestion
        const { options } = this.opts;
```

lib/utils/zero-config.js
```diff
@@ -1,8 +1,8 @@
 function getEnvFromOptionsAndMode(mode, optionsObject) {
     const { dev, prod } = optionsObject;
```
jamesgeorge007:
```js
function getEnvFromOptionsAndMode(mode, {dev, prod}) {
```

ematipico: That's not my favourite suggestion. I would keep it like this. Destructuring is fine when there's only one single object passed to the function. Here, we don't know what's inside the second parameter (variable names help)

lib/groups/AdvancedGroup.js
ematipico: At this point you should suggest:

```js
const { args, opts: { options } } = this;
```

:)

jamesgeorge007: Deep nesting might affect code readability.

lib/utils/zero-config.js
jamesgeorge007: Yeah, makes sense.

ematipico:
Thank you!

PR #1141, opened 2019-12-07
feat: help for core flags

rishabh3112:
Feature

No, will need

No

**Summary**
Adds help of flags as we have help for command from #1098.

No

**Other information**
#1098 https://github.com/webpack/webpack-cli/issues/1053


lib/groups/help.js
```diff
@@ -3,13 +3,13 @@ const chalk = require('chalk');
 const commandLineUsage = require('command-line-usage');
 
 class HelpGroup {
-    outputHelp(command = null) {
-        if (command !== null) {
-            const commands = require('../utils/commands').info;
+    outputHelp(subject = null, isCommand = true) {
+        if (subject !== null) {
+            const info = isCommand ? require('../utils/commands').info : require('../utils/core-flags').info;
+            const options = isCommand ? info.find(cmd => cmd.name === subject || cmd.alias === subject) : info.find(flg => flg.name === subject.slice(2) || flg.alias === subject.slice(1));
```
jamesgeorge007: This one seems a bit hard to read. Maybe use an if/else statement instead.

jamesgeorge007: This one seems a bit hard to read. Maybe use an if/else statement instead.

lib/utils/core-flags.js
```diff
@@ -0,0 +1,14 @@
...
+        if (name === 'help') return [];
+        if (alias) {
+            return [`--${name}`, `-${alias}`];
+        }
+        return [`--${name}`];
+    })
+    .reduce((arr, val) => arr.concat(val), []);
+
+module.exports.names = names;
+module.exports.info = core;
```
jamesgeorge007:
```js
module.exports = {names, core};
```

lib/webpack-cli.js
```diff
@@ -162,11 +162,17 @@ class WebpackCLI extends GroupHelper {
...
-        return new HelpGroup().outputHelp(command);
+        const commandNames = require('./utils/commands').names;
+        const flagNames = require('./utils/core-flags').names;
+        const subject =
+            commandNames.filter(name => {
+                return args.includes(name);
+            })[0] ||
+            flagNames.filter(name => {
+                return args.includes(name);
+            })[0];
```
jamesgeorge007: Maybe there would be a potential scope for refactoring.

jamesgeorge007: Also, why not make use of the inline approach?

lib/groups/help.js
rishabh3112: `if/else` would make `options` mutable. Instead used arrow function so as to not lose immutability.

lib/groups/help.js
```diff
@@ -3,13 +3,18 @@ const chalk = require('chalk');
...
-        if (command !== null) {
-            const commands = require('../utils/commands').info;
+    outputHelp(subject = null, isCommand = true) {
+        if (subject !== null) {
+            const info = isCommand ? require('../utils/commands').info : require('../utils/core-flags').info;
+            const options = () => {
+                if (isCommand) {
+                    return info.find(cmd => cmd.name === subject || cmd.alias === subject);
+                }
+                return info.find(flg => flg.name === subject.slice(2) || flg.alias === subject.slice(1));
```
jamesgeorge007:
```suggestion
                return info.find(flag => flag.name === subject.slice(2) || flag.alias === subject.slice(1));
```

lib/groups/help.js
jamesgeorge007: Yeah, makes sense :+1:

lib/groups/help.js
rishabh3112: Made this change, also updated `cmd` to `command` too.

lib/webpack-cli.js
rishabh3112: @jamesgeorge007 refactored in latest commit.

lib/utils/core-flags.js
rishabh3112: I prefer current ones as I want to keep `info` and variable name instead of `core`.

rishabh3112:
@jamesgeorge007 awesome quick reviews 🐱‍👤👍

lib/utils/core-flags.js
pranshuchittora:
```

module.exports = {
    names,
    info:core
}

```

lib/groups/help.js
```diff
@@ -3,13 +3,18 @@ const chalk = require('chalk');
 const commandLineUsage = require('command-line-usage');
 
 class HelpGroup {
-    outputHelp(command = null) {
-        if (command !== null) {
-            const commands = require('../utils/commands').info;
+    outputHelp(subject = null, isCommand = true) {
+        if (subject !== null) {
+            const info = isCommand ? require('../utils/commands').info : require('../utils/core-flags').info;
```
jamesgeorge007: https://github.com/webpack/webpack-cli/blob/3d6c24b8e345e7db48a5970d778f29168d5d3ba6/lib/utils/commands.js#L12

https://github.com/webpack/webpack-cli/blob/3d6c24b8e345e7db48a5970d778f29168d5d3ba6/lib/groups/help.js#L8

#1122

lib/webpack-cli.js
jamesgeorge007: Much better :+1:

lib/utils/core-flags.js
rishabh3112: Cool. It's night here, will update tommorow 👍

jamesgeorge007: We're all from the same time zone :sweat_smile:

lib/webpack-cli.js
```diff
@@ -162,11 +162,14 @@ class WebpackCLI extends GroupHelper {
...
     runHelp(args) {
         const HelpGroup = require('./groups/help');
-        const { names } = require('./utils/commands');
-        const command = names.filter(name => {
+        const commandNames = require('./utils/commands').names;
+        const flagNames = require('./utils/core-flags').names;
+        const allNames = [...commandNames, ...flagNames];
+        const subject = allNames.filter(name => {
             return args.includes(name);
         })[0];
```
jamesgeorge007:
```js
        const subject = allNames.filter(name => args.includes(name))[0];
```
Maybe :thinking:

pranshuchittora: A little bit difficult to understand at first glance ...

rishabh3112:
## ℹ Status 
- All CI errors related to this PR has been resolved.
- All expected features implemented

lib/utils/cli-flags.js
```diff
@@ -182,27 +192,31 @@ module.exports = {
...
             name: 'hot',
+            usage: '--hot',
             alias: 'h',
             type: Boolean,
             group: ADVANCED_GROUP,
             description: 'Enables Hot Module Replacement',
         },
         {
             name: 'sourcemap',
+            usage: '--sourcemap <sourcemap>',
```
ematipico: Maybe we should show the alternatives provided by webpack:
`--sourcemap <sourcemap|eval|>` etc.

Basically what we have here: https://webpack.js.org/configuration/devtool/#devtool

lib/utils/cli-flags.js
```diff
@@ -182,27 +192,31 @@ module.exports = {
         {
             name: 'target',
             alias: 't',
+            usage: '--target <build target>',
```
ematipico: Maybe we should show the targets available: https://webpack.js.org/configuration/target/#target

lib/utils/cli-flags.js
```diff
@@ -129,50 +131,58 @@ module.exports = {
...
         {
             name: 'plugin',
+            usage: '--plugin',
             group: ADVANCED_GROUP,
             type: String,
             description: 'Load a given plugin',
         },
         {
             name: 'global',
+            usage: '--global',
```
ematipico: The usage should be different:
`--global myVar ./global.js`

lib/groups/help.js
```diff
@@ -3,13 +3,18 @@ const chalk = require('chalk');
 const commandLineUsage = require('command-line-usage');
 
 class HelpGroup {
-    outputHelp(command = null) {
-        if (command !== null) {
-            const commands = require('../utils/commands').info;
+    outputHelp(subject = null, isCommand = true) {
```
ematipico: You should revert the properties. Usually the ones that are not needed (the ones that are `undefined` or `null`) should go after the one that have mandatory values.

Also, are you use you want to have a `null`? I mean, if you don't pass anything `isCommand` will be `undefined` which is good for this case, right?

lib/groups/help.js
```diff
@@ -3,13 +3,18 @@ const chalk = require('chalk');
...
+        if (subject !== null) {
+            const info = isCommand ? require('../utils/commands').info : require('../utils/core-flags').info;
+            const options = (() => {
```
ematipico: Why this is and inline IIFE? I don't think we need it.

```js
const options = info.find(commandOrFlag => {
   if (isCommand) {
     // return 
   }
   // return
})
```

lib/utils/core-flags.js
```diff
@@ -0,0 +1,13 @@
+const { core } = require('./cli-flags');
+
+const names = core
```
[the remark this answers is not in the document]
ematipico: Maybe also a test would help. It's up to you! We should improve the readability/usage of the code :)

lib/groups/help.js
rishabh3112: > Why this is

`options` is the `object` with details about `subject`.

> inline IIFE?

I was recommended to maintain immutability at each step and I didn't had any other implementation in my mind. Thus Opted for inline iffe. Will update it as per your suggestion in next commit.

> I don't think we need it.

Is there any other way to do this?

rishabh3112:
I updated the help and added documentation links for some of those.
image

lib/groups/help.js
ematipico: You can use `.reduce` if you really want to keep a functional approach :)

ematipico:
@rishabh3112 if you resolved the conflicts, I'd say we're ready to merge it!

rishabh3112:
@ematipico rebased, and we are ready to go I think.

Issue #941, opened 2019-06-07
[Enhancement]: getPackageManager considered flaky


evenstensberg:
**Describe the bug**

When determining the current package version, the lookup path for yarn is wrong and causes sometimes misleading information

**What is the current behavior?**

Npm is almost used everytime

**To Reproduce**
1. Go to `packages/utils/package-manager` L77
2. Reproduce by creating new tests or using the init generator

**Expected behavior**
Both npm and yarn is respected, both locally and globally.

**Screenshots**
N/A

**Please paste the results of `webpack-cli info` here, and mention other relevant information**
N/A

**Additional context**
N/A

jamesgeorge007:
@evenstensberg I would take this up :+1: Can you point me to the required files to be worked upon?

evenstensberg:
Sure! https://github.com/webpack/webpack-cli/blob/master/packages/utils/package-manager.ts#L55 needs to have better logic. Currently it doesn't work as expected

Deathrow77:
@evenstensberg Can I take up this issue?

Deathrow77:
@webpack-bot claim

Deathrow77:
@evenstensberg Please correct me if I am wrong. The default package manager has to npm. Only if we have yarn installed locally, will we use yarn. In every other case, NPM will be the package manager. Right?

ematipico:
@Deathrow77 yes, npm should be the default

Deathrow77:
@ematipico So, there's no need to check for local NPM. We can set it to default package manager, check for local yarn and then, try spawning yarn. If local yarn exists, we use that. Else We try the spawn and if it returns an error, npm is the returned.
        ```
const hasLocalYarn = fs.existsSync(path.resolve(process.cwd(), "yarn.lock"));
	let defaultPackageManager = "npm";
	if(hasLocalYarn)
	{
		defaultPackageManager = "yarn";
	}
	else if(spawn.sync("yarn", [" --version"], { stdio: "ignore" }).error)
		return defaultPackageManager;
	else
		return defaultPackageManager;
```

Deathrow77:
@ematipico I don't think there's a need to check for localNPM here. We just need to check for localYarn. I am using the following logic -

1. Set npm as the defaultPackageManager.
2. Check for local Yarn. If present, update the defaultPackageManager to yarn. If absent, return npm.
3. Try spawning yarn -v as well. If it return error, return npm.

[...]

PR #1161, opened 2020-01-01
fix: updated getPackageManager logic

Deathrow77:
The getPackageManager function was ideally supposed to return yarn if present locally, else use npm.
The function also tried to spawn yarn in addition, which was not needed. Hence, as per the current
logic, the function will check for localYarn. If present, use yarn, else use npm. This also renders
the test cases related to spawnChild useless, so those test have also been removed.

ISSUES CLOSED: #941

Bugfix

No

**Summary**

**Other information**

### DIFF
--- a/packages/utils/__tests__/package-manager.test.ts
+++ b/packages/utils/__tests__/package-manager.test.ts
@@ -67,14 +67,14 @@ describe("package-manager", () => {
 		expect(spawn.sync).toHaveBeenLastCalledWith("npm", ["install", "-g", packageName], { stdio: "inherit" });
 	});
 
-	it("should spawn npm update from spawnChild", () => {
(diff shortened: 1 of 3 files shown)


packages/utils/__tests__/package-manager.test.ts
```diff
@@ -67,14 +67,14 @@ describe("package-manager", () => {
 		expect(spawn.sync).toHaveBeenLastCalledWith("npm", ["install", "-g", packageName], { stdio: "inherit" });
 	});
 
-	it("should spawn npm update from spawnChild", () => {
-		const packageName = "some-pkg";
+	// it("should spawn npm update from spawnChild", () => {
```
rishabh3112: Please remove comments from here (we have git to handle old code)

PR #1162, opened 2020-01-01
fix: updated getPackageManager logic

Deathrow77:
The getPackageManager function of the package-manager module, which was supposed to resolve the
default Package Manager. The expected behaviour is that if yarn is present, yarn is used else npm is
used. However, the logic didnt allow the ideal behaviour. As per the updated logic, we check for
local yarn. If present, we use yarn, else npm is used. Also, the spawn statement is eliminated and
so are the related test statements.

ISSUES CLOSED: #941

**Summary**

**Other information**


packages/utils/package-manager.ts
```diff
@@ -13,17 +13,13 @@ import * as path from "path";
...
-	} else if (hasLocalYarn) {
-		return "yarn";
-	} else if (spawn.sync("yarn", [" --version"], { stdio: "ignore" }).error) {
-		return "npm";
-	} else {
-		return "npm";
+	let defaultpackageManager = "npm";
+	if(hasLocalYarn)
+	{
+		defaultpackageManager = "yarn";
```
ematipico: Here you can simply `return "yarn"`

packages/utils/package-manager.ts
```diff
@@ -13,17 +13,13 @@ import * as path from "path";
...
 	const hasLocalYarn = fs.existsSync(path.resolve(process.cwd(), "yarn.lock"));
-	if (hasLocalNPM) {
-		return "npm";
-	} else if (hasLocalYarn) {
-		return "yarn";
-	} else if (spawn.sync("yarn", [" --version"], { stdio: "ignore" }).error) {
-		return "npm";
-	} else {
-		return "npm";
+	let defaultpackageManager = "npm";
```
ematipico: The logic is basically the same. What if `yarn` is installed but the user hasn't done a first installation yet? Checking if it is installed means trying to execute `yarn` command, I think

ematipico:
The logic has been updated inside another PR but thank you!

PR #1177, opened 2020-01-19
Add init-generator test for yeoman generator

Loonride:
fix and test

Yes

N/A

**Summary**

This makes use of the `yeoman-test` package, explained here: https://yeoman.io/authoring/testing.html.

It tests that the init generator is working as expected. For now, I only created one test, but I will add more to experiment with different configurations.

I also made changes to ensure that the init generator is behaving as expected, as there were some issues I found.

No, the changes are all fixes

**Other information**


ematipico:
Thank you!

Issue #717, opened 2018-12-25
[ROADMAP] Webpack-CLI v4 Roadmap


evenstensberg:
## Summary

This is the roadmap to keep progress of webpack-cli v4 which is on the `next` branch of webpack.

### Stable
- [ ] Choose a lead developer **ALWAYS TO READ** 

In order to avoid conflicts and misunderstandings what happened before, together we select a leading developer. This person makes the final decision, when a dispute arises. If a situation arises when the whole team believes that the lead developer is wrong, then one of the developers from @webpack/core-team is involved in the dispute, his/her/it decision is last and not subject to dispute.

- [x] CLI arguments from core https://github.com/webpack/webpack/pull/10515 + tests
- [x] Allow multiple types for arguments, i.e. 
`webpack --stats` (means `stats: true`), 
`webpack --stats=verbose` (means: `stats: verbose`), 
- [x] Allow negative boolean args, i.e. `webpack --no-stats` (means: `stats: false`)
- [x] Allow multiple same arguments, i.e. `webpack --entry=./src/one.js --entry=./src/two.js`
- [x] To solve the points above, we may need to seach good CLI helper, like `yargs-parser` and etc. Update - we decided to go with commander.js
- [x] Verify webpack-dev-server + webpack-cli: 
  - [x] Tests for `webpack serve`
  - [x] Relax dependencies, i.e. webpack-cli should works with webpack-dev-server@3 and webpack-dev-server@next 
- [x] Compatibility with webpack@4 + tests (**always**)
- [x] Write Documentation, **need clarify** What we should write
- [x] Refactor https://github.com/webpack/webpack-cli/blob/next/packages/webpack-cli/lib/utils/cli-flags.js . We need to determine what arguments we are leaving, what we remove in favor CLI arguments from core, for example `--sourcemap` is not good candidate (https://github.com/webpack/webpack-cli/blob/next/packages/webpack-cli/lib/utils/cli-flags.js#L236), because `--devtool` used in webpack and most developers are used, but `verbose` is used in most of CLI tools and will be great to keep it for better DX (it is alias for `--stats=verbose`), not sure, it can be misleading, need dicussion
- [x] 0CJS: `webpack-cli` should be shipped with `info` and `init` commands out of box, need discussion about `serve` command 
- [x] 0CJS: `webpack-cli` should works without the `--config` option + tests
- [x] Update all dependencies to latest and setup dependabot
- [x] Part of the code in JS, part in TypeScript, let's choose one
[...]

PR #1196, opened 2020-01-30
Better output

ematipico:
Work in progress of  #717

This PR introduces the following changes:
- `compiler.js` => `Compiler.js` as class. I holds all the same methods as before. It also holds a class `CompilerOutput.js` which is responsible to generate the wanted output
- `CompilerOutput.js`: this class is responsible to output based on certain options. At the moment I left the "fancy" compilation but I think this fancy way of showing the results should be "opt-in" by passing an argument. Or we should provide a way to switch back to the old output. I also prepared some other functions for emit the normal output and the JSON one.
- I fixed eslint around some files. We were using a plugin that was not needed for the JS files.

Not yet, will do

No, I am awaiting feedback

**Summary**
The output must contain at least all the information emitted by the method `stats.getOutput`. 
Hopefully I managed to put all of them but I need to try it with some other project.

Nope

**Other information**
Screenshot 2020-01-30 at 15 42 20


ematipico:
@rishabh3112 I can remove the header's width so the table will be narrower:

Screenshot 2020-01-31 at 09 16 38

ematipico:
Honestly, I'd want to propose an argument called `--output-pretty` which will emit this output. And  if we don't pass anything, we keep the standard webpack output. What do you think?

rishabh3112:
> Honestly, I'd want to propose an argument called --output-pretty which will emit this output. And if we don't pass anything, we keep the standard webpack output. What do you think?

Yes, we should do this work this way.

I experimented with your work a little to make pretty output more compact.
image
image
WDYT?

ematipico:
I love it!!

rishabh3112:
@ematipico I added that change to your branch :)

rishabh3112:
I thought and I think we can do other way around. Like add a `--verbose` flag for standard output and give this output by default.

ematipico:
`--verbose` is not the standard output. The standard output is always stripped/limited to a certain number of lines (it doesn't contain all the modules). So we should get a default, standard and verbose. We are release a major version, so we could use this as default and we need also a way to opt-in the standard one using another argument that is not `--verbose`

rishabh3112:
So what could be a possible flag name?
`--output-standard`?

ematipico:
Sounds good!

ematipico:
@rishabh3112 I'd say we can make a different PR to add that new flag

rishabh3112:
Thanks and Cheers!

PR #1204, opened 2020-02-05
chore(cli): better group handling

ematipico:

A refactor of how configuration and arguments are handled and creating the final configuration to pass to webpack compiler

It's a refactor and all the existing tests are still passing

Yes, I added JSDoc to help the refactor

**Summary**

The problem of the current logic is that all the group helpers are randomly created and with a simple loop their logic is run. Which mean there's no order and we don't know which options comes first and which after.

For example, we have a group helper responsible to handle the configuration object (`ConfigHelper`) and we have a group helper responsible of the output (`OutputHelper`).
If `OutputHelper` is run first and `ConfigHelper` second, the information that we have inside the configuration will override the one inside the output, which is not the behaviour that the CLI should have.

The arguments passed via CLI have to override the one that we have inside the configuration. So now the order is the following:
- defaults
- webpack configuration
- CLI arguments

Nope

**Other information**


lib/groups/BasicGroup.js
```diff
@@ -1,9 +1,20 @@
-const GroupHelper = require('../utils/group-helper');
+const GroupHelper = require('../utils/GroupHelper');
+const { core, groups } = require('../utils/cli-flags');
 
 class BasicGroup extends GroupHelper {
     constructor(options) {
         super(options);
-        this.WEBPACK_OPTION_FLAGS = ['prod', 'dev', 'watch', 'w', 'prod', 'p', 'interactive', 'i', 'defaults', 'progress'];
+        this.WEBPACK_OPTION_FLAGS = core
```
ematipico: Reading from `cli-flags` is better because we keep everything in one single place

lib/utils/GroupHelper.js
```diff
@@ -24,14 +25,11 @@ class GroupHelper {
...
         }
-        const result = {};
-        const arrLength = arr.length;
-        for (let i = 0; i < arrLength; i++) {
-            const key = Object.keys(arr[i])[0];
-            const val = arr[i][key];
-            result[this.hyphenToUpperCase(key)] = val;
-        }
-        return result;
+        return arr.reduce((result, currentItem, index) => {
```
ematipico: A `.reduce` is far better

lib/utils/GroupHelper.js
```diff
@@ -100,6 +98,10 @@ class GroupHelper {
         }
         return configPathExists ? predefinedConfigPath : configPath;
     }
+
+    run() {
```
ematipico: A class that extends `GroupHelper` *must* implement the `run` function

lib/utils/development-config.js
```diff
@@ -10,12 +9,4 @@ module.exports = () => ({
         path: defaultOutputPath,
         filename: 'bundle.js',
     },
-
```
ematipico: Development should not have the minimizer

lib/groups/StatsGroup.js
```diff
@@ -1,5 +1,8 @@
-const GroupHelper = require('../utils/group-helper');
+const GroupHelper = require('../utils/GroupHelper');
 
+/**
+ * I gathers information about the stats options
```
rishabh3112:
```suggestion
 * StatsGroup gathers information about the stats options
```

lib/groups/ZeroConfigGroup.js
```diff
@@ -0,0 +1,45 @@
+const GroupHelper = require('../utils/GroupHelper');
+
+/**
+ * It is responsible to create a zero configuration based on the environment
```
rishabh3112:
```suggestion
 * ZeroConfigGroup creates a zero configuration based on the environment
```

lib/groups/ZeroConfigGroup.js
```diff
@@ -0,0 +1,45 @@
...
+    constructor(options) {
+        super(options);
+    }
+
+    /**
+     * It determines the mode to pass to webpack compiler
+     * @returns {string} The mode
+     */
+    getEnvFromOptionsAndMode() {
+        if ((process.env.NODE_ENV && process.env.NODE_ENV === 'production') || process.env.NODE_ENV === 'development') {
```
rishabh3112:
```suggestion
        if (process.env.NODE_ENV && (process.env.NODE_ENV === 'production' || process.env.NODE_ENV === 'development')) {
```

lib/utils/GroupHelper.js
rishabh3112: This check should be in constructor of the class.

lib/utils/GroupHelper.js
```diff
@@ -100,6 +98,10 @@ class GroupHelper {
...
+        throw new Error('Not implemented');
```
rishabh3112:
```suggestion
        throw new Error('Abstract class GroupHelper do not have implementation for function run');
```

bin/cli.js
```diff
@@ -36,7 +36,16 @@ const version = packageJson.engines.node;
 
 if (!semver.satisfies(process.version, version)) {
     const rawVersion = version.replace(/[^\d\.]*/, '');
-    logger.error('webpack CLI requires at least Node v' + rawVersion + '. ' + 'You have ' + process.version + '.\n' + 'See https://webpack.js.org/ ' + 'for migration help and similar.');
+    logger.error(
+        'webpack CLI requires at least Node v' +
```
rishabh3112: I would recommend using template literal here.

lib/groups/BasicGroup.js
rishabh3112: May leverage `core-flags.js` here to avoid redundant filter and reduce.

ematipico: Here we do different transformation. I don't think we need this here. I just kept it to avoid possible regressions

lib/utils/GroupHelper.js
ematipico: No, this implementation is "a la TypeScript". `GroupHelper` is like our abstract class. If we create a new group and we forget to implement `run` an error is thrown.

rishabh3112: Oh, I see now `run()` is like a pure virtual function (right?). Lets keep this then.

lib/groups/BasicGroup.js
rishabh3112: Yeah, they are different

lib/utils/GroupHelper.js
ematipico: Yes exactly!! :)

lib/webpack-cli.js
```diff
@@ -72,89 +82,195 @@ class WebpackCLI extends GroupHelper {
...
+        /**
+         * options is an array (multiple configuration) so we create a new
+         * configuration where each element is individually merged
+         */
+        if (Array.isArray(options)) {
+            this.compilerConfiguration = options.map(configuration => {
+                if (strategy) {
+                    return webpackMerge.strategy(strategy)(this.compilerConfiguration, configuration);
+                } else {
+                    return webpackMerge(this.compilerConfiguration, configuration);
```
jamesgeorge007:
```js
if () {
  return ...
}
return ...
```
The `else` block is redundant. We've been following the above approach across the codebase.

lib/webpack-cli.js
```diff
@@ -72,89 +82,195 @@ class WebpackCLI extends GroupHelper {
...
+             * The compiler configuration is already an array, so for each element
+             * we merge the options
+             */
+            if (Array.isArray(this.compilerConfiguration)) {
+                this.compilerConfiguration = this.compilerConfiguration.map(thisConfiguration => {
+                    if (strategy) {
+                        return webpackMerge.strategy(strategy)(thisConfiguration, options);
+                    } else {
+                        return webpackMerge(thisConfiguration, options);
+                    }
```
jamesgeorge007: Please see the above comment.

lib/webpack-cli.js
```diff
@@ -72,89 +82,195 @@ class WebpackCLI extends GroupHelper {
...
-            wrappedConfig.options = this.checkDefaults(wrappedConfig.options, wrappedConfig.outputOptions);
-            return wrappedConfig;
         }
-        return groupResult;
     }
 
-    async processArgs(args, yargsOptions) {
-        this.setMappedGroups(args, yargsOptions);
+    /**
+     * Responsible of creating and updating the new  output configuration
```
jamesgeorge007:
```suggestion
     * Responsible for creating and updating the new output configuration
```

lib/webpack-cli.js
```diff
@@ -72,89 +82,195 @@ class WebpackCLI extends GroupHelper {
...
+        if (groupHelper) {
+            const result = groupHelper.run();
+            this._mergeOptionsToConfiguration(result.options, groupHelper.strategy);
+            this._mergeOptionsToOutputConfiguration(result.outputOptions);
+            this._mergeProcessingMessageBuffer(result.processingMessageBuffer);
+        }
+    }
+
+    /**
+     * Responsible of apply defaults, if necessary
```
jamesgeorge007:
```suggestion
     * Responsible for applying defaults, if necessary
```

rishabh3112:
Thanks everyone!

PR #1207, opened 2020-02-06
chore: test cases for --max-old-space-size passed along with --node-args flag

jamesgeorge007:
test update

Yes

Nope

**Summary**
Added test cases for the following scenarios:-

- If the user didn't supplied a value to the `max-old-space-size` flag.
- If the user supplied an invalid value to the `max-old-space-size` flag.

Nope

**Other information**
N/A

### DIFF
--- a/test/node/node.test.js
+++ b/test/node/node.test.js
@@ -73,12 +73,14 @@ describe('node flags', () => {
     });
 
     it('throws an error if no values were supplied with --max-old-space-size', () => {
-        const { stderr } = run(__dirname, ['--node-args', '--max-old-space-size']);
-        expect(stderr).toContain('Error: missing value for flag --max-old-space-size of type size_t');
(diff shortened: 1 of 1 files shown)


test/node/node.test.js
```diff
@@ -71,4 +71,14 @@ describe('node flags', () => {
         const { stderr } = run(__dirname, ['--node-args', '--unknown']);
         expect(stderr).toContain('node: bad option:');
     });
+
+    it('throws an error if no values were supplied with --max-old-space-size', () => {
+        const { stderr } = run(__dirname, ['--node-args', '--max-old-space-size']);
+        expect(stderr).toContain('Error: missing value for flag --max-old-space-size of type size_t');
```
ematipico: If by chance node decide to change error, this test would fail. Just extract stderr and stdout and check if one is thruthy and the other one is falsy. I think that would be enough

PR #1209, opened 2020-02-07
init generator: read name property from package.json

Loonride:
refactoring

Not yet

N/A

**Summary**
The way that yeoman generator works is that it reads from the user's `package.json` to get the "name" property of this file. When it writes data to `.yo-rc.json`, it sets that data such that the key is that "name" property. Therefore, I have changed the reading of the yeoman configuration data such that we grab the desired configuration from that "name" key.

N/A

**Other information**


rishabh3112:
Thanks!

PR #1223, opened 2020-02-12
docs(cli): changed README to notify the beta version

ematipico:
Documentation

Not needed

Yes

**Summary**
I updated the README file to notify that the documentation will be for only for BETA, with a link to feedback.

**Other information**
[Rendered format](https://github.com/webpack/webpack-cli/blob/194b224695a04c595f0b84a40646e427dd371da1/README.md)


rishabh3112:
Are we keeping init as the name?

ematipico:
I honestly didn't understand what's the plan. I thought it was "create" but the package itself is still called "init". If we have to move it to be "create" so we need to change the package itself. Sounds good?

anshumanv:
init sounds better to me tbh, any motivations for changing it btw?

rishabh3112:
> I honestly didn't understand what's the plan. I thought it was "create" but the package itself is still called "init". If we have to move it to be "create" so we need to change the package itself. Sounds good?

Yup

> init sounds better to me tbh, any motivations for changing it btw?

I don't know. Even changed that and I don't know any reasons on why he did that. Will ask him for more info.

PR #1225, opened 2020-02-14
chore: move cli under lerna package

ematipico:

Refactor. This PR moves the `cli` code under a lerna package. This will help the symlinks between different packages and it will help the release process.

No. Tests should all pass after this refactor.

No. It should be done in later commits.

**Summary**
The cli should be under the `lerna` supervision. This is useful because it will help the creation of changelog, tagging, committing and publishing using one single tool.

Also, I changed the way the `serve` package retrieves the arguments from the CLI. Instead of referring with a relative path, it now import `webpack-cli` package (note the change in `package.json` of the cli):

```json
"main": "./lib/webpack-cli.js"
```

And it exposes a getter (`getCoreFlags`) that are passed to the serve command.

> ### Moved to `yarn`. It's essential to use the **workspaces** features so we can link locally all the packages together and we can test them seamlessly!
 

It should not.

**Other information**


test/serve/serve-basic.test.js
```diff
@@ -7,12 +7,19 @@ const { runWatch } = require('../utils/test-utils');
...
+        setOutput: false,
+        outputKillStr: 'main',
+    });
 };
 
 describe('basic serve usage', () => {
     it('compiles without flags', async () => {
         const { stdout, stderr } = await runServe([]);
+        console.log(stderr);
+        console.log(stdout);
```
jamesgeorge007: Why are you logging these to the console?

ematipico: Good catch!

packages/cli/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "webpack-cli",
```
rishabh3112:
```suggestion
  "name": "@webpack-cli/core",
```

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "webpack-cli",
+  "name": "webpack-cli-monorepo",
```
rishabh3112: Keep it the same as before.
We can follow babel here https://github.com/babel/babel/blob/master/package.json

ematipico: This is just a placeholder name. This won't be published to `npm`. In fact I am going to make a further change, I forgot to add `"private": true` to the `package.json`

packages/cli/package.json
ematipico: We can do it later with another PR. 
Anyway, we need at least a package called `webpack-cli` because that's what the users will download.

packages/cli/README.md
```diff
@@ -0,0 +1,63 @@
+# `webpack-cli`
+
+## About
+
+webpack CLI provides a flexible set of commands for developers to increase speed when setting up a custom webpack project. As of webpack v4, webpack is not expecting a configuration file, but often developers want to create a more custom webpack configuration based on their use-cases and needs. webpack CLI addresses these needs by providing a set of tools to improve the setup of custom webpack configuration.
+
+## How to install
+
+When you have followed the [Getting Started](https://webpack.js.org/guides/getting-started/) guide of webpack then webpack CLI is already installed!
```
jamesgeorge007:
```suggestion
If you have followed the [Getting Started](https://webpack.js.org/guides/getting-started/) guide of webpack then webpack CLI is already installed!
```

README.md
```diff
@@ -51,51 +51,18 @@ When you have followed the [Getting Started](https://webpack.js.org/guides/getti
...
-  -s, --sourcemap string   Determine source maps to use
-  --prefetch string        Prefetch this request
-  -j, --json               Prints result as JSON
-  --standard               Prints standard output
-  -d, --dev                Run development build
-  -p, --prod               Run production build
-  --version                Get current version
-  --node-args string[]     NodeJS flags
-```
+Please read [here](./packages/cli/README.md) what are the available commands and arguments.
```
jamesgeorge007: Ig this statement could be rephrased somehow.

ematipico: Please give me a suggestion. I am not good in this :)

ematipico:
@rishabh3112 

> Instead of naming sub package "webpack-cli" we should call it "@webpack-cli/core" as it will go good with the conventional naming of sub packages. Like @babel/cli, @webpack-cli/core, etc.

We still need a package called `webpack-cli` because that's what we are going to publish and that's what the users are installing (`npm i -D webpack-cli`). We could have a `core` package but we still need the `webpack-cli` one.

> Also we should keep bin in the main package itself and spawn cli from "core" sub package.

The main package won't be published on the registry (it has to be private). The `cli` folder (I will rename it `webpack-cli` in my next PR) because the actual CLI and that's where the binary has to be defined.

README.md
jamesgeorge007:
```suggestion
 Get to know what are the available commands and arguments [here](./packages/cli/README.md).
```

jamesgeorge007:
```suggestion
Get to know what are the available commands and arguments [here](./packages/cli/README.md).
```

anshumanv:
This is superb, makes it way easy to setup the project now that it leverages yarn workspaces. Kudos! 🎉

Issue #1230, opened 2020-02-16
[init]: invoking init command doesn't create a webpack config anymore


anshumanv:
**Describe the bug**

I don't see a webpack config scaffolded after running the init command. Just the `.yo-rc.json` is being created.

**What is the current behavior?**
init command is no longer creating webpack config 

**To Reproduce**
`webpack-cli init`

**Expected behavior**
It should create a webpack config from the `.yo-rc.json` that the generator creates

**Screenshots**
NA

**Please paste the results of `webpack-cli info` here, and mention other relevant information**

**Additional context**

anshumanv:
Looking into it.

SampsonCrowley:
@anshumanv this is still happening with 3.3.12

anshumanv:
Can you try with `4.0.0-rc.1`?

SampsonCrowley:
Doesn't that only support webpack 5?

anshumanv:
Supports 4 as well

PR #1231, opened 2020-02-16
fix(init): fix webpack config scaffold in init package

anshumanv:
fix webpack config scaffold

ISSUES CLOSED: #1230

bugfix, sync lock files

Yes

NA

**Summary**
Function to generate webpack config file was being skipped, fixed it.

Fixes instead.

**Other information**


packages/utils/modify-config-helper.ts
```diff
@@ -130,8 +144,8 @@ export default function modifyHelperUtil(action: string, generator: typeof Gener
 
                 const successMessage = `\nYou can now run ${chalk.green(runCommand)} to bundle your application!\n\n`;
                 process.stdout.write(`\n${successMessage}`);
-                return;
```
anshumanv: We were exiting before actually creating the config.

anshumanv:
Lock files seem out of sync and need to be updated eventually, should I do that in a separate PR @ematipico ?

ematipico:
That's strange because I don't get all these changes locally. Though, it would be better to do it on a separate PR

anshumanv:
Sounds good! Will push another commit.

> That's strange because I don't get all these changes locally.

Can you try running npm run bootstrap? I got these when I did that I believe.

ematipico:
OK no problem, we will soon switch to `yarn` because we need workspaces feature in order to test different packages together

PR #1237, opened 2020-02-18
chore: post refactor CLI

ematipico:

Refactor:
- moved the cli package under `webpack-cli`, more intuitive
- removed libraries `update-notifier`, `semver` and `terser-webpack-plugin`
  - `semver`: not needed at the moment. We use it just for one check. We already specify the engine inside the `package.json`
  - `update-notifier`:  it's nice but not essential. Let's save space
  - `terser-webpack-plugin`: webpack already applies minification when the `mode` is production. Not needed
- moved from `inquirer` to `enquirer`: smaller, less dependencies and same functionality (thank you @rishabh3112 for the suggestion)

No

Not needed

**Summary**
The `cli` folder doesn't respect the name inside `package.json`. That's why I moved it under `webpack-cli` folder. Also, I decided to remove some libraries that are not essential to the CLI and that could save us some space.

I also ditched `inquirer` for a more lighter version: `enquirer`. It has the same functionalities and is lighter is terms of dependencies and weight. 

No

**Other information**


rishabh3112:
Thanks!

PR #1238, opened 2020-02-18
feat: add progress bar for progress flag

rishabh3112:
feature

No

No

**Summary**
Add a real progress bar for`--progress` flag.
image

Yes

**Other information**

### DIFF
--- a/packages/webpack-cli/lib/utils/Compiler.js
+++ b/packages/webpack-cli/lib/utils/Compiler.js
@@ -1,6 +1,5 @@
 const webpack = require('webpack');
 const logger = require('./logger');
-const readline = require('readline');
 const { cyanBright, greenBright } = require('chalk');
 const { CompilerOutput } = require('./CompilerOutput');
 
--- a/packages/webpack-cli/lib/utils/Compiler.js
+++ b/packages/webpack-cli/lib/utils/Compiler.js
@@ -16,23 +16,25 @@ class Compiler {
 
         compilation.hooks.beforeRun.tap('webpackProgress', () => {
             if (outputOptions.progress) {
-                process.stdout.write("\n");
+                process.stdout.write('\n');
                 const defaultProgressPluginHandler = (percent, msg) => {
                     percent = Math.floor(percent * 100);
                     process.stdout.clearLine();
                     process.stdout.cursorTo(0);
                     if (percent !== undefined) {
-                        process.stdout.write(' (')
-                        for(let i = 0; i <= 100; i+=10) {
+                        process.stdout.write(' (');
+                        for (let i = 0; i <= 100; i += 10) {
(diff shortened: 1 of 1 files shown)


packages/webpack-cli/lib/utils/Compiler.js
```diff
@@ -1,5 +1,7 @@
 const webpack = require('webpack');
 const logger = require('./logger');
+const readline = require('readline');
```
ematipico: Looks like this is not used

packages/webpack-cli/lib/utils/Compiler.js
```diff
@@ -15,17 +16,24 @@ class Compiler {
...
             if (outputOptions.progress) {
-                let tmpMsg;
+                process.stdout.write("\n");
                 const defaultProgressPluginHandler = (percent, msg) => {
                     percent = Math.floor(percent * 100);
-                    if (percent === 100) {
-                        msg = 'Compilation completed';
+                    process.stdout.clearLine();
+                    process.stdout.cursorTo(0);
+                    if (percent !== undefined) {
```
anshumanv: Will `if (percent)` not work? Are there caveats to that?

packages/webpack-cli/lib/utils/Compiler.js
```diff
@@ -15,17 +16,24 @@ class Compiler {
...
+                        process.stdout.write(' (')
+                        for(let i = 0; i <= 100; i+=10) {
```
anshumanv:
```suggestion
                        for (let i = 0; i <= 100; i+=10) {
```

anshumanv: Is prettier not running for this file?

packages/webpack-cli/lib/utils/Compiler.js
rishabh3112: Zero as percent is a valid value, thus we should print when value is not strictly undefined.

packages/webpack-cli/lib/utils/Compiler.js
rishabh3112: Yup I think so, let me check.

packages/webpack-cli/lib/utils/Compiler.js
anshumanv: Ah cool, my bad. lgtm. 👍

packages/webpack-cli/lib/utils/Compiler.js
ematipico: Probably you did a commit and skipped the git hooks?

rishabh3112: I had to run eslint with `--no-ignore` for it to lint.

PR #1250, opened 2020-02-22
Fix loader-generator and plugin-generator tests

Loonride:
test fix

Yes

N/A

**Summary**
There were some problems with the `loader-generator` and `plugin-generator` test files which I fixed.

No

**Other information**


Loonride:
> We got rid of lock files since we use yarn workspaces now, please use yarn and remove the lock files.

fixed

Loonride:
> It looks like it's not fixed yet.
> 
> Look at the logs in the CI. I believe the assertion library doesn't work well with jest OR the assertions are not correct.
> 
> ```
> npm ERR! code ENOENT
> npm ERR! syscall open
> npm ERR! path /tmp/c12aae677bfdb92438557a5218666476c5e2f591/my-loader/package.json
> npm ERR! errno -2
> npm ERR! enoent ENOENT: no such file or directory, open '/tmp/c12aae677bfdb92438557a5218666476c5e2f591/my-loader/package.json'
> npm ERR! enoent This is related to npm not being able to find a file.
> npm ERR! enoent 
> 
> npm ERR! A complete log of this run can be found in:
> npm ERR!     /home/runner/.npm/_logs/2020-02-22T08_18_01_051Z-debug.log
> ```

It turns out this had nothing to do with the test itself, the generators needed to be writing a package.json but were not.

anshumanv:
> It turns out this had nothing to do with the test itself, the generators needed to be writing a package.json but were not.

Exactly, and the weird bit it that the tests are passing, it should fail in case it fails to write the files.

ematipico:
So we've found a bug! Nice!! 🙏

packages/generators/src/plugin-generator.ts
```diff
@@ -21,7 +21,7 @@ const PluginGenerator = addonGenerator(
 			validate: (str: string): boolean => str.length > 0
 		}
 	],
-	path.resolve(__dirname, "..", "generate-plugin"),
+	path.resolve(__dirname, "../../generate-plugin/templates"),
```
ematipico: Thank you for fixing this! That was my fault

PR #1254, opened 2020-02-22
tests(output): add test case for defaults flag

anshumanv:
Added a test case for --defaults flag

Yes

NA

**Summary**
* Added a case when --defaults flag is used to indicate compiler to use webpack defaults.

No

**Other information**

### DIFF
--- a/test/output/defaults/index.js
+++ b/test/output/defaults/index.js
@@ -1,0 +1,1 @@
+module.export = 'output-flag-test';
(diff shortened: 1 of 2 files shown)


anshumanv:
I think defaults tests can be moved a folder up since they're not really output but rather a way to indicate compiler about the config, what say @ematipico?

ematipico:
Yes it makes totally sense! Feel free to move them around

anshumanv:
Thanks for the heads up! Will do. ✨

anshumanv:
Will do it in a separate PR, have more tests and cleanup in mind.

PR #1255, opened 2020-02-23
feat: create a cli executer

rishabh3112:
feature

No

No

**Summary**
I am creating a CLI executer that will be used in case of invalid/unknown args.
Currently it looks like this
image

image
After this is merged, I would add this in appropriate places

No

**Other information**
Improving intractability with cli

### DIFF
--- a/packages/webpack-cli/lib/utils/cli-executer.js
+++ b/packages/webpack-cli/lib/utils/cli-executer.js
@@ -1,4 +1,4 @@
-const { MultiSelect, Input} = require('enquirer');
+const { MultiSelect, Input } = require('enquirer');
 const runner = require('../runner');
 const logger = require('./logger');
 const cliArgs = require('./cli-flags').core;
@@ -10,8 +10,11 @@ async function prompter() {
         name: 'type',
         message: 'Which flags do you want to use?',
         choices: cliArgs.reduce((prev, curr) => {
-            return [...prev, `--${curr.name}`];
-        }, [])
+            return [...prev, `--${curr.name}: ${curr.description}`];
+        }, []),
+        result: (value) => {
+            return value.map((flag) => flag.split(":")[0]);
+        }
     });
 
     const selections = await typePrompt.run();
@@ -53,4 +56,4 @@ async function run() {
     runner([], args);
 }
 
-run();
+module.exports = run;
--- a/packages/webpack-cli/lib/utils/cli-executer.js
+++ b/packages/webpack-cli/lib/utils/cli-executer.js
@@ -1,31 +1,31 @@
 const { MultiSelect, Input} = require('enquirer');
 const runner = require('../runner');
 const logger = require('./logger');
-const allArgs = require('./cli-flags');
+const cliArgs = require('./cli-flags').core;
 
-logger.info(`Initiating webpack-cli executer \n`);
-
-const cliArgs = [];
-const nodeArgs = [];
 async function prompter() {
+    const args = [];
+
     const typePrompt = new MultiSelect({
         name: 'type',
-        message: 'Which flags do you want to use ?',
-        choices: allArgs.core.reduce((prev, curr) => {
+        message: 'Which flags do you want to use?',
+        choices: cliArgs.reduce((prev, curr) => {
             return [...prev, `--${curr.name}`];
         }, [])
     });
 
     const selections = await typePrompt.run();
-    const questions = selections.map(selection => {
-        const options = allArgs.core.find(flag => {
+
+    const boolArgs = [];
+    const questions = [];
+    selections.forEach(selection => {
+        const options = cliArgs.find(flag => {
             return flag.name === selection.slice(2);
         });
 
-        if (options.type == Boolean) {
-            return {
-                run: () => Promise.resolve([selection])
-            };
+        if (options.type === Boolean) {
+            boolArgs.push(selection);
+            return;
         }
 
         const valuePrompt = new Input({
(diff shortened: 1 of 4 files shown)


anshumanv:
> I am creating a CLI executer that will be used in case of invalid/unknown args.

Interesting! What are we planning though? Throw this prompt when an invalid flag is supplied?

rishabh3112:
@anshumanv 
Yes, that what I am planing for now. we may also assign it it's own flag if it seems useful.

anshumanv:
> Yes, that what I am planing for now. we may also assign it it's own flag if it seems useful.

Sounds fair, though I think adding it just as a separate flag would be better since we for now show help on detecting any invalid flag, so user can correct any typos they made, it would be slower to show a prompt and ask them to enter the flags over again rather than correct just some invalid flag.

rishabh3112:
Sure it slows down but as a CLI our first focus should be on user's successful interaction with the CLI   i.e. helping user getting their work done instead of keep on correcting them.

anshumanv:
> Sure it slows down but as a CLI our first focus should be on user's successful interaction with the CLI i.e. helping user getting their work done instead of keep on correcting them.

Ah didn't see the flag picker screenshot before, that can help surely, let's see how this turns out.

packages/webpack-cli/lib/utils/cli-executer.js
```diff
@@ -0,0 +1,53 @@
...
+    const typePrompt = new MultiSelect({
+        name: 'type',
+        message: 'Which flags do you want to use ?',
+        choices: allArgs.core.reduce((prev, curr) => {
+            return [...prev, `--${curr.name}`];
+        }, [])
+    });
+
+    const selections = await typePrompt.run();
+    const questions = selections.map(selection => {
```
ematipico: Change approach in this function:
1. create a local variable where you store args and `Input`s
2. iterate `selections` using a `.forEach` (you need a local variable to store `Input` instances) or `.reduce` (here you return an array of `Input`)
3. If you find a `Boolean` you don't return anything and you its value inside the local variable for storing the args. You don't create that `run: () =>` which I don't like. It's basically smell code.
4. Iterate over the `Input` array using a `async .map` where you return the cli arg value, together with a `Promise.all`

```js
const args = await Promise.all(questions.map(async question => {
   const value = await question.run();
  // do something here
   return valueAsArg;
}));

return [
  ...boolArgs,
  ...args
]
```

rishabh3112: Can't use `Promise.all` as it will force all the promises to execute in parallel ( which in turn will be blocked by await ), that is why, I used reduce to create chain for synchrunous execution.
I have taken care of all other issues you have mentioned.

packages/webpack-cli/lib/utils/cli-executer.js
```diff
@@ -0,0 +1,53 @@
+const { MultiSelect, Input} = require('enquirer');
+const runner = require('../runner');
+const logger = require('./logger');
+const allArgs = require('./cli-flags');
+
+logger.info(`Initiating webpack-cli executer \n`);
```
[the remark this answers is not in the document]
rishabh3112: I have removed it.

packages/webpack-cli/lib/utils/cli-executer.js
```diff
@@ -0,0 +1,53 @@
...
+
+const cliArgs = [];
+const nodeArgs = [];
+async function prompter() {
+    const typePrompt = new MultiSelect({
+        name: 'type',
+        message: 'Which flags do you want to use ?',
```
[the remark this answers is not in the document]
rishabh3112: Done!

packages/webpack-cli/lib/utils/cli-executer.js
```diff
@@ -0,0 +1,53 @@
...
+        }, [])
+    });
+
+    const selections = await typePrompt.run();
+    const questions = selections.map(selection => {
+        const options = allArgs.core.find(flag => {
+            return flag.name === selection.slice(2);
+        });
+
+        if (options.type == Boolean) {
```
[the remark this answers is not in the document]
rishabh3112: Yes, I have changed this.

packages/webpack-cli/lib/utils/cli-executer.js
```diff
@@ -0,0 +1,53 @@
...
+    });
+    await questions.reduce((prev, curr) => {
+        return prev.then(() => curr.run().then((args) => {
+            cliArgs.push(...args);
+        }));
+    }, Promise.resolve(null))
+}
+
+async function run() {
+    await prompter();
```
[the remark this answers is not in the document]
rishabh3112: Yes, sorry. I have changed this.

rishabh3112:
Thanks for the detailed review :)

packages/webpack-cli/lib/utils/cli-executer.js
```diff
@@ -0,0 +1,53 @@
+const { MultiSelect, Input} = require('enquirer');
+const runner = require('../runner');
+const logger = require('./logger');
+const allArgs = require('./cli-flags');
```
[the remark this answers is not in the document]
rishabh3112: Done!

packages/webpack-cli/lib/utils/cli-executer.js
ematipico: It makes. So at this point consider the following approaches:
1. A queue
2. An asynchronous loop (you will have to use a map/set)

rishabh3112:
Thanks! updated the code now.

packages/webpack-cli/lib/utils/cli-executer.js
rishabh3112: Just curious, what you meant by 2nd approach?

packages/webpack-cli/lib/utils/cli-executer.js
```diff
@@ -0,0 +1,56 @@
+const { MultiSelect, Input} = require('enquirer');
```
jamesgeorge007:
```suggestion
const { MultiSelect, Input } = require('enquirer');
```

rishabh3112:
Added descriptions as per suggestion by @ematipico 

image

packages/webpack-cli/lib/utils/cli-executer.js
rishabh3112: Thanks!

PR #1258, opened 2020-02-23
tests: fix folder structure, add more default tests

anshumanv:
Fix test arch, add more default case tests

Yay

NA

**Summary**
* Fix arch
* Add more tests for default flag

No

**Other information**
Follow up of https://github.com/webpack/webpack-cli/pull/1254


anshumanv:
@ematipico seems good to go, can you think of any other cases that I can cover?

test/defaults/with-config-and-entry/default-with-config.test.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+const { stat } = require('fs');
+const { resolve } = require('path');
+const { run, extractSummary } = require('../../utils/test-utils');
+
+describe('output flag defaults with config', () => {
+    it('should use default entry if config entry file is not present', done => {
```
ematipico: I don't understand this test. Here there is a config file, so why the test says "not present"?

test/defaults/with-config-and-entry/default-with-config.test.js
```diff
@@ -0,0 +1,23 @@
...
+        const { stdout, stderr } = run(__dirname, ['--defaults'], false);
+        const summary = extractSummary(stdout);
+        // Should use the output dir specified in the config
+        const outputDir = 'with-config-and-entry/binary';
+        // eslint-disable-next-line quotes
```
ematipico: Can be removed

test/defaults/without-config-and-entry/default-without-config.test.js
```diff
@@ -0,0 +1,20 @@
...
+const { stat } = require('fs');
+const { resolve } = require('path');
+const { run, extractSummary } = require('../../utils/test-utils');
+
+describe('output flag defaults without config', () => {
+    it('should throw if the entry file is not present', done => {
+        const { stdout, stderr } = run(__dirname, ['--defaults'], false);
+        const summary = extractSummary(stdout);
+        const outputDir = 'without-config-and-entry/dist';
+        // eslint-disable-next-line quotes
```
ematipico: Remove this line. Use the template literal (`)

ematipico:
And what's `(arch)` by the way? 😊 
Usually we should specify the name of the package we are modifying. If there's no package involved, let's not put anything

test/defaults/with-config-and-entry/default-with-config.test.js
anshumanv: The entry file specified in the config doesn't exist, so it will pick the default entry file `./index.js`.

Should I improve the wording here?

test/defaults/without-config-and-entry/default-without-config.test.js
anshumanv: Cool, will do 👍

test/defaults/with-config-and-entry/default-with-config.test.js
anshumanv: Maybe `.. if entry file from config is not present` ?

ematipico: Yes, we should improve the wording: "should override the [entry] property of the config file" We don't care if the file really exists or not

anshumanv: Actually, if the entry file exists, it picks up the entry specified in the config instead of the default entry but in case it doesn't then it falls back to using the default entry

anshumanv:
> And what's (arch) by the way? 😊
Usually we should specify the name of the package we are modifying. If there's no package involved, let's not put anything

Ah, I meant to put architecture, but yes it should be removed good catch! 😄

test/defaults/with-config-and-entry/default-with-config.test.js
ematipico: This should NOT be the case. The point of enforcing defaults is to use OUR defaults and override everything that already exists.

ematipico: Which means there's a bug in our strategy when we apply the defaults

anshumanv: Yes, will take a look at that too. 👍

anshumanv: Skipping this test for now, will enable it once I fix the above problem.

test/defaults/without-config-and-entry/default-without-config.test.js
anshumanv: Actually we're enforcing single quotes in eslint config.

anshumanv:
@ematipico, I'm getting a huge change on running watch command, these files aren't ignored in the respective .gitignore of their packages, should I modify the configs to ignore transpiled stuff too?

image

ematipico:
Have you tried to delete those files and build again? The project is already setup correctly, where we have the src folder and the lib folder as destination where compiled files are stored and ignored in git.

anshumanv:
> Have you tried to delete those files and build again? The project is already setup correctly, where we have the src folder and the lib folder as destination where compiled files are stored and ignored in git.

Ah yes, not seeing them anymore, thanks @ematipico!

Anything else that I need to take care of here? Seems good from my side.

Issue #1288, opened 2020-03-04
webpack entry config doesn't work if defaultEntry file existed


Mistyyyy:
**What is the current behavior?**

webpack.config.js

```javascript
module.exports = {
  mode: 'development',

  entry: {
    index: './src/entry.js',
  },

  output: {
    path: path.resolve(__dirname, 'dist'),
    filename: '[name].bundle.js',
  },

  plugins: [
    new CleanWebpackPlugin()
  ]
}
```
below is project Directory Structure
```
├── index.js
├── package.json
├── src
│   ├── a.js
│   └── entry.js
└── webpack.config.js
```

when running` npx webpack`, the output is` main.bundle.js  `, the content of the final bundle includes index.js module and excludes entry.js module

**how I know this is a webpack-cli bug rather webpack bug**

I debug the program locally

**To Reproduce**

https://github.com/Mistyyyy/webpack-cli-entry-repo

**Expected behavior**

`output: index.bundle.js`

**Screenshots**

![image](https://user-images.githubusercontent.com/37856353/75847913-92e79580-5e1b-11ea-939c-6a1d608875e8.png)

I had fix it locally, the reason is  BasicGroup has a default entry value and override the config entry value if find defaultEntry file.

If appropriate, I am willing to submit a PR

**version**

Node: v12.10.0
webpack: 5.0.0-beta.13
webpack-cli: 4.0.0-beta.8

anshumanv:
This is a bug surely, please feel free to submit a PR.

ematipico:
This is a nice chance for first contributors to tackle the issue. A good way to tackle issue would be to create a test first and make that fails.

PR #1289, opened 2020-03-04
fix(webpack-cli): to void defaultEntry override the webpack config entry

Mistyyyy:
if not pass the entry option through command line, webpack-cli would look for defaultEntry
The defaultEntry would override the entry in webpack config files

ISSUES CLOSED: #1288

a bugFix

yes

no relevant

#1288

Nothing


test/entry/config-entry/entry-with-config/entry-with-config.test.js
```diff
@@ -0,0 +1,23 @@
...
+describe('single entry with entry in config', () => {
+    it('should use config entry if config entry existed', done => {
+        const { stdout, stderr } = run(__dirname, ['-c', '../1.js'], false);
+        const summary = extractSummary(stdout);
+        const outputDir = 'entry-with-config/binary';
+        // eslint-disable-next-line quotes
+        expect(summary['Output Directory']).toContain(outputDir);
+
+        // eslint-disable-next-line quotes
+        console.log(stderr);
```
jamesgeorge007: Please remove this line.

packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -248,6 +248,7 @@ class WebpackCLI extends GroupHelper {
     async runOptionGroups() {
         await Promise.resolve()
             .then(() => this._handleGroupHelper(this.zeroConfigGroup))
+            .then(() => this._handleGroupHelper(this.basicGroup))
```
ematipico: Why executing the basic group twice?

packages/webpack-cli/lib/groups/BasicGroup.js
```diff
@@ -15,6 +15,7 @@ class BasicGroup extends GroupHelper {
                 }
                 return result;
             }, []);
+        this._excuted = false;
```
ematipico: Each `GroupHelper` class can have a `this.strategy` parameter. Check `OutputGroup`.
The strategy tell to `webpack-merge` how to handle specific bits of the configuration:

```js
this.strategy = {
   'output.filename': 'prepend',
   'output.path': 'prepend',
};

```

packages/webpack-cli/lib/webpack-cli.js
Mistyyyy: > I think the solution should be different. I left some feedback on how to tackle the problem. Let me know what you think

ok.

## Entry rules

to get the right entry, I should consider the different entry priority

- command entry exists, use the command entry
- no command entry, config entry exists, use the config entry
- no command entry, no config entry, use defaultEntry

The origin BasicGroup would get the defaultEntry or command entry, which can't be distinguished when executing webpackMerge, caused defaultEntry override config entry.
because BasicGoup execute after ConfigGroup

## Why execute BasicGroup twice?

The first time, to get defaultEntry, and BasicGroup execute before ConfigGoup, the config entry can override defaultEntry correctly.

The second time, to get command entry, and BasscGroup execute after ConfigGoup, the command entry can override config entry correctly.

## Why not using webpackMerge strategies

because entry may be string, while strategies only impact object or array merge. Maybe strategies can do but I don't know. if so, please let me know.

## improvement

The implemented method is not good, I am considering how to improve it, maybe I can pass the configuration to BasicGroup to decide to use defaultEntry or command entry. if you have a good idea, please let me know.

Thanks for reviewing.

test/entry/config-entry/entry-with-config/entry-with-config.test.js
Mistyyyy: done

test/entry/config-entry/entry-with-config/entry-with-config.test.js
```diff
@@ -0,0 +1,23 @@
...
+const { stat } = require('fs');
+const { resolve } = require('path');
+const { run, extractSummary } = require('../../../utils/test-utils');
+
+describe('single entry with entry in config', () => {
+    it('should use config entry if config entry existed', done => {
+        const { stdout, stderr } = run(__dirname, ['-c', '../1.js'], false);
+        const summary = extractSummary(stdout);
+        const outputDir = 'entry-with-config/binary';
+        // eslint-disable-next-line quotes
```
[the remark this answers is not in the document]
Mistyyyy: done

packages/webpack-cli/lib/webpack-cli.js
ematipico: I had the issue while working with the `output` property. Try to apply the same strategy. Strategies work for all type of entries

Mistyyyy: Thanks, I will try it in this way.

Mistyyyy:
@ematipico 

> I had the same issue while working with the `output` property. Try to apply the same strategy. Strategies work for all type of entries

After diving into the source code of webpack-merge, I found strategies can't work for the `primitives` type. It can only work for `Array`、`Object`、`Function` type.

The core code of webpack-merge is [here](https://github.com/survivejs/webpack-merge/blob/master/src/join-arrays.js)

For example
```javascript
const source = {
  entry: 'foo'
}

const target = {
  entry: 'bar'
}

const strategy = {
  entry: 'append', // or replace, the default is append
}

const result = merge.strategy(strategy)(source, target);
```

No matter what strategies are applied,  `target.entry` always override `source.entry`. 

So, if `config entry` and `defaultEntry` all exist, the `defaultEntry` will always override config entry. because `this.basicGroup.run` execute later than `this.configGroup.run`

Additional, I can't figure out the effect of the strategies applied to `output.filename` and `output.path`, I try to comment out this line, all tests passed. because the value of `output.path` and `output.filename` are all primitives type, the strategies don't work.

```javascript
this.strategy = {
   'output.filename': 'prepend',
   'output.path': 'prepend',
};
```
I have updated my code, please review that. Thanks.

ematipico:
Hey @Mistyyyy, thank you very much for your extensive explanation! 
If your findings are true, I now wonder if this is a `webpack-merge` bug of if it is intended by design.

If it is a bug, I think we should file an issue in webpack-merge repository. If it is intended, this is a a big problem. 

When I was refactoring the code few weeks back, I was facing a similar issue with `output.filename` and using the strategies actually worked! So now I am actually confused. I am going to take a look at the code and see if we are good to land the PR (it looks good though, nice work here, thank you)

ematipico:
@Mistyyyy  I did a quick try and the strategy actually works... This is the tests that I run structure that I run. Let me know if I missed something.

### Folder contents of the test
`webpack.config.js`

```js
const { resolve } = require('path');

module.exports = {
    mode: 'development',
    entry: {
        index: './file.js',
    },
    output: {
        path: resolve('./dist'),
        filename: '[name].bundle.js',
    },
};
```

`file.js`
```js
console.log('js');
```

`custom.entry.js`
```js
'use strict';
const { stat, readFileSync } = require('fs');
const { resolve } = require('path');
const { run } = require('../../utils/test-utils');

describe('default entry and config entry all exist', () => {
    it('should use config entry if config entry existed', done => {
        const { stdout, stderr } = run(__dirname, ['-c', 'webpack.config.js'], false);

        expect(stderr).toBeFalsy();
        expect(stdout).toBeTruthy();
        stat(resolve(__dirname, './dist/index.bundle.js'), (err, stats) => {
            expect(err).toBeFalsy();
            expect(stats.isFile()).toBe(true);
            done();
        });
        const data = readFileSync(resolve(__dirname, './dist/index.bundle.js'), 'utf-8');
        expect(data).toContain('console.log');
    });
});
```

### Changes applied

`BaseGroup.js`

```diff
constructor() {
+  this.strategy = {
+     'entry': 'prepend'
+  }
}

```

Mistyyyy:
Yeah, I think you should add some defaultEntry file which cli can reach. like `./index.js` `./src/index.js`, and test it again. the purpose is to test defaultEntry would override config entry and strategy takes effect or not.

Mistyyyy:
> @ematipico Yeah, I think you should add some defaultEntry file which cli can reach. like `./index.js` `./src/index.js`, and test it again. the purpose is to test defaultEntry would override config entry and strategy takes effect or not.

The project structure should look like this
```
├── src
│   ├── file.js
│   └── index.js
├── webpack.config.js
```

ematipico:
Issue found. The problem is inside `BasicGroup`

`resolveFlags` should be something like this:
```js
   resolveFlags() {
        const { args } = this;
        const { outputOptions, options } = this.opts;
        Object.keys(args).forEach(arg => {
            if (this.WEBPACK_OPTION_FLAGS.includes(arg)) {
                outputOptions[arg] = args[arg];
            }

            if (arg === 'entry') {
                if (args[arg]) {
                    options.entry = this.resolveFilePath(args[arg], 'index.js');

                } else {
                    options.entry = this.resolveFilePath(args.entry, 'index.js');
                }
            }
        });

        // TODO: to add once webpack bundle analyzer is installed, which is not at the moment
        // if (arg == 'analyze') {
        // const { BundleAnalyzerPlugin } = require('webpack-bundle-analyzer');
        // this.opts.options.plugins = [new BundleAnalyzerPlugin()];
        // }
        if (args.sourcemap) {
            options.devtool = args.sourcemap || 'eval';
            outputOptions.devtool = args.sourcemap;
        }

        if (outputOptions['dev']) {
            outputOptions['prod'] = undefined;
        }
    }
```

Your solution was correct but here I basically shifted the responsibility to `BasicGroup`. It checks of the argument name is `"entry"`. After that **it checks its value** (this was the bug, it wasn't doing it and it was creating an entry point property every time) and if it has a value, a resolve path is called. I hope it makes sense

Mistyyyy:
Yes, you are right. It's the `BasicGroup's` responsibility to get the right entry for the whole project. However, I can't get the `this.compilerConfiguration` which may contain the config entry In the instance of `BasicGroup`. It should be passed as a parameter when  Initialize the `BasicGroup` instance.

and the `resolveFlags` should like

```javascript
 // The configuration is the result of this.configuration after ConfigGroup executing
 const configuration = obj  
 if (arg === 'entry') {
     // if the boolean of args[arg] is truthy, the command entry exists
     // it should override the config entry
      if (args[arg]) {
          options.entry = this.resolveFilePath(args[arg], 'index.js');
      } else if (!configuration.entry) {
          // the config entry doesn't exist, use defaultEntry
          options.entry = this.resolveFilePath(args[arg], 'index.js');
     }
}
```

ematipico:
I don't think you need the configuration object. I tested that code against your tests and they pass (plus all the current ones pass too)

Mistyyyy:
```javascript
if (arg === 'entry') {
     if (args[arg]) {
        options.entry = this.resolveFilePath(args[arg], 'index.js');
     } else {
          options.entry = this.resolveFilePath(args.entry, 'index.js');
      }
}
```
I think the code above has no difference with the code below, because the `arg` is always `entry`.

```javascript
if (arg === 'entry') {
        options.entry = this.resolveFilePath(args[arg], 'index.js');
}
```

The reason for all tests passes maybe you don't clean the output directory which contain previous test output bundle.

I run this test failed.

```javascript
    it('should use config entry if config entry existed', done => {
        const { stdout } = run(__dirname, ['-c', '../webpack.config.js'], false);
        const outputDir = 'entry-with-config/binary';
        expect(summary['Output Directory']).toContain(outputDir);

        expect(stdout).toContain('./a.js');
        stat(resolve(__dirname, './binary/index.bundle.js'), (err, stats) => {
            expect(err).toBeFalsy();
            expect(stats.isFile()).toBe(true);
            done();
        });
    });

// test Structure
├── a.js
├── entry-with-config
│   ├── entry-with-config.test.js
│   └── index.js
└── webpack.config.js
```

Why I need the configuration object? because the `BasicGroup` only executes once. How I know choose command entry or defaultEntry.

If has config entry, I shouldn't get the defaultEntry and get command entry if `args[arg]` exists.
If no config entry, I should get the defaultEntry or command entry. which depends on if `args[arg]`  exists.

In addition, I think Strategy is the best method to fix this issue if webpack-merge supports it.

ematipico:
> The reason for all tests passes maybe you don't clean the output directory which contain previous test output bundle

You might be right!!

And now that I think about the whole logic, I believe your solution, for now, it's the best option!

ematipico:
@Mistyyyy I merged the PR too soon. Would you mind fixing a failing test? We got rid of the "extractSummary" bit. Sorry man, this was a mistake of mine!

Mistyyyy:
@ematipico Ok, I will fix it.

PR #1299, opened 2020-03-06
feat(webpack-cli): stats

### LINKED ISSUE #1298: [ENHANCEMENT]: Use stats for outputting

We're currently overriding stats to output compilation stats, and we should rather change stats than do this in the CLI.

Stats being refactored and CLI using stats

N/A

**Additional context**

N/A

ematipico:

It implements the `stats` option for webpack via CLI.
Closes #1298

Yes

Yes

**Summary**
These options https://webpack.js.org/configuration/stats/#stats are now passed to the compiler via CLI.

I also removed the "fancy" output with the table. I had a discussion with the core maintainers about the library. The CLI should not alter the output of the compiler. I left `--pretty` as option, so we can still use but it is experimental.

Kind of because it removed `--standard` flag.

**Other information**

### DIFF
--- a/packages/webpack-cli/lib/utils/cli-flags.js
+++ b/packages/webpack-cli/lib/utils/cli-flags.js
@@ -1,4 +1,6 @@
 const { logger } = require('@webpack-cli/logger');
+const StatsGroup = require('../groups/StatsGroup');
+
 const HELP_GROUP = 'help';
 const CONFIG_GROUP = 'config';
 const BASIC_GROUP = 'basic';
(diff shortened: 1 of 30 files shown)


packages/webpack-cli/lib/utils/cli-flags.js
```diff
@@ -315,6 +318,27 @@ module.exports = {
...
+            group: DISPLAY_GROUP,
+            description: 'It instructs webpack on how to treat the stats',
+            link: 'https://webpack.js.org/configuration/stats/#stats',
+        },
+        {
+            name: 'verbose',
+            usage: 'webpack --verbose',
+            type: Boolean,
+            group: DISPLAY_GROUP,
+            description: 'It tells to webpack to output all the information',
```
jamesgeorge007:
```suggestion
            description: 'It tells webpack to output all the information',
```

PR #1313, opened 2020-03-10
Remove the extractSummary methods in test case

Mistyyyy:

Test cases about entry and stats

already exist

NA

**Summary**

No

**Other information**

NA

### DIFF
--- a/test/entry/config-entry/entry-with-command/entry-with-command.test.js
+++ b/test/entry/config-entry/entry-with-command/entry-with-command.test.js
@@ -1,14 +1,11 @@
 'use strict';
 const { stat } = require('fs');
 const { resolve } = require('path');
-const { run, extractSummary } = require('../../../utils/test-utils');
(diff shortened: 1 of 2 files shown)


Mistyyyy:
I think before each test run, we should clean the output info which may contain the previous test result. I encountered this problem when I run the test cases, I get the incorrect test results.

anshumanv:
Hi, thanks for your contribution. 

What's the motivation for this change? Is there some discussion that I've missed? :)

Mistyyyy:
@anshumanv The discussion is here #1289

ematipico:
@anshumanv basically the default output of webpack is changed. That nice output should not be here in the CLI (it should be in webpack side). So in one of my previous PRs I removed moved into `--pretty` and doing so I had to revisit most of the tests. These ones slipped from my eyesight and this PR is to fix them!

I am going to make a new PR to remove `--pretty` because we don't need it.

PR #1314, opened 2020-03-11
chore: reduce install size

evilebottnawi:

Reduce package size

No need

No need

**Summary**

We don't use that package in our code

No

**Other information**

We don't need extra logger because webpack has own loggers


evilebottnawi:
@ematipico Do we already have a PR with fix tests?

PR #1316, opened 2020-03-11
chore(webpack-cli): remove pretty output functionality

ematipico:

I kind of breaking change but it's needed for webpack ecosystem.

I removed the tests that are not needed

Yes

**Summary**
This PR removes `--pretty` argument and all the code behind it. Also, it removes `cli-table3` which is not maintained anymore (it still runs on node 6 and there are no updates on future releases).

It could, but I don't the `--pretty` argument was used so much. An alternative will be available inside webpack core

**Other information**


packages/webpack-cli/lib/groups/StatsGroup.js
```diff
@@ -14,7 +14,7 @@ class StatsGroup extends GroupHelper {
 
     resolveOptions() {
         if (this.args.verbose && this.args.stats) {
-            logger.warn('Conflict between "verbose" and "stats" options. Using verbose.');
+            logger.warn('Conflict between "verbqose" and "stats" options. Using verbose.');
```
evilebottnawi: Typo?

PR #1325, opened 2020-03-15
fix: enable passing silent flag to the compiler

### LINKED ISSUE #1324: webpack-cli: Passing silent flag doesn't work
**Describe the bug**

Still seeing webpack output when the silent flag is passed.

Passing silent flag still logs webpack build data.

**To Reproduce**
`webpack-cli --silent`

**Expected behavior**

Should disable all logs. We should still log errors right?

**Screenshots**
when there are no errs - 
image

**Additional context**

anshumanv:
Fixes #1324 

fix

Yes

NA

**Summary**
- Allow passing the silent flag to the compiler

No

**Other information**
NA

### DIFF
--- a/packages/webpack-cli/lib/groups/StatsGroup.js
+++ b/packages/webpack-cli/lib/groups/StatsGroup.js
@@ -27,20 +27,12 @@ class StatsGroup extends GroupHelper {
                 this.opts.options.stats = this.args.stats;
             }
         }
-<<<<<<< HEAD
-=======
-
-        if (this.args.pretty) {
-            this.opts.outputOptions.pretty = true;
-        }
-
         if (this.args.silent) {
             if (this.args.stats) {
                 logger.warn('"--stats"  will be ignored since you have supplied "--silent" flag');
             }
             this.opts.outputOptions.silent = true;
         }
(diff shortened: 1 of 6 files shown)


anshumanv:
This is like a shorthand for `--stats none` like `--verbose` is for `--stats verbose`

anshumanv:
> I would suggest not to block the flag for this purpose. Instead we can implement --silent flag for the purpose stated above (if we need it).

Did you test it? What are the changes needed here then?

rishabh3112:
Currently it disables only webpack's output. IMO it should disable every possible output
( For example, warnings displayed by CLI, disable progress bar when `--progress` is supplied, etc.). That way consistency isn't compromised.

snitin315:
We already support `webpack --stats=none` and `webpack --no-stats` both to disable output. Maybe we don't need `--silent`? was it present in the old CLI?

anshumanv:
```js
silent: {
				type: "boolean",
				describe: "Prevent output from being displayed in stdout"
			},
```

I think this overlaps with stats-none, let's proceed to remove this, we can add it later if users ask. Will do in near future.

/cc @evilebottnawi

PR #1345, opened 2020-03-17
chore(webpack-cli): improve import-local

snitin315:
**What kind of change does this PR introduce?**

removed stale import-local

**Did you add tests for your changes?**
NA
**If relevant, did you update the documentation?**
NA
**Summary**
Fixes #1318 

**Does this PR introduce a breaking change?**

NO

**Other information**

PR #1348, opened 2020-03-18
chore: fix eslint configuration

evilebottnawi:

chore

No need

No

**Summary**

For https://github.com/webpack/webpack-cli/pull/1345/

No

**Other information**

webpack-cli package do not use modules, don't know why we use that configuration

### DIFF
--- a/packages/webpack-cli/.eslintrc.js
+++ b/packages/webpack-cli/.eslintrc.js
@@ -1,4 +1,4 @@
 module.exports = {
     root: false,
-    parserOptions: { ecmaVersion: 2020, sourceType: 'module' },
+    parserOptions: { ecmaVersion: 2020, sourceType: 'script' },
 };


packages/webpack-cli/.eslintrc.js
```diff
@@ -1,4 +1,4 @@
 module.exports = {
     root: false,
-    parserOptions: { ecmaVersion: 2020, sourceType: 'module' },
+    parserOptions: { ecmaVersion: 2020, sourceType: 'script' },
```
anshumanv: script is default, removing also works

evilebottnawi: Default root config contains `parserOptions: { ecmaVersion: 2020, sourceType: 'module' },` so I override it, in future we should revisit eslint configuration, I think now we can avoid it to continue development

anshumanv: Ah yes, good catch!

PR #1351, opened 2020-03-18
chore: remove stale code and improve logic

snitin315:

Refactoring. 
Removed the stale imports, unused code & improved logic.

NA
NA
**Summary**
The motivation for this change -

Screenshot from 2020-03-18 17-34-51

NO
**Other information**


snitin315:
> Would suggest keeping this open for a while, go through the code and remove all stale code and logic improvement at once.

@anshumanv There are two more warnings in `/package-utils/__tests__`  regarding the use of type `any` -

Screenshot from 2020-03-18 17-49-34

I'm not sure what should be used instead of `any`.

snitin315:
Rest all code looks good to me.

anshumanv:
> @anshumanv There are two more warnings in /package-utils/__tests__ regarding the use of type any -

@snitin315 it should be `jest.Mock` 👍 

Let me know if that works.

snitin315:
> Let me know if that works.

It worked thanks.

anshumanv:
Will take a look locally in a while.

packages/info/src/index.ts
```diff
@@ -4,7 +4,6 @@ import process from 'process';
 import { argv } from './options';
 
 import { AVAILABLE_COMMANDS, AVAILABLE_FORMATS, IGNORE_FLAGS } from './commands';
-import { configReader, fetchConfig, resolveFilePath, getNameFromPath } from './configParser';
```
anshumanv: This file is also unused, can be safely removed. Not really sure why it was introduced, @rishabh3112 any ideas?

snitin315: @anshumanv  this file is being used in some test suites. , 

Screenshot from 2020-03-22 10-59-57

anshumanv: I think you forogt to rebuild, passing fine for me after removing the file.

anshumanv: And no, not used, the fn you showed is in `index.ts` in info, not in `configParser`.

snitin315: Accidently removed some extra code. My bad ;-)

jamesgeorge007:
Thanks

PR #1355, opened 2020-03-20
chore: remove old, unused CI configuration files

ryanclark:
Travis is super slow, and we're no longer using Azure pipelines.

We're fine to rely on GitHub workflows, so I'm removing the old configuration files for these CI providers


evilebottnawi:
I can't find `lint` job https://github.com/webpack/webpack-cli/blob/next/.github/workflows/nodejs.yml + no windows and macos tests + no caches and etc, let's improve it in that PR, here simple example what I can use for many private projects https://github.com/itgalaxy/eslint-plugin-itgalaxy/blob/master/.github/workflows/ci.yml

PR #1358, opened 2020-03-21
chore(logger): remove logger package, inline webpack logger

anshumanv:
Remove logger package.

Yes

Yes

**Summary**
- Remove logger package, inline webpack logger

No

**Other information**
NA


anshumanv:
@evilebottnawi rebased which removed your review, PTAL again. ✨

rishabh3112:
Needs to fix conflicts, rest is fine.

anshumanv:
Fixing in a bit. 👍

anshumanv:
PTAL @evilebottnawi @rishabh3112, let's get this in. 🌟

anshumanv:
Uhoh, taking look at tests.

packages/webpack-cli/lib/utils/logger.js
```diff
@@ -1,3 +1,3 @@
-const { logger } = require('@webpack-cli/logger');
+const { getLogger } = require('webpack/lib/logging/runtime');
```
sokra, line 1: `webpack/lib/logging/runtime` should not be used here. Use `compiler.getInfrastructureLogger("webpack-cli")` instead.

anshumanv: Will update

evilebottnawi: @sokra we use logging before run webpack and after, and when we do other actions which don't need webpack (like `init`/`info`/`migrate`/`generate`), so we can't use infrastructure logger

PR #1359, opened 2020-03-21
feat: add stats detailed option

anshumanv:
added stats detailed option which was unrecognized.

Yes

Yes

**Summary**
- Added detailed stats option which we missed.

No

**Other information**
NA


anshumanv:
@rishabh3112 here - https://github.com/webpack/webpack/blob/f0b1a71d5a4ba7a685e76d08281195364d7dd9b3/lib/stats/DefaultStatsPresetPlugin.js#L47

Probably it needs to be documented too, I'll open an issue in the docs?

packages/webpack-cli/lib/groups/StatsGroup.js
```diff
@@ -5,7 +5,7 @@ const { logger } = require('@webpack-cli/logger');
  */
 class StatsGroup extends GroupHelper {
     static validOptions() {
-        return ['minimal', 'none', 'normal', 'verbose', 'errors-warnings', 'errors-only'];
+        return ['minimal', 'none', 'normal', 'verbose', 'errors-warnings', 'errors-only', 'detailed'];
```
evilebottnawi: @rishabh3112 All values are valid https://github.com/webpack/webpack/blob/master/schemas/WebpackOptions.json#L2605

evilebottnawi: @anshumanv Can you reorder values as here https://github.com/webpack/webpack/blob/master/schemas/WebpackOptions.json#L2605?

rishabh3112:
Thanks for in site @anshumanv. It is good now :)

packages/webpack-cli/lib/groups/StatsGroup.js
rishabh3112: Why not extract it from there ?

evilebottnawi: @rishabh3112 We will do it late, it is part of task - `Args from webpack`

anshumanv:
@evilebottnawi @rishabh3112, should be good now. :)

packages/webpack-cli/lib/groups/StatsGroup.js
rishabh3112: Okay!

Issue #1297, opened 2020-03-06
docs: Adding readme.md in folders


himanshu010:
**Do you want to request a *feature* or report a *bug*?**

**What is the current behavior?**
The readme file does not have the file system describing the locations of files. Sometimes it is very tedious to find a particular file, therefore projects like this needs a fully descriptive file system.

**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**
The readme file should contain file system. 

**If this is a feature request, what is motivation or use case for changing the behavior?**

**Please paste the results of `webpack-cli info` here, and mention other relevant information such as programming language.**

I would Like to work on this issue

ematipico:
Could you help me here and explain me what you'd like to see?

himanshu010:
it is something like this : 

![Screenshot (27)](https://user-images.githubusercontent.com/43860289/76080754-9f135480-5fcd-11ea-8649-b0889137067e.png)

maria99williams:
Yeah I think that it would be great for someone like me as I am new to open source. I am looking forward to contribute to webpack.

anshumanv:
+1 for this as this could provide a general overview of the code.

snitin315:
It's Nice to have a folder structure.

ematipico:
Which section would you think makes sense to have it? The contribution section?

snitin315:
Yes Contribution .

himanshu010:
> Which section would you think makes sense to have it? The contribution section?

Yeah

ematipico:
Thank you! Feel free to make a PR!

rishabh3112:
@himanshu010 I didn't get the exact benefit it would bring to the first time contributors as we have individual READMEs in each package which explains their functionalities.

Instead I would like to propose to add READMEs in the folders which you think you faced issues to understand, something like this:
https://github.com/webpack/webpack/blob/master/examples/README.md

himanshu010:
okay, I would work on this.

adrijshikhar:
Is this issue still open? I have been using webpack for quite a while now and I absolutely love it. I feel like this will be a nice starting point. I will go through the existing PRs and post a final conclusion if that suits or if anyone else has an idea of what has to be done here, please guide me through that.

snitin315:
I think we already have Readme files at appropriate  places.
/cc @webpack/cli-team

evilebottnawi:
Yep, we don't need readme at each directory

[...]

PR #1360, opened 2020-03-21
docs: adding new readme files

himanshu010:
fixes: #1297 

docs related change
N/A
yes
**Summary**

**Other information**


himanshu010:
@rishabh3112 can you help me with the readme files of smoketest and docs

scripts/README.md
```diff
@@ -0,0 +1,98 @@
+# scripts
+
+This directory contains javascripts file to build and prepare packages.
+
+## buildPackages.js
+
+```javascript
+'use strict';
+
+const fs = require('fs');
```
anshumanv: Do we really need the entire code in README?

snitin315: This is just 1 line readme otherwise.

himanshu010: I have seen some readme files which contains the entire code of important scripts files.
Screenshot (16)
https://github.com/webpack/webpack/tree/master/examples/aggressive-merging

therefore I inserted the whole code in the readme.

anshumanv: Can do that for important files like config as your example, but not sure if everyone needs to read the buildPackages (an example) code in a README, if they want they can always refer the actual file fwiw.

himanshu010: np i will remove these

himanshu010: Screenshot (17)
@anshumanv however these are also some of the files which are displayed and i think that they are not as important as the config file

anshumanv: We have a separate documentation site altogether with additional docs, for devs who want to actually get started with the codebase, what we have right now should suffice IMO. Anyways, that's just my say, would wait for thoughts from others. 😄

snitin315: I agree with @anshumanv

scripts/README.md
```diff
@@ -0,0 +1,98 @@
+# scripts
```
rishabh3112: Having whole code don't make sense here. Any one can look up to the file and read the code.

rishabh3112: Please add short descriptions about the folders only, not the files as they are bound to change in logic in every PR.

.github/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+### Pull Request
+
+    -   [PULL_REQUEST_TEMPLATE](https://github.com/webpack/webpack-cli/tree/next/.github/PULL_REQUEST_TEMPLATE) : For new pull request
+
+## Others
+
+    -   [Workflows](https://github.com/webpack/webpack-cli/tree/next/.github/workflows)
+    -   [CONTRIBUTING.md](https://github.com/webpack/webpack-cli/tree/next/.github/CONTRIBUTING.md) : Contains informations regarding project setup and contribution
+    -   [FUNDING.yml](FUNDING.yml) : Contains information regarding funds collected by webpack organisation
```
evilebottnawi, line 26: No need

test/README.md
```diff
@@ -0,0 +1,8 @@
+# test
+
+This directory contains all the neccesary tests to check to check that whether the project runs perfectly or is there some error.
+
+## Testing is essential because:
+
+-   To point out the defects and errors that were made during the development phases.
+-   It’s important to ensure that the application should not result into any failures
```
evilebottnawi, line 8: No need

PR #1366, opened 2020-03-23
fix(packages): make packages have correct main paths to index

Loonride:
fix

N/A

N/A

**Summary**

Some of the paths to main index scripts were incorrect, causing the modules to fail

No

**Other information**

Edit: I went through all the packages to make sure everything was correct, but certainly double-check that I didn't miss anything.


yarn.lock
```diff
@@ -12467,80 +12467,80 @@ webpack-dev-middleware@^3.7.0, webpack-dev-middleware@^3.7.2:
...
+    webpack-dev-middleware "^3.7.2"
     webpack-log "^2.0.0"
     ws "^6.2.1"
     yargs "12.0.5"
 
-webpack-dev-server@^3.9.0:
-  version "3.10.3"
-  resolved "https://registry.yarnpkg.com/webpack-dev-server/-/webpack-dev-server-3.10.3.tgz#f35945036813e57ef582c2420ef7b470e14d3af0"
-  integrity sha512-e4nWev8YzEVNdOMcNzNeCN947sWJNd43E5XvsJzbAL08kGc2frm1tQ32hTJslRS+H65LCb/AaUCYU7fjHCpDeQ==
+webpack-dev-server@3.8.0:
```
evilebottnawi: I think here something wrong, we don't update lock file

Loonride: I reverted it back to original. It looks like the lock file was updated in a recent commit anyway, so there should be no problems either way.

PR #1375, opened 2020-03-24
tests(utils): add packageUtils tests

Loonride:
tests/fix

Yes

N/A

**Summary**
I added tests for the `packageUtils` file.

I also made some fixes and changes to the `getPackageManager` function. If a `yarn.lock` is not present but a `package-lock.json` is present, I think we should assume the user wants to use npm.

Yes, as noted above in the `getPackageManager` function

**Other information**


evilebottnawi:
Thanks!

PR #1382, opened 2020-03-26
chore: refactor info package

rishabh3112:
refactor/feature.
- Display all info and remove unnecessary flags

WIP

WIP

**Summary**
Would be adding info package as direct dependency as per roadmap

No

**Other information**
No


test/info/info-help.test.js
```diff
@@ -28,13 +28,6 @@ describe('should print help for info command', () => {
...
-    it('should output help even if other flags are supplied with help', () => {
-        const { stdout, stderr } = runInfo(['info', '--help', '--system']);
-        expect(stdout).toContain(usageText);
-        expect(stdout).toContain(descriptionText);
-        expect(stdout).not.toContain('System:');
-        expect(stderr).toHaveLength(0);
-    });
 
     it('should respect the --color=false flag', () => {
         const { stdout, stderr } = runInfo(['info', 'help', '--color=false']);
```
evilebottnawi: Please improve tests, we should check what it is included `npm`/`yarn`/`webpack`/etc strings

packages/info/src/index.ts
```diff
@@ -13,13 +13,6 @@ interface Information {
...
-interface ArgvI {
-    _?: string[];
-    bin?: boolean;
-    binaries?: boolean;
-    config?: string;
-}
-
 const DEFAULT_DETAILS: Information = {
     Binaries: ['Node', 'Yarn', 'npm'],
     Browsers: ['Chrome', 'Firefox', 'Safari'],
```
evilebottnawi: Need add all possible browsers

packages/info/src/options.ts
```diff
@@ -1,42 +1,7 @@
...
-		demandOption: false,
-		describe: "To get the output as markdown",
-		type: "boolean"
-	})
-	.group(["output-json", "output-markdown"], `Output format`).argv;
+export default [
+    {
+        name: 'output',
+        type: String,
+        description: 'To get the output in specified format ( accept json or markdown )',
```
evilebottnawi: Avoid spaces after `(` and before `)`

test/info/info-output.test.js
```diff
@@ -18,7 +18,7 @@ describe('basic info usage', () => {
     });
 
     it('gets info as json', async () => {
-        const { stdout, stderr } = await runInfo(['--output-json']);
+        const { stdout, stderr } = await runInfo(['--output="json"']);
         expect(stdout).toContain('"System":');
         expect(stderr).toHaveLength(0);
```
evilebottnawi: Please add test what output is valid JSON - `JSON.parse` should not throw an error and contains basic fields

test/info/info-help.test.js
evilebottnawi: Not only for that test

packages/info/src/index.ts
```diff
@@ -28,55 +21,30 @@ const DEFAULT_DETAILS: Information = {
...
-            return { npmGlobalPackages: DEFAULT_DETAILS.npmGlobalPackages };
-        case 'npmPackages':
-            return { npmPackages: DEFAULT_DETAILS.npmPackages };
-    }
-}
-
-export default async function info(customArgv: object): Promise<string[]> {
-    let detailsObj = {};
+export default async function info(args: object): Promise<string[]> {
+    console.log(args);
```
anshumanv: rm logs

test/info/info-help.test.js
rishabh3112: This tests is for help flag used with `info`. But will add these in the tests for `info` output.

test/info/info-output.test.js
rishabh3112: JSON parse test exist already ( Just below were you have made comment. )
Would add tests for basic fields.

packages/info/src/index.ts
```diff
@@ -13,70 +12,50 @@ interface Information {
...
-    bin?: boolean;
-    binaries?: boolean;
-    config?: string;
-}
-
 const DEFAULT_DETAILS: Information = {
     Binaries: ['Node', 'Yarn', 'npm'],
-    Browsers: ['Chrome', 'Firefox', 'Safari'],
+    Browsers: [
+        'Brave Browser',
```
evilebottnawi: `Brave Browser`? What is it? :smile:

rishabh3112: https://brave.com/
A browser form a company of Brendan Eich ✨

rishabh3112:
Snapshots for testing is a bad idea IMO. On different testing environments (windows, linux, macOS, ...) output would be different on each one of them.

Here is output:
image

image

image

I use WSL and don't have any browser installed over there, so no browsers :)

evilebottnawi:
> Snapshots for testing is a bad idea IMO. On different testing environments (windows, linux, macOS, ...) output would be different on each one of them.

No, it is good idea, we should remove platform specific output from snapshot (like do it webpack)

rishabh3112:
@evilebottnawi **System** output is very important for info command, and it is system specific IMO.

evilebottnawi:
Now I have to spar every time what the output looks like, it’s extremely inconvenient and I can skip some regressions, run it locally every time and look very time-consuming, so we need to simplify this

evilebottnawi:
@rishabh3112 I will improve tests in future and you see how it is better, let's avoid it here, just provide example current default output

rishabh3112:
Here it is https://github.com/webpack/webpack-cli/pull/1382#issuecomment-606588072

rishabh3112:
@evilebottnawi also should I add the `info` as direct dependency in this PR itself, or should we release it `info` first to npm with current changes?

evilebottnawi:
Let's rename `npmPackages:` to `Packages:` and `globalNpmPackages` to `Global packages:`

> @evilebottnawi also should I add the info as direct dependency in this PR itself, or should we release it info first to npm with current changes?

In the next PR

rishabh3112:
Replaced output:
image

packages/info/src/index.ts
evilebottnawi: Do you know who use it? :smile:

rishabh3112: I use it (+edge chromium).
https://brave.com/brave-passes-10m-mau/

rishabh3112: Nevertheless, not a blocker for this PR anyways.

evilebottnawi:
Thanks, feel free to integrate env package into CLI

PR #1399, opened 2020-04-01
chore: refactor typescript configuration

evilebottnawi:

chore and refactor

No need

No need

**Summary**

Simplify configuration + avoid source maps for dist (no need it, because it is increase package size) + migrate on one script to build without code + clean directories before build (avoid for files to be part of release but they do not used)

No

**Other information**

No


packages/generate-loader/package.json
```diff
@@ -19,6 +19,7 @@
     "yeoman-environment": "2.8.1"
   },
   "scripts": {
+    "clean": "del-cli \"lib/**/!(*.tpl)\" \"*.tsbuildinfo\"",
```
evilebottnawi, line 22: Workaround + https://github.com/microsoft/TypeScript/issues/30602

PR #1418, opened 2020-04-04
tests: add test for zero config

anshumanv:

tests

Yes

NA

**Summary**
- Add some test cases for zero-config.

No

**Other information**
- Any suggestion on how to strengthen are appreciated.

### DIFF
--- a/test/zero-config/entry-present/zero-config.test.js
+++ b/test/zero-config/entry-present/zero-config.test.js
@@ -1,3 +1,5 @@
+const fs = require('fs');
+const path = require('path');
 const { run } = require('../../utils/test-utils');
 
 describe('Zero Config tests', () => {
@@ -7,6 +9,8 @@ describe('Zero Config tests', () => {
         expect(stdout).toContain('./src/index.js');
         // Should output at the default output dir and filename
         expect(stdout).toContain('Entrypoint main = main.js');
+        // check that the output file exists
+        expect(fs.existsSync(path.join(__dirname, '/dist/main.js'))).toBeTruthy();
         expect(stderr).toBeFalsy();
     });
 });
--- a/test/zero-config/entry-present/zero-config.test.js
+++ b/test/zero-config/entry-present/zero-config.test.js
@@ -5,7 +5,7 @@ describe('Zero Config tests', () => {
         const { stdout, stderr } = run(__dirname, [], false);
         // Should be able to find the entry file
         expect(stdout).not.toContain("Module not found: Error: Can't resolve './src'");
-        expect(stdout).toContain('./src/index.js 21 bytes [built]');
+        expect(stdout).toContain('./src/index.js');
         // Should output at the default output dir and filename
         expect(stdout).toContain('Entrypoint main = main.js');
         expect(stderr).toBeFalsy();
--- a/test/zero-config/entry-present/zero-config.test.js
+++ b/test/zero-config/entry-present/zero-config.test.js
@@ -4,7 +4,6 @@ describe('Zero Config tests', () => {
     it('runs when no config is supplied but entry is present', () => {
         const { stdout, stderr } = run(__dirname, [], false);
         // Should be able to find the entry file
-        expect(stdout).not.toContain("Module not found: Error: Can't resolve './src'");
         expect(stdout).toContain('./src/index.js');
         // Should output at the default output dir and filename
         expect(stdout).toContain('Entrypoint main = main.js');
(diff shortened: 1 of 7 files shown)


test/zero-config/with-config/zero-config-entry.test.js
```diff
@@ -0,0 +1,15 @@
+const { run } = require('../../utils/test-utils');
+
+describe('Zero Config', () => {
+    it('runs when config is present but not supplied via flag', () => {
+        const { stdout, stderr } = run(__dirname, [], false);
+        console.log({ stdout, stderr });
```
jamesgeorge007, line 6: Unnecessary log.

test/zero-config/entry-present/zero-config.test.js
```diff
@@ -0,0 +1,13 @@
+const { run } = require('../../utils/test-utils');
+
+describe('Zero Config tests', () => {
+    it('runs when no config is supplied but entry is present', () => {
+        const { stdout, stderr } = run(__dirname, [], false);
+        // Should be able to find the entry file
+        expect(stdout).not.toContain("Module not found: Error: Can't resolve './src'");
```
evilebottnawi, line 7: No need test it

test/zero-config/with-config/zero-config-entry.test.js
```diff
@@ -0,0 +1,14 @@
+const { run } = require('../../utils/test-utils');
+
+describe('Zero Config', () => {
+    it('runs when config is present but not supplied via flag', () => {
+        const { stdout, stderr } = run(__dirname, [], false);
+        // Should find the entry file, thus not throw the below warning
+        expect(stdout).not.toContain("Module not found: Error: Can't resolve './src'");
```
evilebottnawi, line 7: No need test it

test/zero-config/entry-present/zero-config.test.js
```diff
@@ -0,0 +1,12 @@
+const { run } = require('../../utils/test-utils');
+
+describe('Zero Config tests', () => {
+    it('runs when no config is supplied but entry is present', () => {
+        const { stdout, stderr } = run(__dirname, [], false);
+        // Should be able to find the entry file
+        expect(stdout).toContain('./src/index.js');
+        // Should output at the default output dir and filename
+        expect(stdout).toContain('Entrypoint main = main.js');
```
anshumanv, line 9: We already have it, only entry present and config absent. @evilebottnawi

evilebottnawi:
Can you provide here structure what we generate on init by default?

anshumanv:
on init? as in?

anshumanv:
The default config you mean?

evilebottnawi:
@anshumanv `init`, we need investigate unnecessary stuff

anshumanv:
> @anshumanv init, we need investigate unnecessary stuff

init tests are here - https://github.com/webpack/webpack-cli/pull/1419

evilebottnawi:
@anshumanv Feel free to merge after CI green

anshumanv:
Cool, need another review then I'll merge once green. 👍

test/zero-config/entry-present/zero-config.test.js
```diff
@@ -0,0 +1,12 @@
+const { run } = require('../../utils/test-utils');
+
+describe('Zero Config tests', () => {
+    it('runs when no config is supplied but entry is present', () => {
```
rishabh3112, line 4: We could also test for existence of bundle output file.

anshumanv: We can def do that, thinking if this would be the right scope for it as output file check is something which is config/flag agnostic. wdyt? Since we're not checking for it anywhere else, should be also checked for it in all other tests?

anshumanv: Or maybe we can just check the test util since it's used everywhere, will add tests for the `run` fn that we use in all tests separately. Would make more sense to check file existence there.

rishabh3112: In this particular case we should test it either way imo.  Because the test is for checking whether without config we are bundling or not.

Tests for run as well in a different PR would be great :)

jamesgeorge007: > Or maybe we can just check the test util since it's used everywhere, will add tests for the `run` fn that we use in all tests separately. Would make more sense to check file existence there.

We're consuming the `run` fn for couple of test cases where bundle output is not generated, say the `migrate` command. It should be taken into account while writing the respective tests.

anshumanv: Sounds fair, added. 👍

PR #1419, opened 2020-04-05
tests: add tests for init

rishabh3112:
Tests for init package

Yes

Not Applicable

**Summary**
Add tests for init package as part of roadmap.

No

**Other information**
After this we would discuss more about the scaffold strategy and integrate it with CLI


test/init/auto/init-auto.test.js
```diff
@@ -0,0 +1,30 @@
...
+const firstPrompt = 'Will your application have multiple bundles?';
+const fs = require('fs');
+const { join } = require('path');
+
+describe('init auto flag', () => {
+    it('should prompt with w/o auto flag', () => {
+        const { stdout, stderr } = run(__dirname, ['init'], false);
+        expect(stdout).toBeTruthy();
+        expect(stderr).toBeFalsy();
+        expect(stdout).toContain(firstPrompt);
```
evilebottnawi, line 14: Can we improve it, and answer on all questions?

rishabh3112: @evilebottnawi the future questions changes when we answer with different values. So should should I test all possible question flows or one of them?

evilebottnawi: Ideally - yes

rishabh3112: This would take a lot of CI time and memory.
This may result in being flaky, as if we would add/change/remove one question from generator, Or we change the functionality used by generator, it would break the while test suite for init.
In past, tests for init were removed because of this reason
Rest I would do as you suggest :)

anshumanv: I guess one full example would be good, we can check to the point it generates the config.

evilebottnawi: > @evilebottnawi This would take a lot of CI time and memory.

> This may also result in tests being flaky, as if we would add/change/remove one question from generator, Or we change the external functionality which is some where used by generator, it would break the whole test suite for init.

> In past, these type of tests for init were removed because of this reason

> Rest I would do as you suggest :)

When something removed we need remove it from tests, this is testing, which allows us to understand that everything is fine, the tests are stable, and if something doesn’t work, then we have an error, I don’t know why someone decided to delete them, this is a very, very bad idea

evilebottnawi: > I guess one full example would be good, we can check to the point it generates the config.

Yes, as minimum we should have test for it

rishabh3112: Great, thanks for suggestions.

packages/webpack-cli/__tests__/init/generator/init-inquirer.test.js
```diff
@@ -0,0 +1,73 @@
...
+    });
+
+    return new Promise((resolve) => {
+        runner.stdout.pipe(
+            concat((result) => {
+                resolve(result.toString());
+            }),
+        );
+    });
+};
```
evilebottnawi, line 55: Move it to `test/utils/test-utils.js`

packages/webpack-cli/__tests__/init/auto/init-auto.test.js
```diff
@@ -0,0 +1,30 @@
+/* eslint-disable node/no-unpublished-require */
+'use strict';
+
+const firstPrompt = 'Will your application have multiple bundles?';
+const fs = require('fs');
+const { join } = require('path');
+const { run } = require('../../../../../test/utils/test-utils');
```
anshumanv, line 7: Won't it make more sense to add these directly in `test` dir?

.gitignore
```diff
@@ -54,3 +54,7 @@ packages/**/*.map
 
 # temporary test files
 test-assets/
+
+# ignore scaffolded files due to tests of init
+test/init/*
```
anshumanv, line 59: Can get rid of this.

test/utils/test-utils.js
```diff
@@ -80,6 +82,35 @@ function runAndGetWatchProc(testCase, args = [], setOutput = true) {
     return webpackProc;
 }
 
+const runInitWithAnswers = async (location, answers) => {
```
anshumanv, line 85: Let's make this generic and not tied to init, maybe we can have a third param which takes in a runner.

packages/webpack-cli/__tests__/init/auto/init-auto.test.js
rishabh3112: Having them here makes more sense as they are integration tests with `webpack-cli`, at root we should have integration tests for `webpack` as suggested by @evilebottnawi above.

I think we should re structure `test-utils.js`, I would suggest to have it it's package of it own.

.gitignore
rishabh3112: No, but we need to add path to tests' new location.

test/utils/test-utils.js
rishabh3112: Yes we can, good idea! But instead of taking runner should take params to be passed to the CLI.
This function then could be used for e2e tests :)

packages/webpack-cli/__tests__/init/auto/init-auto.test.js
anshumanv: Sounds good.

.gitignore
anshumanv: Yes need to be moved there.

packages/webpack-cli/__tests__/init/auto/init-auto.test.js
anshumanv: But in a way root tests also use webpack-cli, and this is like testing the whole command from start to end so still makes more sense at root to me, @evilebottnawi what was your motivation?

test/utils/test-utils.js
anshumanv: Yes, that would be better

packages/webpack-cli/__tests__/init/auto/init-auto.test.js
jamesgeorge007: Yes indeed, having integration tests for `webpack-cli` at the root would be better. Infact this is the case for other packages.

test/utils/test-utils.js
```diff
@@ -80,6 +82,35 @@ function runAndGetWatchProc(testCase, args = [], setOutput = true) {
...
+    const runner = runAndGetWatchProc(location, ['init'], false);
+    runner.stdin.setDefaultEncoding('utf-8');
+
+    // Simulate answers buy sending the answers after waiting for 1s
```
jamesgeorge007, line 89:
```suggestion
    // Simulate answers by sending the answers after waiting for 1s
```

wizardofhogwarts: @rishabh3112 this suggestion is correct

rishabh3112: Thanks for pointing typo :)

test/utils/test-utils.js
```diff
@@ -80,6 +82,35 @@ function runAndGetWatchProc(testCase, args = [], setOutput = true) {
     return webpackProc;
 }
 
+const runInitWithAnswers = async (location, answers) => {
+    const runner = runAndGetWatchProc(location, ['init'], false);
+    runner.stdin.setDefaultEncoding('utf-8');
+
+    // Simulate answers by sending the answers after waiting for 1s
```
jamesgeorge007, line 89: It seems the wait time is `2s`, why not update the comment accordingly?

rishabh3112: Did that in latest commit. Thanks!

jamesgeorge007:
Thanks

PR #1442, opened 2020-04-11
chore(serve): webpack-dev-server@3 deprecated warnings

Loonride:

chore: deprecation warnings

Not yet

No

**Summary**

Adds a warning when user runs `webpack-cli serve` with a webpack-dev-server version below v4. This warning will be pushed to them before and after their webpack compilation to ensure they see it.

Edit: Should I go into greater detail in the warning as to what unexpected behavior may occur?

No

**Other information**


packages/serve/src/index.ts
```diff
@@ -11,6 +13,10 @@ import argsToCamelCase from './args-to-camel-case';
  * @returns {Function} invokes the devServer API
  */
 export default function serve(...args): void {
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    const devServerPackage = require('webpack-dev-server/package.json');
```
evilebottnawi, line 17: try/catch

packages/serve/src/index.ts
```diff
@@ -27,6 +33,23 @@ export default function serve(...args): void {
...
+            const warning = chalk.yellow.bold(
+                'Using webpack-dev-server v3 with ' +
+                    `${chalk.green('webpack-cli serve')} is deprecated and will lead to ` +
+                    'unexpected behavior. Please upgrade to webpack-dev-server v4.',
+            );
+            server.log.warn(warning);
+
+            // send a warning after the compilation to ensure that the message
+            // reaches the user
+            compiler.hooks.done.tap('serve', () => {
```
evilebottnawi, line 48: Why?

Loonride: When the user compiles something, all the compilation stats will be printed and the message will be lost unless it's at the bottom

Loonride: dev-server-deprecated-warning

evilebottnawi: Interesting place, I think we don't need emit warning(s) to stats in such situation

/cc @webpack/cli-team

Loonride: Personally I think it will help the user realize why there might be unexpected behavior.

snitin315 (changes requested):
closing as stale.

PR #1446, opened 2020-04-11
tests: add more tests cases for serve

anshumanv:

tests

Yes

NA

**Summary**
- Adds more tests cases for serve

No

**Other information**


test/utils/test-utils.js
```diff
@@ -190,9 +190,19 @@ async function runInstall(cwd) {
     });
 }
 
+const runServe = (args, testPath) => {
```
jamesgeorge007, line 193: I would say the former implementation was good. Why should `testPath` be passed in  explcitly?

anshumanv: Because runServe needs to be shared, and needs the path of the test to run

jamesgeorge007: Yeah, makes sense 👍

PR #1448, opened 2020-04-12
tests(generators): plugins util tests

Loonride:

tests

Yes

No

**Summary**

generators plugins util tests

No

**Other information**


Loonride:
> Some tests seem to fail, can you check?

It seems like a bad test run, nothing in my changes is causing the problem as far as I can tell

anshumanv:
> It seems like a bad test run, nothing in my changes is causing the problem as far as I can tell

Yes just saw, unrelated serve test is failing. Re-running should fix.

Loonride:
Let me push another commit, I forgot to test one case anyway

anshumanv:
Unusual, init-generator tests seem to be timing out now.

PR #1450, opened 2020-04-12
chore(init): remove unused transform files

Loonride:

chore

No

No

**Summary**

Correct me if I'm wrong, but I think these are old files scaffold files that were never removed. The new scaffold file that init-generator uses can be found here: https://github.com/webpack/webpack-cli/blob/next/packages/utils/src/scaffold.ts

It was moved because it should eventually function as a general-purpose scaffold tool that isn't just for initializing projects, but also modifying them

No, because none of these files are used anywhere

**Other information**


rishabh3112:
> Correct me if I'm wrong, but I think these are old files scaffold files that were never removed. The new scaffold file that init-generator uses can be found here: https://github.com/webpack/webpack-cli/blob/next/packages/utils/src/scaffold.ts

Yes, you are right.

PR #1452, opened 2020-04-12
tests: increase coverage for webpack-scaffold

jamesgeorge007:
add test cases

Yep

Nope

**Summary**
Added missing out test cases for `webpack-scaffold`

Nope

**Other information**
N/A

### DIFF
--- a/packages/webpack-scaffold/__tests__/index.test.ts
+++ b/packages/webpack-scaffold/__tests__/index.test.ts
@@ -65,6 +65,16 @@ describe('utils', () => {
         });
     });
     describe('Inquirer', () => {
+        it('should emulate a prompt for List', () => {
+            expect(List(this.mockSelf, 'entry', 'does it work?', ['Yes', 'Maybe'], 'Yes')).toEqual({
+                choices: ['Yes', 'Maybe'],
+                type: 'list',
+                name: 'entry',
+                message: 'does it work?',
+                default: 'Yes',
+            });
+        });
+
         it('should make default value for a List', () => {
             expect(List(this.mockSelf, 'entry', 'does it work?', ['Yes', 'Maybe'], 'Yes', true)).toEqual({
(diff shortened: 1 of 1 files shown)


packages/webpack-scaffold/src/index.ts
```diff
@@ -109,7 +109,7 @@ export function InputValidate(
     message: string,
     cb?: (input: string) => string | boolean,
     defaultChoice?: string,
-    skip?: boolean,
+    skip = false,
```
anshumanv, line 112: ??

jamesgeorge007: To have a consistent style, if you take a look at other `prompts` we've an initial value for the `skip` argument. Also, there is type inference in place.

anshumanv: Consistent style doesn't mean it's correct, you can give default fn params while maintaining types, please do that.

anshumanv: In all other instances too

evilebottnawi, line 112: Why do you change it?

jamesgeorge007: Reverted.

jamesgeorge007: Yeah that would capture type mismatch errors AOT.

jamesgeorge007: I would say, having it as a default parameter would be a better idea such that it doesn't stay `undefined`.

evilebottnawi: Please avoid it, that PR is only tests, if you want to refactor it, do it in the other PR

jamesgeorge007: In that case, I guess we're good to land

PR #1460, opened 2020-04-14
fix: json flag, enable tests

### LINKED ISSUE #1459: cli[BUG]: --json flag doesn't work
**Describe the bug**

--json flag does not work.

--json flag is not respected by the cli, tests are also skipped.

**To Reproduce**

`webpack-cli --json`

**Expected behavior**

Should output the JSON data from the compiler.

**Screenshots**

NA

**Additional context**
- Fix is wip.

anshumanv:

- fix json flag
- enable tests 

Yes

NA

**Summary**
Fixes #1459 

No

**Other information**
Please suggest if I can add any other tests cases.


test/json/json.test.js
```diff
@@ -1,23 +1,39 @@
...
-            entry: './index.js',
-            output: {
-                filename: 'main.js',
-                path: join(__dirname, 'bin'),
-            },
-        });
-        compiler.run((err, stats) => {
-            expect(err).toBeFalsy();
-            const webpackStats = stats.toJson({ json: true });
-            expect(jsonstdout).toEqual(webpackStats);
```
anshumanv, line 20: This was invalid from the start, since 2 compilations can never have same hashes and times are also different in most cases, so this should always fail ideally.

rishabh3112: Yes, you are right!

PR #1464, opened 2020-04-15
chore(webpack-scaffold): use default params

jamesgeorge007:
refactoring

Nope

Nope

**Summary**
Follow up of #1452

Nope

**Other information**
N/A

### DIFF
--- a/packages/webpack-scaffold/src/index.ts
+++ b/packages/webpack-scaffold/src/index.ts
@@ -109,7 +109,7 @@ export function InputValidate(
     message: string,
     cb?: (input: string) => string | boolean,
     defaultChoice?: string,
-    skip?: boolean,
+    skip = false,
 ): object | any {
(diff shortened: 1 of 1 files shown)


packages/webpack-scaffold/src/index.ts
```diff
@@ -58,7 +58,7 @@ export function List(
     message: string,
     choices: string[],
     defaultChoice?: string,
-    skip = false,
+    skip: boolean = false, // eslint-disable-line @typescript-eslint/no-inferrable-types
```
evilebottnawi, line 61: We have rule for it to avoid it, why you want to change it? It doesn't have sense because it is already boolean

jamesgeorge007: It was put forward by as a suggestion in the previous PR. But yeah as you said it won't be a good idea.

PR #1470, opened 2020-04-16
 fix: pass devServer args to server options

### LINKED ISSUE #1469: Serve command not passing webpack `devServer` config to webpack-dev-server
**Describe the bug**

Serve command not passing webpack `devServer` config to webpack-dev-server

#### webpack.dev.js
```
  devServer: {
    https: isUsingHTTPS && secureLocalDomain,
    host: "myreactapp.test",
    hot: true,
    port: 8080,
    publicPath: "/myreactapp",
    quiet: true,
    proxy: {
      '**': {
        target: localURL,
        secure: false,
        headers: {
          'X-Dev-Server-Proxy': localURL
        }
      }
    }
  }
```

#### command used
`webpack serve --config webpack.dev.js`

The webpack server runs but doesn't pick up all the options from the `devServer` config. You can see this by `webpack output` being undefined in the CLI output, but also the proxy was also not working.

```
ℹ ｢wds｣: Project is running at http://myreactapp.test:8080/
ℹ ｢wds｣: webpack output is served from undefined
ℹ ｢wds｣: Content not from webpack is served from /path/to/my/reactapp
```

**To Reproduce**

#### Relevant entries from `package.json`
```json
"@webpack-cli/serve": "^1.0.1-alpha.5",
"webpack": "^5.0.0-beta.13",
"webpack-cli": "^4.0.0-beta.8",
```

1. Define a `devServer` config in your webpack config file
2. Run `webpack serve --config webpack.dev.js`

**Expected behavior**

Previously the dev server would pick up `devServer` config object from your webpack config file 

**Additional context**

If I change [line 28](https://github.com/webpack/webpack-cli/blob/1c8d0ce94c1ab55f2b0ec99f4c84ca677c0fc24d/packages/serve/src/startDevServer.ts#L28) from `startDevServer.ts` it works as expected but obviously this ignores the cli overrides.

```diff
- const server = new Server(compiler, options);
+ const server = new Server(compiler, devServerOptions);
```

CLI output
```
ℹ ｢wds｣: Project is running at http://myreactapp.test:8080/
ℹ ｢wds｣: webpack output is served from /eccreactapp/dist
ℹ ｢wds｣: Content not from webpack is served from /path/to/my/reactapp
```

anshumanv:

fix

Not yet.

NA

**Summary**
- pass additional devServer config from config file to the server 

No

**Other information**
Fixes #1469


packages/serve/src/startDevServer.ts
```diff
@@ -11,21 +11,25 @@ import Server from 'webpack-dev-server/lib/Server';
...
 
-    options.host = host;
-    options.port = port;
     if (socket) {
         options.socket = socket;
     }
 
-    const server = new Server(compiler, options);
+    // Compose config from the CLI and devServer object from webpack config
+    const serverConfig = { ...devServerOptions, ...options };
```
anshumanv, line 28: Problem - this way we won't be able to override default options unless specified via flag. eg. `clientLogLevel` by default is set to 'info', so if we pass any other option in devServer, it won't take precedence, but we also need to support flags like `--hot` and override them with the devServer config so that's a dilemma

evilebottnawi: CLI args should be override config values, what is a problem, can you clarify with examples?

anshumanv: If we don't need to override config values then I think we're fine with devServer options taking precedence.

packages/serve/src/startDevServer.ts
```diff
@@ -11,21 +11,25 @@ import Server from 'webpack-dev-server/lib/Server';
...
 
-    options.host = host;
-    options.port = port;
     if (socket) {
         options.socket = socket;
     }
 
-    const server = new Server(compiler, options);
+    // Compose config from the CLI and devServer object from webpack config
+    const serverConfig = { ...options, ...devServerOptions };
```
Loonride, line 28: I think I understand what you are saying. Do you mean the issue is that if a CLI flag has a default value, it will override `devServer` regardless of what the user has set for that value in `devServer`? Yes, this is an issue.

Our goal is that CLI args should take precedence over `devServer`, so this line here is not the perfect solution.

What we need to do is find a way to check if the user has actually provided the CLI flag, or if it is just getting passed in as a default value. It might be easier to hold off on this until we swap over to `commander` as the CLI flags parser. Ideally we can pass a CLI options object into `startDevServer` which indicates whether each flag was defaulted to, or whether the user explicitly provided the flag.

Alternatively, we should just not have dev server CLI flag default values.

anshumanv: > I think I understand what you are saying. Do you mean the issue is that if a CLI flag has a default value, it will override devServer regardless of what the user has set for that value in devServer? Yes, this is an issue.

Precisely.

That makes sense to me, lets revisit this after we're on the new parser.

anshumanv: Can you find a case when we don't want the devServer config to override? I could think of `--hot` flag, when passed will appear in the CLI options object, and if `devServer` contains `hot: false` then the CLI option won't override this way, but nobody does will explicitly pass `hot: false` in their config right, since it's default?

anshumanv: /cc @Loonride how do we proceed here?

Loonride: I think the goal may be to just make dev server flags not set default values. If we do that, this shouldn't be an issue. But in that case it would still be nice for users to see what values are defaulted to later on when they do like `webpack-cli help serve`. Will think about this a bit and come back to it

anshumanv: Did you get any findings about this @Loonride ?

Loonride: Sorry for the delay, I can look at this more closely now that we are close to webpack-dev-server v4

anshumanv: Awesome, let me know your thoughts :)

packages/serve/src/startDevServer.ts
```diff
@@ -11,21 +11,25 @@ import Server from 'webpack-dev-server/lib/Server';
  */
 export default function startDevServer(compiler, options): void {
     const firstWpOpt = compiler.compilers ? compiler.compilers[0].options : compiler.options;
-    const devServerOptions = firstWpOpt.devServer || {};
+    let devServerOptions = firstWpOpt.devServer || {};
```
evilebottnawi, line 14: This is not entirely true. Second compiler also can contains `devServer`. To be honestly both can contain `devServer`, we need to think how to handle it, do we have `--name` option to choose compiler?

anshumanv: What would be the merge strategy?

rishabh3112: Assuming you are referring to merger of CLI options with config then I think you may use same strategy for each complier. Just CLI provided options would be same for each of them.

evilebottnawi: I think we should look at all compilers, if you have two `devServer` output `warning` about you need specify `name` of compiler, in future we will dev server as plugin and we don't need this hack

anshumanv:
Closing in favour of https://github.com/webpack/webpack-cli/pull/1649

PR #1347, opened 2020-03-18
[WIP] chore: migrate to commander

rishabh3112:
Refers #717 

**What kind of change does this PR introduce?**
Migration to new arguments parser

**Did you add tests for your changes?**
No, needs verification

**If relevant, did you update the documentation?**
No, ideally shouldn't be needed

**Summary**
Use commander as decided at https://github.com/webpack/webpack-cli/issues/717#issuecomment-600083131

**Does this PR introduce a breaking change?**
Don't mean to, attempting as seamless transition as possible

**Other information**
Tasks needed to be done.
- [x] ~Handle custom types~ https://github.com/webpack/webpack-cli/blob/6eecb26ae8e1b2e032c925c21b2cca5e7f37a510/packages/webpack-cli/lib/utils/cli-flags.js#L284 ( flag values should be sanitised where their functionality is implemented )
- [ ] Migrate serve package to use `commander.js`
- [x] Add support for `webpack serve help`
- [ ] Verify tests
- [ ] Remove `command-line-args` completely
- [x] ~Process global flag correctly~ ( cli-team decided to drop support for it.)

Issue #1240, opened 2020-02-19
Negated flags


ematipico:
**Is your feature request related to a problem? Please describe.**
The CLI should support negated flags.

**Describe the solution you'd like**
The CLI should support same negated flags
- `--no-hot`
- `--no-colors`
- `--no-mode`

snitin315:
@ematipico I would like to work on this issue. It would be great if you can point me to some reference or share any tip/prerequisite .

rishabh3112:
Thanks for trying to work on the issue.
You can look at how other flags are added in the CLI, that would be enough to get you working on the issue.
You can ask for more help here if you need in future 🙌

snitin315:
@rishabh3112 thanks , Looking into it.

snitin315:
@rishabh3112 correct me if I'm wrong the `--no-mode` flag should set `mode = "none"` .?

ematipico:
It depends on us. What do you think?

snitin315:
I think it should set `mode = "none"`

ematipico:
Sounds good :)

snitin315:
@ematipico @rishabh3112  , I am able to add `nomode` or `no_mode` but I can't pass `no-mode` in the code below in `zero-config.js` like this  -

```
function getEnvFromOptionsAndMode(mode, optionsObject) {
     const { dev, prod, **no-mode** } = optionsObject;

```
Whats the trick here to do so?

anshumanv:
`-` would be seen as an operator while destructuring since what you're destructuring is not a valid key, syntactically wrong.

Maybe add an alias variable for the flag like `--no-mode` stores a property is stored with key `no_mode`?

snitin315:
@anshumanv 
>INVALID_DEFINITIONS: Invalid option definition: an alias must be a single character

:-(

ematipico:
As the file says, that's a file we use for covering the case the user doesn't use any configuration.

The reading and parsing of the new flag will occur inside the group you decide to put.

PR #1481, opened 2020-04-18
migrate to commander

Loonride:

refactor

Yes, updating tests

No

**Summary**

Picking up from https://github.com/webpack/webpack-cli/pull/1347

Looking to get this done so we can move forward with `webpack-cli serve`.

Breaking changes to adapt to the norms of `commander`:
- An argument that expects a string will throw an error if one is not provided (e.g. `webpack-cli --output` will throw an error because `output` expects a string value following it)
- In the case of our `--mode` and `--no-mode` arguments, the `--no-mode` argument actually functions in commander to set the `--mode` argument to false. Therefore, we can check if `--no-mode` has been set by checking if `opts.mode === false`. The quirk here is that if both are used, the one that comes last takes precedence. So if a user does `--mode production --no-mode`, then mode will be false, but if the user does `--no-mode --mode production`, mode will be production.

**Other information**


evilebottnawi:
> So if a user does --mode production --no-mode, then mode will be false, but if the user does --no-mode --mode production, mode will be production.

Maybe we should throw an error? I think we should not support multiple same arguments for mode (and for some other arguments)

packages/webpack-cli/package.json
```diff
@@ -27,8 +27,8 @@
     "@webpack-cli/info": "^1.0.1-alpha.4",
     "ansi-escapes": "^4.3.1",
     "chalk": "^3.0.0",
-    "command-line-args": "^5.1.1",
     "command-line-usage": "^6.1.0",
```
evilebottnawi, line 30: I think we don't need that package too

packages/serve/src/index.ts
```diff
@@ -10,22 +10,33 @@ import argsToCamelCase from './args-to-camel-case';
...
-    // partial parsing usage: https://github.com/75lb/command-line-args/wiki/Partial-parsing
 
-    // since the webpack flags have the 'entry' option set as it's default option,
-    // we need to parse the dev server args first. Otherwise, the webpack parsing could snatch
-    // one of the dev server's options and set it to this 'entry' option.
-    // see: https://github.com/75lb/command-line-args/blob/master/doc/option-definition.md#optiondefaultoption--boolean
-    const devServerArgs = cli.commandLineArgs(devServer, { argv: args, partial: true });
-    const webpackArgs = cli.commandLineArgs(core, { argv: devServerArgs._unknown || [], stopAtFirstUnknown: false });
-    const finalArgs = argsToCamelCase(devServerArgs._all || {});
+    const filteredArgs = process.argv.filter((arg) => arg != 'serve');
```
evilebottnawi, line 17: Maybe we can implement API method on webpack-cli side, like getFlagsForCommand('serve') and it is remove unnecessary argv?

evilebottnawi: Will be great to avoid working with argv on command side and process on command side. What do you think?

rishabh3112:
> Maybe we should throw an error? I think we should not support multiple same arguments for mode (and for some other arguments)

Instead of error we should display warning. We should complete user's interaction where ever we can.

evilebottnawi:
@rishabh3112 hm, other CLI programs always throw an error on invalid arguments, but I am ok with warning

rishabh3112:
@evilebottnawi My point of view on this (sharing to validate my thinking as I might be wrong):
In this case we don't have invalid arguments, instead we we have dilemma which of the two same flags takes precedence. This can be resolved with a logical decision.
But yes, Many CLIs (including legacy ones) throw error in this case. But I think throwing error in this case is bad for DX, when we can logically imply what user wants to do and warn user with decision we took and what assumptions we took to make this decision.

Loonride:
My personal opinion (feel free to disagree) is that we shouldn't have the `--no-mode` argument. It was introduced recently as described here: https://github.com/webpack/webpack-cli/issues/1240

- The same thing can already be done with `--mode none`
- `--no-argument` generally implies to me that the argument is a boolean and you are switching it off
- It adds ambiguities like if the user is doing `--mode production --no-mode` which need to be dealt with
- `commander` doesn't support parse by telling us that `--no-mode` was provided so we would have to do our own parsing to see if the user has provided `--mode` and `--no-mode` (it simply sets `--mode` to false). Search for `--no-cheese` in here to see the behavior: https://github.com/tj/commander.js/

packages/webpack-cli/package.json
Loonride: @evilebottnawi Good you pointed that out, we actually still rely on it for `webpack-cli help`. It still functions correctly, since we didn't change the `cli-flags` object format at all. I think it is fine actually to continue using it as it formats nicely and we can keep our current CLI flags object format is it is so that it continues to work with `command-line-usage`.

The alternative is to start using the built in commander help, we would just need to do some work making the output look nicer. Here is the difference:

Current webpack-cli help output using `command-line-usage`:
webpack-cli-usage-help

`commander` help:
webpack-cli-commander-help

Loonride:
Notes about commander:
- it seems if you want to negate a boolean flag with `--no-` it has to actually be added as an option along with the original boolean flag
- there is no negating of a boolean via `--flag=false`, though you can set string flags with `--flag=value`.

TODO:
- figure out how to handle negated flags for booleans
- fix `resolveNegatedArgs` in `bootstrap`
- add warning if user includes both `--mode` and `--no-mode`
- add handling for list arguments such as `node-args` and `allowed-hosts` for serve (https://github.com/webpack/webpack-dev-server/blob/master/bin/cli-flags.js) using commander custom option parsing
- decide whether to keep using `command-line-usage` or use the built in command help of `commander`
- add `getFlagsForCommand('serve')`

evilebottnawi:
@Loonride 

> https://github.com/webpack/webpack-cli/pull/1481#issuecomment-616664835

Agree, let's remove `--no-mode` support :+1: We can do it after migration.

> https://github.com/webpack/webpack-cli/pull/1481#issuecomment-616939342

Good job!

packages/webpack-cli/package.json
evilebottnawi: @Loonride Do we have API from commander for customize help output?

jamesgeorge007: > @Loonride Do we have API from commander for customize help output?

Yes indeed, https://github.com/tj/commander.js#custom-help

packages/serve/src/index.ts
Loonride: actually this is not needed, it seems the filtered args are already passed into the serve command and I just wasn't aware of it

Loonride:
@webpack/cli-team @evilebottnawi I would say this is ready to merge, so please review

We will do these things in later PRs:
- enabling `--boolean-flag=false`
- potentially removing `command-line-usage` in favor of built-in `commander` help
- remove `--no-mode`

evilebottnawi:
@Loonride Great job, thanks!

packages/webpack-cli/lib/groups/OutputGroup.js
```diff
@@ -27,7 +27,10 @@ class OutputGroup extends GroupHelper {
         const { args } = this;
         if (args) {
             const { output } = args;
-            if (!output) {
+            // TODO: Remove comment before merge
+            // We need to show warning when empty output flag is supplied
+            // which is set to boolean true by commander
```
evilebottnawi, line 32: /cc @Loonride should we remove it before merge?

packages/webpack-cli/lib/groups/ZeroConfigGroup.js
```diff
@@ -32,17 +34,19 @@ class ZeroConfigGroup extends GroupHelper {
                     return NONE;
                 }
             }
-            if (this.args.noMode && this.args.mode) {
-                logger.warn('You Provided both mode and no-mode arguments. You Should Provide just one. "mode" will be used.');
-            }
+
             if (this.args.mode) {
+                if (this.args.mode !== PRODUCTION && this.args.mode !== DEVELOPMENT && this.args.mode !== NONE) {
+                    logger.warn('You provided an invalid value for "mode" option.');
```
evilebottnawi, line 40: `You provided an invalid value for "mode" option. Using "production" by default` (maybe better)

packages/webpack-cli/lib/bootstrap.js
```diff
@@ -20,39 +20,21 @@ const isCommandUsed = (commands) =>
...
-        return;
-    } else if (versionFlagExists) {
-        cli.runVersion(commandIsUsed);
-        return;
+        process.exit(0);
+    }
+
+    if (parsedArgs.unknownArgs.includes('version')) {
+        runVersion();
+        process.exit(0);
```
evilebottnawi, line 37: Maybe we can using `help` (https://github.com/tj/commander.js/#helpcb and https://github.com/tj/commander.js/#addhelpcommand) for help, same for version? Yes we can do it in future

packages/webpack-cli/lib/groups/OutputGroup.js
Loonride: Yes, removed. I couldn't seem to get it to be set to true as @rishabh3112 wrote. Here is the test I added:

https://github.com/webpack/webpack-cli/pull/1481/files#diff-2cb16055e37ba4584d42ff3630696acdR78-R82

packages/webpack-cli/lib/groups/ZeroConfigGroup.js
Loonride: done :+1:

packages/webpack-cli/lib/bootstrap.js
Loonride: Yes, just don't want to do it here because it will probably involve some significant rewriting of https://github.com/webpack/webpack-cli/blob/next/packages/webpack-cli/lib/groups/HelpGroup.js

Loonride:
> /cc @Loonride Couple questions

@evilebottnawi all done

packages/webpack-cli/lib/groups/OutputGroup.js
rishabh3112: @Loonride The check can be removed now.

P.S.
This was because I was kept value for flag optional when I implemented `arg-parser` initially.
It was this (in `arg-parser.js` where we are registering options),
```
 const flagsWithType = option.type !== Boolean ? flags + ' [type]' : flags;
```
Now it is using required argument `<value>`.

packages/webpack-cli/lib/utils/arg-parser.js
```diff
@@ -0,0 +1,103 @@
...
+
+    // Use customized version output if available
+    if (versionFunction) {
+        parser.on('option:version', () => {
+            versionFunction();
+            process.exit(0);
+        });
+    }
+
+    // Use customised help output is avaliable
```
rishabh3112, line 26:
```suggestion
    // Use customised help output if avaliable
```

snitin315, line 26:
```suggestion
    // Use customised help output is available
```

test/stats/stats.test.js
```diff
@@ -48,7 +48,8 @@ describe('stats flag', () => {
     it('should warn when an unknown flag stats value is passed', () => {
         const { stderr, stdout } = run(__dirname, ['--stats', 'foo']);
         expect(stderr).toBeTruthy();
-        expect(stderr).toContain('No value recognised for "stats" option');
+        console.log(stderr);
```
jamesgeorge007, line 51: Unnecessary log.

PR #1495, opened 2020-04-23
Fix migrate command

Loonride:

fix/tests

Yes (still need to add tests for runPrettier)

I'm not exactly sure how the migrate command is supposed to work. Could someone give an example of a webpack config that it should make modifications to?

No

**Summary**

@jamesgeorge007 Sorry for taking over, just couldn't debug `runPromptWithAnswers` without using your changes. It seems that execa does not like an empty `input` string option.

- Fixes `runPromptWithAnswers`
- Fixes migrate
- Fixes issues with `runPrettier`

Slightly with how `runPrettier` works

**Other information**

### DIFF
--- a/test/utils/test-utils.js
+++ b/test/utils/test-utils.js
@@ -91,7 +91,9 @@ function runAndGetWatchProc(testCase, args = [], setOutput = true, input = '', f
 /**
  * runInitWithAnswers
  * @param {string} location location of current working directory
+ * @param {string[]} args CLI args to pass in
  * @param {string[]} answers answers to be passed to stdout for inquirer question
+ * @param {boolean} waitForOutput whether to wait for stdout before writing the next answer
  */
 const runPromptWithAnswers = (location, args, answers, waitForOutput = false) => {
     const runner = runAndGetWatchProc(location, args, false, '', true);
--- a/test/utils/test-utils.js
+++ b/test/utils/test-utils.js
@@ -93,38 +93,69 @@ function runAndGetWatchProc(testCase, args = [], setOutput = true, input = '', f
  * @param {string} location location of current working directory
  * @param {string[]} answers answers to be passed to stdout for inquirer question
  */
-const runPromptWithAnswers = (location, args, answers) => {
+const runPromptWithAnswers = (location, args, answers, waitForOutput = false) => {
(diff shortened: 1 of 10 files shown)


anshumanv:
@Loonride one test unusually still fails, doesn't show any log, can you debug?

jamesgeorge007:
> I'm not exactly sure how the migrate command is supposed to work. Could someone give an example of a webpack config that it should make modifications to?

You can find everything related to `migrate` command [here](https://github.com/webpack/webpack-cli/blob/next/MIGRATE.md)

test/utils/test-utils.js
```diff
@@ -67,17 +67,24 @@ function runWatch({ testCase, args = [], setOutput = true, outputKillStr = 'Time
...
 
-    const webpackProc = execa(WEBPACK_PATH, argsWithOutput, {
+    const options = {
         cwd,
         reject: false,
         stdio: ENABLE_LOG_COMPILATION ? 'inherit' : 'pipe',
-    });
+    };
+
+    // some tests don't work if the input option is an empty string
```
jamesgeorge007, line 82: Good catch 👏

Loonride:
It seems sending a SIGKILL to the process at the end of `runPromptWithAnswers` made it more stable :woman_shrugging:

evilebottnawi:
hm, ci is not stable in some cases, it is strange

Loonride:
> hm, ci is not stable in some cases, it is strange

whats confusing is it always either seems like a large group of them fail, or none of them fail :thinking:

rishabh3112:
> whats confusing is it always either seems like a large group of them fail, or none of them fail 🤔
Others are cancelled when one of them fails. the macOS Node v13 container is the failing one everywhere :(

Loonride:
> Others are cancelled when one of them fails. the macOS Node v13 container is the failing one everywhere :(

Oh that makes more sense thanks

evilebottnawi:
I think we have racing

Loonride:
Let me try a few things with how `runPromptWithAnswers` works.

Loonride:
@evilebottnawi 4 test runs good in a row, I think stabilization worked!

What I did was forced `runPromptWithAnswers to wait for new stdout before writing the next answer to stdin

evilebottnawi:
Good job!

PR #1502, opened 2020-04-27
tests: add longer jest timeout to migrate test, set other timeouts to 20000

Loonride:
tests

Yes

No

**Summary**

The latest run of the migrate tests seems to have timed out: https://github.com/webpack/webpack-cli/runs/622211039

We should have a longer timeout

No

**Other information**

### DIFF
--- a/test/migrate/config/migrate-config.test.js
+++ b/test/migrate/config/migrate-config.test.js
@@ -11,7 +11,7 @@ const outputPath = path.join(__dirname, outputDir);
 const outputFile = `${outputDir}/updated-webpack.config.js`;
 const outputFilePath = path.join(__dirname, outputFile);
 
-jest.setTimeout(200000);
+jest.setTimeout(20000);
 
 describe('migrate command', () => {
     beforeEach(() => {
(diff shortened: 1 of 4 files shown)


test/migrate/config/migrate-config.test.js
```diff
@@ -11,6 +11,8 @@ const outputPath = path.join(__dirname, outputDir);
 const outputFile = `${outputDir}/updated-webpack.config.js`;
 const outputFilePath = path.join(__dirname, outputFile);
 
+jest.setTimeout(200000);
```
evilebottnawi, line 14: Let's use `20000` otherwise CI really can freeze and we should wait a lot of time

Loonride: I copied and pasted from another test not realizing that they are all 200,000 by mistake.

Loonride:
I think `init-inquirer.test.js` needs long timeout since it installs dependencies

evilebottnawi:
hm, maybe you are right, try to setup it on `1` minute

rishabh3112:
Feel free to open if you feel there is some usecase for it now :)

PR #1523, opened 2020-05-05
fix: ignore type inconsistency for now, fix build

anshumanv:

Fix build, there is type inconsistency with webpack@next types, this should ignore that error for now, we need to enable this once it's fixed.

Yes

NA

**Summary**
- Fix build

No

**Other information**
NA

### DIFF
--- a/packages/migrate/package.json
+++ b/packages/migrate/package.json
@@ -20,7 +20,7 @@
     "p-lazy": "3.0.0"
   },
   "peerDependencies": {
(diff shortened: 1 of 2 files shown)


anshumanv:
Tests seem to pass now, I think we can let them stay

snitin315:
I think its good to merge. 
/cc @anshumanv

evilebottnawi:
Done :+1

PR #1544, opened 2020-05-14
chore(deps): update yargs from 13.2.4 to 13.3.2 to patch vulnerability

patomation:
Updates yargs from 13.2.4 to 13.3.2 to patch prototype pollution vulnerability.

No.

Not necessary.
 
**Summary**
I would like to continue using webpack 3 until 4 is out of beta. It would be nice to fix the yargs audit warning in the mean time. 

Since this is a dependency patch at the minor level there shouldn't be any breaking changes.

**Other information**

### DIFF
--- a/package.json
+++ b/package.json
@@ -123,7 +123,7 @@
     "loader-utils": "1.2.3",
     "supports-color": "6.1.0",
     "v8-compile-cache": "2.0.3",
-    "yargs": "13.2.4"
+    "yargs": "13.3.2"
   },
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


patomation:
Maybe someone can illuminate me about why the azure test are failing. The only error code that I see is exited with a code of 1. Could it be that that version of yargs is not compatible with node 8 10 and 12?

evilebottnawi:
Fixed in the master https://github.com/webpack/webpack-cli/blob/master/package.json#L126

PR #1508, opened 2020-04-29
tests: add more tests for init

rishabh3112:
**What kind of change does this PR introduce?**
tests

**Did you add tests for your changes?**
N/A

**If relevant, did you update the documentation?**
N/A

**Summary**
As per suggestions by evilebottnawi,
we are adding exhaustive tests for `init` command, then refactor it if needed and then integrate it with CLI.

**Does this PR introduce a breaking change?**
No

**Other information**
Part of integrating init with CLI by default (as per roadmap)
More info #1509

PR #1545, opened 2020-05-15
chore: 🤖 remove stale tests for standard flag and instances

anshumanv:
Cleanup instances of standard flag, not supported anymore

Yes

Yes

**Summary**
- standard flag is no longer supported, remove it's instances.

No

**Other information**


test/standard/standard.test.js
```diff
@@ -1,13 +0,0 @@
-const { run } = require('../utils/test-utils');
-
-describe('standard output', () => {
-    it('should print standard output', () => {
-        const { stdout, stderr } = run(__dirname);
```
snitin315, line 5: I think this test is not for `--standard` flag but the default webpack output (referred to as standard). Please correct me if I'm wrong.

anshumanv: No, this was used for --standard, but later --standard was dropped from the test when flag was removed.

snitin315: So, do we have another test for default output or we don't need one?

anshumanv: zero-config

anshumanv: I've added the things missing there, from this test we removed.

rishabh3112:
Let's wait for #1508 to be merged as that PR also has changes for plugin tests and these changes could be conflicting.

PR #1546, opened 2020-05-15
tests: improve --node-args test cases

snitin315:
tests

Yes
NA
**Summary**
Improved test cases to check for the specific error messages.

No

**Other information**


snitin315:
Actually it's hardcoded for --unknown flag case as well, 
https://github.com/webpack/webpack-cli/blob/6a8dd32a64d6428ba6b6aa367e6a1c477b38857b/test/node/node.test.js#L65

So it should changed as well. Let's wait for other opinions as well.

snitin315:
Good to merge?

PR #1547, opened 2020-05-15
chore: 🤖 improve cli-flags usage

snitin315:
Chore

NA
NA
**Summary**
improved flags example usage for some String type flags.

No

**Other information**


packages/webpack-cli/lib/utils/cli-flags.js
```diff
@@ -121,28 +121,28 @@ module.exports = {
...
             link: 'https://webpack.js.org/concepts/#output',
         },
         {
             name: 'plugin',
-            usage: '--plugin',
+            usage: '--plugin <pluginName>',
             group: ADVANCED_GROUP,
             type: String,
-            description: 'Load a given plugin',
+            description: 'Load a given plugin e.g. HtmlWebpackPlugin',
```
anshumanv, line 136: This is only to generate plugin iirc?

jamesgeorge007: I assume you're speaking about the `plugin` command which is used to scaffold a plugin.
https://github.com/webpack/webpack-cli/blob/49620e0070a1be9287b493fe22bfcfab650341f3/packages/webpack-cli/lib/utils/cli-flags.js#L42

jamesgeorge007: It isn't the case here.

anshumanv: Ah okay yes

PR #1552, opened 2020-05-17
Update lerna.json

kaiwalyakoparkar:
Beautification of the program is improved

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -4,11 +4,15 @@
   "version": "independent",
   "npmClient": "yarn",
   "useWorkspaces": true,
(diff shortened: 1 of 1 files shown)


snitin315:
Thanks for the PR, but we are already using  `ESlint` & `Prettier`, and your change is not according to our code style, that's why Lint check is failing.

kaiwalyakoparkar:
> Thanks for the PR, but we are already using `ESlint` & `Prettier`, and your change is not according to our code style, that's why Lint check is failing.

Thankyou for the information

PR #1558, opened 2020-05-19
fix(webpack-cli): --progress flag output

### LINKED ISSUE #1551: bug: --progress doesn't ouput everything
**Describe the bug**

`--progress` flag doesn't output everything.

`--progress` only prints the 100% compilation stage.

**To Reproduce**
`webpack --progress`.

**Expected behavior**

should print out every detail from 0 to 100% of compilation.

**Screenshots**
Screenshot at 2020-05-16 19-17-03

**Additional context**
Looking into it.

snitin315:
FIX

No
No
**Summary**
Fixes #1551 
fixed progress flag output.

## Before 
Screenshot at 2020-05-19 11-38-26

## After
Screenshot at 2020-05-19 11-34-09
Screenshot at 2020-05-19 11-33-47

NO

**Other information**

### DIFF
--- a/packages/webpack-cli/lib/utils/Compiler.js
+++ b/packages/webpack-cli/lib/utils/Compiler.js
@@ -32,9 +32,10 @@ class Compiler {
                             }
                         }
                         process.stdout.write(`) ${percent}% : `);
-                        process.stdout.write(`${cyanBright(msg)}`);
-                        if (percent === 100) {
-                            process.stdout.write(`${cyanBright('Complilation completed\n')}`);
+                        if (percent !== 100) {
+                            process.stdout.write(`${cyanBright(msg)}\n`);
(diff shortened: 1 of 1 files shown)


rishabh3112:
All steps are visible "Before" as well.
The previous line is removed when percentage increases and new line is added to make progress bar look like it is "progressing". In a bigger project (where using `--progress` makes sense) this becomes more evident. So my suggestion would be to keep this particular as it is (or add a new option if it is actually needed for most of the people's use case).

snitin315:
something like `webpack --progress detailed` ?

rishabh3112:
I am -1 on this change as of now. We can do this when we get feature request for this from users.

snitin315:
I agree

PR #1565, opened 2020-05-20
fix: 🐛 do not apply own defaults while setting mode

anshumanv:

bugfix

Yes

NA

**Summary**
- Do not apply own defaults when setting mode, defaults are already in core.
- remove unused configs

No

**Other information**


anshumanv:
Adding in a while

anshumanv:
@evilebottnawi added 👍

packages/webpack-cli/__tests__/ZeroConfigGroup.test.js
```diff
@@ -40,6 +41,8 @@ describe('GroupHelper', function () {
         ]);
 
         const result = group.run();
+        // ensure no other properties are added
+        expect(result.options).toMatchObject({ mode: 'development' });
```
evilebottnawi, line 45: Can we look on plugins here?

anshumanv: Old method is only adding config options and no plugins

evilebottnawi: We should look what we don't have two terser plugins here

anshumanv: Can you suggest ways to check it here? We're just passing mode to the compiler, so since core doesn't have two, we won't have two either.

evilebottnawi: Hm, maybe we can add a test there manually added `Terser` to minimizer with custom options, for extractFilename, not the best solution, but will be work

anshumanv: Added, something wrong with CI, will take a look tomm

anshumanv: @evilebottnawi seems good now, PTAL

test/mode/mode-with-config/mode-with-config.test.js
```diff
@@ -0,0 +1,95 @@
+'use strict';
+const { stat, readFile } = require('fs');
+const { resolve } = require('path');
+// eslint-disable-next-line node/no-unpublished-require
+const { run } = require('../../utils/test-utils');
+
+describe.skip('mode flags with config', () => {
+    it('should run in production mode when --mode=production is passed', (done) => {
```
snitin315, line 8: why `.skip`?

anshumanv: My bad, debug

anshumanv:
Need look at CI, idk what's failing

anshumanv:
@evilebottnawi fixed now, had to do some changes to the install script since it was running into race condition. Looks all good now.

PR #1598, opened 2020-05-30
feat: 🎸 add support for env flag

anshumanv:

feat

Yes

Yet to

**Summary**

- Add support for --env flag back.

No

**Other information**


packages/webpack-cli/lib/utils/cli-flags.js
```diff
@@ -240,6 +240,13 @@ module.exports = {
             group: DISPLAY_GROUP,
             description: 'It tells webpack to output all the information',
         },
+        {
+            name: 'env',
+            usage: '--env',
+            type: String,
+            group: CONFIG_GROUP,
+            description: 'It tells webpack to output all the information',
```
snitin315, line 248:
```suggestion
            group: CONFIG_GROUP,
            description: 'Environment passed to the configuration when it is a function',
```

anshumanv: Lol yes

anshumanv:
@evilebottnawi yes - https://webpack.js.org/configuration/configuration-types/#exporting-a-function

evilebottnawi:
I was wrong, give me time to think

evilebottnawi:
I would like to avoid this because this is very misleading, Maybe we should just pass `argv`, without `env` flag (normalize them before)

evilebottnawi:
What is really diffidence between `env` and `argv`? I see no difference, just pass flags in difference ways

anshumanv:
argv has other things too, env is only what passed to the configuration function.

evilebottnawi:
@anshumanv Alright, let's get this back for compatibility, but we will need to throw depications in the future and pass only `args` + `normalized args`

PR #1604, opened 2020-06-01
tests: add test for progress flag

snitin315:
TEST

yes
No
**Summary**
added test case for `--progress` flag.

no

**Other information**
no

### DIFF
--- a/packages/webpack-cli/lib/utils/Compiler.js
+++ b/packages/webpack-cli/lib/utils/Compiler.js
@@ -2,6 +2,7 @@ const webpack = require('webpack');
 const logger = require('./logger');
 const { cyanBright, greenBright } = require('chalk');
 const { CompilerOutput } = require('./CompilerOutput');
+const readline = require('readline');
 
 class Compiler {
     constructor() {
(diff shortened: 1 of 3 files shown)


packages/webpack-cli/lib/utils/Compiler.js
```diff
@@ -20,8 +21,8 @@ class Compiler {
                 process.stdout.write('\n');
                 const defaultProgressPluginHandler = (percent, msg) => {
                     percent = Math.floor(percent * 100);
-                    process.stdout.clearLine();
-                    process.stdout.cursorTo(0);
+                    readline.clearLine(process.stdout, 0);
+                    readline.cursorTo(process.stdout, 0, null);
```
evilebottnawi, line 25: Why?

snitin315: Because when running with jest test was failing with error `TypeError: process.stdout.clearLine is not a function`. 

Maybe because `clearLine` and `cursorTo` are undefined when `process.stdout` is not a tty

More info => https://stackoverflow.com/questions/34570452/node-js-stdout-clearline-and-cursorto-functions

PR #1454, opened 2020-04-13
feat(webpack-cli): allow multiple entry files

snitin315:
**What kind of change does this PR introduce?**
Feature

**Did you add tests for your changes?**

Yes.

**If relevant, did you update the documentation?**

Yes

**Summary**

Now multiple entry files are possible. `--entry ./src/a.js ./src/b.js` or `--entry ./src/a.js --entry ./src/b.js`

entry2

**Does this PR introduce a breaking change?**

Maybe

**Other information**

Refers #717

PR #1619, opened 2020-06-05
feat(webpack-cli): allow multiple entry files

snitin315:
Feature

YES
YES
**Summary**
Now we can provide multiple entry files via CLI using  -
- `webpack --entry=a.js --entry=b.js` 
- `webpack ./a.js ./b.js`

Refers #717
Fixes #1454 

**Other information**

### DIFF
--- a/packages/webpack-cli/lib/bootstrap.js
+++ b/packages/webpack-cli/lib/bootstrap.js
@@ -49,14 +49,21 @@ async function runCLI(cli, commandIsUsed) {
             // if the unknown arg starts with a '-', it will be considered
             // an unknown flag rather than an entry
             let entry;
-            if (parsedArgs.unknownArgs.length === 1 && !parsedArgs.unknownArgs[0].startsWith('-')) {
-                entry = parsedArgs.unknownArgs[0];
-            } else if (parsedArgs.unknownArgs.length > 0) {
-                parsedArgs.unknownArgs
-                    .filter((e) => e)
-                    .forEach((unknown) => {
-                        logger.warn('Unknown argument:', unknown);
+            if (parsedArgs.unknownArgs.length > 0 && !parsedArgs.unknownArgs[0].startsWith('-')) {
+                if (parsedArgs.unknownArgs.length === 1) {
+                    entry = parsedArgs.unknownArgs[0];
+                } else {
+                    entry = [];
+                    parsedArgs.unknownArgs.forEach((unknown) => {
+                        if (!unknown.startsWith('-')) {
+                            entry.push(unknown);
+                        }
                     });
+                }
+            } else if (parsedArgs.unknownArgs.length > 0) {
+                parsedArgs.unknownArgs.forEach((unknown) => {
+                    logger.warn('Unknown argument:', unknown);
+                });
                 cliExecuter();
                 return;
             }
--- a/packages/webpack-cli/README.md
+++ b/packages/webpack-cli/README.md
@@ -35,7 +35,7 @@ Available Commands
 
 Options
 
-  --entry string           The entry point of your application.
+  --entry string[]         The entry point(s) of your application.
   -c, --config string      Provide path to a webpack configuration file
   -m, --merge string       Merge a configuration file using webpack-merge
   --progress               Print compilation progress during build
--- a/packages/webpack-cli/__tests__/arg-parser.test.js
+++ b/packages/webpack-cli/__tests__/arg-parser.test.js
@@ -51,6 +51,13 @@ const basicOptions = [
         },
         description: 'custom type flag',
     },
+    {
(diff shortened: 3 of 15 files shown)


snitin315:
> webpack src/one.js src/two.js src/three.js

@evilebottnawi we can do this for entry but what about other flags that can take multiple values?

evilebottnawi:
@snitin315 Let's start with entry, it is most popular use case

packages/webpack-cli/lib/bootstrap.js
```diff
@@ -49,14 +49,21 @@ async function runCLI(cli, commandIsUsed) {
...
-                parsedArgs.unknownArgs
-                    .filter((e) => e)
-                    .forEach((unknown) => {
-                        logger.warn('Unknown argument:', unknown);
+            if (parsedArgs.unknownArgs.length > 0 && !parsedArgs.unknownArgs[0].startsWith('-')) {
+                if (parsedArgs.unknownArgs.length === 1) {
+                    entry = parsedArgs.unknownArgs[0];
+                } else {
+                    entry = [];
+                    parsedArgs.unknownArgs.forEach((unknown) => {
```
rishabh3112, line 57: `webpack --entry a.js b.js c.js`  is what we should be supporting imo.

snitin315: It will be a bit tricky to parse `--entry` like  that.

snitin315: We need to discuss which to support or support both.
/cc @webpack/cli-team

packages/webpack-cli/lib/bootstrap.js
```diff
@@ -49,14 +49,21 @@ async function runCLI(cli, commandIsUsed) {
...
             // an unknown flag rather than an entry
             let entry;
-            if (parsedArgs.unknownArgs.length === 1 && !parsedArgs.unknownArgs[0].startsWith('-')) {
-                entry = parsedArgs.unknownArgs[0];
-            } else if (parsedArgs.unknownArgs.length > 0) {
-                parsedArgs.unknownArgs
-                    .filter((e) => e)
-                    .forEach((unknown) => {
-                        logger.warn('Unknown argument:', unknown);
+            if (parsedArgs.unknownArgs.length > 0 && !parsedArgs.unknownArgs[0].startsWith('-')) {
```
evenstensberg, line 52: This is not a way to go, set multipleValues in cliFlags instead

snitin315: That way we will be able to support  `webpack --entry a.js --entry b.js`

The idea is to support multiple entries by default without passing `--entry` i.e. `webpack a.js b.js c.js`

snitin315: Or we can support , need discussion for same.

packages/webpack-cli/lib/bootstrap.js
evilebottnawi: @rishabh3112 I think we support two cases here for better 0CJS:
```
webpack --entry ./src/index.js --entry ./src/other.js
webpack ./src/index.js ./src/other.js
```

Second just sugar.

snitin315: I agree with @evilebottnawi we should support both.

rishabh3112: Lets do both then. But this would be a breaking change so should be mentioned as such during release.

evilebottnawi: @rishabh3112 Why? It is breaking change?

rishabh3112: Yeah it's not, my bad. I wrongly thought that we were supporting configurations to be passed without flags instead on entry in CLI v3.

evilebottnawi: Configuration should loaded by default (0CJS) or using `--config` flag

snitin315:
Ready for review :raised_hands:

octaharon:
Hey lads. What about named entries? Webpack 4 allowed passing `<name>=request>` as entry, as of now it doesn't seem to work, and "--entry" argument is treated like a filepath regardless of format. There seems to be no deprecation warning about it, will this functionality be restored?

petemill:
> Hey lads. What about named entries? Webpack 4 allowed passing <name>=<request> as entry, as of now it doesn't seem to work, and "--entry" argument is treated like a filepath regardless of format. There seems to be no deprecation warning about it, will this functionality be restored?

Confirming this is broken now in webpack-cli. You cannot define an entry's name in the cli, preventing different compiled file names than entry file names, where this is possible in webpack config.

PR #1627, opened 2020-06-10
feat(webpack-cli): allow multiple types for --stats

snitin315:
Feature.

WIP
WIP
**Summary**
Refers #717 

Allow multiple types for stats argument, i.e.
`webpack --stats (means stats: true)`,
`webpack --stats=verbose (means: stats: verbose)`

Screenshot at 2020-06-10 09-18-27

Screenshot at 2020-06-10 09-18-52

No

**Other information**

### DIFF
--- a/packages/webpack-cli/lib/utils/cli-flags.js
+++ b/packages/webpack-cli/lib/utils/cli-flags.js
@@ -240,7 +240,7 @@ module.exports = {
         {
             name: 'stats',
             usage: '--stats <value>',
-            type: String,
+            type: [String, Boolean],
             group: DISPLAY_GROUP,
             description: 'It instructs webpack on how to treat the stats e.g. verbose',
             link: 'https://webpack.js.org/configuration/stats/#stats',
--- a/packages/webpack-cli/__tests__/arg-parser.test.js
+++ b/packages/webpack-cli/__tests__/arg-parser.test.js
@@ -44,6 +44,12 @@ const basicOptions = [
         defaultValue: 'default-value',
     },
     {
+        name: 'multi-type',
+        usage: '--multi-type | --multi-type <value>',
+        type: [String, Boolean],
+        description: 'flag with multiple types',
+    },
+    {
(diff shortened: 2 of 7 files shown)


packages/webpack-cli/lib/utils/cli-flags.js
```diff
@@ -237,6 +237,7 @@ module.exports = {
             name: 'stats',
             usage: '--stats <value>',
             type: String,
```
evilebottnawi, line 239: Will be great to allow setup it like `type: [String, Boolean]`, what do you think, it is possible do on args-parser?

snitin315: Yes we can do this.

evilebottnawi: Will be great :+1:

snitin315:
@evilebottnawi there is one issue, `webpack@4` and `webpack@5` parses `stats` option differently.

|  version  | webpack config  |  compiler.options  | 
|--------------|-----------------------|-------------------------| 
|  v4 |  `stats: true` |  `stats: true` |
|  v4 |  `stats: normal` |  `stats: normal` |
|  v5 |  `stats: true` |  `stats: { }` |
|  v5 |  `stats: normal` |  `stats: {preset: 'normal'}` |
 
but I see both versions allow the `type: boolean` 

`v4`
https://github.com/webpack/webpack/blob/dfd20210c9441c309859d29d3b256474ae5101e5/schemas/WebpackOptions.json#L2189

`v5` 
https://github.com/webpack/webpack/blob/6f5a7ea9e158d83dbbf03a5ceb1538da436935ee/schemas/WebpackOptions.json#L2951

Is it a bug in v5? Should it be `stats: { preset: true }` with `stats: true` in the configuration as webpack 5 doesn't throw invalid configuration error for this case?

evilebottnawi:
Good job, it is good features for webpack core flags

PR #1629, opened 2020-06-14
feat: import flags from webpack core [WIP]

anshumanv:
**What kind of change does this PR introduce?**
feature

**Did you add tests for your changes?**
Not yet

**If relevant, did you update the documentation?**
Not yet

**Summary**
- Import flags from webpack core

**Does this PR introduce a breaking change?**
Shouldn't

**Other information**

PR #1391, opened 2020-03-31
Coverage

evilebottnawi:
**What kind of change does this PR introduce?**

Enable coverage

**Did you add tests for your changes?**

No

**If relevant, did you update the documentation?**

No need

**Summary**

Enable coverage

**Does this PR introduce a breaking change?**

No

**Other information**

We need nyc because https://github.com/facebook/jest/issues/3190

PR #1630, opened 2020-06-14
feat(webpack-cli): import flags from webpack core

### LINKED ISSUE #1241: Import flags from webpack core
At the moment we defined the supported flags inside the CLI (`cli-flags.js`). This could create discrepancy between webpack core supported options and webpack CLI supported arguments.

The solution would be to import this flags straight from webpack core (missing at the moment). @evilebottnawi will help us here, he volunteered to integrate them inside webpack core. Later, we will import them here and decide which arguments we want to support. 

N/A

**Additional context**
N/A

snitin315:
Feature.

Yes
No
**Summary**

Follow up #1629
Refers #717 
Closes #1241 

**Other information**

## TODO

- [x] Write tests
- [x] remove duplicate flags
- [x] register `--no-name` negated flags.
- [x] fix help output ( currently showing type `string` for every flag )
- [x] allow parsing type: 'number'


snitin315:
Right now works well -

Screenshot at 2020-06-15 09-42-48
Screenshot at 2020-06-15 09-43-09

packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -213,6 +215,30 @@ class WebpackCLI extends GroupHelper {
...
+     */
+    _handleCoreFlags() {
+        if (this.groupMap.has('core')) {
+            const coreFlags = this.groupMap.get('core');
+
+            // convert all the flags from map to single object
+            const coreConfig = coreFlags.reduce((allFlag, curFlag) => (allFlag = { ...allFlag, ...curFlag }), {});
+
+            // for some reason these flags are always appended now, remove this later
+            delete coreConfig['module-no-parse'];
```
anshumanv, line 231: Can you also investigate why these properties are added when importing core flags?

snitin315: looking into it.

packages/webpack-cli/lib/utils/cli-flags.js
```diff
@@ -6,6 +8,16 @@ const ADVANCED_GROUP = 'advanced';
...
 
+const flagsFromCore = version.startsWith(5)
+    ? Object.entries(require('webpack').cli.getArguments()).map(([flag, meta]) => {
+          return {
+              ...meta,
+              name: flag,
+              group: 'core',
+          };
+      })
+    : [];
```
evilebottnawi, line 19: Can you provide list of all options here?

snitin315: 100+ options are exported from the core as listed here - 

https://github.com/webpack/webpack/blob/master/test/__snapshots__/Cli.test.js.snap

anshumanv: Yep not possible to list options, some 230 flags are present in core, link is better

evilebottnawi: Ideally we should test all of them, so I prefer to write one test like:
```
const optionsAndValues = [...];


for (const case of (optionsAndValues)) {
  it(`${case}`, () => {
   // Base test with `WebpackCLITestPlugin` and check we set options correctly
  })
}

```

Maybe we should put it in `integration` tests

evilebottnawi: After this, we will catch all broken options on our and webpack side, maybe some of them should be excluded, because webpack in `beate`

snitin315: We can write one test with 3 test cases one each for type `string` `boolean` and `number`

evilebottnawi: Yep :+1:

snitin315:
Draft list of broken flags (updating the list as testing) - 
- `amd`

packages/webpack-cli/lib/utils/cli-flags.js
snitin315: @evilebottnawi there are a total of 290 flags and it is very difficult to write a base test for all of them because we have to write different tests assertions for most flags as some flags depends on other flags, for some property names are not same as the flag name, for example, if we want to write a base test for all boolean flags -

- for `--bail` option name is `bail` as well and it should set `bail: true`.
- for `--dependencies-reset` option name is `dependencies` and it should set `dependencies: []`  but for `output-enabled-library-types-reset` property name is `enabledLibraryTypes`
- for `--node-global` option is global => `global: true`.

Hence instead we can group flags on the basis of primary config option, for example, there are 56 flags for `resolve` related config we can tests these together. similarly, around 50 flags related to `output`, 46 flags related to `optimization`.

snitin315:
Draft list of flags we can drop ( need discussion ) - 
- all flags related to `module.rules` i.e, `--module-rules-use` `module-rules-use-options` etc. (total 18 flags) because it ts no feasible to configure such options via CLI https://webpack.js.org/configuration/module/#ruleuse

anshumanv:
I'm not sure of the feasibility of removing this on the core, should we keep a list of unsupported flags and remove them after importing core flags?

package.json
```diff
@@ -126,7 +126,7 @@
     "rimraf": "^3.0.2",
     "ts-jest": "^25.3.0",
     "typescript": "^3.9.5",
-    "webpack": "^4.42.0",
+    "webpack": "https://github.com/webpack/webpack",
```
anshumanv, line 129: Leaving a comment here, this to be updated before we merge

snitin315:
> I'm not sure of the feasibility of removing this on the core, should we keep a list of unsupported flags and remove them after importing core flags?

Yes, I think we can filter flags on the CLI side.

packages/webpack-cli/lib/utils/cli-flags.js
```diff
@@ -6,6 +8,31 @@ const ADVANCED_GROUP = 'advanced';
...
+
+// remove entry flag to avoid duplicate entry
+flagsFromCore = flagsFromCore.filter((flag) => flag.name !== 'entry');
```
anshumanv, line 34: Stats is also present on core

snitin315: Yes, we can filter here all the duplicate flags and the flags we don't wish to support.

packages/webpack-cli/lib/webpack-cli.js
snitin315: @anshumanv corresponding flags are broken as well :/

snitin315:
`--amd` is broken on our side. it accepts only `false` value and right now CLI doesn't support parsing of boolean flags like `webpack-cli --amd=false`. 
Looking into it if it's possible to parse it like this with `commander.js`

packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -213,6 +215,30 @@ class WebpackCLI extends GroupHelper {
...
+     * Responsible for handling flags coming from webpack/webpack
+     * @private\
+     * @returns {void}
+     */
+    _handleCoreFlags() {
+        if (this.groupMap.has('core')) {
+            const coreFlags = this.groupMap.get('core');
+
+            // convert all the flags from map to single object
+            const coreConfig = coreFlags.reduce((allFlag, curFlag) => (allFlag = { ...allFlag, ...curFlag }), {});
```
jamesgeorge007, line 228:
```suggestion
            const coreConfig = coreFlags.reduce((allFlag, curFlag) => ({ ...allFlag, ...curFlag }), {});
```

evilebottnawi:
@snitin315 

> --amd is broken on our side. it accepts only false value and right now CLI doesn't support parsing of boolean flags like webpack-cli --amd=false.

Not sure we should support it, because we have `--amd` and `--no-amd` support, boolean as string maybe some misleading, for example `--entry=false` can be mean:
- loading `false.js` as entry point
- no entry configuration

snitin315:
@evilebottnawi I think we can specify `--no-amd` in `cli-flags` and we will be good.

evilebottnawi:
@snitin315 For all (not sure for all but for most of them) `boolean` flag we should register `--no-name` flag

snitin315:
Let's implement `--no-name` flag for all now, we can filter them later (if needed).

packages/webpack-cli/lib/utils/cli-flags.js
```diff
@@ -34,8 +34,11 @@ let flagsFromCore = version.startsWith(5)
 const duplicateFlags = ['entry', 'stats', 'watch', 'target'];
```
evilebottnawi, line 34: We should remove duplicate flags from our flags in favor core flags, but we can postpone it and do it in the other PR

snitin315: I'm afraid if we can do this we are still supporting webpack 4, and if we remove these flags from our flags we will lose compatibility with v4, and users using v4 will no longer be able to access these flags.

evilebottnawi: @snitin315 We can do it only for webpack@5

snitin315: oh yes, will implement in separate PR :+1:

evilebottnawi:
@snitin315 Can we move these changes:
- allow parsing type: 'number'
- fix help output ( currently showing type string for every flag )

to other PRs, to avoid unnecessary changes in this PR?

snitin315:
We can move this one `allow parsing type: 'number' to another PR.  (should be mergers before this PR).

help output was wrong for core-flags and is related so should keep it in this PR.

evilebottnawi:
> We can move this one allow parsing type: 'number' to another PR. (should be merged before this PR).

:+1:

snitin315:
I can see `--output-jsonp-script-type` flag being imported from core, but it's not present in snapshot. https://github.com/webpack/webpack/blob/master/test/__snapshots__/Cli.test.js.snap
it's `--output-script-type` is in snapshot instead.

Screenshot at 2020-07-03 18-24-01

anshumanv:
@snitin315 probably needs snapshot update, let me see

snitin315:
@anshumanv using `webpack@next` instead of `master` branch somehow fixed it.

package.json
snitin315: done

anshumanv: nicenice

packages/webpack-cli/lib/groups/BasicGroup.js
```diff
@@ -48,6 +48,7 @@ class BasicGroup extends GroupHelper {
 
     run() {
         this.resolveFlags();
+        // console.log(this);
```
anshumanv, line 51: rm log, added while debugging

packages/webpack-cli/lib/groups/StatsGroup.js
```diff
@@ -33,6 +33,7 @@ class StatsGroup extends GroupHelper {
 
     run() {
         this.resolveOptions();
+        // console.log(this);
```
anshumanv, line 36: same

packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -122,12 +123,13 @@ class WebpackCLI extends GroupHelper {
          * options is an array (multiple configuration) so we create a new
          * configuration where each element is individually merged
          */
+        // console.log({ options, strategy });
```
anshumanv, line 126: same

packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -122,12 +123,13 @@ class WebpackCLI extends GroupHelper {
...
         if (Array.isArray(options)) {
             this.compilerConfiguration = options.map((configuration) => {
                 if (strategy) {
                     return webpackMerge.strategy(strategy)(this.compilerConfiguration, configuration);
                 }
-                return webpackMerge(this.compilerConfiguration, configuration);
+                return webpackMerge(configuration, this.compilerConfiguration);
```
anshumanv, line 132: I think we should revert this, not quite sure why I made this change 😅

anshumanv: But this seems logically incorrect, configuration should take precedence

test/utils/test-utils.js
```diff
@@ -292,6 +292,15 @@ const runServe = (args, testPath) => {
     });
 };
 
+const hyphenToUpperCase = (name) => {
```
anshumanv, line 295: Seems duplicate, possible to import it?

packages/webpack-cli/lib/utils/arg-parser.js
```diff
@@ -63,6 +64,11 @@ function argParser(options, args, argsOnly = false, name = '', helpFunction = un
             if (!option.multiple) {
                 // Prevent default behavior for standalone options
                 parserInstance.option(flagsWithType, option.description, option.defaultValue).action(() => {});
+                if (option.group === 'core' && option.type === Boolean) {
+                    // commander requires explicitly adding the negated version of boolean flags
+                    const negatedFlag = `--no-${option.name}`;
+                    parserInstance.option(negatedFlag, `negates ${option.name}`);
+                }
```
evilebottnawi, line 71: I think we can improve it like:
```js
{
            name: 'hot',
            usage: '--hot',
            alias: 'h',
            type: Boolean,
            negative: true
            group: ADVANCED_GROUP,
            link: 'https://webpack.js.org/concepts/hot-module-replacement/',
},      
```

Allow core flags have `negative: true`, I think default webpack-cli flags also should have `negative: true`

snitin315: Yes, It's better.

snitin315: should I make a PR to core or it's okay to handle this CLI side.?

anshumanv: I think it's okay to handle on CLI side, since we're already adding groups and other things, should be fine.

snitin315: then we can move this change to a seperate PR.

test/utils/test-utils.js
snitin315: It's defined as a method in `GroupHelper` util,

packages/webpack-cli/lib/utils/arg-parser.js
anshumanv: yes, ideally, are you on it? If not, I have bandwidth to pick up

snitin315: @anshumanv WIP

anshumanv: Great 👍

snitin315: done :raised_hands:

test/utils/test-utils.js
anshumanv: possible to move it to utils package? any downsides?

snitin315: Ye we can move, we just have to import the `GroupHelper` util and create it's new instance. I will make the necessary changes.

snitin315: but this may cause noise in this PR, can we do this in a future PR ?

snitin315:
/cc @webpack/cli-team  there is one issue with `--watch-options-poll`,  it's `simpleType` should be `number` as it can take only boolean on number values but it's `simpleType` is defined as `strirng` instead.

https://github.com/webpack/webpack/blob/e3efb7b46cc45d97622cbb6a3b8fda3f21143829/test/__snapshots__/Cli.test.js.snap#L4365-L4383

snitin315:
> @snitin315 is handling it as string throws errors?

Yes. It can't be handled as string. 

Screenshot at 2020-07-07 11-27-03

snitin315:
Ideally we should fix it on core side, but we can override it's type to `Number` as a workaround for now. 
Or we can skip test for this one :smile: until it's fixed in core.

rishabh3112:
@snitin315, we should fix it on core side first then.

snitin315:
till then i will add to the list of broken flags.

evilebottnawi:
@snitin315 Yep, all broken options should be fixed on webpack side :+1:

packages/webpack-cli/lib/webpack-cli.js
snitin315: Let's keep this change.

snitin315:
Ready for review :raised_hands: 

Screenshot at 2020-07-07 18-33-39

evilebottnawi:
@snitin315 What options are broken from webpack side?

rishabh3112:
@evilebottnawi here -> https://github.com/webpack/webpack-cli/pull/1630#issuecomment-646407523

packages/webpack-cli/lib/webpack-cli.js
anshumanv: Did you confirm?

test/core-flags/.gitignore
```diff
@@ -0,0 +1,3 @@
+/test
+/bin
+/dist
```
anshumanv, line 3: We don't need this once rebased with master

snitin315:
> @evilebottnawi here -> [#1630 (comment)](https://github.com/webpack/webpack-cli/pull/1630#issuecomment-646407523)

Theese are broken flags some might be broken on cli side need to investigate.

evilebottnawi:
@snitin315 Thanks, no we need to focus on fix bugs on webpack side

packages/webpack-cli/lib/utils/cli-flags.js
```diff
@@ -6,6 +8,43 @@ const ADVANCED_GROUP = 'advanced';
 const DISPLAY_GROUP = 'stats';
 const ZERO_CONFIG_GROUP = 'zero-config';
 
+// Extract all the flages being exported from core. A list of cli flags generated by core
+// can be found here https://github.com/webpack/webpack/blob/master/test/__snapshots__/Cli.test.js.snap
+let flagsFromCore = version.startsWith(5)
```
evilebottnawi, line 13: I think we can simplify this, using `typeof webpack.cli !== 'undefined'`

snitin315: yes we can

packages/webpack-cli/lib/webpack-cli.js
snitin315: well it's not necessary to write it like this, but I see the `commonConfig` is passed as first argument in docs.

test/core-flags/.gitignore
snitin315: removed :+1:

packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -209,6 +210,25 @@ class WebpackCLI extends GroupHelper {
...
+     * @returns {void}
+     */
+    _handleCoreFlags() {
+        if (this.groupMap.has('core')) {
+            const coreFlags = this.groupMap.get('core');
+
+            // convert all the flags from map to single object
+            const coreConfig = coreFlags.reduce((allFlag, curFlag) => ({ ...allFlag, ...curFlag }), {});
+
+            const coreCliHelper = require('webpack').cli;
```
evilebottnawi, line 225: Need to move on top of file, just rewrite this (fast exit):
```js
if (!this.groupMap.has('core')) {
  return;
}
```

packages/webpack-cli/lib/webpack-cli.js
evilebottnawi: Do we have tests on this change?

snitin315: No but I see this change is not affecting anything, can be safely reverted.

evilebottnawi: Let's add tests on this

snitin315: how should we test it.?

evilebottnawi: Need investigate, right now we have:
```
return webpackMerge.strategy(strategy)(this.compilerConfiguration, configuration);
```

and

```
return webpackMerge(configuration, this.compilerConfiguration);
```

I think here bug

snitin315: @anshumanv any suggestions?

jest.config.js
```diff
@@ -1,5 +1,10 @@
+const { version } = require('webpack');
+
+//ignore core-flags test for webpack@4
+const ignorePattern = version.startsWith('5') ? ['<rootDir>/node_modules/'] : ['<rootDir>/node_modules/', '<rootDir>/test/core-flags'];
```
anshumanv, line 4: You can use the same check here, if .cli exists

packages/webpack-cli/lib/webpack-cli.js
anshumanv: Yep let's revert this for now, I think I changed while debugging, we can revisit later to add tests

test/utils/test-utils.js
anshumanv: Um okay

anshumanv:
Apart from this I think we're mostly good 🎉

package.json
```diff
@@ -126,7 +126,7 @@
     "rimraf": "^3.0.2",
     "ts-jest": "^25.3.0",
     "typescript": "^3.9.5",
-    "webpack": "^4.43.0",
+    "webpack": "^5.0.0-beta.20",
```
evilebottnawi, line 129: Let's change it to `^4`, we automatically install webpack@5 in CI

packages/webpack-cli/lib/utils/cli-flags.js
```diff
@@ -6,6 +8,44 @@ const ADVANCED_GROUP = 'advanced';
...
+// duplicate flags
+const duplicateFlags = ['entry', 'mode', 'stats', 'watch', 'target'];
+// broken flags
+const brokenFlags = ['module-no-parse', 'module-no-parse-reset', 'optimization-no-emit-on-errors', 'watch-options-poll'];
+// regExp Flags
+const regExpFlags = ['module-expr-context-reg-exp', 'module-unknown-context-reg-exp', 'module-wrapped-context-reg-exp'];
+// inFeasibleFlags https://github.com/webpack/webpack-cli/pull/1630#issuecomment-646992322
+const inFeasibleFlags = flagsFromCore.filter(({ name }) => {
+    return name.includes('module-rules-') || name.includes('alias-') || name.startsWith('output-library');
+});
```
evilebottnawi, line 44: /cc @snitin315 Why we can't handle `output-library` and `alias-`?

packages/webpack-cli/lib/utils/cli-flags.js
```diff
@@ -6,6 +8,44 @@ const ADVANCED_GROUP = 'advanced';
...
+                  name: flag,
+                  group: 'core',
+              };
+          })
+        : [];
+
+// duplicate flags
+const duplicateFlags = ['entry', 'mode', 'stats', 'watch', 'target'];
+// broken flags
+const brokenFlags = ['module-no-parse', 'module-no-parse-reset', 'optimization-no-emit-on-errors', 'watch-options-poll'];
```
evilebottnawi, line 38: `watch-options-poll` fixed

evilebottnawi: `optimization-no-emit-on-errors` - what is bug with this?

snitin315: `webpack --optimization-no-emit-errors` should set `optimization.noEmitErrors: true` but it's not doing so.
Screenshot at 2020-07-09 18-13-15

packages/webpack-cli/lib/utils/cli-flags.js
snitin315: `output-library` related flags some I think are broken (like `--output-library-amd`) some depends on other flags for example we can't use `--output-library-name` alone if did webpack will throw error it should be done like  
`webpack --output-library-name testName --outpyt-library-type var` .

snitin315: I think such nested webpack config not feasable via CLI hence I have refered them as inFeasableFlags. We can discuss more about it. https://github.com/webpack/webpack-cli/pull/1630#issuecomment-646992322

packages/webpack-cli/lib/utils/cli-flags.js
evilebottnawi: @snitin315 Let's send a fix to webpack, this option is very popular

packages/webpack-cli/lib/utils/cli-flags.js
evilebottnawi: /cc @sokra What do you think about it?

sokra: I think the cli should not apply additional filtering. If some arguments exposed by webpack doesn't make sense that should be filtered on webpack side. Webpack allows to filter arguments in the Schema with `cli.exclude`. This is already done for some options, but we can add more.

evilebottnawi: @sokra Agree, but I think some options from array still can be handled

@snitin315 Can you provide full list of filtered options from webpack? I think we can handle aliases too, but in some difference way

snitin315: Below is the full list -
```js
module-rules-compiler
module-rules-enforce
module-rules-exclude
module-rules-include
module-rules-issuer
module-rules-loader
module-rules-mimetype
module-rules-options
module-rules-real-resource
module-rules-resource
module-rules-resource-query
module-rules-side-effects
module-rules-test
module-rules-type
module-rules-use-ident
module-rules-use-loader
module-rules-use-options
module-rules-use
module-rules-reset
output-library
output-library-reset
output-library-amd
output-library-commonjs
output-library-root
output-library-root-reset
output-library-auxiliary-comment
output-library-auxiliary-comment-amd
output-library-auxiliary-comment-commonjs
output-library-auxiliary-comment-commonjs2
output-library-auxiliary-comment-root
output-library-export
output-library-export-reset
output-library-name
output-library-name-reset
output-library-name-amd
output-library-name-commonjs
output-library-name-root
output-library-name-root-reset
output-library-type
output-library-umd-named-define
resolve-alias-alias
resolve-alias-name
resolve-alias-only-module
resolve-alias-reset
resolve-alias-fields
resolve-alias-fields-reset
resolve-loader-alias-alias
resolve-loader-alias-name
resolve-loader-alias-only-module
resolve-loader-alias-reset
resolve-loader-alias-fields
resolve-loader-alias-fields-reset
```

packages/webpack-cli/lib/utils/cli-flags.js
snitin315: @evilebottnawi I investigated and found `optimization-no-emit-on-errors` comes with pre-appended value = true while importing and when we do `webpack --optimization-no-emit-on-errors` it sets it to flase.

snitin315: opposite behavior from rest of the boolean flags.

evilebottnawi: Why we set it to false?

packages/webpack-cli/lib/utils/cli-flags.js
evilebottnawi: I think `resolve-*` can be supported with some limitations, we need implement `separator`, for example:
```

webpack resolve-alias=alias|name
```

`resolve-alias-fields` also can be easy implemented, it is just `webpack --resolve-alias-fields=foo --resolve-alias-fields=bar`, not sure what mean `resolve-alias-fields-reset`? Can you clarify?

Same for `output-library`

snitin315: `resolve-alias-fields-reset` will reset `resolve.aliasFields` to empty array. 
```
resolve: {
aliasFields: []
}
```

packages/webpack-cli/lib/utils/cli-flags.js
snitin315: we don't do this intentionally, I think somewhere we have bug I'm looking into it.

packages/webpack-cli/lib/utils/cli-flags.js
evilebottnawi: Yes, but here other two cases:
```
webpack --resolve-alias=alias1|name1 --resolve-alias=alias2|name2
```

Adding `alias1` and `alias2` to aliases, you can already have aliases in your configuration

```
webpack --resolve-alias=alias1|name1 --resolve-alias=alias2|name2 --resolve-alias-fields-reset
```

Adding `alias1` and `alias2` to aliases, but reset them before, so aliases from the configuration are not respected

snitin315: > webpack --resolve-alias=alias1|name1 --resolve-alias=alias2|name2

**It's result will be -
`resolve : { alias: [ { alias: ['alias1',  'alias2'], name: ['alias1', 'alias2'] } ] }`** 

> webpack --resolve-alias=alias1|name1 --resolve-alias=alias2|name2 --resolve-alias-fields-reset

**It's result will be -
`resolve : { alias: [ { alias: ['alias1',  'alias2'], name: ['alias1', 'alias2'] } ] , allFields: [] }`**

evilebottnawi: We need improve our arg parser, we have case where you have complex options and you can use it multiple times, any ideas?

snitin315: maybe something like `webpack --mupltiple-values=value1,value2,value3` ?

snitin315: /cc @webpack/cli-team need suggestions.

snitin315: @evilebottnawi are these options popular?

test/core-flags/output-flags.test.js
```diff
@@ -3,9 +3,9 @@
 const { run, hyphenToUpperCase } = require('../utils/test-utils');
 const { flagsFromCore } = require('../../packages/webpack-cli/lib/utils/cli-flags');
 
-const outputFlags = flagsFromCore.filter(({ name }) => name.startsWith('output-'));
+const outputFlags = flagsFromCore.filter(({ name }) => name.startsWith('output-') && !name.includes('output-library'));
```
evilebottnawi, line 6: Why we exclude `output-library'`? It is just a string

snitin315: I'm adding separate tests for `output-library` related flags, as they are dependent on each other and creates noise with these test cases.

evilebottnawi: I don't understand, developer can use `webpack ./src/index.js --output-library=myLibrary`

snitin315: Actually we are filtering around 20 flags here,
```
output-library
output-library-reset
output-library-amd
output-library-commonjs
output-library-root
output-library-root-reset
output-library-auxiliary-comment
output-library-auxiliary-comment-amd
output-library-auxiliary-comment-commonjs
output-library-auxiliary-comment-commonjs2
output-library-auxiliary-comment-root
output-library-export
output-library-export-reset
output-library-name
output-library-name-reset
output-library-name-amd
output-library-name-commonjs
output-library-name-root
output-library-name-root-reset
output-library-type
output-library-umd-named-define
```

evilebottnawi: Yes, we should fix, `output-library-*` is very popular flags, so we should support them, please focus on this

packages/webpack-cli/lib/utils/cli-flags.js
snitin315: >We need improve our arg parser, we have case where you have complex options and you can use it multiple times, any ideas?

I found commander.js will support `variadic args` https://github.com/tj/commander.js/pull/1250  in next release, This is maybe helpful.

packages/webpack-cli/lib/utils/cli-flags.js
sokra: I think that's probably caused by the `no-` instead of the name. Sounds like a bug in the args parsing.

Anyway it might make sense to change this in webpack core to `emitOnErrors`, as negative flags are bad practice anyway...

snitin315: @sokra yes, It was caused by `no-` it was a bug in `commander.js`, I have submitted a fix for same https://github.com/tj/commander.js/pull/1302 .

> Anyway it might make sense to change this in webpack core to emitOnErrors, as negative flags are bad practice anyway...

I Agree.

snitin315: This will be fixed with next release of `commander.js`.

packages/webpack-cli/lib/utils/cli-flags.js
```diff
@@ -6,6 +8,38 @@ const ADVANCED_GROUP = 'advanced';
...
+              };
+          })
+        : [];
+
+// duplicate flags
+const duplicateFlags = ['entry', 'mode', 'stats', 'watch', 'target'];
+// regExp Flags
+const regExpFlags = ['module-expr-context-reg-exp', 'module-unknown-context-reg-exp', 'module-wrapped-context-reg-exp'];
+// filter all duplicate, broken and regExp flags
+const flagsToFilter = [...duplicateFlags, ...regExpFlags];
```
evilebottnawi, line 40: Why we ignore `regexp` flags? We can implement `regexp: true` for some options and use `new RegExp(options.something)` and set it for webpack

sokra: You can also pass these are strings to webpack processArguments, it will convert them to regexp.

The webpack processArguments method allows strings for all arguments and converts them to the correct format.

test/core-flags/module-flags.test.js
```diff
@@ -0,0 +1,55 @@
...
+
+const moduleFlags = flagsFromCore.filter(({ name }) => name.startsWith('module-'));
+
+describe('module config related flag', () => {
+    moduleFlags.forEach((flag) => {
+        // extract property name from flag name
+        const property = flag.name.split('module-')[1];
+        const propName = hyphenToUpperCase(property);
+
+        //TODO: improve webpackCLITestPlugin for testing of nested options
```
evilebottnawi, line 14: Let's resolve it here

snitin315:
> /cc @snitin315 Can we write script to automatically generate markdown docs with all supported flags for v4 and v5 and put it he
re?

Like we have listed here https://github.com/webpack/webpack-cli/blob/next/packages/webpack-cli/README.md#supported-arguments-and-commands

evilebottnawi:
Yes :+1: Two sections `Webpack 4` and `Webpack 5`

packages/webpack-cli/README.md
```diff
@@ -58,6 +58,932 @@ Options
...
+  --no-stats-reasons                                                                 Negates stats-reasons
+  --no-stats-runtime                                                                 Negates stats-runtime
+  --no-stats-source                                                                  Negates stats-source
+  --no-stats-timings                                                                 Negates stats-timings
+  --no-stats-used-exports                                                            Negates stats-used-exports
+  --no-stats-version                                                                 Negates stats-version
+  --no-stats-warnings                                                                Negates stats-warnings
+  --no-stats-warnings-filter-reset                                                   Negates stats-warnings-filter-reset
+  --no-watch-options-ignored-reset                                                   Negates watch-options-ignored-reset
+  --no-watch-options-stdin                                                           Negates watch-options-stdin
```
evilebottnawi, line 986: Let's use

### webpack 4

```
...
```

### webpack 5

```
...
````

Also we have very many spaces between flags and description. Can we fix it?

Comment (snitin315) [reply packages/webpack-cli/README.md:986]: there were many spaces because some flags have a very very long name like 
`--optimization-split-chunks-fallback-cache-group-automatic-name-delimiter`
`--optimization-split-chunks-fallback-cache-group-max-initial-size.` 

I managed to reduce down spaces to some extent.

Comment (snitin315) [reply packages/webpack-cli/README.md:986]: I will push my changes soon.

File: test/core-flags/cache-flags.test.js [hunk 24]
```diff
@@ -0,0 +1,100 @@
...
+
+        expect(stderr).toBeFalsy();
+        expect(stdout).toContain(`hashAlgorithm: 'sha256'`);
+    });
+
+    it('should set cache.managedPaths with --cache-managed-paths', () => {
+        const { stderr, stdout } = run(__dirname, ['--cache-type', 'memory', '--cache-managed-paths', '/test-manage-path']);
+
+        expect(stderr).toBeFalsy();
+        if (process.platform === 'win32') {
```
evilebottnawi, line 65: We should avoid rely on `process.platform` in tests, can we fix it?

snitin315: Yes

snitin315:
> What about flags which we want to exclude?

PR here https://github.com/webpack/webpack/pull/11171

evilebottnawi:
@snitin315 Let's exclude them from tests to avoid problems

snitin315:
@evilebottnawi Done 👍🏻

snitin315:
Yes, we should start documenting the changes the new changes CLI offers.

> disable logging and stats output (like --quiet)

Yes, we should support `--quiet`, though our logging isn't consistent somewhere we have used `logger` sometimes `console.error` and sometimes `process.stdout.write`.

evilebottnawi:
@snitin315 I think couple popular examples is enough, let's do it after merge 

We need restore coverage https://github.com/webpack/webpack-cli/pull/1391

test/core-flags/cache-flags.test.js
```diff
@@ -0,0 +1,82 @@
+'use strict';
+
+const { run } = require('../utils/test-utils');
+
+describe('cahche related flags from core', () => {
```
jamesgeorge007, line 5:
```suggestion
describe('cache related flags from core', () => {
```

packages/webpack-cli/lib/utils/cli-flags.js
```diff
@@ -241,4 +273,5 @@ module.exports = {
 			group: BASIC_GROUP
 		} */
     ],
+    flagsFromCore: [...flagsFromCore],
```
anshumanv, line 276: no need spread in blank array

snitin315:
```suggestion
    flagsFromCore: ...flagsFromCore,
```

packages/webpack-cli/lib/utils/cli-flags.js
```diff
@@ -241,4 +273,5 @@ module.exports = {
 			group: BASIC_GROUP
 		} */
     ],
+    flagsFromCore: ...flagsFromCore,
```
snitin315, line 276:
```suggestion
    flagsFromCore: flagsFromCore,
```

test/core-flags/module-flags.test.js
```diff
@@ -0,0 +1,74 @@
+'use strict';
+
+const { run, hyphenToUpperCase } = require('../utils/test-utils');
+const { flagsFromCore } = require('../../packages/webpack-cli/lib/utils/cli-flags');
+//--module-rules-options will be exluded for cli.
```
jamesgeorge007, line 5:
```suggestion
//--module-rules-options will be excluded for cli.
```

packages/webpack-cli/lib/utils/cli-flags.js
```diff
@@ -6,6 +8,35 @@ const ADVANCED_GROUP = 'advanced';
 const DISPLAY_GROUP = 'stats';
 const ZERO_CONFIG_GROUP = 'zero-config';
 
+// Extract all the flages being exported from core. A list of cli flags generated by core
```
jamesgeorge007, line 11:
```suggestion
// Extract all the flags being exported from core. A list of cli flags generated by core
```

packages/webpack-cli/lib/utils/cli-flags.js
```diff
@@ -241,4 +273,5 @@ module.exports = {
 			group: BASIC_GROUP
 		} */
     ],
+    flagsFromCore: flagsFromCore,
```
jamesgeorge007, line 276:
```suggestion
    flagsFromCore,
```

anshumanv:
@jamesgeorge007 please don't commit yourself, suggestions suffice

evilebottnawi:
@webpack/cli-team Merge?

anshumanv:
> @webpack/cli-team Merge?

yep we can merge 👍

evilebottnawi:
Big thanks for all, it is great feature, I think we should resolve `serve` command and do release, when focus on perf and refactor

PR #1631, opened 2020-06-14
tests: add test for multiple config in dotfolder

anshumanv:

tests

Yes

NA

**Summary**
- Add tests when there are multiple configs present in dotfolder

No

**Other information**

### DIFF
--- a/test/config/multiple/.webpack/webpack.config.dev.js
+++ b/test/config/multiple/.webpack/webpack.config.dev.js
@@ -1,0 +1,9 @@
+const { resolve } = require('path');
+
+module.exports = {
+    entry: './index.js',
+    output: {
+        path: resolve(__dirname, '../binary'),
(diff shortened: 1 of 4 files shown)


anshumanv:
@evilebottnawi we presently only support reading `.dev.js` `.prod.js` files from `.webpack` folder, do we wish to support reading from root as well?

anshumanv:
Presently root only supports webpack.config.*

evilebottnawi:
Yes, need support too

anshumanv:
Will fix 👍

evilebottnawi:
@anshumanv Also webpack has `none` env, so I think we should support `webpack.config.none.js`, very rare case, but I think we should supports + I think we should support full naming like `webpack.config.development.js` and `webpack.config.production.js`

anshumanv:
Sounds good, will fix 👍

PR #1636, opened 2020-06-16
chore: upgrades typescript to the latest

maraisr:
With this PR I aim to upgrade typescript to the latest. 

Testing notes:
- All linters passed ✅
- All builds passed ✅
- Unit tests 9 failed, but believe this to be a windows issue. 
image
- There was only 1 code change required, and was due to a fundamental type mismatch. I've noted that in the [line-comment](https://github.com/webpack/webpack-cli/pull/1636/files#r441183817). 

fixes #1617

### DIFF
--- a/packages/utils/src/modify-config-helper.ts
+++ b/packages/utils/src/modify-config-helper.ts
@@ -133,13 +133,13 @@ export function modifyHelperUtil(
                 process.exitCode = -1;
             }
 
-            const transformConfig: TransformConfig = Object.assign(
+            const transformConfig = Object.assign(
                 {
                     configFile: !configPath ? null : fs.readFileSync(configPath, 'utf8'),
                     configPath,
                 },
                 finalConfig,
-            );
+            ) as TransformConfig;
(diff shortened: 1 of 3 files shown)


packages/utils/src/modify-config-helper.ts
```diff
@@ -133,13 +133,13 @@ export function modifyHelperUtil(
                 process.exitCode = -1;
             }
 
-            const transformConfig: TransformConfig = Object.assign(
+            const transformConfig = Object.assign(
```
maraisr, line 136: The `TransformConfig` type and the `WebpackScaffoldObject` type both had the field `config`, but both config fields were different. When the object.assign kicked in, it changed the type to `TransformConfig & WebpackScaffoldObject`, but was then trying to put it into a `TransformConfig`. Which then complained that the ScaffoldObject one didnt have a required merge field. So without understanding these types fully, ive opted to explicitly cast the `transformConfig` as the type it originally was. Perhaps someone can give me more light what these different types mean, and if the `config` field can be shared perhaps, or handle the case typescript is highlighting?

PR #1637, opened 2020-06-17
feat: add support for none config in dotfolder

anshumanv:
feat

Yes

NA

**Summary**
- Add support for reading none config from dotfolder
- Add relevant tests

No

**Other information**
No


anshumanv:
> Can we add tests where we have multiple configuration and load them based on mode?

Config resolves doesn't respect mode, will add the support

packages/webpack-cli/lib/groups/ConfigGroup.js
```diff
@@ -4,14 +4,25 @@ const { extensions } = require('interpret');
...
     '.webpack/webpackfile',
+    'webpack.config.development',
     'webpack.config',
+    'webpack.config.production',
 ];
 
+const modeAlias = {
+    production: 'prod',
+    development: 'dev',
+};
```
evilebottnawi, line 24: What about none?

anshumanv: none has no alias, no need

anshumanv: It matches either mode or it's alias, none will match the mode, no need alias

anshumanv: @evilebottnawi

anshumanv:
Yes, in near future @evilebottnawi 👍

anshumanv:
@evilebottnawi added 👍

anshumanv:
> So we prefer webpack.config.develoment.js over .webpack/webpack.config.development.js?


Yes

evilebottnawi:
@webpack/cli-team we need small discussion about it, should we prefer config from `.webpack` directory over root configuration or not?

rishabh3112:
`.webpack/*` as this now what we intend to be the default directory for configurations by design.

anshumanv:
Seems fair

anshumanv:
Updated

packages/webpack-cli/lib/groups/ConfigGroup.js
evilebottnawi: Can you add comment how we prefer configurations here for future developers?

PR #1644, opened 2020-06-22
feat: add aliases to all available commands

jamesgeorge007:
enhancement

Nope

Nope

**Summary**
- Added aliases for all available commands.
- Updated the help information to show up command aliases.
Screenshot 2020-06-23 at 4 39 27 PM

Nope

**Other information**
N/A

### DIFF
--- a/packages/webpack-cli/lib/utils/cli-flags.js
+++ b/packages/webpack-cli/lib/utils/cli-flags.js
@@ -26,6 +26,7 @@ module.exports = {
         },
         {
             name: 'migrate',
+            alias: 'm',
             type: String,
             usage: 'migrate',
             description: 'Migrate a configuration to a new version',
@@ -49,6 +50,7 @@ module.exports = {
         {
             name: 'info',
             scope: 'external',
+            alias: 'i',
             type: String,
             usage: 'info [options]',
             description: 'Outputs information about your system and dependencies',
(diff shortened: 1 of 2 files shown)


evilebottnawi:
Let's do for each command small alias (first letter of work):
- init | i
- migrate | m
- loader | l
- etc

packages/webpack-cli/lib/utils/cli-flags.js
```diff
@@ -63,6 +65,7 @@ module.exports = {
         },
         {
             name: 'serve',
+            alias: 's',
             scope: 'external',
```
snitin315, line 69: Alias `s` is already occupied by `sourcemap`, should be removed from there. 
https://github.com/webpack/webpack-cli/blob/ac2e52c2ef8aa74aa795476be94ce7c968b8d1c5/packages/webpack-cli/lib/utils/cli-flags.js#L168-L175

jamesgeorge007: `sourcemap` is a `cli-flag` as oppoesd to `serve` which is a command.
`webpack s` vs `webpack -s <value>`

snitin315: oh yes, my bad :-)

Issue #866, opened 2019-04-28
[ENHANCEMENT]: move types to @types/jscodeshift


ematipico:
At the moment we use internal interfaces for our AST functions. This is not working as we have duplicates and this will hit really bad on the long run.

We should move to [`@types/jscodeshift`](https://www.npmjs.com/package/@types/jscodeshift)

For this one, ideally it would be better to create PRs for each package, starting from the small ones.

anshumanv:
Will take a look at this. :)

evenstensberg:
@anshumanv thanks :)

omginbd:
Seems like this was about 6 months ago and I'm looking for a "Good First Contribution" ticket, does anyone mind if I start looking into this?

evenstensberg:
Yes!! If you want help, we’re here for you!

omginbd:
👍 Great, I'll get started!

hemal7735:
@ematipico shouldn't we do the same for Webpack options?

should we move [WebpackOptions](https://github.com/webpack/webpack-cli/blob/next/packages/generators/types/index.ts) to [webpack-type definitions](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/webpack/index.d.ts) ?

snitin315:
bump

Cronus1007:
@snitin Please assign this issue to me.

snitin315:
Feel free to send a PR.

Cronus1007:
@snitin @EugeneHlushko  @ematipico  Sir in my opinion rather than moving to @types/jscodeshift we can apply the cohesion principle.

Cronus1007:
#1933 @anshumanv   Sir I think the linked PR would close this issue for the particular Info package.

shreyan-invalid:
Hey, I am new to open source contribution, I would like to work with this issue, can you guide me through it?

alexander-akait:
Close in favor https://github.com/webpack/webpack-cli/issues/2430

PR #1649, opened 2020-06-24
fix(serve): merge CLI and devServer options correctly

Loonride:

fix/refactor

Not yet: This will not exactly work without a new webpack-dev-server release

no

**Summary**

@anshumanv Sorry for taking over on this from https://github.com/webpack/webpack-cli/pull/1470, just wanted to show what I'm thinking. The merge strategy for CLI options and the user's devServer options in their config should be that CLI options take precedence over devServer options.

This merge strategy relies on: https://github.com/webpack/webpack-dev-server/pull/2659, where default values are removed from serve CLI flags, then default values are set later if needed.

We need to keep the `createConfig` stuff here as minimal as possible, because we want all of the heavy configuration work to be done on the `webpack-dev-server` side, but some config creation is still necessary on this side.

TODO: even though default CLI flag values are removed internally, we still need to make a way to display the values that are defaulted to when somebody runs `webpack-cli help serve`.

Also: need to remove the `any` types from this PR still.

**Other information**

Things to think about:
- How should types like the dev server options type be handled? Currently I made a `types.ts` file in the serve package, but such types should probably be in a separate package
- We should add some e2e tests ensuring the dev server works with special flag cases (like `--no-static`, `--static dir1 --static dir2`, etc.), but we need to wait for webpack-dev-server v4 release for these CLI flag settings to be available to webpack-cli


packages/serve/src/getDevServerOptions.ts
```diff
@@ -0,0 +1,24 @@
...
+ * @param {Object} compiler - webpack compiler
+ *
+ * @returns {Object}
+ */
+export default function getDevServerOptions(compiler): any {
+    let devServerOptions;
+    if (compiler.compilers) {
+        // devServer options could be found in any of the compilers,
+        // so simply find the first instance and use it, if there is one
+        const comp = compiler.compilers.find((comp) => comp.options.devServer);
```
rishabh3112, line 14: We should let user decide which one to use. There are two possible solutions for this.
- Add an inquirer question with the names of compilers and let user select.
- Add a `--name` flag where user can supply the name of the compiler that should be used. (as suggested by @evilebottnawi)

Using both is also a solution (a better one imo).

packages/serve/src/mergeOptions.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ *
+ * Merges CLI options and devServer options from config file
+ *
+ * @param {Object} cliOptions - devServer args
+ *
+ * @returns {Object} merged options object
+ */
+export default function mergeOptions(cliOptions, devServerOptions): any {
```
rishabh3112, line 9:
```suggestion
export default function mergeOptions(cliOptions, devServerOptions): object {
```

packages/serve/src/getDevServerOptions.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ *
+ * Get the devServer option from the user's compiler options
+ *
+ * @param {Object} compiler - webpack compiler
+ *
+ * @returns {Object}
+ */
+export default function getDevServerOptions(compiler): any {
```
rishabh3112, line 9:
```suggestion
export default function getDevServerOptions(compiler): object {
```

packages/serve/src/createConfig.ts
```diff
@@ -0,0 +1,27 @@
+/**
+ *
+ * Creates a devServer config from CLI args
+ *
+ * @param {Object} args - devServer args
+ *
+ * @returns {Object} valid devServer options object
+ */
+export default function createConfig(args): any {
```
rishabh3112, line 9:
```suggestion
export default function createConfig(args): object {
```

packages/serve/src/getDevServerOptions.ts
evilebottnawi: Think the `name` is better here, but in theory developers can run `two` dev servers, we need thing about it

rishabh3112: Yeah @evilebottnawi, I have also suggested to run all possible `devServer`s in parallel before, [here](https://github.com/webpack/webpack-cli/pull/1470#discussion_r443180108) is a small thread on the discussion.

Friendly ping @Loonride :)

Loonride: Will get back to this soon, there is still quite a few things to do on the dev server side first :smile:

Loonride: I agree with the `name` idea. Maybe it should also take a number index for the compiler being used, in case the user has not given their compilers names?

We don't currently have a way for the devServer to take multiple settings objects and run them all in parallel. Are you suggesting that we create this feature, or are you suggesting that we run multiple devServers on different ports and pass each individual devServer options in?

rishabh3112: > I agree with the name idea. Maybe it should also take a number index for the compiler being used, in case the user has not given their compilers names?

I think we should enforce `name` only for `name` flag instead of indices here as they are bound to change if we change order of compilers in subsequent calls. 

> Are you suggesting that we run multiple devServers on different ports and pass each individual devServer options in? I think it would make sense to throw an error if they don't specify unique ports for each config.

For a particular `devServer`, we assign it a port if it is there else look for a free port. Throwing errors in this case can be avoided leading to better UX.

P.S.
A possible issue (on CLI side) in this case would handling logging in CLI such that we know which compiler's devServer is logging on `stdout` of terminal, ( compiler `name` would be helpful here ).

Loonride: Ok, I switched `name` to not act as an index.

I'm just concerned with ports that if there is a lot of CLI output, it will be difficult for the user to even figure out which ports the server is running on when ports are assigned automatically

packages/serve/src/types.ts
```diff
@@ -0,0 +1,5 @@
+export type devServerOptionsType = {
```
Loonride, line 1: It would be ideal to have a full dev server config type once this is worked on: https://github.com/webpack/webpack-cli/issues/866 and once we have finished major dev server options changes

We already have it here but needs to be updated, and I don't want to be importing the `generators` package just for this https://github.com/webpack/webpack-cli/blob/next/packages/generators/src/types/index.ts

packages/serve/src/getDevServerOptions.ts
rishabh3112: > I'm just concerned with ports that if there is a lot of CLI output, it will be difficult for the user to even figure out which ports the server is running on when ports are assigned automatically

Yeah, this would be problem if `silent`/`quiet` is configured by a user (in other cases we can log this metadata to terminal).
Should we log this metadata (like port, name, etc) any way in that case as well 😅?

Regarding finding output of a particular compiler/port among a lot CLI output, we can configure logging using `name` of compiler as it's identifier and have output something like [concurrently](https://www.npmjs.com/package/concurrently).

packages/serve/src/types.ts
rishabh3112: @Loonride friendly ping

Loonride: Sorry, need to focus on webpack-dev-server v4 first, will get back to this

gabriel-peracio:
This is a bit of a blocker for my workflow, is there anything I can do to help move this PR along?

Loonride:
My apologies, if anyone desperately needs this done feel free to take it over, otherwise I will look at it again in a few days

snitin315:
I will take this up.

evilebottnawi:
@Loonride Ready for review?

Loonride:
> @Loonride Ready for review?

@evilebottnawi Not quite yet, still adding tests

Loonride:
/cc @evilebottnawi Ready for review

Basically what I have done is:
- broken everything into helper functions for easier testing, and added full testing
- make multiple flags and a String or Boolean flag work correctly, as needed here: https://github.com/webpack/webpack-dev-server/pull/2722
- update tests for parse-args of webpack-cli

evilebottnawi (commented):
Anyway, it looks good

rishabh3112 (approved):

evilebottnawi (approved):

snitin315 (approved):

anshumanv (approved):

jamesgeorge007 (approved):

snitin315 (approved):

PR #1667, opened 2020-07-04
feat: add scripts to cleanup stale test data

anshumanv:
feat/cleanup/test not sure

Yes

NA

**Summary**
- We're manually cleaning up data in multiple tests, this will take care of it at once after all tests run.

No

**Other information**

### DIFF
--- a/scripts/utils.js
+++ b/scripts/utils.js
@@ -25,7 +25,4 @@ function extractFolder(folderToRead, folders = [], folderStrategy) {
     return folders;
 }
 
-module.exports = {
-    collectTestFolders,
-    extractFolder,
-};
+module.exports = collectTestFolders;
--- a/scripts/utils.js
+++ b/scripts/utils.js
@@ -1,0 +1,31 @@
+const fs = require('fs');
+const path = require('path');
+
+const BASE_DIR = 'test/';
+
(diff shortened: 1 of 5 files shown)


scripts/utils.js
```diff
@@ -0,0 +1,31 @@
...
+
+const BASE_DIR = 'test/';
+
+function collectTestFolders(strategy) {
+    const testFolder = path.resolve(path.join(process.cwd(), BASE_DIR));
+
+    return extractFolder(testFolder, [], strategy);
+}
+
+function extractFolder(folderToRead, folders = [], folderStrategy) {
```
jamesgeorge007, line 12: `extractFolder()` is supposed to be invoked via `collectTestFolders()`. It would be better to have a default export - `collectTestFolders()` (`extracFolder()` is not consumed elsewhere).

jamesgeorge007, line 12: `extractFolder()` is supposed to be invoked via `collectTestFolders()`. It would be better to have a default export - `collectTestFolders()` (`extractFolder()` is not consumed elsewhere).

anshumanv: in near future

anshumanv:
I don't think we need any package level removals, will take care of them in a while

PR #1668, opened 2020-07-04
feat(webpack-cli): allow negative property for cli-flags

### LINKED ISSUE #1240: Negated flags

snitin315:
Feature

Yes
NA
**Summary**
allow cli-flags to have a `negative` property. if `negative: true` for `flag-name` then `--no-flag-name` is valid.
for #1630 

Resolve #1240 

**Other information**


snitin315:
Yes, but they will be no longer available in help output as well. Should I remove then ?

anshumanv:
Can't we filter the flags which are negative and provide a list to show up in help?

snitin315:
Yes we can list all the negative flags together in help output.

anshumanv:
Not sure if bunching them all would be ideal, let's see

/cc @webpack/cli-team need small discussion

jamesgeorge007:
I think a better approach would be to not register the negated entries listed in `cli-flags`
https://github.com/webpack/webpack-cli/blob/62d77da8eaaa0e4cea4093d58d38dd24ae272a63/packages/webpack-cli/lib/utils/arg-parser.js#L62

```js
        if ((option.type === Boolean || option.type === String) && !option.name.startsWith('no-')) {
```

IMHO removing those entries from `cli-flags` and writing custom logic to show up `negated flags` as part of the help information would be too much of a work as compared to this tweak.

evilebottnawi:
@jamesgeorge007 We are still show negative flags in help, it is just sugar

jamesgeorge007:
@evilebottnawi the workaround that I suggested above won't affect the help information. Instead, it makes sure that `negated flags` aren't registered for the second time.

packages/webpack-cli/lib/utils/arg-parser.js
```diff
@@ -63,6 +63,11 @@ function argParser(options, args, argsOnly = false, name = '', helpFunction = un
             if (!option.multiple) {
                 // Prevent default behavior for standalone options
                 parserInstance.option(flagsWithType, option.description, option.defaultValue).action(() => {});
+                if (option.negative) {
+                    // commander requires explicitly adding the negated version of boolean flags
+                    const negatedFlag = `--no-${option.name}`;
+                    parserInstance.option(negatedFlag, `negates ${option.name}`);
+                }
```
jamesgeorge007, line 70: Add the check here so that `--no-stats` is recognized.
https://github.com/webpack/webpack-cli/blob/7489e639d13e8b89690a50595eb48214e9cdb1d9/packages/webpack-cli/lib/utils/arg-parser.js#L74-L76

jamesgeorge007:
Both the strategies seem to be fine, I don't see any harm in having negated flags listed within `cli-flags` considering the fact that they are standalone. But anyway if we're gonna remove it would be better to have them in a separate section as part of the help information:-

```bash
Negated Flags

  --no-hot      negates hot
  --no-stats    negates stats

```

snitin315:
Let's list down negated flags in help output together, we can always improve help output later if needed or as per feedback.

evilebottnawi:
@jamesgeorge007 I agree with @snitin315 , we can improve it late, you can be champion :1st_place_medal: Maybe we should not list all the negative flags. We can describe the fact that all boolean flags can be negative and provide small example

jamesgeorge007:
> @jamesgeorge007 I agree with @snitin315 , we can improve it late, you can be champion :1st_place_medal: Maybe we should not list all the negative flags. We can describe the fact that all boolean flags can be negative and provide small example

Sounds good 👍

snitin315:
Screenshot at 2020-07-06 20-36-16

packages/webpack-cli/lib/utils/arg-parser.js
jamesgeorge007: `--no-stats` is not recognized with the changes made as part of afadcd1

snitin315: fixed

packages/webpack-cli/lib/groups/HelpGroup.js
```diff
@@ -84,6 +84,7 @@ class HelpGroup {
         const { underline, bold } = chalk.white;
         const o = (s) => chalk.keyword('orange')(s);
         const options = require('../utils/cli-flags');
+        const negatedFlags = options.core.filter((flag) => flag.negative === true);
```
jamesgeorge007, line 87:
```suggestion
        const negatedFlags = options.core.filter((flag) => flag.negative);
```

jamesgeorge007: I don't think we need an explicit check here.

packages/webpack-cli/lib/utils/arg-parser.js
```diff
@@ -63,6 +63,12 @@ function argParser(options, args, argsOnly = false, name = '', helpFunction = un
             if (!option.multiple) {
                 // Prevent default behavior for standalone options
                 parserInstance.option(flagsWithType, option.description, option.defaultValue).action(() => {});
+
+                if (option.negative) {
+                    // commander requires explicitly adding the negated version of boolean flags
+                    const negatedFlag = `--no-${option.name}`;
+                    parserInstance.option(negatedFlag, `negates ${option.name}`).action(() => {});
+                }
```
jamesgeorge007, line 71: Since this check is not limited to any flags of a certain kind why not moving it outside? That way we can also remove the redundancy.

packages/webpack-cli/lib/groups/HelpGroup.js
```diff
@@ -115,6 +116,16 @@ class HelpGroup {
...
             },
+            {
+                header: 'Negated Flags',
+                optionList: negatedFlags.map((e) => {
+                    e.description = `Negates ${e.name}`;
+                    e.name = `no-${e.name}`;
+                    e.alias = null;
+                    e.type = Boolean;
+                    return e;
+                }),
```
jamesgeorge007, line 127: I think it would be better to have It in the first place.

```js
const negatedFlags = options.core
            .filter((flag) => flag.negative)
            .reduce((allFlags, flag) => {
                return [...allFlags, { name: `no-${flag.name}`, description: `Negates ${flag.name}`, type: Boolean }];
            }, [])
            .map(({ alias, ...rest }) => rest); // remove alias or set it to null
```

packages/webpack-cli/lib/utils/arg-parser.js
snitin315: done

packages/webpack-cli/lib/groups/HelpGroup.js
```diff
@@ -115,6 +120,10 @@ class HelpGroup {
                     return e;
                 }),
             },
+            {
+                header: 'Negated Flags',
+                optionList: negatedFlags.map((e) => e),
```
jamesgeorge007, line 125:
```suggestion
                optionList: negatedFlags,
```

packages/webpack-cli/lib/groups/HelpGroup.js
```diff
@@ -115,6 +120,10 @@ class HelpGroup {
                     return e;
                 }),
             },
+            {
```
rishabh3112, line 123: I think we should have negated flags with `Options` only, as to the user they are flags only.

rishabh3112: To be clear, spreading these with `optionList` of `Options` header is what I am suggesting here.

snitin315: Screenshot at 2020-07-07 11-16-07

snitin315: like this ?

rishabh3112: Yes

PR #1674, opened 2020-07-08
fix: warn about merge config resolution cases

anshumanv:

fix

Yes

NA

**Summary**
- Right now if merge config is absent, no feedback is given to user when it's absent or when we fallback to default merge config, fix it

No

**Other information**


anshumanv:
> Maybe we should not fallback to webpack.base.js here. Just throw an error what we don't find the configuration, because it can be misleading for developers too

Agree, but .base is usually used to keep shared config for prod and dev env, which is eventually merged with the respective config hence this fallback was added in past I think, we're already notifying user in case we fallback, what do you think? @evilebottnawi

evilebottnawi:
@anshumanv Do we have this logic for v3?

anshumanv:
> @anshumanv Do we have this logic for v3?

No, we don't. Should we remove fallback?

rishabh3112:
I think we shouldn't fallback as well, Developer never intends to use `.base` config as a standalone configuration.

evilebottnawi:
@anshumanv Let's discuss about it. What configuration files do we have and how are they inherited and them priorities? Maybe we can do simple drawing/picture?

anshumanv:
> Developer never intends to use .base config as a standalone configuration.

Using as standalone is out of the picture

anshumanv:
> @anshumanv Let's discuss about it. What configuration files do we have and how are they inherited and them priorities? Maybe we can do simple drawing/picture?

Yep we can, I'll come up with it in a while 👍

anshumanv:
```js
[
    'webpack.config',
    'webpack.config.dev',
    'webpack.config.development',
    'webpack.config.prod',
    'webpack.config.production',
    '.webpack/webpack.config',
    '.webpack/webpack.config.none',
    '.webpack/webpack.config.dev',
    '.webpack/webpack.config.development',
    '.webpack/webpack.config.prod',
    '.webpack/webpack.config.production',
    '.webpack/webpackfile',
]
```

@evilebottnawi this the order of priority of files, in increasing order of priority.

eg if `webpack.config.*` is present and `.webpack/webpackfile.*` are present, then `webpackfile` will be resolved during config lookup as it has max priority while `webpack.config.*` has least priority in this list

We do not resolve for `.base` while config lookup, hence I think that was added as a fallback file for merge strategy when the supplied config in the merge flag doesn't exist.

anshumanv:
Also, should we document it somewhere for future reference, it's present in code only for now.

anshumanv:
```
I think we should change logic for this, it is the interesting case, when I use

webpack --config' './1.js', '--config' './2.js
I would think that we want to run both configurations, like multi compiler mode.

But, if i write

webpack --merge --config' './1.js', '--config' './2.js
I want to merge configurations.

I think this logic is simpler and clearer.

If configuration doesn't exists in --config we should always throw an error.

Also provide --config options disable auto resolving configurations, because I have clearly indicated that I want to use.

And yes we should documented configurations and their priority
```

@evilebottnawi that makes sense, there are a several steps involved to achieve that, let's do them bit by bit.

I'll start by removing the fallback to base config from here, just the error message that the merge config is not found should be good for now.

anshumanv:
@evilebottnawi are there any cases when the user only wants to merge to one config and use another as is?

Eg - `webpack --config webpack1.config.js --config webpack2.config.js --merge webpack.base.js`

In this case the user only intends to merge base config with `webpack2.config.js`, are such cases possible?

evilebottnawi:
> Eg - webpack --config webpack1.config.js --config webpack2.config.js --merge webpack.base.js

I don't think we should support is, because it is misleading, if you want to more three configuration, you can use 
```
webpack --config webpack1.config.js --config webpack2.config.js --config webpack.base.js --merge
```

If you want to run 3 configuration:

```
webpack --config webpack1.config.js --config webpack2.config.js --config webpack.base.js --merge
```

If you want to merge first and second configurations, use code for this, no need handle very complex logic

anshumanv:
> If you want to merge first and second configurations, use code for this, no need handle very complex logic

Seems fair, I've dropped the support for fallback to default config in merge, will add support for multi config after this. 👍

packages/webpack-cli/lib/groups/ConfigGroup.js
```diff
@@ -167,22 +169,26 @@ class ConfigGroup extends GroupHelper {
...
-                const foundConfig = configFiles[0];
-                const resolvedConfig = this.requireConfig(foundConfig);
-                const newConfigurationsObject = await this.finalize(resolvedConfig);
-                const webpackMerge = require('webpack-merge');
-                this.opts['options'] = webpackMerge(this.opts['options'], newConfigurationsObject.options);
+            // Try to resolve the given merge config else fallback to `webpack.base.js`
+            const newConfigPath = this.resolveFilePath(merge);
+
+            if (!newConfigPath) {
+                return logger.warn("The supplied merge config doesn't exist.");
```
evilebottnawi, line 176: It should be error, not logging

packages/webpack-cli/lib/groups/ConfigGroup.js
```diff
@@ -167,22 +169,26 @@ class ConfigGroup extends GroupHelper {
...
+                this.opts.processingMessageBuffer.push({
+                    lvl: 'warn',
+                    msg: 'Could not find file to merge configuration with...',
+                });
+                return;
             }
+            const foundConfig = configFiles[0];
+            const resolvedConfig = this.requireConfig(foundConfig);
+            const newConfigurationsObject = await this.finalize(resolvedConfig);
+            const webpackMerge = require('webpack-merge');
```
evilebottnawi, line 190: Move on top of file

anshumanv: But then we'll be loading it when we might not even need it, is that fine?

packages/webpack-cli/lib/groups/ConfigGroup.js
anshumanv: Right, exiting the process

packages/webpack-cli/lib/groups/ConfigGroup.js
anshumanv: moved to top

evilebottnawi: Fine, v8 cache` require`

packages/webpack-cli/lib/groups/ConfigGroup.js
```diff
@@ -167,22 +170,26 @@ class ConfigGroup extends GroupHelper {
...
+                process.exit(1);
+            }
+
+            const configFiles = getConfigInfoFromFileName(newConfigPath);
+            if (!configFiles.length) {
+                this.opts.processingMessageBuffer.push({
+                    lvl: 'warn',
+                    msg: 'Could not find file to merge configuration with...',
+                });
+                return;
```
evilebottnawi, line 187: Can you clarify? What means `configFiles.length`? Maybe we need an error here too?

anshumanv: Good catch, this is redundant, already throwing above.

packages/webpack-cli/lib/groups/ConfigGroup.js
```diff
@@ -167,22 +170,19 @@ class ConfigGroup extends GroupHelper {
...
-                const foundConfig = configFiles[0];
-                const resolvedConfig = this.requireConfig(foundConfig);
-                const newConfigurationsObject = await this.finalize(resolvedConfig);
-                const webpackMerge = require('webpack-merge');
-                this.opts['options'] = webpackMerge(this.opts['options'], newConfigurationsObject.options);
+            // try resolving merge config
+            const newConfigPath = this.resolveFilePath(merge);
+
+            if (!newConfigPath) {
+                logger.error("The supplied merge config doesn't exist.");
```
evilebottnawi, line 177: We don't need logging error in this case, because it is duplicate, you already see the error in console

test/merge/config-absent/merge-config-absent.test.js
```diff
@@ -0,0 +1,22 @@
...
+        const { stdout, stderr } = run(__dirname, ['--config', './1.js', '--merge', './2.js'], false);
+
+        // Since the process will exit, nothing on stdout
+        expect(stdout).toBeFalsy();
+        // Confirm that the user is notified
+        expect(stderr).toContain(`[webpack-cli] The supplied merge config doesn't exist.`);
+        // Default config would be used
+        expect(fs.existsSync(join(__dirname, './dist/merged.js'))).toBeFalsy();
+        // Since the process will exit so no compilation will be done
+        expect(fs.existsSync(join(__dirname, './dist/main.js'))).toBeFalsy();
```
evilebottnawi, line 20: Please test `code` on `-1`

test/merge/config-absent/merge-config-absent.test.js
```diff
@@ -0,0 +1,22 @@
...
+        const { stdout, stderr } = run(__dirname, ['--config', './1.js', '--merge', './2.js'], false);
+
+        // Since the process will exit, nothing on stdout
+        expect(stdout).toBeFalsy();
+        // Confirm that the user is notified
+        expect(stderr).toContain(`MergeError: The supplied merge config doesn't exist.`);
+        // Default config would be used
+        expect(fs.existsSync(join(__dirname, './dist/merged.js'))).toBeFalsy();
+        // Since the process will exit so no compilation will be done
+        expect(fs.existsSync(join(__dirname, './dist/main.js'))).toBeFalsy();
```
evilebottnawi, line 20: Can we add test there we have the same options in both configuration, to ensure we override it from the second configuration?

anshumanv: Added 👍

evilebottnawi: Can't find this test :confused: What is overridden?

anshumanv: Added in this commit - https://github.com/webpack/webpack-cli/pull/1674/commits/9b34377c562e60eeb6e59d7d4a6500fefb234746

test/merge/config/merge-config.test.js
```diff
@@ -15,4 +15,13 @@ describe('merge flag configuration', () => {
...
     });
+    it('merges two configurations together with flag alias', (done) => {
+        const { stdout } = run(__dirname, ['--config', './1.js', '-m', './2.js'], false);
+        expect(stdout).toContain('merged.js');
+        stat(resolve(__dirname, './dist/merged.js'), (err, stats) => {
+            expect(err).toBe(null);
+            expect(stats.isFile()).toBe(true);
+            done();
+        });
+    });
```
evilebottnawi, line 26: Let's override the other option, for example `filename`

anshumanv: added https://github.com/webpack/webpack-cli/pull/1674/commits/1fdf2ae68039941fd47496bfc086fb5be77146a0 👍

Issue #1261, opened 2020-02-24
Make CLI more empathic to the user


rishabh3112:
**Is your feature request related to a problem? Please describe.**
Currently CLI gives error or gives whole help in case of invalid interaction with the CLI.
I propose to make the CLI respond to errors and invalid flags with prompts to help user to get their work done instead of forcing user to study errors and correct their command to CLI.

**Describe the solution you'd like**
I would like user to answer prompts in case of invalid value of a flag or invalid flag altogether to let user complete it's task.

I have started working on this with #1255 which is aimed for invalid flags.
For further work on this issue we need to validate the value of args passed by user before passing their value to the webpack complier so as to avoid errors.

**Describe alternatives you've considered**
Alternatives are already in place, i.e. printing full help and showing just errors.
 

**Additional context**
https://github.com/lirantal/nodejs-cli-apps-best-practices/blob/master/README.md#12-build-empathic-clis

aquibbaig:
Hey @rishabh3112. Any progress on this yet?

rishabh3112:
Hi, I am currently working on polishing the handling of the invalid flag case.

aquibbaig:
@rishabh3112 Cool.

snitin315:
validated mode argument in #1290

aman29271:
I really like this git features, when user gives some invalid command then it shows the list of possible  commands that the user may want to use.
For example : 
```bash
~/Desktop/webpack-cli$ git statu
git: 'statu' is not a git command. See 'git --help'.

The most similar commands are
        status
        stage
        stash
```
It really helps users understanding the CLI in easy way. This could be one of the way to handle invalid commands and flags.

snitin315:
@rishabh3112 I read the docs above and found that now we have covered major of the things in recent months like -

1. colorful CLI.
2. Zero configuration.
3. Appropriate and informational errors and warnings are in place.
4. Better error handling in case of invalid args.
5. compatibility with webpack 4 and webpack 5.
6. compatible with different versions of Node.js and different platforms (Windows, macOS, Linux )
7. proper use of exit codes.
8. better integration with the webpack ecosystem. 
9. more tests.

anything else you want to specify.

rishabh3112:
There is definitely an improvement but there are couple of this still left to be addressed:
- We need to give users information how to resolve a particular error. Like, when we throw Errors, we should also provide remedial information on how can we recover from this error.
[...]

PR #1675, opened 2020-07-08
feat(webpack-cli): validate --target flag

snitin315:
Feature

yes
NA
**Summary**
Refers #1261 
Added `target` option validation. warn the user and set `target: 'web'` for invalid values. 

No

**Other information**

### DIFF
--- a/packages/webpack-cli/lib/groups/AdvancedGroup.js
+++ b/packages/webpack-cli/lib/groups/AdvancedGroup.js
@@ -1,4 +1,6 @@
 const GroupHelper = require('../utils/GroupHelper');
+const logger = require('../utils/logger');
+const { yellow } = require('chalk');
 
 class AdvancedGroup extends GroupHelper {
     constructor(options) {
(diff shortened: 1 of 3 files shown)


packages/webpack-cli/lib/groups/AdvancedGroup.js
```diff
@@ -29,7 +31,22 @@ class AdvancedGroup extends GroupHelper {
...
+                'async-node',
+                'node-webkit',
+                'electron-main',
+                'electron-renderer',
+                'electron-preload',
+            ];
+            if (validTargets.includes(args.target)) {
+                options.target = args.target;
+            } else {
+                logger.warn(yellow(`"${args.target}" is an invalid value for "target" option. Using "web" by default.`));
```
anshumanv, line 47: Yep we can't assume about the platform for which they;re bundling, throwing err and displaying the possible values would be ideal

snitin315: makes sense :+1:

snitin315:
throw error and exit the process ?

anshumanv:
> throw error and exit the process ?

Yes

evilebottnawi:
Webpack already throw an error, we should add tests that we exit with same code too

snitin315:
My take was to validate target before passing it's value to compiler , i.e, passing only valid values to compiler.

evilebottnawi:
@snitin315 Webpack already validate options, we don't need to do double job, let's add tests on exit code and errors

evilebottnawi:
I think we can close it, we should not validate options, because it is webpack job

anshumanv:
yep core already throws err we can close

PR #1681, opened 2020-07-16
feat: make errors and warnings colorful

snitin315:
enhancement.

no
no
**Summary**
More colorful errors and warnings.

no

**Other information**

### DIFF
--- a/packages/webpack-cli/lib/groups/StatsGroup.js
+++ b/packages/webpack-cli/lib/groups/StatsGroup.js
@@ -1,3 +1,4 @@
+const { yellow } = require('chalk');
 const GroupHelper = require('../utils/GroupHelper');
 const logger = require('../utils/logger');
 /**
@@ -14,13 +15,13 @@ class StatsGroup extends GroupHelper {
 
     resolveOptions() {
         if (this.args.verbose && this.args.stats) {
-            logger.warn('Conflict between "verbose" and "stats" options. Using verbose.');
+            logger.warn(yellow('Conflict between "verbose" and "stats" options. Using verbose.'));
             this.opts.options.stats = 'verbose';
         } else {
             if (this.args.verbose) {
                 this.opts.options.stats = 'verbose';
             } else if (this.args.stats && !StatsGroup.validOptions().includes(this.args.stats)) {
-                logger.warn(`'${this.args.stats}' is invalid value for stats. Using 'normal' option for stats`);
+                logger.warn(yellow(`'${this.args.stats}' is invalid value for stats. Using 'normal' option for stats`));
                 this.opts.options.stats = 'normal';
             } else {
                 this.opts.options.stats = this.args.stats;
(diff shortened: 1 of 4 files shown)


packages/webpack-cli/lib/groups/StatsGroup.js
```diff
@@ -14,13 +15,13 @@ class StatsGroup extends GroupHelper {
 
     resolveOptions() {
         if (this.args.verbose && this.args.stats) {
-            logger.warn('Conflict between "verbose" and "stats" options. Using verbose.');
+            logger.warn(yellow('Conflict between "verbose" and "stats" options. Using verbose.'));
```
evilebottnawi, line 18: I think wepback logger should colors errors and warnings out of box, we don't need do this

Also `stats` has `color` option, what we should respect

snitin315: > I think wepback logger should colors errors and warnings out of box, we don't need do this

I think the same, should I create PR at the core ?

> Also stats has the color option, what we should respect

I think we should enable color output by default. Users can disable output if needed with `--no-stats-colors`.

evilebottnawi: @snitin315 

> I think the same, should I create PR at the core ?

yes, it is good idea

> I think we should enable color output by default. Users can disable output if needed with --no-stats-colors.

I think we can use `https://github.com/jorgebucaran/colorette` instead of `chalk`, it is faster, less and has automatic color support detection

snitin315: https://github.com/jorgebucaran/colorette looks good to me, I think we can use it. 
/cc @webpack/cli-team

rishabh3112: `chalk` is already a dependency in `webpack-cli`. For the scope of this PR I would recommend either to use chalk only, or first migrate the chalk to `colorette` and then continue with this PR.

evilebottnawi: > first migrate the chalk to colorette

Yes think it is better to first

snitin315: We have now migrated to `colorette` and the runtime logger we are using is meant for browsers and not for Node.js. So we have to do it like this `logger.warn(yellow('warning'))` until webpack exposes a runtime logger API for Node. 

Refer https://github.com/webpack/webpack/issues/11215#issuecomment-667217564

evilebottnawi: Yes, let's use `console` for logging while webpack doesn't implement node logger

evilebottnawi:
@snitin315 Should we close it in favor https://github.com/webpack/webpack-cli/pull/1740?

snitin315:
Yes.

PR #1687, opened 2020-07-17
docs: add --env type

snitin315:
Docs

NA
yes
**Summary**
added type `string` of `--env` flag

no

**Other information**


packages/webpack-cli/README.md
```diff
@@ -39,7 +39,7 @@ Options
   -c, --config string      Provide path to a webpack configuration file
   -m, --merge string       Merge a configuration file using webpack-merge
   --progress               Print compilation progress during build
-  --env                    Environment passed to the configuration when it is a function
+  --env string             Environment passed to the configuration when it is a function
```
anshumanv, line 42: We should maybe also mention that only values are `production`, 'development', 'none'

snitin315: I think we can add that info in help with `webpack --env --help`. Like we have for `--mode`

anshumanv: Yeah we can do that, and in case the env is invalid we can also show the `env help`

anshumanv:
@snitin315 are you gonna make the changes here?

packages/webpack-cli/README.md
snitin315: I see we can pass `webpack --env invalid` and it doesn't throw a warning or something else to let the user know it's invalid. We should implement this. showing `env help` seems good to me.

snitin315:
> @snitin315 are you gonna make the changes here?

No, in future PR.

packages/webpack-cli/README.md
anshumanv: Sure

snitin315: @anshumanv maybe you want to pick this up? I'm caught up with some other issues and PRs :smile:

anshumanv: Cool np, focus on core 👍 
I'll PR by tomm

anshumanv: Ah my bad I confused env with mode, env can accept any string so we're good here 👍

Issue #1222, opened 2020-02-12
webpack-cli v4 beta feedback and changes


ematipico:
# About v4

With webpack-cli `v4` we will support only webpack `v5` and possibly `v4` but not officially, so if there are any problems we suggest to use `v3.*` if you still want to keep the best compatibility.

We dropped Node.js 8 from our CI. We support at least Node.js 10.

# What's changing

## Less arguments

We drastically reduced the number of arguments that we support. This is because we think most of them are not used by anyone (or used inside a configuration file). Run

```bash
webpack --help
```

To check the arguments supported.

**If you think that we wrongly removed an argument that you thought it was useful, please leave a comment here and we can discuss its implementation. (add a 👍 if someone already requested an argument you need)**

## `.webpack` folder support

The CLI now supports the `.webpack` folder by default. This feature is still in beta so it would be great if people could start using it. Files that are automatically imported:
- `.webpack/webpack.config.dev`
- `.webpack/webpack.config.prod`
 - `.webpack/webpackfile`
 - `webpack.config`

## Changed default output

We wanted to change a bit the output emitted via CLI. 

```bash
✅ Compilation Results

🌏 Version: 5.0.0-beta.12
⚒️  Built: Wed Feb 12 2020 09:05:53 GMT+0100 (Central European Standard Time)
⏱  Compile Time: 265ms
📂 Output Directory: /Users/***/www/webpack-cli/test/entry/defaults-index/dist
#️⃣ Hash: 0b145a012a5ce8805b14

┌────────────┬───────────────────────────────┐
│ Entrypoint │ Bundle                        │
├────────────┼───────────────────────────────┤
│ main       │  Bundle Name       │ main.js  │
│            │  Compared For Emit │ true     │
│            │  Bundle size       │ 0.00 kb  │
│            │                               │
│            │ Modules:                      │
│            │ ./index.js [size: 1 bytes]    │
└────────────┴───────────────────────────────┘
```

Don't worry, you can still go to the old output using the `--standard` flag.

## Arguments `--dev` and `--prod`

Now you can pass the mode to the compiler using these two simple arguments.

## Help for single arguments

Running the following command

```bash
webpack help --dev
```

Will output

```bash
Usage: webpack --dev
Description: Run development build
Documentation: https://webpack.js.org/concepts/#mode
```

tfritsch-km:
Hello,

[...]

PR #1688, opened 2020-07-17
fix: set mode=production by default

### LINKED ISSUE #1678: Webpack default mode not set to production
# Bug report

As stated in https://webpack.js.org/configuration/mode/, "If not set, webpack sets production as the default value for mode."

`argv` does not contain `mode` property set to `'production'`, when invoking `webpack` without any mode set.

file webpack.config.js:
``` 
module.exports = (env, argv) => { console.log(argv); }
```
now run `webpack`, without setting any mode, and see that the `argv` object logged does not contain `mode` property set to `'production'`.

`argv` object should have `mode` property set to `'production'`.

**Other relevant information:**
webpack version:  4.43.0
Node.js version: 14.3.0
Operating System: Mac OSX 10.15.5
Additional tools: None

snitin315:
bugfix

yes
NA
**Summary**
Resolve #1678 

no

**Other information**


packages/webpack-cli/lib/utils/cli-flags.js
```diff
@@ -201,7 +201,7 @@ module.exports = {
             alias: 'p',
             usage: '--prod',
             type: Boolean,
-            defaultValue: undefined,
+            defaultValue: false,
```
evilebottnawi, line 204: I've always wondered why we have flags `--dev`/`--prod`, we already have `--mode`, it can be misleading for new developers and some of them will be write `webpack --prod --mode=production`, but it is wrong

snitin315: > I've always wondered why we have flags --dev/--prod, we already have --mode, it can be misleading for new developers and some of them will be write webpack --prod --mode=production, but it is wrong

It's just alias for `--mode=development or production` like we have `--verbose` for `--stats=verbose`. 
CLI throws a warning if user do something like `webpack --prod --mode=production` 
`You provided both 'mode' and --dev arguments. You should provide just one. "mode" will be used`

evilebottnawi: I think we should remove them in the future, but it is not high priority

snitin315: /cc @webpack/cli-team anyone in favor of `--dev` and `--prod`.? 
> Not a big deal but even if I understand the point of having clean and concise --dev and --prod options, I think the --mode= option better matches the way it works in configuration file and is more straightforward when you already know configuration file. Just my 2 cents, but I'm not sure a breaking change in this area is really useful ?

from feedback too https://github.com/webpack/webpack-cli/issues/1222#issuecomment-585678755

PR #1691, opened 2020-07-18
chore: migrate to colorette

snitin315:
chore

yes
NA
**Summary**
Migrate to `colorette `

No

**Other information**


test/info/info-help.test.js
```diff
@@ -19,10 +19,11 @@ describe('should print help for info command', () => {
 
     it('should respect the --color=false flag', () => {
         const { stdout, stderr } = runInfo(['help', '--color=false'], __dirname);
-        chalk.enabled = true;
-        chalk.level = 3;
-        const orange = chalk.keyword('orange');
-        expect(stdout).not.toContain(orange(usageText));
+        //unusual behavior for windows CI
+        if (process.platform !== 'win32') {
```
rishabh3112, line 23: We should respect `--color=false` when explicitly stated in windows too.
What issue is being faced here?

snitin315: I think `--color=false` is not working for windows.
Screenshot at 2020-07-19 19-55-47

Unfortunately, I don't have a windows machine to test it locally :disappointed:

snitin315: Can you provide some additional info on how this `--color` flag works? I see it isn't defined anywhere?

rishabh3112: > I think --color=false is not working for windows.

I was asking why it isn't working :). Now we know it was removed somehow in some changes.
Reason I think it was not caught till now is `chalk` seems to be respecting [`NODE_DISABLE_COLOR`](https://nodejs.org/dist/latest-v12.x/docs/api/cli.html#cli_node_disable_colors_1) supplied by CI's environment and `colorette` is not.

> Can you provide some additional info on how this --color flag works? I see it isn't defined anywhere?

I wasn't aware it was removed from `cli-flags.js`. Here is the commit tree and flag description where you would find what it is supposed to do 😅️
https://github.com/webpack/webpack-cli/blob/530b4f76a45d7eb20648e9ee2981cbf4a707854d/lib/utils/cli-flags.js#L84:L89

snitin315: Dont know how but `webpack --help --color=false` works as expected in my local env.

snitin315: We don't have the flag better we should remove the tests too.

rishabh3112: > We don't have the flag better we should remove the tests too.

+1 from my side. /cc @webpack/cli-team for more opinions.

anshumanv: +1 for remove, we can already tune stats color from core flag, no need here.
/cc @evilebottnawi

evilebottnawi: Just clarify? On window it is not work out of box?

test/info/info-help.test.js
```diff
@@ -17,16 +16,6 @@ describe('should print help for info command', () => {
...
 
-    it('should respect the --color=false flag', () => {
-        const { stdout, stderr } = runInfo(['help', '--color=false'], __dirname);
-        chalk.enabled = true;
-        chalk.level = 3;
-        const orange = chalk.keyword('orange');
-        expect(stdout).not.toContain(orange(usageText));
-        expect(stdout).toContain(descriptionText);
-        expect(stderr).toHaveLength(0);
-    });
```
evilebottnawi, line 28: I think we should not release it, due v4 is not supported core flags, so webpack@4 developers can't disable colors using CLI, some developers disable it only for CI and etc, and they probably don't want to do it in config

snitin315: I think we should be able to disable colors with `colorette` too, https://github.com/jorgebucaran/colorette#optionsenabled

evilebottnawi: Can you clarify?

evilebottnawi: What is the problem, in this PR nothing was changed, only migration

snitin315: We have test with `--color=false`, but we don't have this flag defined in `cli-flag.js`.

snitin315: It was removed in some commit.

evilebottnawi: Let's do migration here and discussion about `--color` in the other PR, now keep logic as is

evilebottnawi:
Good job, let's fix tests

evilebottnawi:
Something with CI :disappointed:

snitin315:
I found the catch with the `--color` flag. For v3 we were using a dep [supports-color](https://github.com/chalk/supports-color) which gives us the ability to Enable/Disable colors with chalk by use of `--color`.
https://github.com/webpack/webpack-cli/blob/33574ec82e1f6d8b71e872db086bbfaaf77707ba/package.json#L124

snitin315:
But for v4 `--color` flag was removed in https://github.com/webpack/webpack-cli/commit/b4cd29fbad26e291687e90d10b3087331193b9f4 , but the corresponding tests were not removed.

evilebottnawi:
@snitin315 can you rewrite it?

evilebottnawi:
Need fix CI

snitin315:
CI FIxed.

Now we support `--color` and `--no-color`

evilebottnawi:
Freeze on github actions, ignore

snitin315:
Good to merge?

yarn.lock
```diff
@@ -3949,6 +3949,11 @@ color-name@~1.1.4:
   resolved "https://registry.yarnpkg.com/color-name/-/color-name-1.1.4.tgz#c2a09a87acbde69543de6f63fa3995c826c536a2"
   integrity sha512-dOy+3AuW3a2wNbZHIuMZpTcgjGuLU/uBL/ubcZF9OXbDo8ff4O8yVp5Bf0efS8uEoYo5q4Fx7dY9OgQGXgAsQA==
 
+colorette@^1.2.1:
```
anshumanv, line 3952: There should've been diff to remove chalk too right?

anshumanv: No worries handled - https://github.com/webpack/webpack-cli/pull/1710

snitin315: Strange, it was auto generated.

anshumanv: probably you removed package from `package.json` directly instead of doing `yarn remove ..`

PR #1694, opened 2020-07-19
fix(serve): supplying help or version as an arg should throw

jamesgeorge007:
bugfix

yup

N/A

**Summary**
- `webpack-cli serve help` should throw an error.
- Added a test case to ensure that `serve` respects `--no-color` flag.

Nope

**Other information**
N/A

### DIFF
--- a/test/help/help-commands.test.js
+++ b/test/help/help-commands.test.js
@@ -4,6 +4,11 @@ const { run } = require('../utils/test-utils');
 const helpHeader = 'The build tool for modern web applications';
 
 describe('commands help', () => {
+    it('recognizes subcommands', () => {
+        const { stderr } = run(__dirname, ['help', 'serve'], false);
+        expect(stderr).not.toContain('Unknown argument: help');
+    });
+
     it('throws error for invalid command with --help flag', () => {
         const { stderr } = run(__dirname, ['--help', 'myCommand'], false);
         expect(stderr).toContain(`You provided an invalid command 'myCommand'`);
--- a/test/help/help-commands.test.js
+++ b/test/help/help-commands.test.js
@@ -4,9 +4,15 @@ const { run } = require('../utils/test-utils');
 const helpHeader = 'The build tool for modern web applications';
 
 describe('commands help', () => {
-    it('recognizes subcommands', () => {
-        const { stderr } = run(__dirname, ['help', 'serve'], false);
-        expect(stderr).not.toContain('Unknown argument: help');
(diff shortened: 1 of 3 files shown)


jamesgeorge007:
> Seems to work for me, can you give a failing case?

`webpack-cli serve help`

anshumanv:
> webpack-cli serve help

It's a `command`, they already recognize args `--help`, not sure if we should allow user to write 2 commands.

/cc @webpack/cli-team

evilebottnawi:
@anshumanv do you mean `webpack server --help`?

anshumanv:
No @evilebottnawi, like `webpack serve help` -> 2 commands

We already support `webpack serve --help`

snitin315:
Actually we support two help commands `webpack --help` and `webpack help` both works the same. Same is the case with `version`

anshumanv:
> Actually we support two help commands webpack --help and webpack help both works the same. Same is the case with version

Different point, talking about whether we need 2 commands *simultaneously*

snitin315:
It's really confusing and misleading. I think we should only support `--help` & `--version`. 
Hence we won't have to support `webpack serve help`.
`webpack serve --help` is sufficient.

anshumanv:
> It's really confusing and misleading. I think we should only support --help & --version.
Hence we won't have to support webpack serve help.
webpack serve --help is sufficient.

+1 only creates confusion, /cc @evilebottnawi

evilebottnawi:
Good question, I don't have the strong opinion here, but `git` support
```
git help fetch
git fetch --help
```

anshumanv:
> Good question, I don't have the strong opinion here, but git supports

In this case `fetch` is the arg for the help command => only 1 command, we are supporting `serve help` to which is actually misleading

evilebottnawi:
Yes we should remove `webpack serve help` supporting

anshumanv:
@jamesgeorge007 can you update the PR to limit command count to 1?

jamesgeorge007:
> Yes we should remove `webpack serve help` supporting

Couldn't agree more.

test/help/help-commands.test.js
```diff
@@ -4,6 +4,11 @@ const { run } = require('../utils/test-utils');
 const helpHeader = 'The build tool for modern web applications';
 
 describe('commands help', () => {
+    it('throws error if supplied as an argument for subcommands', () => {
+        const { stderr } = run(__dirname, ['serve', 'help'], false);
```
anshumanv, line 8: test for `help serve` also needed if we wish to support

PR #1702, opened 2020-07-25
fix: remove silent flag

anshumanv:

chore

Yes

Yes

**Summary**

- Delet silent flag.
- Silent flag doesn't work now and flags from core can handle it.

Yes, this was present in v3

**Other information**

Ref - https://github.com/webpack/webpack-cli/pull/1325#issuecomment-660466342


anshumanv:
> How we can disable all output now?

this doesn't work right now anyways so, no way to do it right now, once core flags are in we can use `--no-stats`

snitin315:
> > How we can disable all output now?
> 
> this doesn't work right now anyways so, no way to do it right now, once core flags are in we can use `--no-stats`

We already support `--no-stats` + we can always do `--stats=none`, something I'm missing?

evilebottnawi:
Just not, here is other problem to disable all output you need to disable logging too

anshumanv:
> Just note, here is other problem to disable all output you need to disable logging too

Yes, we should rethink and intercept logging to not log anything

anshumanv:
For that I think we should first align all the logging, that we do.

In some places we're doing console.error, in some we're writing to process.stdout and in some places we're using the logger.

What do you suggest @evilebottnawi @snitin315?

snitin315:
I think we should use `logger` only.

evilebottnawi:
Agree, we should use `logger`, it is allow to disable logging on CLI side too

anshumanv:
> Agree, we should use logger, it is allow to disable logging on CLI side too

Will handle it 👍

PR #1703, opened 2020-07-25
chore: leverage arg-parser for subcommands

jamesgeorge007:
enhancement

Yup

Nope

**Summary**
Onboard validations to handle unknown args for subcommands.

Nope

**Other information**
N/A


packages/init/src/index.ts
```diff
@@ -15,11 +16,20 @@ const CONFIG_PREFIX = '--force';
...
-    if (packages.length === 0 || includesDefaultPrefix || generateConfig) {
+    const cli = new WebpackCLI();
+    const { opts, unknownArgs } = cli.argParser(options, args, true);
+    const { auto: includesDefaultPrefix, force: generateConfig } = opts;
+
+    // Filter out cli-flags from unknownArgs
+    const unknownFlags = unknownArgs.filter((arg) => arg.startsWith('--'));
+
+    if (unknownFlags.length > 0) {
+        logger.warn(`Unknown ${unknownFlags.length === 1 ? 'argument' : 'arguments'}: ${unknownFlags}`);
```
evilebottnawi, line 27: I think we should throw an error in this case and exit with code `-1`

jamesgeorge007: `serve` seems to follow a similar approach. Should we change it?

https://github.com/webpack/webpack-cli/blob/f1876c125be90e0b0ae96e6e9dbaf91ed447c779/packages/serve/src/index.ts#L27-L34

evilebottnawi: Yes, throw an error for all unknown arguments:
```
node_modules/.bin/eslint --unknown
```

I was wrong about exit codes:
- 0 - all fine
- 1 - errors in the compilation
- 2 - problems with configurations/cli flags/etc (we even don't run build)

packages/init/src/index.ts
```diff
@@ -15,11 +16,20 @@ const CONFIG_PREFIX = '--force';
  */
 
 export default function initializeInquirer(...args: string[]): Function | void {
-    const packages = args;
-    const includesDefaultPrefix = packages.includes(AUTO_PREFIX);
-    const generateConfig = packages.includes(CONFIG_PREFIX);
-    if (packages.length === 0 || includesDefaultPrefix || generateConfig) {
+    const cli = new WebpackCLI();
+    const { opts, unknownArgs } = cli.argParser(options, args, true);
```
evilebottnawi, line 20: Why do not fix it all problems here? It is very bad to handle this in the each package?

jamesgeorge007: Yeah, makes sense.

packages/init/src/index.ts
```diff
@@ -15,11 +16,17 @@ const CONFIG_PREFIX = '--force';
...
```
evilebottnawi, line 20: Why do not throw error inside `cli.argParser` on unknown flags?

jamesgeorge007: `serve` supports the use of standalone flags, `webpack-cli serve --port <port> --hot`. If we were to handle `unknownArgs` within `arg-parser` standalone flags (when used with `serve`) won't be recognized by default.
https://github.com/webpack/webpack-cli/blob/f1876c125be90e0b0ae96e6e9dbaf91ed447c779/packages/serve/src/index.ts#L17-L20

When it comes to `init`, we've to filter out the `cli-flags` from `unknownArgs`, or else `webpack-scaffold-<name>` would be treated as an unknown argument.

Also, handling unknown args within `arg-parser` seems to interfere with the standalone flags.

packages/info/src/index.ts
```diff
@@ -38,10 +38,7 @@ export default async function info(...args): Promise<string[]> {
     const infoArgs = parsedArgs.opts;
     const envinfoConfig = {};
 
-    if (parsedArgs.unknownArgs.length > 0) {
-        logger.error(`Unknown argument: ${parsedArgs.unknownArgs}`);
-        process.exit(2);
-    }
+    handleUnknownArgs(parsedArgs.unknownArgs);
```
evilebottnawi, line 41: We should avoid handling this in sub packages, unknown args and commands should be handled in `webpack-cli`

jamesgeorge007: https://github.com/webpack/webpack-cli/pull/1703#discussion_r461366612

evilebottnawi: So we need to refactor our commands, and register their command names and flags before parsing args

evilebottnawi: We can do it separately

alexander-akait:
Fixed in master

PR #1712, opened 2020-07-31
feat: serve integration

rishabh3112:
Serve integration

Yes
No
**Summary**
Integrated serve regardless some issues left to be addressed.

No

**Other information**
@evilebottnawi 's comment https://github.com/webpack/webpack-cli/issues/717#issuecomment-664997156


packages/webpack-cli/lib/commands/ExternalCommand.js
```diff
@@ -34,6 +34,15 @@ class ExternalCommand {
...
         }
+        // Check dev server availability before running serve
+        if (name === 'serve') {
+            try {
+                // eslint-disable-next-line node/no-extraneous-require
+                require.resolve('webpack-dev-server');
+            } catch (error) {
+                throw new Error(`You need to install 'webpack-dev-server' for running 'webpack serve'.\n${error}`);
+            }
+        }
```
evilebottnawi, line 45: We should put this logic to `serve` command, not here

packages/serve/src/serve.ts
```diff
@@ -0,0 +1,39 @@
...
+            .forEach((unknown) => {
+                logger.warn('Unknown argument:', unknown);
+            });
+        return;
+    }
+
+    cli.getCompiler(webpackArgs, core).then((compiler): void => {
+        startDevServer(compiler, devServerArgs);
+    });
+}
```
evilebottnawi, line 39: Please do not create extra files, put it in one,

packages/serve/src/index.ts
```diff
@@ -1,39 +1,10 @@
...
+// Test dev-server availability before importing it
+try {
+    // eslint-disable-next-line node/no-extraneous-require
+    require.resolve('webpack-dev-server');
+} catch (error) {
+    throw new Error(`You need to install 'webpack-dev-server' for running 'webpack serve'.\n${error}`);
 }
+
+import serve from './serve';
+export default serve;
```
evilebottnawi, line 10: Before import can not be something, it is work because we use ts transpiler, but it is bad practice

rishabh3112: Yeah, I wasn't happy with the solution either, but we should not import from 'webpack-dev-server', before checking it's existance and `serve.ts`( original index file) is dependent on that. So, can you point me where do you like to have this check then?

evilebottnawi: Just put it in one file and do it inside `server`, using here is not safe, because it will output error for all developers who uses webpack-cli

rishabh3112: Okay, just be clear, is this what you are suggesting in the single file?
```js
import something from 'webpack-dev-server';
 // Inside serve method...
try {
   require.resolve('webpack-dev-server');
} catch (err){
// Handle err
}
//....
```

evilebottnawi:
```js
let devServer;

try {
   devServer = require('webpack-dev-server');
} catch (err){
  // Handle err
}

devServer.listen()
```

packages/serve/src/index.ts
```diff
@@ -1,10 +1,45 @@
...
+
+let devServerFlags: object[];
 try {
-    // eslint-disable-next-line node/no-extraneous-require
-    require.resolve('webpack-dev-server');
-} catch (error) {
-    throw new Error(`You need to install 'webpack-dev-server' for running 'webpack serve'.\n${error}`);
+    devServerFlags = require('webpack-dev-server/bin/cli-flags').devServer;
+} catch (err) {
+    throw new Error(`You need to install 'webpack-dev-server' for running 'webpack serve'.\n${err}`);
```
evilebottnawi, line 9: Put this inside `export default function serve`, because it is break webpack-cli for developers who don't want to use `webpack-dev-server` (and don't install this)

rishabh3112: Done

packages/serve/src/startDevServer.ts
```diff
@@ -10,6 +8,13 @@ import Server from 'webpack-dev-server/lib/Server';
  * @returns {Void}
  */
 export default function startDevServer(compiler, options): void {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    let Server: any;
+    try {
+        Server = require('webpack-dev-server/lib/Server');
+    } catch (err) {
+        throw new Error(`You need to install 'webpack-dev-server' for running 'webpack serve'.\n${err}`);
+    }
```
evilebottnawi, line 17: I think it is unnecessary, we already do check before

rishabh3112: We are importing this file before the `serve` function, so I think it is kind of necessary as well. Am I missing something?

evilebottnawi: We run `startDevServer` after first check so we don't need it :smile:

rishabh3112: Yeah, my bad. Updated now.

test/serve/basic/serve-basic.test.js
```diff
@@ -19,62 +19,58 @@ describe('basic serve usage', () => {
 
     const isWindows = process.platform === 'win32';
 
+    // TODO fix me on windows
     if (isWindows) {
-        // TODO fix me on windows
-        it('compiles without flags', () => {
-            expect(true).toBe(true);
+        return;
```
evilebottnawi, line 27: Can you open issue about windows?

PR #1715, opened 2020-07-31
feat: support multiple env params

anshumanv:

feat

Yes

Yes

**Summary**
Having multiple env is a common use case so a single arg won't suffice, hence added support for multiple

No

**Other information**

### DIFF
--- a/test/config/type/function-with-env/a.js
+++ b/test/config/type/function-with-env/a.js
@@ -1,3 +1,5 @@
 console.log('chuntaro');
 // eslint-disable-next-line no-undef
-console.log(envMessage || 'env message absent');
+if (envMessage) {
+    console.log('env message present');
+}
--- a/test/config/type/function-with-env/a.js
+++ b/test/config/type/function-with-env/a.js
@@ -1,1 +1,3 @@
 console.log('chuntaro');
+// eslint-disable-next-line no-undef
+console.log(envMessage || 'env message absent');
--- a/test/config/type/function-with-env/function-with-env.test.js
+++ b/test/config/type/function-with-env/function-with-env.test.js
@@ -18,4 +18,12 @@ describe('function configuration', () => {
         // Should generate the appropriate files
         expect(existsSync(resolve(__dirname, './bin/dev.js'))).toBeTruthy();
     });
+    it('is able to understand multiple env flags', () => {
+        const { stderr, stdout } = run(__dirname, ['--env', 'isDev', '--env', 'verboseStats']);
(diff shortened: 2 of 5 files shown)


anshumanv:
in near future

test/config/type/function-with-env/function-with-env.test.js
```diff
@@ -9,19 +9,21 @@ describe('function configuration', () => {
...
-        });
+        expect(existsSync(resolve(__dirname, './bin/dev.js'))).toBeTruthy();
+    });
+    it('is able to understand multiple env flags', () => {
+        const { stderr, stdout } = run(__dirname, ['--env', 'isDev', '--env', 'verboseStats']);
+        expect(stderr).toBeFalsy();
+        expect(stdout).toBeTruthy();
+        expect(stdout).toContain('LOG from webpack.buildChunkGraph.visitModules');
+        // Should generate the appropriate files
+        expect(existsSync(resolve(__dirname, './bin/dev.js'))).toBeTruthy();
```
evilebottnawi, line 27: Can we improve testing them here?

anshumanv: What can we do? we're already checking for both the env flags

anshumanv: Can you give example?

evilebottnawi: I think for better testing will be great to use `DefinePlugin`, insert this variables to source code and test what we contains them in generated bundle

anshumanv: Thanks for the suggestion @evilebottnawi 👍 Done 💯

test/config/type/function-with-env/a.js
```diff
@@ -1 +1,3 @@
 console.log('chuntaro');
+// eslint-disable-next-line no-undef
```
anshumanv, line 2: need disable because envMessage is not present in scope but instead coming from env

test/config/type/function-with-env/webpack.config.js
```diff
@@ -10,6 +13,8 @@ module.exports = (env) => {
     return {
         entry: './a.js',
         mode: 'development',
+        stats: env.verboseStats ? 'verbose' : 'normal',
+        plugins: [new DefinePlugin({ envMessage: env.envMessage ? JSON.stringify('env message present') : false })],
```
anshumanv, line 17: need JSON.stringify, can't pass string directly 😞

Issue #1581, opened 2020-05-25
webpack-cli v4-beta.8 ignoring loader's error.


mstssk:
**Describe the bug**

webpack-cli v4-beta.8 ignoring loader's error.

Reproduction code: https://github.com/mstssk/webpack-cli-v4-beta8-ignore-loader-error

I wrote reproduction code of ts-loader, but this problem occurs with other loaders. e.g. file-loader.

**What is the current behavior?**

webpack-cli v3.3.11 fails and returns exit code 2 when an error occurs in loaders.
But webpack-cli v4.0.0-beta.8 does not fail and return exit code 0.

**To Reproduce**

1. Setup webpack@4.43.0, webpack-cli@4.0.0-beta.8, and some loaders
2. Write codes which raise error in loaders.
3. Run webpack
4. See error logs and status code.

**Expected behavior**

webpack fails and returns non-zero exit code when an error occurs in loaders.

**Screenshots**

**Please paste the results of `webpack-cli info` here, and mention other relevant information**

`webpack-cli info` fails in my env.

<details>
<summary>`webpack-cli info` Log</summary>

```
 $ npx webpack-cli info
[webpack-cli] Promise rejection: TypeError: Cannot convert undefined or null to object
[webpack-cli] TypeError: Cannot convert undefined or null to object
    at Function.entries (<anonymous>)
    at Object.<anonymous> (/private/tmp/webpack-cli-v4-beta8-ignore-loader-error/node_modules/@webpack-cli/info/index.js:45:41)
    at Generator.next (<anonymous>)
    at /private/tmp/webpack-cli-v4-beta8-ignore-loader-error/node_modules/@webpack-cli/info/index.js:8:71
    at new Promise (<anonymous>)
    at __awaiter (/private/tmp/webpack-cli-v4-beta8-ignore-loader-error/node_modules/@webpack-cli/info/index.js:4:12)
    at Object.info [as default] (/private/tmp/webpack-cli-v4-beta8-ignore-loader-error/node_modules/@webpack-cli/info/index.js:44:12)
    at Function.run (/private/tmp/webpack-cli-v4-beta8-ignore-loader-error/node_modules/webpack-cli/lib/commands/ExternalCommand.js:34:51)
    at WebpackCLI.runCommand (/private/tmp/webpack-cli-v4-beta8-ignore-loader-error/node_modules/webpack-cli/lib/webpack-cli.js:286:60)
[...]

PR #1720, opened 2020-08-01
fix: generated loader template

### LINKED ISSUE #1719: Error using a loader generated via CLI
**Describe the bug**
Error using a loader generated via CLI

Syntax Error.
Screenshot at 2020-08-01 16-47-33

**To Reproduce**

1. generate loader via CLI `webpack-cli loader`
2. Go to `cd ./examples/simple`
3. Run `webpack`

**Expected behavior**
Should run without any syntax error.

**Screenshots**
See above

**Additional context**
Fix in progress.

snitin315:
bugfix

Yes
NA
**Summary**
Fixes #1719 

no

**Other information**


packages/generate-loader/templates/src/_index.js.tpl
```diff
@@ -3,9 +3,9 @@
  * https://webpack.js.org/contribute/writing-a-loader
  */
 
-export default function loader(source) {
+module.exports = function loader(source) {
```
evilebottnawi, line 6: I think we should implement `--babel` option to this in future (put babel config and generate es module syntax)

anshumanv:
Great that we're at this,

We should also get back to https://github.com/webpack/webpack-cli/issues/1581 with proper handling, @snitin315 can you handle it next?

snitin315:
> We should also get back to #1581 with proper handling, @snitin315 can you handle it next?

Sure 👍🏻

packages/generate-loader/templates/src/_index.js.tpl
snitin315: Will implement in future PR.

test/loader/loader.test.js
```diff
@@ -46,5 +46,10 @@ describe('loader command', () => {
         files.forEach((file) => {
             expect(existsSync(join(__dirname, `${loaderName}/${file}`))).toBeTruthy();
         });
+
+        //check if the the generated plugin works successfully
+        const path = resolve(__dirname, './test-loader/examples/simple/');
+        stdout = run(path, [], false).stdout;
```
jamesgeorge007, line 52:
```suggestion
        ({ stdout } = run(path, [], false));
```
What about the destructuring approach?

snitin315: I think its just personal preferences, I have no problem writing it like that.

snitin315: Should I change this @jamesgeorge007 ?

jamesgeorge007: I would've went ahead with object destructuring. But yeah, as you said it's just personal preference.
/cc @webpack/cli-team

anshumanv: you can do it for consistency sake

Issue #1126, opened 2019-11-19
[BUG]: webpack-cli@4.0.0-beta.1 no longer supports function-export-based webpack.config.js


wood1986:
**Describe the bug**

When my webpack.config.js is exporting a function instead of a simple object. Webpack won't recognize my config anymore. I have checked your test case on your next branch and I do not see any webpack.config.js which is exporting a function. When I export an object, it works Are you work in progress? Or Will you deprecate that kind of export?

**What is the current behavior?**

See the previous

**To Reproduce**

use this webpack.config.js
```js
module.exports = (env, argv) => {
  return {
    "entry": {
      "web.index": path.resolve(__dirname, "src", "web.index.jsx")
    },
    "output": {
      "filename": `[name].${argv.mode}.[chunkhash].js`,
      "libraryTarget": "umd",
      "path": path.resolve(__dirname, "dist")
    }
  }
}
```

and then run
```
webpack-cli --config=./webpack.config.js
```

**Expected behavior**
/

**Screenshots**
/

**Please paste the results of `webpack-cli info` here, and mention other relevant information**

webpack: next
webpack-cli: 4.0.0-beta.1
node: 13.1.0
OS: macOS 10.15.1

**Additional context**
/

wood1986:
+ @TheLarkInn

TheLarkInn:
cc @evenstensberg this is considered a release-blocking bug.

evenstensberg:
Considered!

evenstensberg:
Thx for your bug report @wood1986 , appreciate you putting in the time ❤️

wood1986:
I have one more request. webpack-cli 3 has a private function https://github.com/webpack/webpack-cli/blob/f9f08604a1ce5272b20d4f942e6861f3dc170dc0/bin/cli.js#L125 to convert function-export-base to the real object. I want to reuse but it is not exposed. In webpack-cli 4, is it possible to expose that?

evenstensberg:
I think that the edit needs to be done here:

https://github.com/webpack/webpack-cli/blob/next/lib/groups/config.js#L86-L88

anikethsaha:
Not sure why this bug is happening but the tests seems fine and passing for config type function

https://github.com/webpack/webpack-cli/tree/next/test/config/type/function

[...]

PR #1721, opened 2020-08-01
fix: supply argv to config with functions

anshumanv:

fix

Yes

NA

**Summary**

- Pass argv to config function

No

**Other information**
Fixes https://github.com/webpack/webpack-cli/issues/1704
Fixes https://github.com/webpack/webpack-cli/issues/1126


test/config/type/function-with-argv/function-with-argv.test.js
```diff
@@ -0,0 +1,14 @@
...
+const { existsSync } = require('fs');
+const { resolve } = require('path');
+const { run } = require('../../../utils/test-utils');
+
+describe('function configuration', () => {
+    it('is able to understand a configuration file as a function', () => {
+        const { stderr, stdout } = run(__dirname, ['--mode', 'development']);
+        expect(stderr).toBeFalsy();
+        expect(stdout).toBeTruthy();
+        // Should generate the appropriate files
```
anshumanv, line 11: @evilebottnawi any way to make these tests better?

snitin315: is it possible to log `argv` and check its value? It should not be `undefined`

evilebottnawi: I think testing is enough in this case

anshumanv: oops I added a commit for checking argv too, I guess it's fine

milizhang:
I think this does not actually work - while it seems like the arguments now can be passed to config function, when you actually tries to pass custom arguments, user will stuck at "Which flags do you want to use", likely as the custom argument is unknown to the cli.

anshumanv:
@milizhang can you give example?

milizhang:
@anshumanv Please take a look at this: https://github.com/milizhang/webpack-cli-4-argv-repro

If you use webpack-cli 3, `console.log(argv.testArg);` will give you "Test" if you run `webpack --testArg=Test`, while webpack-cli 4 will stuck asking you to clarify which argument you want to place. (Unless it is an argument that already exists in core webpack).

anshumanv:
Yep unknown arguments are not registered, you should use `--env` for it

milizhang:
Is this something that webpack-cli explicitly decided to deprecate in V4? Are there anyway to register custom arguments in configurations?

anshumanv:
> Is this something that webpack-cli explicitly decided to deprecate in V4? Are there anyway to register custom arguments in configurations?

Yes, passing arbitrary values was deprecated. You should be able to achieve it using `--env` or does it not fit your use case?

You can just do - 

`--env testArg=test`

```js
env = { testArg: 'test' } // in config 
```

milizhang:
Got it. Picking custom argv vs. env was more of a preference issue. Since it is deprecated I will move toward env.

Thanks for the info.

PR #1727, opened 2020-08-02
feat: add support for .cjs config

anshumanv:

feat

Yes

Not yet

**Summary**
- Add support for cjs config, handled manually now since .cjs is not available on the interpret side but will update when interpret adds it to the list https://github.com/gulpjs/interpret/issues/68#issuecomment-667638420

No

**Other information**


packages/webpack-cli/lib/groups/ConfigGroup.js
```diff
@@ -42,6 +42,16 @@ const getDefaultConfigFiles = () => {
 
 const getConfigInfoFromFileName = (filename) => {
     const fileMetaData = parse(filename);
+    // .cjs is not available on interpret side, handle it manually for now
```
evilebottnawi, line 45: Can you clarify?

anshumanv: @evilebottnawi check https://github.com/gulpjs/interpret/blob/57cd5973c313c420b1ad24a922d1c95ebec3c532/index.js#L181 doesn't have `.cjs` hence need manual handling

PR #1737, opened 2020-08-07
fix(info): throw an error if help or version is passed as an arg

jamesgeorge007:
bugfix

Generic test cases were added as part of #1694

N/A

**Summary**
Follow up of #1694

Nope

**Other information**
N/A

### DIFF
--- a/packages/info/src/index.ts
+++ b/packages/info/src/index.ts
@@ -36,10 +36,9 @@ export default async function info(...args): Promise<string[]> {
     const infoArgs = parsedArgs.opts;
     const envinfoConfig = {};
 
-    if (parsedArgs.unknownArgs.some((arg) => ['help', 'version', 'color'].includes(arg))) return;
-
     if (parsedArgs.unknownArgs.length > 0) {
         process.stderr.write(`Unknown argument: ${red(parsedArgs.unknownArgs)}\n`);
+        process.exit(2);
     }
 
     if (infoArgs.output) {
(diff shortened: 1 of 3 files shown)


jamesgeorge007:
> Can we fix it in all packages in this PR?

It is just `serve` and `info` that leverages `arg-parser` as of now.

packages/info/src/index.ts
```diff
@@ -36,10 +36,9 @@ export default async function info(...args): Promise<string[]> {
     const infoArgs = parsedArgs.opts;
     const envinfoConfig = {};
 
-    if (parsedArgs.unknownArgs.some((arg) => ['help', 'version', 'color'].includes(arg))) return;
-
     if (parsedArgs.unknownArgs.length > 0) {
         process.stderr.write(`Unknown argument: ${red(parsedArgs.unknownArgs)}\n`);
```
evilebottnawi, line 40: Should we use `logger` here?

packages/info/src/index.ts
```diff
@@ -36,10 +36,9 @@ export default async function info(...args): Promise<string[]> {
     const infoArgs = parsedArgs.opts;
     const envinfoConfig = {};
 
-    if (parsedArgs.unknownArgs.some((arg) => ['help', 'version', 'color'].includes(arg))) return;
-
     if (parsedArgs.unknownArgs.length > 0) {
-        process.stderr.write(`Unknown argument: ${red(parsedArgs.unknownArgs)}\n`);
+        logger.error(`Unknown argument: ${parsedArgs.unknownArgs}`);
```
evilebottnawi, line 40: Can you check other code on same problems? looks we need more `logger` :smile:

PR #1741, opened 2020-08-12
perf: do not spawn new process for running webpack

anshumanv:

fix

Not yet

No

**Summary**
Do not spawn new process for running the CLI

Shouldn't 

**Other information**


evilebottnawi:
Do you need help with tests?

snitin315:
Yes, I'm helping with the node flags support. I will push changes soon.

snitin315:
Using [NODE_OPTIONS](https://nodejs.org/api/cli.html#cli_node_options_options) gives desired behavior with `webpack@next` only and valid `node flags` but if we use `invalid node flags` it throws the following error  -
Screenshot at 2020-08-13 18-39-33

evilebottnawi:
I think we should remove `--node-args` in favor `NODE_OPTIONS`, because all tools uses `NODE_OPTIONS`, no one implement `--node-args`

anshumanv:
> I think we should remove --node-args in favor NODE_OPTIONS, because all tools uses NODE_OPTIONS, no one implement --node-args

Yeah we can remove, also doing unnecessary parsing

evilebottnawi:
Let's do it here :+1:

anshumanv:
Okay, on it

evilebottnawi:
Here we improve starting webpack around 50-200ms on difference computer (already tested), it is very good improvements

anshumanv:
> Here we improve starting webpack around 50-200ms on difference computer (already tested), it is very good improvements

Thanks for the benchmark!

test/node/node.test.js
```diff
@@ -2,78 +2,37 @@
...
         expect(stderr).toContain('missing value for flag --max-old-space-size');
         expect(stdout).toBeFalsy();
     });
 
-    it('throws an error if an illegal value was supplied with --max-old-space-size', () => {
-        const { stderr, stdout } = run(__dirname, ['--node-args', '--max-old-space-size=1024a']);
-        expect(stderr).toContain('illegal value for flag --max-old-space-size');
+    it('throws an error if an illegal value was supplied with --max-old-space-size', async () => {
+        const { stderr, stdout } = await run(__dirname, [], true, ['--max_old_space_size=1024a']);
+        expect(stderr).toContain('Error: illegal value for flag --max_old_space_size=1024a of type size_t');
```
evilebottnawi, line 35: Can we add tests for `NODE_OPTIONS` env variable?

anshumanv: Using inbuilt env to pass `NODE_OPTIONS` from execa throws different error messages since it sets on process.env so I'm guessing it's past the validation stage. https://github.com/sindresorhus/execa#env hence set NODE_OPTIONS via the `nodeOption` param

Directly passing NODE_OPTIONS doesn't seem to work correctly 
image

anshumanv: Different err when passing it via execa env param, I'm guessing it does some validation for the passed NODE_OPTIONS before setting to env, but here we're setting in env directly bypassing the validation hence different err
image

evilebottnawi: interesting, need investigate, can you add `TODO` with `NODE_OPTIONS` to add test on this stuff and test it manually?

anshumanv: Yep added comment 👍 

Working fine otherwise 
image

anshumanv:
CI freeze 😞

anshumanv:
Something wrong with commitlint 😞

evilebottnawi:
@anshumanv Don't worry about `commitlint` I will fix commit message before merge, anyway we need look why it is broken

PR #1747, opened 2020-08-15
fix: allow unknown files to use default require as fallback

anshumanv:
fix

Yes

No need

**Summary**
- Remove unnecessary filtering
- We filter all the config based on the available extensions in interpret which is not right and files should fallback to the default node require when they don't have a module needed.

No

**Other information**
Question -> when we pass some file which doesn't exist ex `-c webpack.testconfig.js` we fallback to default config `webpack.config.js`, should we allow that considering the user already passed a config they want to use or should we throw err that config not found?

### DIFF
--- a/packages/webpack-cli/lib/groups/ConfigGroup.js
+++ b/packages/webpack-cli/lib/groups/ConfigGroup.js
@@ -42,25 +42,16 @@ const getDefaultConfigFiles = () => {
 };
 
 const getConfigInfoFromFileName = (filename) => {
-    const fileMetaData = parse(filename);
-    // .cjs is not available on interpret side, handle it manually for now
-    if (filename.endsWith('.cjs')) {
-        return [
-            {
-                path: resolve(filename),
-                ext: '.cjs',
-                module: null,
-            },
-        ];
-    }
-    return Object.keys(extensions)
-        .filter((ext) => ext.includes(fileMetaData.ext))
-        .filter((ext) => fileMetaData.base.substr(fileMetaData.base.length - ext.length) === ext)
-        .map((ext) => {
+    const { ext } = parse(filename);
+    // since we support only one config for now
+    const allFiles = [filename];
+    // return all the file metadata
+    return allFiles
+        .map((file) => {
             return {
-                path: resolve(filename),
+                path: resolve(file),
                 ext: ext,
-                module: extensions[ext],
+                module: extensions[ext] || null,
             };
         })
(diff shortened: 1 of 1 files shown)


evenstensberg:
Considering a file is a small set, I think this is ok

anshumanv:
This is how gulp also uses it, prepares env which need specific module, rest fallback to the default node loader, no use of filtering

anshumanv:
> I think this removes cjs support, could you make a test file first?

Nope doesn't remove, I already added tests for cjs when I added it's support https://github.com/webpack/webpack-cli/pull/1727

evilebottnawi:
> Question -> when we pass some file which doesn't exist ex -c webpack.testconfig.js we fallback to default config webpack.config.js, should we allow that considering the user already passed a config they want to use or should we throw err that config not found?

I think without extensions we should confider it as a JS file, it is edge case, but it doesn't mean we should ignore it

anshumanv:
> I think without extensions we should confider it as a JS file, it is edge case, but it doesn't mean we should ignore it

@evilebottnawi talking about config file that doesn't exist eg `webpack.doesntexist.js`, should we throw error or fallback to default config?

evilebottnawi:
@anshumanv if config doesn't exists we should throw an error

anshumanv:
> @anshumanv if config doesn't exists we should throw an error

Gotcha, will implement in near future

anshumanv:
I think we're good here then

packages/webpack-cli/lib/groups/ConfigGroup.js
```diff
@@ -42,25 +42,16 @@ const getDefaultConfigFiles = () => {
...
-                ext: '.cjs',
-                module: null,
-            },
-        ];
-    }
-    return Object.keys(extensions)
-        .filter((ext) => ext.includes(fileMetaData.ext))
-        .filter((ext) => fileMetaData.base.substr(fileMetaData.base.length - ext.length) === ext)
-        .map((ext) => {
+    const { ext } = parse(filename);
```
evilebottnawi, line 45: Let's use `path.extname`, it faster https://github.com/nodejs/node/blob/master/lib/path.js#L751

anshumanv: Done 💯

evilebottnawi (approved):
/cc @webpack/cli-team

evenstensberg (approved):

snitin315 (approved):

PR #1751, opened 2020-08-16
tests(info): --output flag handles invalid arg

jamesgeorge007:
test update

Yup

N/A

**Summary**
Added a test case to ensure that a warning shows up in case an invalid value is supplied with `webpack-cli info --output`.

Nope

**Other information**
N/A

### DIFF
--- a/test/info/info-output.test.js
+++ b/test/info/info-output.test.js
@@ -36,4 +36,10 @@ describe('basic info usage', () => {
         expect(stdout).toContain('## System:');
         expect(stderr).toHaveLength(0);
     });
+
+    it('shows a warning if an invalid value is supplied', () => {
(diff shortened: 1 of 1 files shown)


evenstensberg:
yaml could be a directory so it's fine, we should support directories and make an entry based on the directory, this would be `yaml/main.js`

evilebottnawi:
@jamesgeorge007 `yaml` is the output format, if you want to write it file please use `command &> file.ext`

evenstensberg (approved):

evilebottnawi (approved):

snitin315 (approved):

anshumanv (approved):

snitin315 (approved):

PR #1753, opened 2020-08-18
feat: add --config-name flag

### LINKED ISSUE #1657: Support setting the config name for multiple configs
Operating System: Ubuntu 18.04
Node Version: 8.11.1
NPM Version: 6.0.0
webpack Version: 4.12.0
webpack-hot-client Version: 4.0.3

- [x] This is a **feature** request
- [ ] This is a **bug**

### What is the motivation and/or use-case for the feature?
Webpack supports multiple configurations in a single file by exporting an array.
[Info Here](https://webpack.js.org/configuration/configuration-types/#exporting-multiple-configurations)
A new feature that was added recently (but currently lacks total documentation) is that you can set a 'name' property in your webpack config and pass `--config-name` to the cli to choose a specific config to build if you don't want to build all of them.

This feature request is to allow support of this functionality in the webpack-dev-middleware and related plugins.

Implementing this could be done in a couple ways:
- Support passing webpack-cli options into the middleware somehow
- Support choosing a config-name in the middleware options

I think the first solution is going to be the best approach if it's possible. This is because it would allow any webpack-cli arguments to be used with middlewares. If for some reason there's no way to go down this route, the second option would suffice for my use case, but it might be a good idea to look at other options listed [here](https://webpack.js.org/api/cli/) to see if they would merit being included as well

snitin315:
Feature

Yes
Yes
**Summary**
Addition of `--config-name` flag.

No

**Other information**
Fixes #1657

### DIFF
--- a/packages/webpack-cli/lib/groups/ConfigGroup.js
+++ b/packages/webpack-cli/lib/groups/ConfigGroup.js
@@ -124,7 +124,7 @@ class ConfigGroup extends GroupHelper {
             const namedOptions = configOptions.filter((opt) => this.args.configName.includes(opt.name));
             if (namedOptions.length === 0) {
                 logger.error(`Configuration with name "${this.args.configName}" was not found.`);
-                process.exit(1);
+                process.exit(2);
             } else {
                 newOptionsObject['options'] = namedOptions;
             }
--- a/packages/webpack-cli/lib/groups/ConfigGroup.js
+++ b/packages/webpack-cli/lib/groups/ConfigGroup.js
@@ -5,6 +5,7 @@ const { extensions, jsVariants } = require('interpret');
 const GroupHelper = require('../utils/GroupHelper');
 const rechoir = require('rechoir');
 const MergeError = require('../utils/errors/MergeError');
+const logger = require('../utils/logger');
 
 // Order defines the priority, in increasing order
 // example - config file lookup will be in order of .webpack/webpack.config.development.js -> webpack.config.development.js -> webpack.config.js
@@ -117,6 +118,16 @@ class ConfigGroup extends GroupHelper {
             const newOptions = configOptions(formattedEnv, argv);
             // When config function returns a promise, resolve it, if not it's resolved by default
             newOptionsObject['options'] = await Promise.resolve(newOptions);
+        } else if (Array.isArray(configOptions) && this.args.configName) {
+            // In case of exporting multiple configurations, If you pass a name to --config-name flag,
+            // webpack will only build that specific configuration.
+            const namedOptions = configOptions.filter((opt) => this.args.configName.includes(opt.name));
+            if (namedOptions.length === 0) {
+                logger.error(`Configuration with name "${this.args.configName}" was not found.`);
+                process.exit(1);
+            } else {
+                newOptionsObject['options'] = namedOptions;
+            }
         } else {
             if (Array.isArray(configOptions) && !configOptions.length) {
                 newOptionsObject['options'] = {};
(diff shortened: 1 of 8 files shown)


anshumanv:
I think it makes sense to add multiple config support using flags first.

snitin315:
`--config-name` works with a single config file exporting an array of configurations.

test/config-name/config-name.test.js
```diff
@@ -0,0 +1,45 @@
...
+        stat(resolve(__dirname, './dist/dist-third.js'), (err, stats) => {
+            expect(err).toBe(null);
+            expect(stats.isFile()).toBe(true);
+            done();
+        });
+    });
+
+    it('should log error if invalid config name is provided', () => {
+        const { stderr, stdout } = run(__dirname, ['--config-name', 'test'], false);
+        expect(stderr).toContain('Configuration with name test was not found.');
```
snitin315, line 42:
```suggestion
        expect(stderr).toContain('Configuration with name "test" was not found.');
```

packages/webpack-cli/lib/groups/ConfigGroup.js
```diff
@@ -117,6 +118,14 @@ class ConfigGroup extends GroupHelper {
             const newOptions = configOptions(formattedEnv, argv);
             // When config function returns a promise, resolve it, if not it's resolved by default
             newOptionsObject['options'] = await Promise.resolve(newOptions);
+        } else if (Array.isArray(configOptions) && this.args.configName) {
```
anshumanv, line 121: Add a comment when this block is executed, can give example, looks good otherwise

packages/webpack-cli/lib/groups/ConfigGroup.js
```diff
@@ -117,6 +118,16 @@ class ConfigGroup extends GroupHelper {
...
+            // In case of exporting multiple configurations, If you pass a name to --config-name flag,
+            // webpack will only build that specific configuration.
+            const namedOptions = configOptions.filter((opt) => this.args.configName.includes(opt.name));
+            if (namedOptions.length === 0) {
+                logger.error(`Configuration with name "${this.args.configName}" was not found.`);
+                process.exit(1);
```
evilebottnawi, line 127: `0` - fine
`1` - warnings/errors from webpack 
`2` - configuration/options problem or an internal error

snitin315:
```suggestion
                process.exit(2);
```

evenstensberg: +1

evilebottnawi: We need check our code based on this and documented it

evilebottnawi: Somebody can take care?

snitin315: I will take care of this.

snitin315:
This `commitlint` check is unstable these days. Can we remove it as a `required` check?

evilebottnawi:
@snitin315 maybe we should update `commitlint` action, need look at this

anshumanv:
> @snitin315 maybe we should update commitlint action, need look at this

I think it's setup externally, not defined in the workflow file. @evenstensberg do you know how it's setup?

jamesgeorge007 (approved):

jamesgeorge007:
It seems we're relying on [commitlint-bot](https://github.com/z0al/commitlint-bot) which is not maintained anymore, it is advised to use [semantic-pull-requests](https://github.com/zeke/semantic-pull-requests) instead.

rishabh3112 (approved):

PR #1755, opened 2020-08-18
fix: use appropriate exit codes

snitin315:
fix, docs,

Yes
Yes
**Summary**
Use appropriate exit codes and document their meanings.

No

**Other information**
NA

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@
     -   [Utilities](#utilities)
 -   [Getting started](#getting-started)
 -   [webpack CLI Scaffolds](#webpack-cli-scaffolds)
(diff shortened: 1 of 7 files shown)


anshumanv:
Seems good @snitin315, just run a check once more that we're covering all the exit code statements to prevent unnecessary follow up for the same thing.

rishabh3112 (approved):

snitin315:
> Seems good @snitin315, just run a check once more that we're covering all the exit code statements to prevent unnecessary follow up for the same thing.

I did a check again, these were the only places I was able to found via GitHub search.

PR #1756, opened 2020-08-18
feat: smoketests for serve

evenstensberg:
Smoketests for devserver

Kinda
n/a
**Summary**

Needs to fix symlinks / close connection before running another tests, other than that okay

No
**Other information**


smoketests/dev-server.sh
```diff
@@ -0,0 +1,65 @@
...
+        ls ./dev-server/*.smoketest.js | xargs -I{} echo "Running:" {} | tee /dev/tty | cut -d':' -f 2 | xargs -I{} node {}
+        sudo lsof -i :3000 | awk '{print $2}' | grep -v "PID" | xargs -I{} kill {} -9
+        if [ "$?" != "0" ]; then
+            teardown
+        fi
+    done
+    echo "============================ DONE RUNNING $i $iterations times ============================"
+done
+
+teardown
```
evilebottnawi, line 65: Please avoid using bash, it is unnecessary, you can rewrite it using JavaScript

evenstensberg: No

evilebottnawi: Why?

evilebottnawi: /cc @webpack/cli-team Somebody love `bash`? I think using `JavaScript` everywhere is good idea, windows always have bugs with `bash`, we are potentially creating flaky tests, before we have same tests for webpack-dev-server and it was big pain

evenstensberg: I think bash is fine.. The setup and teardown is written in js, that's good enough, don't you think? scripting with bash is made for these tasks

evilebottnawi: What is the need for a bash if we have javascript, what is benefits? All I see is reading problems, especially for new developers.

evilebottnawi (changes requested):
Smoke tests are not running on CI, look at output, and it is broken in windows as I written above

smoketests/dev-server.sh
evenstensberg: Let's use bash now and change to js when we are done with smoketests, yes?

evilebottnawi: Why do not rewrite it to JS now before it grows into a big rubbish heap?

evenstensberg:
Only thing left is spawning these proceeses async

smoketests/dev-server.sh
evenstensberg: Concurrency is pretty hard with js, could you have a look @evilebottnawi ? I think it's better to handle spawning and cleaning up with bash, gave js a chance but with bash it's easier to control the processes

evilebottnawi: @evenstensberg What's easier? You can describe the problem?

evenstensberg: Every spawn is async, so each script will be done at a different time, in addition these operations in the scripts won't be in order, scripting is easier cause they will enforce concurrency by default

evilebottnawi: you can use async/await, or you can rewrite then on jest and using `--runInBand`

snitin315:
Stale

PR #1757, opened 2020-08-18
feat: add name flag

anshumanv:
feat

Yes

Yes

**Summary**

Add name flag.

No

**Other information**


rishabh3112 (commented):
Looks good. Just one doubt.
How `name` assignment would work in case of multiple configurations through the flag you created?

anshumanv:
> How name assignment would work in case of multiple configurations through the flag you created?

yeah we should have a test with multiple config

evilebottnawi (commented):
Do you want to backport is for webpack@4? I think core flags have the `name` flag

anshumanv:
Yes this is for v4, https://v4.webpack.js.org/configuration/other-options/#name

v5 already supports this via core flags @evilebottnawi

evilebottnawi:
> Looks good. Just one doubt.
How name assignment would work in case of multiple configurations through the flag you created?

Interesting case, ideas?

anshumanv:
> Interesting case, ideas?

Default behaviour right now is to apply it to all config, not sure if that's right for name

anshumanv:
Devtool flag applies to all the configs

evilebottnawi:
This options make sense only with single `--config-name`

anshumanv:
@snitin315 how does the core flag works in case of multiple config?

snitin315:
> @snitin315 how does the core flag works in case of multiple configs?

`webpack --name=web` purpose is just to set `compiler.options.name = 'web'` (which sets `compiler.name = web`) for a single webpack config file just like other cli-flags (for example `--mode`).

> Name of the configuration. Used when loading multiple configurations. 

This property `options.name` is useful in case if we export an array of multiple configs to the webpack compiler. It shows us output stats with `name` of the child compiler, so we can easily differentiate the stats as per configs  - 
```js
module.exports = [{
     name: 'first'
  }, 
  {
     name: 'second'
  }, 
  {
     name: 'third'
}]
```
Screenshot at 2020-08-20 07-09-31

In case if you don't provide `options.name` in config the output will be like - 

Screenshot at 2020-08-20 07-19-36

snitin315:
But we don't support setting options for multiple configurations.

evilebottnawi:
I think we need discussion about multiple configurations and flags, what is the logic now?

snitin315:
I think right now we only support -
-  choosing a config file from multiple config files via `--config`.
-  falling back to a default config based on priority in case of multiple config files.
-  choosing a config via `--config-name` if an array of configs is exported. If we don't provide `--config-name` all the configurations will run.

evilebottnawi:
Here interesting case, some options can't be overridden when you have multiple configurations, for example `--name` potential break build. So i think, we need list of this options and throw an error about this. I think same for `UniqueName` https://github.com/webpack/webpack/blob/master/schemas/WebpackOptions.json#L3165

snitin315:
+1

evilebottnawi (approved):
Let's open an issue about forbidden flags for multiple compilations and implement it in the other PR

packages/webpack-cli/lib/utils/cli-flags.js
```diff
@@ -253,6 +253,13 @@ module.exports = {
             group: CONFIG_GROUP,
             description: 'Environment passed to the configuration when it is a function',
         },
+        {
+            name: 'name',
+            usage: '--name',
```
snitin315, line 258: Need to add in duplicate flags.

anshumanv: Yep added 💯

packages/webpack-cli/lib/utils/cli-flags.js
```diff
@@ -33,7 +33,7 @@ let flagsFromCore =
         : [];
 
 // duplicate flags
-const duplicateFlags = ['entry', 'mode', 'stats', 'watch', 'target', 'devtool'];
+const duplicateFlags = ['entry', 'mode', 'stats', 'watch', 'target', 'devtool', 'name'];
```
evilebottnawi, line 36: Can we use `Object.keys()` on array flags for webpack@4 to avoid it in future?

anshumanv: Yeah we should do that to avoid manual work every time

anshumanv: Done 💯

Issue #1752, opened 2020-08-18
refactor: utils and package-utils


rishabh3112:
**Is your feature request related to a problem? Please describe.**
Currently we have shared code that is being used by different sub-packages in these packages. But this way, It allows for duplication of dependencies as `webpack-cli` package(dependendt on `package-utils`) package that is a peer dependency of all other sub-packages. Also, there may be a situation in which there would be multiple versions of these packages in user's codebase.

**Describe the solution you'd like**
Simplest solution would be to keep all the utils `webpack-cli` package.
(caveat: we would be increasing package size with code it may never use if some particular sub-package isn't installed by the user)
![image](https://user-images.githubusercontent.com/15812317/90607069-56b7be00-e21e-11ea-962b-2b12247a5a7e.png)

**Describe alternatives you've considered**
N/A

**Additional context**
/cc @evilebottnawi

evilebottnawi:
package-utils was removed, I think we need move `utils` package inside webpack-cli, but need think how we can reduce size

evilebottnawi:
Let's close, `package-utils` unions with `utils`

PR #1758, opened 2020-08-19
chore: update generator deps

rishabh3112:
chore, avoid duplication of package utils **(Just for release)**

N/A

**Summary**
Before:
image

After:
image

No

**Other information**
**Not an intended solution. Follow #1752 for it.**

### DIFF
--- a/packages/generators/__tests__/addon-generator.test.ts
+++ b/packages/generators/__tests__/addon-generator.test.ts
@@ -6,6 +6,7 @@ import fs from 'fs';
 import path from 'path';
 import mkdirp from 'mkdirp';
 import rimraf from 'rimraf';
+// eslint-disable-next-line node/no-extraneous-import
 import { getPackageManager } from '@webpack-cli/package-utils';
 import addonGenerator from '../src/addon-generator';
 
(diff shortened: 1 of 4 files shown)


evilebottnawi (commented):
It is break yarn PnP

rishabh3112:
Would adding `@webpack-cli/package-utils`package as a peer dependency solve this?

evilebottnawi:
I think we can solve it using steps:
1. Union `package-utils` and `utils` into `utils`
2. Move out utils for generators into `generators` package
3. Additional step - move `utils` into `webpack-cli`, because we use almost of them in `webpack-cli`

rishabh3112:
> Additional step - move utils into webpack-cli, because we use almost of them in webpack-cli
we don't, we just use `package-utils` in CLI.

evilebottnawi:
I mean this packages unnecessary, because we always install then, no need to keep them in deps

rishabh3112:
Yes, these packages are unnecessary. As this would take long time to refactor it, I thought to doing this so that we can release within this week. Rest, I also support removing these packages.

rishabh3112:
Closing as full refactor is in progress at #1822

PR #1765, opened 2020-08-21
feat: add support to spawn multiple compilers with different configs

anshumanv:

feat

yes

yes

**Summary**
- Make config flag support multiple config
- when multiple configs are supplied they are passed to the compiler as array `webpack([{...}, {...}])` which spawn multiple compilers
- In case of single config, resolved config object is passed directly

No

**Other information**

### DIFF
--- a/packages/webpack-cli/lib/groups/ConfigGroup.js
+++ b/packages/webpack-cli/lib/groups/ConfigGroup.js
@@ -146,7 +146,7 @@ class ConfigGroup extends GroupHelper {
         if (config.length > 0) {
             const resolvedOptions = [];
             const finalizedConfigs = config.map(async (webpackConfig) => {
-                const configPath = resolve(process.cwd(), webpackConfig);
+                const configPath = resolve(webpackConfig);
                 const configFiles = getConfigInfoFromFileName(configPath);
                 if (!configFiles.length) {
                     throw new ConfigError(`The specified config file doesn't exist in ${configPath}`);
@@ -163,7 +163,7 @@ class ConfigGroup extends GroupHelper {
                     resolvedOptions.push(resolvedOption.options);
                 }
             }
-            // When the resolved configs are more then 1, then pass them as Array [{...}, {...}] else pass the first config object {...}
+            // When the resolved configs are more than 1, then pass them as Array [{...}, {...}] else pass the first config object {...}
             const finalOptions = resolvedOptions.length > 1 ? resolvedOptions : resolvedOptions[0] || {};
 
             this.opts['options'] = finalOptions;


evenstensberg (approved):

packages/webpack-cli/lib/utils/cli-flags.js
```diff
@@ -91,8 +91,9 @@ const core = [
         usage: '--config <path to webpack configuration file>',
         alias: 'c',
         type: String,
-        defaultValue: null,
+        defaultValue: [],
```
evilebottnawi, line 94: It is still `[]` by default?

snitin315: I think it is because having a `defaultValue` ensures that `ConfigGroup`  will always run even without passing `--config` hence leading to the selection of default configuration file i.e, `webpack.config.js`.
So as of right now if we remove the `defaultValue` option here cli will not load `ConfigGroup` and hence `webpack.config.js` will not be selected by default.

/cc @anshumanv am I right?

anshumanv: > It is still [] by default?

Yes we need [] to append all config values because of arg parser https://github.com/webpack/webpack-cli/pull/1765#discussion_r475077654, when we do `webpack -c 1.js -c 2.js` in config group we'll get `{config: ['1.js', '2.js']}`

It is only for arg parsing, not passed to the compiler when no config is supplied

anshumanv: > I think it is because having a defaultValue ensures that ConfigGroup will always run even without passing --config hence leading to the selection of default configuration file i.e, webpack.config.js.

Nope @snitin315 all groups are always run 😄

snitin315: >Yes we need [] to append all config values because of arg parser #1765 (comment), when we do webpack -c 1.js -c 2.js in config group we'll get {config: ['1.js', '2.js']}

We don't have `defaultValue: []` for `entry` but it works perfectly as expected. 

>Nope @snitin315 all groups are always run 

interesting I remove the `defaultValue: null`  line and it wasn't selecting `webpack.config.js`

evilebottnawi: I think we need improve arg-parser to avoid this misleading, but we can do it in separate PR

evilebottnawi (dismissed):

packages/webpack-cli/lib/utils/cli-flags.js
anshumanv: > We don't have defaultValue: [] for entry but it works perfectly as expected.

Interesting, will check the exact behaviour during improvement 💯

snitin315 (dismissed):

packages/webpack-cli/README.md
```diff
@@ -37,7 +37,7 @@ yarn add webpack-cli --dev
 
 ```
   --entry string[]         The entry point(s) of your application.
-  -c, --config string      Provide path to a webpack configuration file
+  -c, --config string[]    Provide path to a webpack configuration file
```
rishabh3112, line 40: Need to change description

rishabh3112 (dismissed):

packages/webpack-cli/README.md
anshumanv, line 40:
```suggestion
  -c, --config string[]    Provide path to a webpack configuration file(s)
```

packages/webpack-cli/README.md
```diff
@@ -37,7 +37,7 @@ yarn add webpack-cli --dev
 
 ```
   --entry string[]         The entry point(s) of your application.
-  -c, --config string      Provide path to a webpack configuration file
+  -c, --config string[]    Provide path to a webpack configuration file(s)
```
rishabh3112, line 40:
```suggestion
  -c, --config string[]    Provide path to webpack configuration file(s)
```

snitin315 (approved):

packages/webpack-cli/lib/groups/ConfigGroup.js
```diff
@@ -143,18 +143,34 @@ class ConfigGroup extends GroupHelper {
...
         const { config, mode } = this.args;
-        if (config) {
-            const configPath = resolve(process.cwd(), config);
-            const configFiles = getConfigInfoFromFileName(configPath);
-            if (!configFiles.length) {
-                throw new ConfigError(`The specified config file doesn't exist in ${configPath}`);
+        if (config.length > 0) {
+            const resolvedOptions = [];
+            const finalizedConfigs = config.map(async (webpackConfig) => {
+                const configPath = resolve(process.cwd(), webpackConfig);
```
jamesgeorge007, line 149:
```suggestion
                const configPath = resolve(webpackConfig);
```

packages/webpack-cli/lib/groups/ConfigGroup.js
```diff
@@ -143,18 +143,34 @@ class ConfigGroup extends GroupHelper {
...
+                if (Array.isArray(resolvedOption.options)) {
+                    resolvedOptions.push(...resolvedOption.options);
+                } else {
+                    resolvedOptions.push(resolvedOption.options);
+                }
             }
-            const foundConfig = configFiles[0];
-            const resolvedConfig = this.requireConfig(foundConfig);
-            this.opts = await this.finalize(resolvedConfig);
+            // When the resolved configs are more then 1, then pass them as Array [{...}, {...}] else pass the first config object {...}
```
jamesgeorge007, line 166:
```suggestion
            // When the resolved configs are more than 1, then pass them as Array [{...}, {...}] else pass the first config object {...}
```

jamesgeorge007 (changes requested):
Left a couple of suggestions.

packages/webpack-cli/lib/groups/ConfigGroup.js
```diff
@@ -143,18 +143,34 @@ class ConfigGroup extends GroupHelper {
...
-            const configFiles = getConfigInfoFromFileName(configPath);
-            if (!configFiles.length) {
-                throw new ConfigError(`The specified config file doesn't exist in ${configPath}`);
+        if (config.length > 0) {
+            const resolvedOptions = [];
+            const finalizedConfigs = config.map(async (webpackConfig) => {
+                const configPath = resolve(process.cwd(), webpackConfig);
+                const configFiles = getConfigInfoFromFileName(configPath);
+                if (!configFiles.length) {
+                    throw new ConfigError(`The specified config file doesn't exist in ${configPath}`);
```
jamesgeorge007, line 152: I think we can improve the message.
```js
if (!configFiles.length) {
    throw new ConfigError(`The specified config file ${basename(configPath)} doesn't exist in ${process.cwd()}`);
                }
```

WDYT?

anshumanv: we shouldn't do this, user can pass config path as well, will look redundant

evilebottnawi (approved):

rishabh3112 (approved):

snitin315 (approved):

evenstensberg (approved):

anshumanv:
Merging, thanks for the feedback everyone! 💯

PR #1786, opened 2020-09-04
feat: add mjs support

rishabh3112:
Support for `.mjs` configs

**Summary**
It is extension of @anshumanv's PR with support for esm configs using `esm` package.

**Other information**
The standard `await import()` will not work as https://github.com/nodejs/node/pull/32985 has been reverted.
Using `esm` package in this PR to have esm interop


test/config-format/mjs/webpack.config.mjs
```diff
@@ -0,0 +1,15 @@
+// import { dirname, resolve } from 'path';
+// import { fileURLToPath } from 'url';
+
```
evenstensberg, line 3: remove these

test/config-format/mjs/webpack.config.mjs
```diff
@@ -0,0 +1,15 @@
...
+// const __dirname = dirname(fileURLToPath(import.meta.url));
+
+const config = {
```
evenstensberg, line 6: export default { ... }

evenstensberg (changes requested):
Tests are failing

rishabh3112:
> Tests are failing

@evenstensberg Yes, because `esm` package is not working properly here. I couldn't find upon investigation on how to make it work.
So, I need help.

evilebottnawi:
I think we don't need package, we can use this idea https://github.com/gulpjs/gulp-cli/blob/master/lib/shared/require-or-import.js

rishabh3112:
> I think we don't need package, we can use this idea https://github.com/gulpjs/gulp-cli/blob/master/lib/shared/require-or-import.js

Same not working again locally.

evilebottnawi:
@rishabh3112 what is node you are use (version)?

rishabh3112:
Tried on 12, 13 and 14. (I use fnm)

evilebottnawi:
Maybe something broken in `fnm`? I use `nvm` and tests for gulp-cli worked with mjs

rishabh3112:
> Maybe something broken in fnm? .

Can't say for sure. Lets see CI, if still not passed then can you try checking out this branch and testing locally once?

anshumanv:
I tried with nvm in the past, no luck 😞

anshumanv:
Yep would be great if @evilebottnawi can check once

Issue #586, opened 2018-09-15
[Feature]: webpack interactive command


evenstensberg:
**Is your feature request related to a problem? Please describe.**
As apart of the new CLI, one thing that would be nice to have is a `webpack interactive` command. This way, users that are in the middle of a git rebase or want to pause compilation by any means, can do that. In addition, it would be nice to have a interactive terminal where you can specify watch mode, what files to compile and such.

**Describe the solution you'd like**
Implement a functional interactive CLI first, look at how [`jest watch`](https://jestjs.io/docs/en/cli.html#running-from-the-command-line) is doing this.

![skjermbilde 2018-09-15 kl 10 15 23](https://user-images.githubusercontent.com/16735925/45584235-58f59d00-b8d0-11e8-887b-38b7a18e4316.png)
 
The features that should be available would be:

• m: filter modules
• p: pause watching (i. e. when doing git actions)
• r: reload configuration and restart webpack
• c: clear cache and recompile
• e: explore mode to interactively walk the module/chunk graph to find out why a module is included
• w: select a module and print out why it's included
• q: quit

**Describe alternatives you've considered**
N/A

**Additional context**

This PR should go against the branch #527 is pointed towards. Feel free to reach out to any of the CLI members to get assistance in implementing this!

The only thing left to do, is to implement the logic for the command, which should be done as a method [here](https://github.com/webpack/webpack-cli/blob/refactor-bin-folder/lib/commands/internal.js#L1)

Happy hacking! 💙

Rohanhacker:
can I work on this ? This will be my first contribution to webpack cli

ev1stensberg:
Gladly!

Rohanhacker:
@ev1stensberg thanks

ev1stensberg:
Yes, but you should be able to do the commands as mentioned in the issue. No, these options are new, so you'll have to experiment a bit yourself!

LakshSingla:
@ev1stensberg Is this issue closed or is there any way I can contribute towards it ?

LakshSingla:
@ev1stensberg `lib/webpack-cli.js` uses `internal.js` but passes arguments in the constructor (which is not yet implemented in `internal.js`). I am presuming that `internal.js` is WIP. Should I implement the `constructor` in `internal.js`? Also many tests currently fail. Is that part any of my concern, or should I write my own tests for the `interactive` command.

LakshSingla:
[...]

PR #1796, opened 2020-09-11
feat: add support for interactive output mode

rishabh3112:
**Different from `interactive` at #586**: This PR just adds support for interactive output mode.

Feature

Not currently, will do.

Not applicable

**Summary**
There are two output modes for `--watch` and `webpack serve`, standard and interactive.
Standard is already supported. This PR aims to implement that.

No

**Other information**
/cc @evilebottnawi


rishabh3112:
Should I implement the plugin from scratch or this is fine?

test/interactive/interactive.test.js
```diff
@@ -0,0 +1,41 @@
...
+            }
+            if (semaphore === 0) {
+                expect(data).toContain('main.js');
+                proc.kill();
+                done();
+                return;
+            }
+        });
+    });
+});
```
evilebottnawi, line 41: Can we use snapshots to test it?

rishabh3112: Added snapshot test for final output with watch.

rishabh3112: @evilebottnawi Build time changes every time in output so snapshot test not working here.

evilebottnawi: And let's add test without `--interactive`, also we need test `webpack server` and we can merge it :+1:

test/interactive/interactive.test.js
```diff
@@ -31,9 +32,30 @@ describe('--interactive flag', () => {
...
+            const data = chunk.toString();
+            if (data.includes('watching files for updates')) {
+                writeFileSync(resolve(__dirname, 'index.js'), `console.log('I am Batman');`);
+                semaphore--;
+                return;
+            }
+            if (semaphore === 0) {
+                expect(data).not.toBe(clear);
+                proc.kill();
+                writeFileSync(resolve(__dirname, 'index.js'), `console.log('I am Iron Man');`);
```
evilebottnawi, line 58: Let's use snapshots too, now we should have multiple outputs, use `.replace()` to remove time and hashes

evenstensberg (commented):
Interactive is not a plugin

evilebottnawi:
@evenstensberg we can't solve it in other way, because webpack-dev-server uses hooks to output information, other plugins can do same, so we need to use hook before output stats to avoid problems with compatibility

evilebottnawi (commented):
Also I found it https://github.com/webpack/webpack-cli/blob/next/packages/webpack-cli/lib/utils/interactive.js, looks we have more features, like `pause`, `enter`, `quit`, I think we should put change in `interactive.js` and reuse logic from there

rishabh3112:
> Also I found it https://github.com/webpack/webpack-cli/blob/next/packages/webpack-cli/lib/utils/interactive.js, looks we have more features, like pause, enter, quit, I think we should put change in interactive.js and reuse logic from there

It spawns new instance of webpack compiler each time new build has to be triggered.
So, Extending it would not work for `--watch` as well as `webpack-dev-server` which watch and  recompile on their own.

rishabh3112:
> Interactive is not a plugin

Check issue description,
This PR aims to implement `interactive` output mode for `--watch` and `webpack serve`, **not** the `interactive` feature implemented in `interactive.js`.

test/interactive/interactive.test.js
rishabh3112: Done regressive phrase testing on stdout instead.

evilebottnawi:
@rishabh3112 maybe we can use hooks to output `pause`/`enter`/`quite` and etc, it should be not hard

rishabh3112:
> we can use hooks to output pause/enter/quite and etc,

Can we restart compilation process using hooks? I couldn't find the way.
`quit` could be implemented nicely.
Not sure about how to implement `pause` atm.

packages/webpack-cli/lib/utils/InteractiveModePlugin.js
```diff
@@ -40,9 +41,17 @@ class InteractiveModePlugin {
         });
 
         // Clear output afterCompile
-        compiler.hooks.afterCompile.tap('InteractiveModePlugin', () => {
+        compiler.hooks.afterCompile.tap(this.name, () => {
             process.stdout.write('\x1B[2J\x1B[3J\x1B[H');
         });
+
+        compiler.hooks.done.tap(this.name, () => {
```
evilebottnawi, line 52: Maybe better to use `afterDone`

packages/webpack-cli/lib/utils/InteractiveModePlugin.js
```diff
@@ -40,9 +41,17 @@ class InteractiveModePlugin {
         });
 
         // Clear output afterCompile
-        compiler.hooks.afterCompile.tap('InteractiveModePlugin', () => {
+        compiler.hooks.afterCompile.tap(this.name, () => {
```
evilebottnawi, line 48: Maybe better to use `compiler.hooks.invalid`

evilebottnawi (commented):
Can we add test(s) for multi compiler mode?

rishabh3112:
> Can we add test(s) for multi compiler mode?

We have a problem here. As interactive mode plugin is added to every compiler, Each one of them operates in parallel thus they obscure the output.

Would be able to fix it if there a way to access sibling compilers through compiler object. Is there a way?

rishabh3112:
Adding test for ideal output in case of multi-compiler. It should fail for now.

rishabh3112:
## Current Status
- Interactive Mode implementation is completed 🎆 
- All tests passing locally, don't know why CI fails 😞

rishabh3112:
# Status
Everything implemented
Waiting for review and merge

alexander-akait (commented):
Let's avoid webpack v4 support, we need to migrate and drop v4, just throw error webpack v4 is not supported

rishabh3112:
> Let's avoid webpack v4 support, we need to migrate and drop v4, just throw error webpack v4 is not supported

Okay, will remove webpack 4 support then

rishabh3112:
> > Let's avoid webpack v4 support, we need to migrate and drop v4, just throw error webpack v4 is not supported
> 
> Okay, will remove webpack 4 support then

Done

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
```diff
@@ -0,0 +1,242 @@
...
+ * Clear the whole terminal
+ */
+const clrscr = () => {
+    process.stdout.write('\x1B[2J\x1B[3J\x1B[H');
+};
+
+/**
+ * Helper plugin for child compilers if MultiCompiler is supplied
+ */
+class InteractiveModeMultiCompilerHelperPlugin {
```
alexander-akait, line 47: Let's avoid it and apply hooks directly on each compiler

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
```diff
@@ -0,0 +1,242 @@
...
+    constructor() {
+        this.name = 'InteractiveModeMultiCompilerHelperPlugin';
+    }
+
+    apply(compiler) {
+        // clear terminal if any one of child starts compilation
+        compiler.hooks.beforeCompile.tap(this.name, () => {
```
alexander-akait, line 54: Let's add test with `--profile` here, anyway I think `beforeCompile` is good hook, I think we need improve logic for multi compiler mode like:
```js
this.isCleanScreen = false;

// Logic

for (const compiler of compilers) {
  compiler.hooks.beforeCompile.tap(this.name, () => {
    if (this.isCleanScreen) { 
      clrscr();
    }
  });
}  
```

to avoid racing between two/three/etc compilations

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
```diff
@@ -0,0 +1,242 @@
...
+        };
+        this.handlers = {
+            quit: this.quitHandler.bind(this),
+            stop: this.stopHandler.bind(this),
+            start: this.startHandler.bind(this),
+        };
+    }
+
+    apply(compiler) {
+        if (!isWebpack5) {
```
alexander-akait, line 85: Avoid it, please use `compiler.webpack` (exist on for webpack v5)

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
```diff
@@ -0,0 +1,242 @@
...
+
+            if (possibleActions.length > 1) {
+                throw new Error('Multiple actions are provided for same key');
+            }
+
+            const action = possibleActions[0];
+            this.handlers[action](compiler);
+        });
+
+        if (compiler.compilers) {
```
alexander-akait, line 115: Maybe better implement same logic for single build and multi compiler mode, using:
```
const compilers = compiler.compilers ? compiler.compilers : [compiler];
```

So we avoid a lot of `if/else`

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
```diff
@@ -0,0 +1,242 @@
...
+        // Register Custom Hook for printing after clrscr
+        if (!compiler.hooks.beforeInteractiveStats) {
+            compiler.hooks = {
+                ...compiler.hooks,
+                beforeInteractiveStats: new SyncHook(),
+            };
+        }
+
+        // Clear for first run as well
+        clrscr();
```
alexander-akait, line 129: maybe bad idea, we can remove extra logs here (from external tools), better clean screen only on `beforeCompile` hooks

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
```diff
@@ -0,0 +1,242 @@
...
+        compiler.hooks.beforeInteractiveStats.call();
```
alexander-akait, line 130: Let's call `interactiveOutput`, so `beforeInteractiveOutput`

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
```diff
@@ -0,0 +1,242 @@
...
+            });
+        } else {
+            const helperPlugin = new InteractiveModeMultiCompilerHelperPlugin();
+
+            // Register helper plugin on each of child compiler
+            for (const childCompiler of this.compilers) {
+                helperPlugin.apply(childCompiler);
+            }
+
+            compiler.hooks.done.tap(this.name, () => {
```
alexander-akait, line 153: Maybe `afterDone`?

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
```diff
@@ -0,0 +1,242 @@
...
+            }
+            process.exit(0);
+        }
+
+        if (compiler.watching === undefined) return;
+        compiler.watching.close(() => {
+            process.exit(0);
+        });
+        return;
+    }
```
alexander-akait, line 182: No need to extra logic here always use `compiler.close()`, we need run it for single build for cache purpose too

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
```diff
@@ -0,0 +1,242 @@
...
+
+        if (compiler.watching.suspended) {
+            spawnCommand('already stoped', false);
+            return;
+        }
+
+        compiler.watching.suspend();
+        spawnCommand('stoped watching', false);
+    }
+}
```
alexander-akait, line 241: Can you clarify logic here?

packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -451,6 +451,14 @@ class WebpackCLI {
...
                 description: 'Print compilation progress during build.',
             },
+            {
+                name: 'interactive',
+                usage: '--interactive',
+                type: Boolean,
+                alias: 'i',
+                multiple: false,
+                description: 'Enable webpack interactive mode',
+            },
```
alexander-akait, line 461: Ideally we should support two arguments:
- `--interactive`, just clean screen before compilation and output stats (no extra information on screen)
- `--interactive=verbose`, like above + interface with information about keys and what they do

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
```diff
@@ -0,0 +1,242 @@
+const readline = require('readline');
+const { red, green, cyanBright, bold } = require('colorette');
+const logger = require('../utils/logger');
```
alexander-akait, line 3: For plugin please use `compiler.getInfrastructureLogger` https://webpack.js.org/api/logging/#examples-of-how-to-get-and-use-webpack-logger-in-loaders-and-plugins

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
```diff
@@ -0,0 +1,242 @@
...
+
+        // Configure stdin for keypress event
+        const stdin = process.stdin;
+        stdin.setEncoding('utf-8');
+        stdin.setRawMode(true);
+        readline.emitKeypressEvents(stdin);
+
+        // Configure keypress event for actions
+        const actions = Object.keys(this.keys);
+        stdin.on('keypress', (_, actionKey) => {
```
alexander-akait, line 98: `CTRL + C` and `CTRL + D` should work

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
```diff
@@ -0,0 +1,242 @@
...
+                return this.keys[action] === actionKey.name;
+            });
+
+            if (possibleActions.length === 0) {
+                return;
+            }
+
+            if (possibleActions.length > 1) {
+                throw new Error('Multiple actions are provided for same key');
+            }
```
alexander-akait, line 109: I don't think we need handle this, because it is internal, right?

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
alexander-akait: Other idea to avoid `isCleanScreen`, `allDone` and etc, to keep status, i.e. `this.status = 'before-compile' | 'compile' | 'after-done' | 'pause' | 'resume'`; (maybe be more, maybe you will know better here), so the plugin will know current status and will not do extra things, also using the one status variables will be comfortably to handle multi compiler mode

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
```diff
@@ -0,0 +1,242 @@
...
+    constructor() {
+        this.isMultiCompiler = false;
+        this.compilers = undefined;
+        this.name = 'InteractiveModePlugin';
+        this.watching = undefined;
+        this.keys = {
+            quit: 'q',
+            stop: 's',
+            start: 'w',
+        };
```
alexander-akait, line 76: When you run `webpack build` (just webpack) we should not register keys, just clear screen and run compilation, this valid only for watch (i.e. `webpack serve` and `webpack watch`)

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
```diff
@@ -0,0 +1,242 @@
...
+                    return result && !childCompiler.watching.running;
+                }, true);
+
+                if (!allDone) return;
+
+                setTimeout(() => {
+                    console.log('\n\n');
+                    spawnCommand('all compilations completed', true);
+                }, 100);
+            });
```
alexander-akait, line 164: Why we need `setTimeout` here? if we need wait v8 look we can use `process.nextTick(() => {})`

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
alexander-akait: Also do not forget handle errors here

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
rishabh3112: Thanks, didn't knew about about nextTick. Updated accordingly.

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
rishabh3112: AfterDone hook is unavaible in MultiCompiler. Here are available hooks:
https://github.com/webpack/webpack/blob/81f244ba2eb51ab0cc09ee136d1765da5617270f/lib/MultiCompiler.js#L58-L73

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
rishabh3112: Compiler recieved here is a multicompiler if multiple compilers are specified in config.

rishabh3112:
# TODO
- Move to manually performing operations on child compilers, instead of multi-compiler
- Implement two modes `standard` and `verbose`
- Handle `Ctrl + C` and `Ctrl + D` in raw mode

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
rishabh3112: Yup, done!

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
rishabh3112: I am hoping to expose mapping of keys to actions to users in future.

packages/webpack-cli/lib/webpack-cli.js
rishabh3112: Done

rishabh3112:
Need review 😄

alexander-akait:
@rishabh3112 Can we exclude from this PR unnecessary changes (changelog/docs/etc)

rishabh3112:
@alexander-akait did that now 😄

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
```diff
@@ -0,0 +1,207 @@
...
+const { red, green, cyanBright, bold, gray } = require("colorette");
+const { SyncHook } = require("tapable");
+const logger = require("../utils/logger");
+
+let version;
+try {
+    version = require("webpack").version;
+} catch (err) {
+    process.exit(2);
+}
```
alexander-akait, line 11: Let's do not read webpack here, we support only v5 so `compiler.webpack` should be used

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
```diff
@@ -0,0 +1,207 @@
...
+        for (const childCompiler of compilers) {
+            if (!childCompiler.watching.suspended) {
+                childCompiler.watching.suspend();
+            }
+        }
+        spawnCommand("stoped watching", false, false, this.verbose);
+        return;
+    }
+}
+module.exports = InteractiveModePlugin;
```
alexander-akait, line 207: A lot of uncovered branches let's add more tests

alexander-akait (commented):
Good looks good, let's add more tests

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
rishabh3112: Okay, on it 👍🏽

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
rishabh3112: Will do.

rishabh3112: Updated tests, let's see coverage once Ci is green.

rishabh3112: CoverageScreen Shot 2021-06-25 at 11 25 08 AMis now fine I guess.

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
```diff
@@ -0,0 +1,201 @@
...
+const { red, green, cyanBright, bold, gray } = require("colorette");
+const { SyncHook } = require("tapable");
+const logger = require("../utils/logger");
+
+/**
+ * Displays command space at bottom of screen
+ * @param {string} msg message to print with command
+ * @param {boolean} status currently watching or not
+ */
+const spawnCommand = (msg, status, toClear = false, verbose = false) => {
```
alexander-akait, line 11: Let's do it method or static method

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
```diff
@@ -0,0 +1,201 @@
...
+
+    readline.cursorTo(process.stdout, 0, totalRows - lines - 1);
+};
+
+/**
+ * Clear the whole terminal
+ */
+const clrscr = () => {
+    process.stdout.write("\x1B[2J\x1B[3J\x1B[H");
+};
```
alexander-akait, line 43: Also let's to it method

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
```diff
@@ -0,0 +1,201 @@
...
+
+/**
+ * Interactive Mode plugin
+ */
+class InteractiveModePlugin {
+    constructor(mode = "verbose") {
```
alexander-akait, line 49: Ideally plugins should accept `options = {}`, because we can have more  options in future

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
```diff
@@ -0,0 +1,201 @@
...
+        this.handlers = {
+            quit: this.quitHandler.bind(this),
+            stop: this.stopHandler.bind(this),
+            start: this.startHandler.bind(this),
+        };
+        this.logger = undefined;
+        this.verbose = mode === "verbose";
+
+        // hide cursor
+        process.stdout.write("\u001B[?25l");
```
alexander-akait, line 65: Avoid doing something extra in configuration, move it to apply

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
rishabh3112: Okay, will do

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
rishabh3112: On it.

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
rishabh3112: Done

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
rishabh3112: Done

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
alexander-akait: Let's fix it too

alexander-akait (commented):
Plugins looks better, let's finish some notes and I think we can merge it :+1:

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
rishabh3112: Sure, on it.

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
rishabh3112: Done.

packages/webpack-cli/lib/plugins/InteractiveModePlugin.js
rishabh3112: done

rishabh3112:
Force exiting jest because `execa` is not exiting cleanly even after it is killed correctly which causes CI to run indefinitely.

package.json
```diff
@@ -37,7 +37,7 @@
     "test": "jest --reporters=default",
     "test:smoketests": "nyc node smoketests",
     "test:coverage": "nyc --no-clean --require ts-node/register jest",
-    "test:cli": "jest test --reporters=default",
+    "test:cli": "jest test --reporters=default --forceExit",
```
alexander-akait, line 40: it is very bad, it means you don't close something, we need to fix code/test

rishabh3112: Yup, it is bad. Had looked into it earlier as well, I came across a bug in `execa` itself, will be exploring more.

rishabh3112: Removed it for now.

rishabh3112:
@webpack/cli-team need help, unable to figure out why failing on windows. Help from anyone with windows would be appreciated!

rishabh3112:
@alexander-akait ?

alexander-akait (commented):
Still in my roadmap :disappointed: All looks good, but I am afraid only about `compiler.hooks.beforeInteractiveOutput`, I think better solution like here https://github.com/webpack/webpack/blob/60d8968cd48cf1677df5f50ed878c6a912e9d692/lib/wasm-async/AsyncWebAssemblyModulesPlugin.js#L61

rishabh3112:
> think better solution like here https://github.com/webpack/webpack/blob/60d8968cd48cf1677df5f50ed878c6a912e9d692/lib/wasm-async/AsyncWebAssemblyModulesPlugin.js#L61

Can you explain a bit the issue? should I be exposing hooks like this function does?

rishabh3112:
Lets wait for all tests to be green now.

rishabh3112:
@alexander-akait we can merge this now imo.

rishabh3112:
@webpack/cli-team This Pull Request has been waiting for a review for 5 months now but haven't got one. Now, the master branch is to far ahead, so solve merge conflicts this only (which by the way I had been doing for over a year). Nonetheless, I am starting a fresh on this feature and closing this PR for above mentioned reason.

P.S.
With 157 commits ranging over 2 years+ time, this have been a journey ⭐

PR #1804, opened 2020-09-17
fix: show warning if bail and watch are used together

### LINKED ISSUE #1794: bail = true should not kill watch mode on ModuleNotFound error

bug

When `bail = true`, a ModuleNotFound error during a change of the files will cause the entire watch process to either abort or get stuck. Subsequent rebuilds will not happen even if the file changes.

* run webpack in watch mode
* add `import './non-existant-module'` to any file
* webpack rebuild fails or exits
* remove the import
* rebuild is not triggered either due to webpack exit (3.10 and below) or due to hang (3.11)

node version: 8.9.4

Webpack 3.11 made one step towards fixing this with https://github.com/webpack/webpack/pull/5797

However even though the process does not die, rebuilds stop happening.

snitin315:
maybe fix

Yes
No
**Summary**
Fixes #1794 

 show a warning if `bail` and `watch` are used together with either config or as cli args.

No

**Other information**
No

### DIFF
--- a/packages/webpack-cli/lib/groups/ConfigGroup.js
+++ b/packages/webpack-cli/lib/groups/ConfigGroup.js
@@ -127,6 +127,12 @@ class ConfigGroup extends GroupHelper {
             newOptionsObject['options'] = configOptions;
         }
 
+        //warn the user if bail and watch both are used together
+        const { bail, watch } = newOptionsObject['options'];
+        if (bail && watch) {
+            logger.warn('You are using "bail" with "watch". "bail" will still exit webpack when the first error is found.');
+        }
+
         if (configOptions && configPath.includes('.webpack')) {
             const currentPath = configPath;
             const parentContext = dirname(currentPath).split(sep).slice(0, -1).join(sep);
--- a/packages/webpack-cli/lib/groups/ConfigGroup.js
+++ b/packages/webpack-cli/lib/groups/ConfigGroup.js
@@ -127,12 +127,6 @@ class ConfigGroup extends GroupHelper {
             newOptionsObject['options'] = configOptions;
         }
 
-        //warn the user if bail and watch both are used together
-        const { bail, watch } = newOptionsObject['options'];
-        if (bail && watch) {
-            logger.warn('You are using "bail" with "watch". "bail" will still exit webpack when the first error is found.');
-        }
-
         if (configOptions && configPath.includes('.webpack')) {
             const currentPath = configPath;
             const parentContext = dirname(currentPath).split(sep).slice(0, -1).join(sep);
--- a/packages/webpack-cli/lib/utils/Compiler.js
+++ b/packages/webpack-cli/lib/utils/Compiler.js
@@ -117,6 +117,11 @@ class Compiler {
             return interactive(options, outputOptions);
         }
 
+        //warn the user if bail and watch both are used together
+        if (this.compiler.options.bail && outputOptions.watch) {
+            logger.warn('You are using "bail" with "watch". "bail" will still exit webpack when the first error is found.');
+        }
+
         if (this.compiler.compilers) {
             this.compiler.compilers.forEach((comp, idx) => {
                 this.setUpHookForCompilation(comp, outputOptions, options[idx]);
--- a/packages/webpack-cli/lib/utils/Compiler.js
+++ b/packages/webpack-cli/lib/utils/Compiler.js
@@ -1,5 +1,6 @@
 const webpack = require('webpack');
 const logger = require('./logger');
+const bailAndWatchWarning = require('./warnings/bailAndWatchWarning');
 const { CompilerOutput } = require('./CompilerOutput');
 
 class Compiler {
@@ -117,16 +118,13 @@ class Compiler {
             return interactive(options, outputOptions);
         }
 
-        //warn the user if bail and watch both are used together
-        if (this.compiler.options.bail && outputOptions.watch) {
-            logger.warn('You are using "bail" with "watch". "bail" will still exit webpack when the first error is found.');
-        }
-
         if (this.compiler.compilers) {
             this.compiler.compilers.forEach((comp, idx) => {
+                bailAndWatchWarning(comp); //warn the user if bail and watch both are used together
                 this.setUpHookForCompilation(comp, outputOptions, options[idx]);
             });
         } else {
+            bailAndWatchWarning(this.compiler);
             this.setUpHookForCompilation(this.compiler, outputOptions, options);
         }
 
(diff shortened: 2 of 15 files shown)


packages/webpack-cli/lib/bootstrap.js
```diff
@@ -71,6 +71,10 @@ async function runCLI(cliArgs) {
             parsedArgsOpts.entry = entry;
         }
 
+        if (parsedArgsOpts.bail && parsedArgsOpts.watch) {
+            logger.warn('You are using "bail" with "watch". "bail" will still exit webpack when the first error is found.');
+        }
```
evilebottnawi, line 76: We should look at `compiler.options.bail` and `parsedArgsOpts.watch`, because other plugins can change options

packages/webpack-cli/lib/groups/ConfigGroup.js
```diff
@@ -127,6 +127,11 @@ class ConfigGroup extends GroupHelper {
             newOptionsObject['options'] = configOptions;
         }
 
+        //warn the user if bail and watch both are used together
+        if (configOptions.bail && configOptions.watch) {
+            logger.warn('You are using "bail" with "watch". "bail" will still exit webpack when the first error is found.');
+        }
```
evilebottnawi, line 133: Ideally we should only one place for this, no need t duplicate code

packages/webpack-cli/lib/utils/Compiler.js
```diff
@@ -118,7 +118,7 @@ class Compiler {
         }
 
         //warn the user if bail and watch both are used together
-        if (this.compiler.options.bail && outputOptions.watch) {
+        if (this.compiler.options.bail && (outputOptions.watch || options.watch)) {
```
evilebottnawi, line 121: `this.compiler.options.bail && this.compiler.watchMode`

packages/webpack-cli/lib/utils/Compiler.js
```diff
@@ -117,6 +117,11 @@ class Compiler {
             return interactive(options, outputOptions);
         }
 
+        //warn the user if bail and watch both are used together
+        if (!this.compiler.compilers && this.compiler.options.bail && this.compiler.watch) {
+            logger.warn('You are using "bail" with "watch". "bail" will still exit webpack when the first error is found.');
+        }
```
evilebottnawi, line 123: `this.compiler` can be array when you use multi compiler mode, we need handle this

packages/webpack-cli/lib/utils/warnings/bailAndWatchWarning.js
```diff
@@ -0,0 +1,7 @@
+const logger = require('../logger');
+
+module.exports = (compiler) => {
+    if (compiler.options.bail && compiler.watch) {
```
evilebottnawi, line 4: Looks good, but why we don't use `compiler.watchMode`?

snitin315: Don't know why but it isn't working with `compiler.watchMode`.

evilebottnawi: https://github.com/webpack/webpack/blob/master/lib/Compiler.js#L336 is always `true`, we need to use `watchMode`, can you try it again?

snitin315: Not working. I logged `compiler` in the console and found `watchMode: false` instead 😕 , 

screenshot-github-com-webpack-webpack-cli-pull-1804-1600434195860

evilebottnawi: oh, yes, I was wrong, `watchMode` is true only when we run `compiler.watch()`, so we need use `watch`

evilebottnawi: Maybe we should use `compiler.options.watch`?

snitin315: with `compiler.options.watch` the process exits too soon without printing the warning in case an error is found.

Screenshot at 2020-09-18 18-54-11

evilebottnawi: We need to search way to detect watch mode, because `compiler.watch` is method and it is always true

snitin315: Ah yes, then we have `compiler.options.bail && (outputOptions.watch || options.watch)`

evilebottnawi: `outputOptions.watch` is `true` when we run `compiler.watch`

snitin315: > `compiler.options.bail && (outputOptions.watch || compiler.options.watch)` 

`outputOptions,watch` is true when we pass `--watch` via `CLI`, so we need to check the existence of `outputOptions.watch` else `webpack --bail --watch` will not log the warning. The rest of all cases will log the warning.

snitin315: Fixed, the issue was with `--watch` flag it was not setting `compiler.options.watch` equal to `true`. Now works perfect.

Screenshot at 2020-09-19 15-37-23

test/bail/bail-and-watch-warning.test.js
```diff
@@ -0,0 +1,19 @@
...
+        expect(stdout).toBeTruthy();
+    });
+
+    it('should log warning in case of multiple compilers', () => {
+        const { stderr, stdout } = run(__dirname, ['-c', 'multi-webpack.config.js']);
+
+        expect(stderr).toContain(`You are using "bail" with "watch". "bail" will still exit webpack when the first error is found.`);
+        expect(stdout).toBeTruthy();
+    });
+});
```
evilebottnawi, line 19: Can we add tests where `bail: false` and `watch: true`/`watch: false` and `bail: true` and check there is not warnings/errors?

snitin315: Added

evenstensberg (dismissed):

packages/webpack-cli/lib/utils/warnings/bailAndWatchWarning.js
```diff
@@ -0,0 +1,7 @@
+const logger = require('../logger');
+
+module.exports = (compiler) => {
```
rishabh3112, line 3: Assign this to variable and then assign it to `module.exports`.
Add a comment about function(or file) as well.

rishabh3112 (dismissed):
Left a small suggestion.
Rest looks good to me.

evenstensberg (approved):

evilebottnawi (approved):

jamesgeorge007 (approved):

PR #1811, opened 2020-09-19
refactor: config group to pure function and move away from group design

anshumanv:

refactor 

Already present

NA

**Summary**
- removed redundant entries from cli flag file
- Refactored config group to pure function

No

**Other information**

https://github.com/webpack/webpack-cli/pull/1765#discussion_r475077508


evenstensberg (commented):
Let's keep entry default option

anshumanv:
> Let's keep entry default option

It's unused

evilebottnawi (commented):
Also default value for `entry` is invalid, so :+1:

anshumanv:
We need to run ConfigGroup needless whether config args are present or nay, I think we should make it a pure function, this should be a good start to move away from group design.

Thoughts? @evilebottnawi @evenstensberg

evilebottnawi:
> We need to run ConfigGroup needless whether config args are present or nay, I think we should make it a pure function, this should be a good start to move away from group design.

Agree, we don't need group design and run only necessary code

rishabh3112 (dismissed):

evenstensberg (approved):

snitin315 (commented):
Good job 🔥

snitin315 (approved):

packages/webpack-cli/lib/groups/ConfigGroup.js
```diff
@@ -29,6 +29,12 @@ const modeAlias = {
     development: 'dev',
 };
 
+let opts = {
+    outputOptions: {},
+    options: {},
+};
+
+// Return a list of default configs in various forrmats
```
jamesgeorge007, line 37: Typo

packages/webpack-cli/lib/groups/ConfigGroup.js
```diff
@@ -58,162 +64,163 @@ const getConfigInfoFromFileName = (filename) => {
...
     }
+    return newOptionsObject;
+};
 
-    async run() {
-        await this.resolveConfigFiles();
-        await this.resolveConfigMerging();
-        return this.opts;
+const resolveConfigMerging = async (args) => {
+    const { merge } = args;
```
jamesgeorge007, line 201: Isn't it better to use destructuring if the function takes in a single argument?

anshumanv: neh I'll prefer this, looks cleaner

snitin315 (approved):

packages/webpack-cli/lib/groups/ConfigGroup.js
```diff
@@ -58,162 +64,163 @@ const getConfigInfoFromFileName = (filename) => {
...
+
+const handleConfigResolution = async (args) => {
+    const parsedArgs = arrayToObject(args);
+    await resolveConfigFiles(parsedArgs);
+    await resolveConfigMerging(parsedArgs);
+    return opts;
+};
 
-module.exports = ConfigGroup;
+module.exports = handleConfigResolution;
```
evilebottnawi, line 226: I think we need move this from `ConfigGroup.js` file, to be honestly we don't need `groups` files :smile: Do you want to do it in this PR?

anshumanv: Yep lets do it here

anshumanv: Or maybe let's migrate all groups first then we can move them all from `groups` to `argResolvers`?

evilebottnawi: Maybe we put each function in own file, anyway let's refactor firstly and when we can think about places

anshumanv: yep once all groups are migrated, we can organize it neatly

evilebottnawi (approved):
Good start

PR #1820, opened 2020-09-21
refactor: move zero-config from group design

anshumanv:

refactor 

Changed

NA

**Summary**
- Refactor zero config group to pure fn, remove it from group design

No

**Other information**


packages/webpack-cli/lib/groups/resolveMode.js
```diff
@@ -0,0 +1,25 @@
...
+    let finalMode;
+    /**
+     * It determines the mode to pass to webpack compiler
+     * @returns {string} The mode
+     */
+    if (NODE_ENV && (NODE_ENV === PRODUCTION || NODE_ENV === DEVELOPMENT)) {
+        finalMode = NODE_ENV;
+    } else if (mode) {
+        finalMode = mode;
+    }
```
evilebottnawi, line 18: Here bug :smile: 

we always set `finalMode` to `NODE_ENV`, even you specify `mode`, it is big bug

```js
if (!mode && NODE_ENV && (NODE_ENV === PRODUCTION || NODE_ENV === DEVELOPMENT)) {
   finalMode = NODE_ENV;
}
```

anshumanv: Damn, it has been this way before, good that we caught this during refactor. We should have a test for this 😄

evilebottnawi: Can you fix it here?

anshumanv: yes wip 😄

anshumanv: Ah so `mode` has a default value set to `production` hence !mode condition is not relevant here so its this way

anshumanv: Hence supplied `NODE_ENV` takes precedence

anshumanv: Right thanks for confirming, fixed 💯

evilebottnawi: Does `args` contain only arguments? Because we should respect the value from config and do not override it

anshumanv: We already respect, added test 👍

evilebottnawi (approved):

rishabh3112 (approved):

PR #1824, opened 2020-09-23
fix: mode behaviour

anshumanv:

fix

Yes

Yes

**Summary**
Fix mode priority 
- mode form flag
- config
- NODE_ENV

Nay

**Other information**

### DIFF
--- a/packages/webpack-cli/lib/groups/resolveMode.js
+++ b/packages/webpack-cli/lib/groups/resolveMode.js
@@ -8,15 +8,17 @@ Mode priority:
     - Mode form NODE_ENV
 */
 
-const resolveMode = (args, configOptions) => {
+/**
+ *
+ * @param {string} mode - mode flag value
+ * @param {Object} configObject - contains relevant loaded config
+ */
+const assignMode = (mode, configObject) => {
     const {
         env: { NODE_ENV },
     } = process;
-    const { mode } = args;
-    const { mode: configMode } = configOptions;
-
+    const { mode: configMode } = configObject;
     let finalMode;
-
     if (mode) {
         finalMode = mode;
     } else if (configMode) {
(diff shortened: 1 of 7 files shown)


packages/webpack-cli/lib/groups/resolveMode.js
```diff
@@ -1,21 +1,30 @@
...
+
+    if (mode) {
+        finalMode = mode;
+    } else if (configMode) {
+        finalMode = configMode;
+    } else if (NODE_ENV && (NODE_ENV === PRODUCTION || NODE_ENV === DEVELOPMENT)) {
         finalMode = NODE_ENV;
     } else {
-        finalMode = mode || PRODUCTION;
+        finalMode = PRODUCTION;
```
evilebottnawi, line 27: No need this, webpack do it itself (default is production)

evilebottnawi (commented):
Also can you add multicompiler tests

packages/webpack-cli/lib/groups/resolveMode.js
anshumanv: gotcha

anshumanv:
> Also can you add multicompiler tests

`-c 1.js -c 2.js`

which config mode should be priority? @evilebottnawi

evilebottnawi:
With `--mode=production` both configurations should set `mode: production`, without `--mode=production` but with `mode: 'production'` in config should keep values, with `NODE_ENV` and without `mode: 'production'` should use `NODE_ENV`, same logic, maybe better to use `module: development` to test, because default value is `production`

packages/webpack-cli/lib/groups/resolveMode.js
anshumanv: Need this because https://github.com/webpack/webpack-cli/issues/1678 hence added

anshumanv:
> With --mode=production both configurations should set mode: production, without --mode=production but with mode: 'production' in config should keep values, with NODE_ENV and without mode: 'production' should use NODE_ENV, same logic, maybe better to use module: development to test, because default value is production

Needed more work than I thought, looks good now

packages/webpack-cli/lib/groups/resolveMode.js
evilebottnawi: Okay, let's refactor it in future

evilebottnawi (approved):

packages/webpack-cli/lib/groups/resolveMode.js
anshumanv: Sure sure

evilebottnawi (approved):

rishabh3112 (approved):

PR #1832, opened 2020-09-24
migrate output group

anshumanv:

refactor

Yes

No need

**Summary**
- Migrate output group out of group design

No

**Other information**


packages/webpack-cli/lib/groups/resolveOutput.js
```diff
@@ -0,0 +1,32 @@
+const path = require('path');
+
+const parseDirectory = (metaData) => {
+    return {
+        // filename: DEFAULT_FILENAME,
```
evilebottnawi, line 5: What is this?

packages/webpack-cli/lib/groups/resolveOutput.js
```diff
@@ -0,0 +1,32 @@
...
+        const outputInfo = path.parse(output);
+        if (!outputInfo.ext.length) {
+            finalOptions.options.output = parseDirectory(outputInfo);
+        } else {
+            finalOptions.options.output.path = path.resolve(outputInfo.dir);
+            finalOptions.options.output.filename = outputInfo.base;
+        }
+    }
+    return finalOptions;
+};
```
evilebottnawi, line 30: Why we need it? I can't understand. Why do not set values from `args`?

packages/webpack-cli/lib/groups/resolveOutput.js
anshumanv: not sure, old code 😄 

Lemme try getting rid of this

packages/webpack-cli/lib/groups/resolveOutput.js
anshumanv: old code, let's try setting from args directly

evenstensberg (dismissed):
Good start

anshumanv:
@evilebottnawi @evenstensberg simplified it a little, PTAL if it looks alright

packages/webpack-cli/lib/groups/resolveOutput.js
```diff
@@ -0,0 +1,21 @@
...
+ */
+const resolveOutput = (args) => {
+    const { output } = args;
+    const finalOptions = {
+        options: { output: {} },
+        outputOptions: {},
+    };
+    if (output) {
+        const { dir, base } = path.parse(output);
+        finalOptions.options.output.path = path.resolve(dir);
```
anshumanv, line 15: this is because `dir` can be both an absolute or relative route

evilebottnawi: To be honestly we have a lot of flags where we can use path, especial in webpack@5, can you open new issue about it, for example `--cache-location`

evilebottnawi (approved):
Good job, please read comment

packages/webpack-cli/lib/groups/resolveOutput.js
anshumanv: here https://github.com/webpack/webpack-cli/issues/1841

packages/webpack-cli/lib/groups/resolveOutput.js
```diff
@@ -0,0 +1,21 @@
...
+const resolveOutput = (args) => {
+    const { output } = args;
+    const finalOptions = {
+        options: { output: {} },
+        outputOptions: {},
+    };
+    if (output) {
+        const { dir, base, ext } = path.parse(output);
+        finalOptions.options.output.path = ext.length === 0 ? path.resolve(dir, base) : path.resolve(dir);
+        if (ext.length > 0) finalOptions.options.output.filename = base;
```
evilebottnawi, line 16: Why we check on `ext`? I think `path.resolve` is enough for test

anshumanv: if we pass absolute path name say `/user/code/dist` then `path.parse` will result in

`{dir: '/user/code', base: 'dist'}`

hence to support that we need this

evilebottnawi: path.isAbsolute(output.path) ? output.path ? path.resolve(process.cwd(), 'output.path')

evilebottnawi: Oh, do you mean `--output=/path/to/file.js`? If yes, it is invalid

anshumanv: yep exactly 👍

evilebottnawi: Output should be only directory, not filename

anshumanv: I think we can support both, won't that be useful for users?

evilebottnawi: It is invalid because we will specify HTML and CSS as entry points in the near future + `dependOn`, if you have HTML entrypoint and use `--output=/path/to/file.js` it was wrong, Ideally we should remove `--output` in favor `--output-path`, also this one will be consistent with webpack@5 cli args

anshumanv: makes sense, will update 👍

anshumanv: actually, we need to change a lot of tests because of this behaviour, let's do this after?

evilebottnawi: Sounds fine, can you open new issue and focus on it?

evilebottnawi (approved):

packages/webpack-cli/lib/groups/resolveOutput.js
anshumanv: https://github.com/webpack/webpack-cli/issues/1844

snitin315 (approved):
Looks Good!

rishabh3112 (approved):

PR #1361, opened 2020-03-22
feat: enable analyze flag

anshumanv:
**What kind of change does this PR introduce?**
feat

**Did you add tests for your changes?**
Not yet, will add tests after we decide that this is needed.

**If relevant, did you update the documentation?**
Not yet.

**Summary**
- re-enable analyze flag which launches webpack-bundle-analyzer to analyze the output bundle.
- Prevents extra steps of installing the plugin, adding it to the config and then running build again i.e convenience.

**Does this PR introduce a breaking change?**
No

**Other information**
Will add tests after we decide to go forward with this.

PR #1837, opened 2020-09-25
migrate basic group

anshumanv:
refactor

Yeaaa

No need

**Summary**
- migrate basic group out of group design

Nayyy

**Other information**


packages/webpack-cli/lib/groups/basicResolver.js
```diff
@@ -0,0 +1,52 @@
...
+    };
+    Object.keys(args).forEach((arg) => {
+        if (WEBPACK_OPTION_FLAGS.includes(arg)) {
+            finalOptions.outputOptions[arg] = args[arg];
+        }
+        // TODO: to add once webpack bundle analyzer is installed, which is not at the moment
+        // if (arg == 'analyze') {
+        // const { BundleAnalyzerPlugin } = require('webpack-bundle-analyzer');
+        // this.opts.options.plugins = [new BundleAnalyzerPlugin()];
+        // }
```
evilebottnawi, line 32: Let's remove this, we need to open new issue and do some work on `webpack-bundle-analyzer` side

packages/webpack-cli/lib/utils/arg-utils.js
```diff
@@ -18,6 +21,37 @@ function arrayToObject(arr) {
...
+            LOOKUP_PATHS.unshift(filename);
+        }
+        LOOKUP_PATHS.forEach((p) => {
+            const lookUpPath = join(process.cwd(), ...p.split('/'));
+            if (existsSync(lookUpPath) && !configPath) {
+                configPath = lookUpPath;
+            }
+        });
+    }
+    return configPathExists ? predefinedConfigPath : configPath;
```
evilebottnawi, line 51: To be honestly it is wrong, webpack uses resolved to resolve `js`/`ts`/`cjs`/`mjs` and other extensions, I think It should be removed

packages/webpack-cli/lib/groups/basicResolver.js
anshumanv: yep, did some work https://github.com/webpack/webpack-cli/pull/1361 will revisit later

packages/webpack-cli/lib/utils/arg-utils.js
evilebottnawi: I think all test will be passed without it

packages/webpack-cli/lib/groups/basicResolver.js
anshumanv: What's the work needed on `webpack-bundle-analyzer` side @evilebottnawi ?

Maybe I'll open an issue about it

packages/webpack-cli/lib/utils/arg-utils.js
anshumanv: makes sense, let's simply do `path.resolve`

packages/webpack-cli/lib/groups/basicResolver.js
evilebottnawi: Let's do it late, just open a new issue

packages/webpack-cli/lib/groups/basicResolver.js
```diff
@@ -0,0 +1,52 @@
...
+    })
+    .reduce((result, flagObject) => {
+        result.push(flagObject.name);
+        if (flagObject.alias) {
+            result.push(flagObject.alias);
+        }
+        return result;
+    }, []);
+
+console.log(WEBPACK_OPTION_FLAGS);
```
evenstensberg, line 17: hm

anshumanv: removed

packages/webpack-cli/lib/utils/arg-utils.js
anshumanv: Ah so we don't even need to do path.resolve since core does this already for relative paths, we just need to pass the entry

anshumanv:
@evilebottnawi PTAL if this is what you had in mind

anshumanv:
CI broken with webpack@next 😞

anshumanv:
@snitin315 can you test the issue with this branch now?

evilebottnawi:
Yes, because we need remove all unnecessary stuff, there are a lot of problems with this due bad code from history

can you look at the problem?

evilebottnawi:
https://github.com/webpack/webpack/issues/11553

evilebottnawi:
We need check bug from webpack-cli or regression from webpack

anshumanv:
@evilebottnawi I think regression in webpack rc-1, works fine with rc-0

snitin315 (commented):
Can you add a test case for a non-js entry point too?

anshumanv:
Done

packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -217,11 +205,13 @@ class WebpackCLI extends GroupHelper {
...
     _handleDefaultEntry() {
-        if (!this.basicGroup) {
-            const BasicGroup = require('./groups/BasicGroup');
-            this.basicGroup = new BasicGroup();
+        let defaultEntry;
+        const rootIndexPath = path.resolve('index.js');
+        if (existsSync(rootIndexPath)) {
+            defaultEntry = './index.js';
+        } else {
+            defaultEntry = './src/index.js';
```
snitin315, line 213: Do we have tests for both of the default entries i.e, `./index.js` and `./src/index.js` ?

evilebottnawi: This should be solved on webpack side, here invalid solution, because index can have `cjs`/`ts`/`mjs` and etc extensions

anshumanv: Need to change a bunch of tests again because of current behaviour, let's fix this separately? Here's the issue for any discussion - https://github.com/webpack/webpack-cli/issues/1852

evilebottnawi (approved):
/cc @webpack/cli-team

snitin315 (approved):

Issue #1844, opened 2020-09-28
Bug: Output flag should only set path


anshumanv:
**Describe the bug**

@evilebottnawi : It is invalid because we will specify HTML and CSS as entry points in the near future + dependOn, if you have HTML entrypoint and use --output=/path/to/file.js it was wrong, Ideally we should remove --output in favor --output-path, also this one will be consistent with webpack@5 cli args

**What is the current behavior?**

Passing a file in output flag sets filename object in output

**To Reproduce**
Pass `/path/to/file.js`

**Expected behavior**
It should only set `output.path`, filename should only be configured using config

**Screenshots**

**Please paste the results of `webpack-cli info` here, and mention other relevant information**

**Additional context**
https://github.com/webpack/webpack-cli/pull/1832#discussion_r495921619

evilebottnawi:
I think better to remove `--output` in favor `--output-path`

anshumanv:
Yep will do that in PR 👍

PR #1855, opened 2020-10-03
fix: only set output path on passing flag

anshumanv:
- convert `output` to `output-path`
- only assign output path when supplied

Yes

Yes

**Summary**
Previously we were assigning output file and path which was not correct so fixed that and converted output flag to output-path

Yes

**Other information**
Fixes https://github.com/webpack/webpack-cli/issues/1844


evilebottnawi (approved):
/cc @webpack/cli-team 

Good work!

snitin315 (approved):

Rob--W:
This should have been called `feat: Rename --output flag to --output-path`.

It is a breaking change, and quite a late one to make after beta, before RC...

Rob--W:
Actually, it seems that `--output` has been split into `--output-path` (=output directory) and `--output-filename` (=file name only). The ability to direct output to a specific absolute location has been lost.

evilebottnawi:
@Rob--W you have configurations so you can setup this, it is breaking change and described

Rob--W:
It seems that there is no way to pass an absolute file path via the command line any more, is there?
I fixed my use case by splitting the path and separately passing `--output-path` and `--output-filename`: https://github.com/mozilla/webextension-polyfill/pull/245/files

This limitation is strange though. Most tools that I can recall supporting a parameter to an output file accept an absolute path, whereas this tool now requires to separate flags. Is there any reason for this change?

evilebottnawi:
@Rob--W because path and filename is different thing, you can use palceholders in filename, so we don't know in `--output path/to/[hash][ext]` `[hash][ext]` is filename or placeholders, so we separate this

Rob--W:
`-o /path/to/plain/file/without/webpack/placeholders.js` is unambiguous. If the `[hash]` and `[ext]` are the primary reason for this breaking change, then I think that restoring support for `--output [absolute path here]` makes sense.

The meaning of the `-o` flag has also changed by this PR. It used to be a file path, now it's supposedly a directory path.

If you consciously accepted these breaking, backwards-incompatible changes, then feel free to ignore my comment.

PR #1874, opened 2020-10-05
fix: log error if --config-name is used without multiple configs

snitin315:
fix

Yes
NA
**Summary**
 log warning if --config-name is used without multiple configs

No

**Other information**
No

### DIFF
--- a/packages/webpack-cli/lib/groups/ConfigGroup.js
+++ b/packages/webpack-cli/lib/groups/ConfigGroup.js
@@ -163,21 +163,27 @@ const finalize = async (moduleObj, args) => {
         const newOptions = configOptions(formattedEnv, args);
         // When config function returns a promise, resolve it, if not it's resolved by default
         newOptionsObject['options'] = await Promise.resolve(newOptions);
-    } else if (Array.isArray(configOptions) && configName) {
-        // In case of exporting multiple configurations, If you pass a name to --config-name flag,
-        // webpack will only build that specific configuration.
-        const namedOptions = configOptions.filter((opt) => configName.includes(opt.name));
-        if (namedOptions.length === 0) {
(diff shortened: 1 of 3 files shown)


packages/webpack-cli/lib/groups/ConfigGroup.js
```diff
@@ -163,21 +163,27 @@ const finalize = async (moduleObj, args) => {
...
+            if (namedOptions.length === 0) {
+                logger.error(`Configuration with name "${configName}" was not found.`);
+                process.exit(2);
+            } else {
+                newOptionsObject['options'] = namedOptions;
+            }
         } else {
-            newOptionsObject['options'] = namedOptions;
+            logger.warn('Multiple configurations not found. Please use "--config-name" with multiple configurations.');
+            newOptionsObject['options'] = configOptions;
```
evilebottnawi, line 179: Why it is a warning?

snitin315: So that build is not interrupted and stats are printed.

snitin315: Or we can throw error and exit process too.

evilebottnawi: I think it should be error and we should exit, because we can't found config

anshumanv: Yep not exiting will set wrong expectations

snitin315: I will update 👍

snitin315: updated 👍🏻

evilebottnawi (approved):
/cc @webpack/cli-team

anshumanv (approved):

PR #1883, opened 2020-10-06
feat: CI - cache lerna dependencies

anshumanv:
feat

Not yet

NA

**Summary**
- https://github.com/actions/cache/blob/main/examples.md#node---lerna
- Try to use cached deps to reduce CI time

No

**Other information**

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -83,7 +83,7 @@ jobs:
           yarn bootstrap
 
       - name: Install webpack ${{ matrix.webpack-version }}
-        if: steps.cache-${{ matrix.webpack-version }}.outputs.cache-hit != 'true'
+        if: steps.cache.outputs.cache-hit != 'true'
         run: yarn add -W webpack@${{ matrix.webpack-version }}
 
       - name: Build
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -83,7 +83,6 @@ jobs:
           yarn bootstrap
 
       - name: Install webpack ${{ matrix.webpack-version }}
-        if: steps.cache.outputs.cache-hit != 'true'
         run: yarn add -W webpack@${{ matrix.webpack-version }}
 
       - name: Build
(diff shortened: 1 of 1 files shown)


anshumanv:
We're hitting cache it seems 
image

Redundant deps install time has significantly dropped 

/cc @evilebottnawi

anshumanv:
My bad, time has not dropped even though we're hitting the cache 😞

evilebottnawi:
@anshumanv strange, based on docs we don't hit caches on windows

anshumanv:
yep we're not hitting on windows, I checked on ubuntu and we're hitting there but install time has not come down

evilebottnawi:
@anshumanv yep, linux is very fast, macos and windows are slow, so we need to look at them

.github/workflows/nodejs.yml
```diff
@@ -67,7 +67,17 @@ jobs:
...
+        id: cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            node_modules
+            */*/node_modules
+          key: ${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}
+
       - name: Install dependencies
+        if: steps.cache.outputs.cache-hit != 'true'
```
anshumanv, line 80: @evilebottnawi we're good now, we needed another step to skip install deps when cache is present

evilebottnawi (commented):
We need improve key, because
- https://github.com/webpack/webpack-cli/pull/1883/checks?check_run_id=1216340642
- https://github.com/webpack/webpack-cli/pull/1883/checks?check_run_id=1216340623

have same key, but we install webpack@4 and webpack@5, I think it should be different

evilebottnawi:
After this we can skip webpack installation too

anshumanv:
> After this we can skip webpack installation too

I did then it didn't install webpack@5 and didn't run core tests, let's try improving key and see if it works

anshumanv:
@evilebottnawi updated key to have webpack version, now it skips webpack installation too and runs all tests fine

rishabh3112 (approved):

snitin315 (approved):
Nice job!

evilebottnawi (approved):

PR #1890, opened 2020-10-07
tests: improve watch tests and enable windows

rishabh3112:
**What kind of change does this PR introduce?**
tests

**Did you add tests for your changes?**
yes

**If relevant, did you update the documentation?**
Not applicable

**Summary**
Added test for which will test simulating user and should run on windows as well.

**Does this PR introduce a breaking change?**
No

**Other information**
No

PR #1891, opened 2020-10-07
chore(deps): move devDependencies from packages to root

snitin315:
chore, deps

NA
NA
**Summary**
move devDependencies from packages to root

no

**Other information**
no

### DIFF
--- a/packages/package-utils/package.json
+++ b/packages/package-utils/package.json
@@ -39,10 +39,6 @@
     "execa": "^4.0.0",
     "global-modules": "^2.0.0"
   },
-  "devDependencies": {
-    "@types/cross-spawn": "^6.0.1",
-    "@webpack-cli/package-utils": "^1.0.1-rc.1"
-  },
   "peerDependenciesMeta": {
     "webpack": {
       "optional": true
--- a/packages/serve/package.json
+++ b/packages/serve/package.json
@@ -9,9 +9,6 @@
     "access": "public"
   },
   "license": "MIT",
-  "devDependencies": {
-    "webpack-dev-server": "3.10.3"
-  },
   "peerDependencies": {
     "webpack-cli": "4.x.x"
   },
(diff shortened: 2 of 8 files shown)


packages/serve/package.json
```diff
@@ -9,9 +9,6 @@
     "access": "public"
   },
   "license": "MIT",
-  "devDependencies": {
-    "webpack-dev-server": "3.10.3"
```
anshumanv, line 13: I think we can let this stay here, let's only move the ones used in >1 package

snitin315: only `serve` and `package-utils` have `devDependencies` field in `package.json`.

packages/package-utils/package.json
```diff
@@ -39,10 +39,6 @@
     "execa": "^4.0.0",
     "global-modules": "^2.0.0"
   },
-  "devDependencies": {
-    "@types/cross-spawn": "^6.0.1",
-    "@webpack-cli/package-utils": "^1.0.1-rc.1"
-  },
```
snitin315, line 45: Don't know why `@webpack-cli/package-utils` is a dev dep of `@webpack-cli/package-utils` 😄

packages/serve/package.json
rishabh3112: We should move all to root imo, this way we would manage all from same place.

packages/package-utils/package.json
rishabh3112: Interesting 😄

evilebottnawi (dismissed):
Good job!

packages/serve/package.json
anshumanv: Sure it's dev dep so doesn't matter

anshumanv (dismissed):

evilebottnawi:
watch tests are flaky need focus on fix them

snitin315:
> watch tests are flaky need focus on fixing them

@rishabh3112 has created a PR for the same #1890

evilebottnawi (approved):

rishabh3112 (approved):

PR #1895, opened 2020-10-08
chore(webpack-cli): reduce install size

snitin315:
chore

**Summary**
remove unused deps and reduce install size.

No

**Other information**
No


packages/webpack-cli/__tests__/info.test.js
```diff
@@ -1,3 +1,4 @@
+//eslint-disable-next-line node/no-extraneous-require
```
evilebottnawi, line 1: I think we need to improve config for this rule

snitin315: Hmm, we have disabled this rule at many places.

evilebottnawi: Yep, it is not good, let's improve this

snitin315: I think we can `disable` it for test files.

evilebottnawi: @snitin315 Ideally we have all tests packages in devDeps, so it should be not triggered, need investigate

snitin315: Maybe we need to use [`allowModules`](https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-extraneous-require.md#allowmodules) ->

```js
'node/no-extraneous-require': ['error', {
            'allowModules': ['execa', 'webpack', 'webpack-dev-server']
        }]
```

evilebottnawi: Why do warning this rule? We have this in our dev deps

snitin315: No idea why it's triggering 😢

evilebottnawi: Can you look at rule logic?

snitin315: This one is not specified explicitly in rules.
https://github.com/webpack/webpack-cli/blob/2c390a1fdcbb85b7596c7b621c8d263da1734690/.eslintrc.js#L11-L18

evilebottnawi: I mean inside code of rule and check why we got error here

snitin315: https://github.com/mysticatea/eslint-plugin-node/blob/4a348e7d24e0d806d7873f293c91a922d0316d14/lib/util/check-extraneous.js#L27-L34

I logged the dependencies set here and found its collecting deps defined inside `packages/webpack-cli/package.json` and not root `package.json` -

Screenshot at 2020-10-09 15-53-40

snitin315: I think we should create an issue for the same in the `eslint-plugin-node` repo.

evilebottnawi: Oh, I think we need setup `resolvePaths` (https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-extraneous-require.md, for `import` too) and setup root `node_modules` here

evilebottnawi:
```
 "settings": {
        "node": {
            "resolvePaths": [__dirname],
        }
    },
```

let try in `eslint` root config

snitin315: still the same :confused:

evilebottnawi: Can you check different values and look at source code of the rule

snitin315: In near future 👍

snitin315:
Closed due to conflicts, will open a new one.

PR #1899, opened 2020-10-08
feat: coverage

rishabh3112:
### WIP

Feature, adds coverage to CI

No

No

**Summary**
Attempt on adding coverage

No

**Other information**
Part of Roadmap


.github/workflows/nodejs.yml
```diff
@@ -118,3 +118,41 @@ jobs:
...
+  coverage:
+    name: Coverage - ${{ matrix.os }} - Node v${{ matrix.node-version }}
+
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        node-version: [12.x]
+        webpack-version: [latest]
```
evilebottnawi, line 131: We have many `if` on `webpack@4` and `webpack@5`, to cover them we need do tests for webpack@5 too

evilebottnawi: Ideally run all tests with coverage, because it allows to catch problems on widnows/macos/linux, sometimes some checks are truthy on linux, but falsy on windows

rishabh3112: Okay, I will run the coverage at all platform + webpack v5 once this simple configuration works as expected.

Also, typescript test isn't working when ran using nyc and jest, but passes on jest only. Tested that locally as well.

tsconfig.json
```diff
@@ -10,6 +10,7 @@
     "outDir": "lib",
     "moduleResolution": "node",
     "resolveJsonModule": true,
+    "sourceMap": true,
```
evilebottnawi, line 13: We need set to true in for CI and false by default, because it is increase package size

scripts/setupBuild.js
```diff
@@ -0,0 +1,10 @@
+const { writeFileSync, readFileSync } = require('fs');
+const { resolve } = require('path');
+
+const tsConfigPath = resolve(__dirname, '../tsconfig.json');
+const tsConfigRaw = readFileSync(tsConfigPath);
+const tsConfig = JSON.parse(tsConfigRaw);
+
+tsConfig.compilerOptions.sourceMap = true;
+
+writeFileSync(tsConfigPath, JSON.stringify(tsConfig, null, 2));
```
evilebottnawi, line 10: I think better create new config and use it as argument for ts compiler

rishabh3112: We have a separate tsconfig for each package 😞 , creating seperate config for each is not practical. This script runs only in ci so shouldn't be a problem.

evilebottnawi:
@rishabh3112 Can you rebase, I think  we should merge it

rishabh3112:
Rebasing.

rishabh3112:
Done, tests on windows never end. Running for 3hr+ https://github.com/webpack/webpack-cli/runs/1346351375

evilebottnawi:
@rishabh3112 let's try to disable watch test

rishabh3112:
Okay, on it.

rishabh3112:
Something weird, all test passed (including watch and serve tests) on windows but `jest` is not exiting after tests ends.
Here is the important part of log:
https://github.com/webpack/webpack-cli/runs/1346351346#step:8:409

evilebottnawi:
I think it is old bug `execa` on windows, working with new processes on windows is problems always, I think we can use `--forceExit` here

rishabh3112:
>  I think we can use --forceExit here
That did the trick 👍🏻

evilebottnawi:
Great, ping me when it will be green I will merge it and fix configuration

rishabh3112:
ping @evilebottnawi

evilebottnawi:
Thanks, I think we need look how configuration codecov for github actions, because not it is not showing in checks

evilebottnawi:
Anyway let's do it in separate PR

test/config-format/typescript/typescript.test.js
```diff
@@ -4,7 +4,7 @@ const { stat } = require('fs');
 const { resolve } = require('path');
 
 describe('webpack cli', () => {
-    it(
+    it.skip(
```
anshumanv, line 7: why skip?

rishabh3112: Not working with `nyc` 🤷🏻‍♂️ , need a fix.

rishabh3112:
> Anyway let's do it in separate PR

Did PR on here #2032

test/config-format/typescript/typescript.test.js
anshumanv: I don't understand why we need nyc when jest has it inbuilt?

/cc @evilebottnawi 

Maybe we should remove it and use `jest --coverage`

rishabh3112: https://github.com/facebook/jest/issues/3190

anshumanv: ahh spawned process mess, we need to get to this

evilebottnawi: Yep spawned process

PR #1907, opened 2020-10-10
tests: windows

evilebottnawi:

test

No need

No need

**Summary**

debug windows test problems

No

**Other information**

No


evilebottnawi:
Can't understand why windows is buggy, locally on windows all works fine, oh, we need to do release, so we return to this after stable release today

PR #1926, opened 2020-10-12
fix: support any config name

### LINKED ISSUE #1925: config file with function doesn't work if path contains `.webpack`
webpack config file where `module.exports`  is a function doesn't work depending on the file path.

If the config file path contains `.webpack`, the result of calling the config function is ignored.

```
// project/config.webpack.js -- note that filename contains ".webpack"
module.exports = (env) => ({
    entry: 'some-file.js'
    output: {
        path: './dist',
        file: 'some-bundle.js'
    }
});
```

```
$/project> webpack --config config.webpack.js
```

This produces an error saying "`./src` cannot be resolved" because the actual config is ignored, and webpack falls back to default configuration.

** Cause **

`lib/groups/ConfigGroup.js` in the `finalize` [function](https://github.com/webpack/webpack-cli/blob/5ef1e7b074390406b76cb3e25dd90f045e1bd8a2/packages/webpack-cli/lib/groups/ConfigGroup.js#L190):

```
    if (configOptions && configPath.includes('.webpack')) {
        // overrides the resolved options
    }
```

this condition is not specific enough.

evilebottnawi:

bugfix

Yes

**Summary**

fixes #1925

No

**Other information**

context should not be changed never

### DIFF
--- a/test/config-lookup/custom-name/custom-name.test.js
+++ b/test/config-lookup/custom-name/custom-name.test.js
@@ -1,0 +1,14 @@
+'use strict';
+const { existsSync } = require('fs');
+const { resolve } = require('path');
+const { run } = require('../../utils/test-utils');
+
+describe('custom config file', () => {
+    it('should work', () => {
+        const { stdout, stderr, exitCode } = run(__dirname, ['--config', resolve(__dirname, 'config.webpack.js')], false);
+        expect(stderr).toBeFalsy();
+        expect(stdout).toBeTruthy();
+        expect(exitCode).toBe(0);
+        expect(existsSync(resolve(__dirname, './binary/a.bundle.js'))).toBeTruthy();
+    });
+});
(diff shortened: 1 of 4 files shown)


test/config-lookup/custom-name/custom-name.test.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+const { existsSync } = require('fs');
+const { resolve } = require('path');
+const { run } = require('../../utils/test-utils');
+
+describe('custom config file', () => {
+    it('should work', () => {
+        const { stdout, stderr, exitCode } = run(
+            __dirname,
+            ['--config', resolve(__dirname, 'config.webpack.js'), '--output-path', './binary'],
```
snitin315, line 10: The config already have `output.path: resolve(__dirname, 'binary')` , why  passing `--output-path ./binary` again ?

evilebottnawi: Just copy/paste, we need refactor tests

snitin315:
Need to fix CI.

Issue #1932, opened 2020-10-12
Enhance env flag support


anshumanv:
**Is your feature request related to a problem? Please describe.**
`env` flag presently supports only booleans, which is not the most verbose thing to work with plus has high migration overhead as everyone is pretty much used to the old format. So we need to support at least custom string value for properties which are passed to env, something like - 

`webpack --env app="test" --env title="start"`

**Describe the solution you'd like**

We'll have to introduce separate arg parsing behaviour for env.

**Describe alternatives you've considered**

**Additional context**

LBRDan:
Any chance of viewing this issue resolved anytime soon?

I think that should be reported on your docs (as a temporary fault on this version) as it can easily drive people nuts. It is also a huge regression (not enhancement as it's marked ATM) and it is against your own docs ( https://webpack.js.org/api/cli/#environment-options )

Please consider switch this issue type both for labels and work priority, for sake of a respectful community using your solution as a multi environment standard.

evilebottnawi:
> Any chance of viewing this issue resolved anytime soon?

No, it is OSS

> Please consider switch this issue type both for labels and work priority, for sake of a respectful community using your solution as a multi environment standard.

You can help and send a PR

LBRDan:
I could for sure, but i cannot change the labels nor webpack team priority neither i have the right skills to do a PR about it (i would happy to do it if i could!)

Anyway, as you avoid to reply to my comment section about it, I will then consider my thoughts accepted because hey... that's OSS spirt too :D

evilebottnawi:
@LBRDan Because we have many issues, if you want this ASAP - send a PR

captainCoolMJ:
If you're looking for a quick snippet to get some env parsing back, you should be able to do something like:

```
env = Object.keys(env).reduce((acc, cur) => {
    const [key, value] = cur.split('=');
    acc[key] = value || env[key];
    return acc;
  }, {});
```

That should help to bring some of the key/value parsing back

```
webpack --env production --env app="test" --env title="start"
>>>
{
  production: true,
  app: 'test'
  title: 'start'
}
```

Hope that helps!

anshumanv:
Thanks for all the input everyone, https://github.com/webpack/webpack-cli/pull/1939 should fix it and should land in the next release, hopefully soon :)

shwetaarshad:
nps is executing `default` : nps webpack
nps is executing `webpack` : nps webpack.server
[...]

PR #1939, opened 2020-10-13
feat: allow passing strings in env flag

anshumanv:
feat/fix

Yes

Need to reflect in docs

**Summary**
- Allow passing strings to env flag which is passed to config
- You can now do - 

`webpack --env platform=app --env production`

which will pass `{ platform: "app", production: true }` to the config

Nope

**Other information**

Fixes https://github.com/webpack/webpack-cli/issues/1932

### DIFF
--- a/packages/webpack-cli/lib/utils/arg-parser.js
+++ b/packages/webpack-cli/lib/utils/arg-parser.js
@@ -97,8 +97,20 @@ const argParser = (options, args, argsOnly = false, name = '') => {
                 // so you can do `--env platform=staging --env production`
                 // { platform: "staging", production: true }
                 const multiArg = (value, previous = {}) => {
-                    const [key, val] = value.split('=');
-                    return { ...previous, [key]: val || true };
+                    const [allKeys, val] = value.split('=');
+                    const splitKeys = allKeys.split('.');
+                    let prevRef = previous;
+                    splitKeys.forEach((someKey, index) => {
+                        if (!prevRef[someKey]) prevRef[someKey] = {};
+                        if (index === splitKeys.length - 1) {
+                            prevRef[someKey] = val || true;
+                        } else {
+                            prevRef[someKey] = typeof prevRef[someKey] === 'string' ? {} : { ...prevRef[someKey] };
+                        }
+                        console.log({ prevRef, previous, someKey });
+                        prevRef = prevRef[someKey];
+                    });
+                    return previous;
                 };
                 parserInstance.option(flagsWithType, option.description, multiArg, option.defaultValue).action(() => {});
             } else {
--- a/packages/webpack-cli/lib/utils/arg-parser.js
+++ b/packages/webpack-cli/lib/utils/arg-parser.js
@@ -92,6 +92,15 @@ const argParser = (options, args, argsOnly = false, name = '') => {
                 // a multiple argument parsing function
                 const multiArg = (value, previous = []) => previous.concat([value]);
                 parserInstance.option(flagsWithType, option.description, multiArg, option.defaultValue).action(() => {});
+            } else if (option.multipleType) {
+                // for options which accept multiple types like env
+                // so you can do `--env platform=staging --env production`
+                // { platform: "staging", production: true }
+                const multiArg = (value, previous = {}) => {
+                    const [key, val] = value.split('=');
+                    return { ...previous, [key]: val || true };
+                };
+                parserInstance.option(flagsWithType, option.description, multiArg, option.defaultValue).action(() => {});
             } else {
                 // Prevent default behavior for standalone options
                 parserInstance.option(flagsWithType, option.description, option.defaultValue).action(() => {});
--- a/packages/webpack-cli/lib/utils/arg-parser.js
+++ b/packages/webpack-cli/lib/utils/arg-parser.js
@@ -97,17 +97,18 @@ const argParser = (options, args, argsOnly = false, name = '') => {
                 // so you can do `--env platform=staging --env production`
                 // { platform: "staging", production: true }
                 const multiArg = (value, previous = {}) => {
-                    const [allKeys, val] = value.split('=');
+                    // this ensures we're only splitting by the first `=`
+                    const [allKeys, val] = value.split(/=(.+)/, 2);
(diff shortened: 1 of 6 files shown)


test/config/type/function-with-env/function-with-env.test.js
```diff
@@ -18,6 +18,13 @@ describe('function configuration', () => {
         // Should generate the appropriate files
         expect(existsSync(resolve(__dirname, './bin/dev.js'))).toBeTruthy();
     });
+    it('Supports passing string in env', () => {
+        const { stderr, stdout } = run(__dirname, ['--env', 'environment=production', '-c', 'webpack.env.config.js']);
```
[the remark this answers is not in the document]
evilebottnawi: I think it will be easy to support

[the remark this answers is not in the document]
anshumanv: Yeah, we can support it, but are there use case for it?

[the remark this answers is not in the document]
evilebottnawi: I think it will be useful, for example, `--env app.title=title --env app.name=name`, so you can combine variables with the same meaning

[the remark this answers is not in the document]
anshumanv: Yeah makes sense, better to have it

paulmackinlay:
It would be good to update the new --env approach at https://webpack.js.org/api/cli/ and also explain how to migrate to version 5 at https://webpack.js.org/migrate/5/

anshumanv:
Yep, @paulmackinlay I have a PR already - https://github.com/webpack/webpack.js.org/pull/4051 hoping it's merged soon.

evilebottnawi (commented):
Can we add tests?

packages/webpack-cli/lib/utils/arg-parser.js
```diff
@@ -90,6 +90,27 @@ const argParser = (options, args, argsOnly = false, name = '') => {
...
+                    const splitKeys = allKeys.split('.');
+                    let prevRef = previous;
+                    splitKeys.forEach((someKey, index) => {
+                        if (!prevRef[someKey]) prevRef[someKey] = {};
+                        if (index === splitKeys.length - 1) {
+                            prevRef[someKey] = val || true;
+                        } else {
+                            prevRef[someKey] = typeof prevRef[someKey] === 'string' ? {} : { ...prevRef[someKey] };
+                        }
+                        console.log({ prevRef, previous, someKey });
```
anshumanv, line 108:
```suggestion
```

packages/webpack-cli/lib/utils/arg-parser.js
```diff
@@ -90,6 +90,26 @@ const argParser = (options, args, argsOnly = false, name = '') => {
                 // a multiple argument parsing function
                 const multiArg = (value, previous = []) => previous.concat([value]);
                 parserInstance.option(flagsWithType, option.description, multiArg, option.defaultValue).action(() => {});
+            } else if (option.multipleType) {
+                // for options which accept multiple types like env
+                // so you can do `--env platform=staging --env production`
+                // { platform: "staging", production: true }
+                const multiArg = (value, previous = {}) => {
+                    const [allKeys, val] = value.split('=');
```
marques-work, line 98: May I suggest:

```javascript
const [allKeys, val] = value.split(/=(.+)/, 2);
```

This can catch edge cases where the value contains `=` as well. For instance, the value might be a Base64 payload with padding, or any other complex value. The above snip will capture everything after the first `=` as the value.

in other words:

- `--env a=b=c` should yield `{ key: "a", value: "b=c" }`

packages/webpack-cli/lib/utils/arg-parser.js
```diff
@@ -90,6 +90,26 @@ const argParser = (options, args, argsOnly = false, name = '') => {
...
+                    const splitKeys = allKeys.split('.');
+                    let prevRef = previous;
+                    splitKeys.forEach((someKey, index) => {
+                        if (!prevRef[someKey]) prevRef[someKey] = {};
+                        if (index === splitKeys.length - 1) {
+                            prevRef[someKey] = val || true;
+                        } else {
+                            prevRef[someKey] = typeof prevRef[someKey] === 'string' ? {} : { ...prevRef[someKey] };
```
marques-work, line 106: This line seems hard to read because I don't think the nuances/side-effects of the spread operator are obvious (like what happens when `prevRef[someKey]` is neither object nor string -- it coerces to an empty object).

This might be easier to understand because it communicates the real intent to create a tree:

```javascript
// ensures that prefRef[someKey] is an object
if (null === prevRef[someKey] || 'object' !== typeof prevRef[someKey]) {
  prevRef[someKey] = {};
}
```

You might even be able to just get away with this since there are probably limits as to what `previous` can be (I'm just guessing, but don't really know):

```javascript
if ("string" === typeof prevRef[someKey]) {
  prevRef[someKey] = {};
}
```

anshumanv: Ah right makes sense, thanks!

packages/webpack-cli/lib/utils/arg-parser.js
anshumanv: Yep missed that, thanks for correcting 💯

test/config/type/function-with-env/function-with-env.test.js
```diff
@@ -18,6 +25,34 @@ describe('function configuration', () => {
...
+            '--env',
+            'environment=production',
+            '-c',
+            'webpack.env.config.js',
+        ]);
+        expect(stderr).toBeFalsy();
+        expect(stdout).toBeTruthy();
+        // Should generate the appropriate files
+        expect(existsSync(resolve(__dirname, './bin/Atsumu.js'))).toBeTruthy();
+    });
```
evilebottnawi, line 55: Can you add test:
- `--env name=foo=bar`
- `--env` (no flags)
- `--env name.` (dot at the end)
- `--env name.=test` (dot at the end)

anshumanv: Done

ntucker:
Release new version? :D

PR #1940, opened 2020-10-13
Log error stack traces to the console

thecrypticace:
Show stack traces for errors logged by webpack-cli.

No.

No.

**Summary**
An error occurring in a plugin (for example calling missing APIs or just throwing an error during `apply()`) results in a displayed error but no stack trace. This in turn makes it nearly impossible to track down the source of the problem and address it.

You may see an example of what I'm trying to address here: https://github.com/thecrypticace/webpack-cli-stack-traces

No.

**Other information**
n/a

### DIFF
--- a/packages/generators/src/addon-generator.ts
+++ b/packages/generators/src/addon-generator.ts
@@ -55,7 +55,8 @@ const addonGenerator = (
                 try {
                     mkdirp.sync(pathToProjectDir);
                 } catch (err) {
-                    logger.error('Failed to create directory', err);
+                    logger.error('Failed to create directory');
+                    logger.error(err);
                 }
                 this.destinationRoot(pathToProjectDir);
             }
--- a/packages/webpack-cli/lib/utils/logger.js
+++ b/packages/webpack-cli/lib/utils/logger.js
@@ -1,7 +1,13 @@
 const { red, cyan, yellow, green } = require('colorette');
 
 module.exports = {
-    error: (val) => console.error(`[webpack-cli] ${red(val)}`),
+    error: (val) => {
+        if (val && val.stack) {
+            val = val.stack;
+        }
+
+        console.error(`[webpack-cli] ${red(val)}`);
+    },
     warn: (val) => console.warn(`[webpack-cli] ${yellow(val)}`),
     info: (val) => console.info(`[webpack-cli] ${cyan(val)}`),
     success: (val) => console.log(`[webpack-cli] ${green(val)}`),


packages/webpack-cli/lib/utils/logger.js
```diff
@@ -1,7 +1,13 @@
 const { red, cyan, yellow, green } = require('colorette');
 
 module.exports = {
-    error: (val) => console.error(`[webpack-cli] ${red(val)}`),
+    error: (val) => {
+        if (val && val.stack) {
+            val = val.stack;
+        }
+
+        console.error(`[webpack-cli] ${red(val)}`);
```
evilebottnawi, line 9: Why? error.toString() always show the stacktrace

packages/generators/src/addon-generator.ts
```diff
@@ -55,7 +55,8 @@ const addonGenerator = (
                 try {
                     mkdirp.sync(pathToProjectDir);
                 } catch (err) {
-                    logger.error('Failed to create directory', err);
+                    logger.error('Failed to create directory');
```
evilebottnawi, line 58: logger.error(`Failed to create directory: ${err}`);

evilebottnawi (changes requested):
Need improve

packages/webpack-cli/lib/utils/logger.js
thecrypticace: In Node 14 `error.toString()` only shows the error message.

packages/generators/src/addon-generator.ts
thecrypticace: See comment above about `error.toString()`

packages/webpack-cli/lib/utils/logger.js
thecrypticace: Ditto in Node 12.x and 13.x as well.

packages/generators/src/addon-generator.ts
evilebottnawi: No:

**error.js**

```js
const error = new Error('test');

console.log(error);
```

and run `node: error.js`

```
Error: test
    at Object.<anonymous> (/home/path/to/test.js:1:77)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
    at Function.Module._load (internal/modules/cjs/loader.js:530:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:742:12)
    at startup (internal/bootstrap/node.js:283:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:743:3)

```

packages/webpack-cli/lib/utils/logger.js
evilebottnawi: https://github.com/webpack/webpack-cli/pull/1940#discussion_r504598344

packages/generators/src/addon-generator.ts
thecrypticace: This is special handling by the console functions.

The code doesn't do that because `logger.error` does this:
```
console.log(`[webpack-cli] ${error}`)
```

Which at runtime is the same as:
```js
console.log('[webpack-cli] ' + error.toString())
```

Which does **not** print the stack trace. Did you check out the test repo which showcases the behavior?

thecrypticace: A more direct test would be to log the error using the logger. You can create a `test.js` file in the root of the repo with this code:
```js
const logger = require("webpack-cli/lib/utils/logger");
logger.error(new Error("test"));
```

And see it not have the expected behavior when you run it with `node test.js`

PR #1950, opened 2020-10-15
fix: move init command to separate package

### LINKED ISSUE #1916: webpack-cli@4 installs 4x number of deps compared to v3 (including entire babel toolchain)
**Describe the bug**

Not really a bug, but an observation.
I have a typescript-based project that uses `webpack@4` and `webpack@3`. When upgrading to `webpack-cli@4` (alongside `webpack@5`), I've noticed a TON of new dependencies. More than v3. More than webpack itself.

Most of those come from cli-related packages installing `jscodeshift` to every consumer.

`webpack-cli@4` installs **668** dependencies.
In comparison, `webpack-cli@3` installs 185 dependencies, `webpack@5` itself installs 121 dependencies.

**To Reproduce**

Create a new project and install the new `webpack-cli` version.

**Expected behavior**

Be light? Avoid installing 600+ extra dependencies for those who just want to execute `webpack` via cli.
 
**Additional context**

One might consider that to be "expected behavior", but I look at it as regression.

@sokra @evilebottnawi `webpack@5` decreased the number of dependencies vastly by dropping all those `node@6` pinned versions. The cli did exactly the opposite. :(

evilebottnawi:

bugfix

No need

No

**Summary**

fixes #1916

No, because we ask to install it if you need, maybe we will refactor it in future and reduce deps more

**Other information**

No

### DIFF
--- a/packages/webpack-cli/lib/utils/__tests__/package-exists.test.js
+++ b/packages/webpack-cli/lib/utils/__tests__/package-exists.test.js
@@ -2,7 +2,7 @@ jest.setMock('../prompt-installation', {
     promptInstallation: jest.fn(),
 });
 
-const ExternalCommand = require('../../commands/resolveCommand');
+const ExternalCommand = require('../resolve-command');
 const { packageExists } = require('../package-exists');
 const { promptInstallation } = require('../prompt-installation');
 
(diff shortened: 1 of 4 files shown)


anshumanv (approved):
lgtm

billyjanitsch:
Thank you @evilebottnawi! 😍

PR #1971, opened 2020-10-19
refactor: code

evilebottnawi:

refactor

No need

No need

**Summary**

- we never throw some errors, but have code for this
- `cli-executer` is broken and doesn't work, I have kept it, we should fix it after refactor our parser
- warning about duplication variables is not working
- allow to use multiple entry before/after unknown flags
- fix exit codes

No

**Other information**

No


packages/webpack-cli/lib/bootstrap.js
```diff
@@ -39,74 +28,45 @@ const runCLI = async (cliArgs) => {
...
+            parsedArgs.unknownArgs.forEach(async (unknown) => {
+                logger.error(`Unknown argument: ${unknown}`);
+            });
+
+            process.exit(2);
+        }
+
         const parsedArgsOpts = parsedArgs.opts;
-        // Enable/Disable color on console
-        options.enabled = parsedArgsOpts.color ? true : false;
```
evilebottnawi, line 70: `color` doesn't work before, because we don't register this option + you need to use this https://github.com/jorgebucaran/colorette/blob/master/index.js#L2 for enabling/disabling colorts

anshumanv (approved):
Looks good

anshumanv:
/cc @webpack/cli-team this PR removes `cli-executor` as it gets really out of hand to use when core flags are imported since we can't search from 250+ flags and it takes infinite time to traverse through them.

@rishabh3112 do you have any ideas on how we can do better here? 

/cc @evilebottnawi

snitin315 (approved):

packages/webpack-cli/lib/bootstrap.js
```diff
@@ -39,74 +28,45 @@ const runCLI = async (cliArgs) => {
...
-            const { opts } = parseArgs(args);
-            await cli.run(opts, core);
-            return;
         }
+
+        if (parsedArgs.unknownArgs.length > 0) {
+            parsedArgs.unknownArgs.forEach(async (unknown) => {
+                logger.error(`Unknown argument: ${unknown}`);
+            });
+
```
rishabh3112, line 56: We should point to documentation for all the flags at least here.

packages/webpack-cli/lib/utils/cli-executer.js
```diff
@@ -1,64 +0,0 @@
-const { MultiSelect, Input } = require('enquirer');
-const { cyan } = require('colorette');
-const logger = require('./logger');
-const cliArgs = require('./cli-flags').core;
-
-const prompter = async () => {
-    const args = [];
-
-    const typePrompt = new MultiSelect({
```
rishabh3112, line 9: We can use autocomplete prompt here to avoid long list.

rishabh3112 (commented):
Left few suggestions

rishabh3112:
> /cc @webpack/cli-team this PR removes cli-executor as it gets really out of hand to use when core flags are imported since we can't search from 250+ flags and it takes infinite time to traverse through them.

But we should have this @evilebottnawi as it is for betterment of user experience for new users of webpack. Left few suggestions in review above.

packages/webpack-cli/lib/bootstrap.js
evilebottnawi: I think no, we have over 250 flags, we should use levenshtein here to say which option is closer

packages/webpack-cli/lib/utils/cli-executer.js
evilebottnawi: We can't because user can't use shell, not each shell support autocomplete

evilebottnawi:
@rishabh3112 This is very bad solution, with very bad DX, it is impossible to specify each flag

packages/webpack-cli/lib/bootstrap.js
jamesgeorge007: > we should use levenshtein here to say which option is closer
Much better

rishabh3112: My point we should help user to get things done, if we can't help them through command line, we should at least point to a place where they can look for how to do it.

That being said, using levenshtein is a good option as well.

packages/webpack-cli/lib/utils/cli-executer.js
rishabh3112: There is also a AutoComplete prompt in enquirer. I talking about it

rishabh3112:
> No one using this, because it is very unhelpful, invalid options should always have exit code 2

No one uses this because it is first time it is released 😃.

Yes, this is not helpful for us or experienced developers, but for new users of webpack.

That being said we need to update this to support increased amount of flags now.

evilebottnawi:
@rishabh3112 

> That being said we need to update this to support increased amount of flags now.

We can do this late, now it is unusable and contains a lot of untested and buggy code, I want to remove this

rishabh3112:
> I would even say more - we need focus on interactive because this is very helpful

I am already working on it and as you already know it is 90% complete with testing, just support for multi compilers is left.

> this feature is bad idea

I would like to think otherwise as we shouldn't ignore beginners' experience when creating a tool, but we can talk this out later. But as for now it seems you have already made up your mind, feel free to remove it 😄.

P.S.
>  nobody will write values of all flags

of course not, user select flags and write values for only for those selected flags. Have you tried using it?

evilebottnawi:
> of course not, user select flags and write values for only for those selected flags. Have you tried using it?

yes, and it is not work, cli asks all flags, instead of one, because we don't know what user means, the best solution is levenshtein here, it is help new developers and uses right exit code + allow to fix it in `package.json` instead of asking

rishabh3112:
Yes, lets go with levenshtein, I have already said in comments above that I like it and is standard among many CLI applications. 😄

evilebottnawi:
@rishabh3112 Can you open an issue about it? Thanks

rishabh3112:
We already have one  https://github.com/webpack/webpack-cli/issues/1261

PR #2006, opened 2020-11-01
fix: --config-name behaviour for fuctional configs

### LINKED ISSUE #1990: --config-name fails for a function returning an array of configurations
When `webpack.config.js` exports a function returning an array of named configurations, `webpack --config-name=NAME` can’t find them. It incorrectly complains `[webpack-cli] Configuration with name "NAME" was not found.`

```console
$ webpack --version

webpack-cli 4.1.0

webpack 5.3.0

$ cat >src.js <<EOF
console.log("Hello, world!")
EOF

$ cat >webpack.config.js <<EOF
module.exports = () => [
  { name: "a", output: { filename: "a.js" } },
  { name: "b", output: { filename: "b.js" } },
];

$ webpack 
[webpack-cli] Compilation finished
a:
  asset a.js 29 bytes [compared for emit] [minimized] (name: main)
  ./src.js 29 bytes [built] [code generated]
  a (webpack 5.3.0) compiled successfully in 147 ms

b:
  asset b.js 29 bytes [emitted] [minimized] (name: main)
  ./src.js 29 bytes [built] [code generated]
  b (webpack 5.3.0) compiled successfully in 139 ms

$ webpack --config-name=a
[webpack-cli] Configuration with name "a" was not found.
```

It works for a function returning a single configuration:

```console
$ cat >webpack.config.js <<EOF
module.exports = () => ({ name: "a", output: { filename: "a.js" } });
EOF

$ webpack --config-name=a
[webpack-cli] Compilation finished
asset a.js 29 bytes [compared for emit] [minimized] (name: main)
./src.js 29 bytes [built] [code generated]
a (webpack 5.3.0) compiled successfully in 157 ms
```

or for an array that’s not wrapped in a function:

```console
$ cat >webpack.config.js <<EOF
module.exports = [
  { name: "a", output: { filename: "a.js" } },
  { name: "b", output: { filename: "b.js" } },
];
EOF

$ webpack --config-name=a
[webpack-cli] Compilation finished
asset a.js 29 bytes [compared for emit] [minimized] (name: main)
./src.js 29 bytes [built] [code generated]
a (webpack 5.3.0) compiled successfully in 142 ms
```

snitin315:
fix

yes
No
**Summary**

Fixes #1990 

No

**Other information**


packages/webpack-cli/lib/groups/resolveConfig.js
```diff
@@ -204,7 +208,7 @@ const finalize = async (moduleObj, args) => {
         process.exit(2);
     }
 
-    newOptionsObject['options'] = isMultiCompilerMode ? configs : configs[0];
+    newOptionsObject['options'] = isMultiCompilerMode || isFunctionalConfig ? configs : configs[0];
```
evilebottnawi, line 211: What if `Function` returns the one/multiple config(s)?

snitin315: Can you give an example?

evilebottnawi:
```js
module.exports = () => {
  return { entry }
}
```

and

```js
module.exports = () => {
  return [{ entry }, { entry }]
}
```

snitin315: I will update PR with tests for both cases.

packages/webpack-cli/lib/groups/resolveConfig.js
```diff
@@ -176,8 +179,9 @@ const finalize = async (moduleObj, args) => {
 
     if (configName) {
         const foundConfigNames = [];
+        const configsToFilter = isFunctionalConfig ? configs[0] : configs;
```
evilebottnawi, line 182: Can you clarify why we need logic here? `configs` already contains configurations, maybe I am missing something?

snitin315: It's needed when config returns an array of configs.

in that case, `configs` is -
```js
[
  [
    {
      output: [Object],
      name: 'first',
      entry: './src/first.js',
      mode: 'development',
      stats: 'minimal'
    },
    {
      output: [Object],
      name: 'second',
      entry: './src/second.js',
      mode: 'production',
      stats: 'minimal'
    }
  ]
]
```

snitin315: I think we need to update this logic as this happens only when an array of configs is returned.

evilebottnawi: Weird, I think we should have `[config, config, config]` here

snitin315: hmm, I'm looking into it.

packages/webpack-cli/lib/groups/resolveConfig.js
```diff
@@ -179,7 +179,7 @@ const finalize = async (moduleObj, args) => {
 
     if (configName) {
         const foundConfigNames = [];
-        const configsToFilter = isFunctionalConfig ? configs[0] : configs;
+        const configsToFilter = Array.isArray[configs[0]] ? configs[0] : configs;
```
evilebottnawi, line 182: I still think we have problems here, maybe `Array.isArray` need here

packages/webpack-cli/lib/groups/resolveConfig.js
```diff
@@ -166,6 +168,7 @@ const finalize = async (moduleObj, args) => {
 
             // `Promise` may return `Function`
             if (typeof rawConfig === 'function') {
+                isFunctionalConfig = true;
                 // when config is a function, pass the env from args to the config function
                 rawConfig = await rawConfig(env, args);
```
snitin315, line 173: I think we need to map again if this returns an array.

evilebottnawi: Do you mean when we return multiple configurations?

snitin315: Yes because when we have `module.exports = () => [config, config]` here it becomes `rawconfig = [ config, config ]` with which the next return statement becomes `[ [ config, config ] ]`.

evilebottnawi: Let's refactor it, just put configuration in array

snitin315: Can you clarify?

evilebottnawi:
```js
const result = [];

await Promise.all(() => { /* async code here */ }).map(item => result.push(item))
```

packages/webpack-cli/lib/groups/resolveConfig.js
andersk: If nothing else, the square brackets are a clear typo: `Array.isArray[…]` → `Array.isArray(…)`.

snitin315:
Should be good now.

rishabh3112 (dismissed):

packages/webpack-cli/lib/groups/resolveConfig.js
```diff
@@ -172,7 +174,15 @@ const finalize = async (moduleObj, args) => {
...
         }),
-    );
+    ).then((allConfigs) => {
+        allConfigs.map((singleConfig) => {
+            if (Array.isArray(singleConfig)) {
+                singleConfig.map((conf) => configs.push(conf));
+            } else {
+                configs.push(singleConfig);
+            }
+        });
```
andersk, line 184: Nit: `map` shouldn’t be used for side effects; there’s `forEach` for that. But more simply,

```suggestion
        configs.push(...allConfigs.flat());
```

snitin315: Thanks.

packages/webpack-cli/lib/groups/resolveConfig.js
```diff
@@ -172,7 +174,15 @@ const finalize = async (moduleObj, args) => {
 
             return rawConfig;
         }),
-    );
+    ).then((allConfigs) => {
```
evilebottnawi, line 177: Avoid using `then`, we can use `async`/`await` safely, I think we need eslint rule for this

evilebottnawi (commented):
Anyway good job, thanks!

evilebottnawi (commented):
Let's fix tests and we can merge

snitin315:
Green CI 😄

rishabh3112 (approved):

evilebottnawi (approved):
/cc @webpack/cli-team

PR #2013, opened 2020-11-01
fix: assign cache value for default configs

anshumanv:

fix

WIP

NA

**Summary**
We assign config path in cache properties passed to the compiler, it didn't take into account the default configs which are resolved so this should fix it.

Nope

**Other information**
Tests WIP

### DIFF
--- a/test/core-flags/cache-flags.test.js
+++ b/test/core-flags/cache-flags.test.js
@@ -136,6 +136,17 @@ describe('cache related flags from core', () => {
         expect(exitCode).toEqual(0);
     });
 
+    it('should assign cache build dependencies with default config', () => {
+        // TODO: Fix on windows
+        if (isWindows) return;
+        const { stderr, stdout, exitCode } = run(__dirname, ['--cache-type', 'filesystem']);
+        expect(stderr).toBeFalsy();
+        expect(stdout).toContain('buildDependencies');
+        expect(stdout).toContain(`'${path.join(__dirname, './webpack.config.js')}'`);
+        expect(stdout).toContain("type: 'filesystem'");
+        expect(exitCode).toEqual(0);
+    });
+
     it('should assign cache build dependencies with merged configs', () => {
         // TODO: Fix on windows
         if (isWindows) return;
--- a/packages/webpack-cli/lib/groups/resolveConfig.js
+++ b/packages/webpack-cli/lib/groups/resolveConfig.js
@@ -140,7 +140,7 @@ const resolveConfigFiles = async (args) => {
 
 // Given config data, determines the type of config and
 // returns final config
-const finalize = async (moduleObj, args, isDefault = false) => {
+const finalize = async (moduleObj, args, isDefaultConfig = false) => {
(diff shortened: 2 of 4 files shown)


packages/webpack-cli/lib/groups/ConfigGroup.js
```diff
@@ -126,21 +126,21 @@ const resolveConfigFiles = async (args) => {
...
+        opts = await finalize(foundConfig, args, true);
 
         return;
     }
 };
 
 // Given config data, determines the type of config and
 // returns final config
-const finalize = async (moduleObj, args) => {
+const finalize = async (moduleObj, args, isDefault = false) => {
```
snitin315, line 143:
```suggestion
const finalize = async (moduleObj, args, isDefaultConfig = false) => {
```

anshumanv: after ci

anshumanv:
@evilebottnawi random failures 😞

evilebottnawi:
@anshumanv Let's do rebase

evilebottnawi:
I think we have a bug in implementation, because:
```
  buildDependencies: {
        defaultWebpack: [
          '/home/runner/work/webpack-cli/webpack-cli/node_modules/webpack/lib/'
        ]
      },
```

test/core-flags/cache-flags.test.js
```diff
@@ -136,6 +136,17 @@ describe('cache related flags from core', () => {
         expect(exitCode).toEqual(0);
     });
 
+    it('should assign cache build dependencies with default config', () => {
+        // TODO: Fix on windows
+        if (isWindows) return;
```
evilebottnawi, line 141: Avoid this here

anshumanv: already skipped for windows, some problems in all tests 😞

anshumanv:
> I think we have a bug in implementation, because:

This doesn't happen in all tests, which is very strange 😞

evilebottnawi:
Can you try it locally?

anshumanv:
passing locally 😞

evilebottnawi:
Can you try to use `node_modules/bin/jest test/path/to/name.test.js` for this test (change it in github actions file), so we will run the only related tests

test/core-flags/cache-flags.test.js
evilebottnawi: Yep, let's focus on it late

packages/webpack-cli/lib/groups/ConfigGroup.js
evilebottnawi: @anshumanv Can you fix it, I want to merge it and do release

anshumanv: fixed

evilebottnawi (approved):
Good, wait CI green, they are very slow :disappointed:

evilebottnawi:
macos is broken today, but should be fine, because linux is green, let's merge and do release

PR #2022, opened 2020-11-02
fix: resolve dev server hot options correctly

### LINKED ISSUE #2004: [serve] --hot-only is broken
**Describe the bug**

`--hot-only` is broken.

Screenshot at 2020-10-31 16-39-01

**To Reproduce**

`webpack serve --hot-only`

**Expected behavior**

Should work properly.

**Screenshots**
pasted above

**Additional context**
No

anshumanv:
fix

Yes

Need update 

**Summary**
- Fix how `hot` flag is resolved in serve package

No

**Other information**
Fixes #2004

### DIFF
--- a/packages/serve/src/createConfig.ts
+++ b/packages/serve/src/createConfig.ts
@@ -10,10 +10,15 @@ import { devServerOptionsType } from './types';
  */
 export default function createConfig(args): devServerOptionsType {
     const options = { ...args };
-    // eslint-disable-next-line node/no-extraneous-require
-    const version = require('webpack-dev-server/package.json').version;
+    let isDevServer4 = false;
+    try {
+        // eslint-disable-next-line node/no-extraneous-require
+        const version = require('webpack-dev-server/package.json').version;
 
-    const isDevServer4 = version.startsWith('4');
+        isDevServer4 = version.startsWith('4');
+    } catch (err) {
+        throw new Error(`You need to install 'webpack-dev-server' for running 'webpack serve'.\n${err}`);
+    }
 
     if (options.clientLogging) {
         options.client = {
--- a/packages/serve/src/createConfig.ts
+++ b/packages/serve/src/createConfig.ts
@@ -1,4 +1,8 @@
 import { devServerOptionsType } from './types';
+// eslint-disable-next-line node/no-extraneous-import
+import { version } from 'webpack-dev-server/package.json';
+
+const isDevServer4 = version.startsWith('4');
 
 /**
  *
@@ -18,10 +22,18 @@ export default function createConfig(args): devServerOptionsType {
         // clientLogging is not a valid devServer option
         delete options.clientLogging;
     }
-
-    // only apply hotOnly when both are supplied
-    if (options.hot && options.hotOnly) {
-        delete options.hot;
+    if (isDevServer4) {
+        if (options.hotOnly) {
+            options.hot = 'only';
+            // hotOnly is not a valid devServer option
+            delete options.hotOnly;
+        }
+    } else {
+        // only apply hotOnly when both are supplied
+        if (options.hot && options.hotOnly) {
+            delete options.hot;
+        }
     }
+
     return options;
 }
--- a/packages/serve/src/createConfig.ts
+++ b/packages/serve/src/createConfig.ts
@@ -22,17 +22,10 @@ export default function createConfig(args): devServerOptionsType {
         // clientLogging is not a valid devServer option
         delete options.clientLogging;
     }
-    if (isDevServer4) {
-        if (options.hotOnly) {
-            options.hot = 'only';
-            // hotOnly is not a valid devServer option
-            delete options.hotOnly;
-        }
-    } else {
-        // only apply hotOnly when both are supplied
-        if (options.hot && options.hotOnly) {
-            delete options.hot;
-        }
+    if (isDevServer4 && options.hotOnly) {
+        options.hot = 'only';
+        // hotOnly is not a valid devServer option
+        delete options.hotOnly;
     }
 
     return options;
--- a/packages/serve/src/createConfig.ts
+++ b/packages/serve/src/createConfig.ts
@@ -1,4 +1,8 @@
+import { utils } from 'webpack-cli';
+
 import { devServerOptionsType } from './types';
+
+const { logger } = utils;
 
 /**
  *
@@ -16,7 +20,8 @@ export default function createConfig(args): devServerOptionsType {
         // eslint-disable-next-line node/no-extraneous-require
         devServerVersion = require('webpack-dev-server/package.json').version;
     } catch (err) {
-        throw new Error(`You need to install 'webpack-dev-server' for running 'webpack serve'.\n${err}`);
+        logger.error(`You need to install 'webpack-dev-server' for running 'webpack serve'.\n${err}`);
+        process.exit(2);
     }
     isDevServer4 = devServerVersion.startsWith('4');
 
--- a/packages/serve/src/createConfig.ts
+++ b/packages/serve/src/createConfig.ts
@@ -10,15 +10,15 @@ import { devServerOptionsType } from './types';
  */
 export default function createConfig(args): devServerOptionsType {
     const options = { ...args };
-    let isDevServer4 = false;
+    let isDevServer4 = false,
+        devServerVersion;
     try {
         // eslint-disable-next-line node/no-extraneous-require
-        const version = require('webpack-dev-server/package.json').version;
-
-        isDevServer4 = version.startsWith('4');
+        devServerVersion = require('webpack-dev-server/package.json').version;
     } catch (err) {
         throw new Error(`You need to install 'webpack-dev-server' for running 'webpack serve'.\n${err}`);
     }
+    isDevServer4 = devServerVersion.startsWith('4');
 
     if (options.clientLogging) {
         options.client = {
--- a/packages/serve/src/createConfig.ts
+++ b/packages/serve/src/createConfig.ts
@@ -19,11 +19,9 @@ export default function createConfig(args): devServerOptionsType {
         delete options.clientLogging;
     }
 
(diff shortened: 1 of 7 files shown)


packages/serve/src/createConfig.ts
```diff
@@ -1,4 +1,8 @@
 import { devServerOptionsType } from './types';
+// eslint-disable-next-line node/no-extraneous-import
+import { version } from 'webpack-dev-server/package.json';
```
evilebottnawi, line 3: We should use `packageExist` or use `try`/`catch`

anshumanv: yep on it

anshumanv: handled at the entry of serve package to avoid redundant checks

packages/serve/src/createConfig.ts
```diff
@@ -18,11 +22,17 @@ export default function createConfig(args): devServerOptionsType {
...
-        delete options.hotOnly;
+    if (isDevServer4) {
+        if (options.hotOnly) {
+            options.hot = 'only';
+            // hotOnly is not a valid devServer option
+            delete options.hotOnly;
+        }
+    } else {
+        // only apply hotOnly when both are supplied
+        if (options.hot && options.hotOnly) {
```
evilebottnawi, line 33: It will break `hot` ability, why we need delete it?

anshumanv: should we allow setting both hot and hotOnly?

anshumanv: won't that apply the plugin twice?

packages/serve/src/createConfig.ts
```diff
@@ -18,11 +22,17 @@ export default function createConfig(args): devServerOptionsType {
...
-        // hotOnly is not a valid devServer option
-        delete options.hotOnly;
+    if (isDevServer4) {
+        if (options.hotOnly) {
+            options.hot = 'only';
+            // hotOnly is not a valid devServer option
+            delete options.hotOnly;
+        }
+    } else {
+        // only apply hotOnly when both are supplied
```
anshumanv, line 32:
```suggestion
        // when both hot flags are supplied, only hotOnly is applied
```

packages/serve/src/createConfig.ts
evilebottnawi: hm, I think you are right here, we should not apply plugin twice, but we have logic on webpack-dev-server side (if you have plugin we don't add it again), removing hot is not safe, because we adding client code for HMR when hot is true

packages/serve/src/createConfig.ts
snitin315: `prefer --hotOnly over --hot`

packages/serve/src/createConfig.ts
anshumanv: ah okay then we can remove this

packages/serve/src/createConfig.ts
anshumanv: server seems to have logic for this so this is not needed anymore

packages/serve/src/createConfig.ts
```diff
@@ -18,8 +22,7 @@ export default function createConfig(args): devServerOptionsType {
         // clientLogging is not a valid devServer option
         delete options.clientLogging;
     }
-
-    if (options.hotOnly) {
+    if (isDevServer4 && options.hotOnly) {
```
evilebottnawi, line 25: I think we don't have `options.hotOnly` on CLI only `hot` and `hot = 'only'`

anshumanv: we have it on v3

evilebottnawi: But options from webpack-cli

anshumanv: nope options here is the dev server flags

anshumanv: printed and checked 😄

packages/serve/src/createConfig.ts
evilebottnawi, line 3: Still can be error, I don't think we should use `import`/`require` on the top of files for peer dependencies, I think we don't have a problem here, because we check it before, but better do not do it, because we can refactoring something and break code

packages/serve/src/createConfig.ts
evilebottnawi: thanks

packages/serve/src/createConfig.ts
anshumanv: agree, changing 👍

packages/serve/src/createConfig.ts
```diff
@@ -14,6 +10,10 @@ const isDevServer4 = version.startsWith('4');
  */
 export default function createConfig(args): devServerOptionsType {
     const options = { ...args };
+    // eslint-disable-next-line node/no-extraneous-require
+    const version = require('webpack-dev-server/package.json').version;
```
evilebottnawi, line 14: And let's use `try`/`catch`, because any `require` can throw and error

packages/serve/src/createConfig.ts
```diff
@@ -10,6 +10,15 @@ import { devServerOptionsType } from './types';
  */
 export default function createConfig(args): devServerOptionsType {
     const options = { ...args };
+    let isDevServer4 = false;
+    try {
+        // eslint-disable-next-line node/no-extraneous-require
+        const version = require('webpack-dev-server/package.json').version;
+
+        isDevServer4 = version.startsWith('4');
```
evilebottnawi, line 18: Let's move out it from `try`/`catch`, because it is unnecessary have here

anshumanv: done

evilebottnawi (commented):
Can we add simple test for `webpack --hot` and `webpack-hot=only` to avoid problems in future

packages/serve/src/createConfig.ts
```diff
@@ -10,6 +10,15 @@ import { devServerOptionsType } from './types';
  */
 export default function createConfig(args): devServerOptionsType {
     const options = { ...args };
+    let isDevServer4 = false,
```
rishabh3112, line 13: Let's keep `webpack-dev-server` check in a single place.
It is in `index.ts` as well as this file.

rishabh3112 (changes requested):
Left a suggestion. Rest looks good to me 😃

packages/serve/src/createConfig.ts
anshumanv: @evilebottnawi suggested to keep it here too

rishabh3112: When availability of `webpack-dev-server` is already tested in `index.ts` which is entry file to `serve` package, why do we need to test it again?

anshumanv: https://github.com/webpack/webpack-cli/pull/2022#discussion_r516001135

anshumanv:
> Can we add simple test for webpack serve --hot and webpack serve --hot=only to avoid problems in future


We already have tests for those flags, also updated the tests for createConfig which was causing this problem

packages/serve/src/createConfig.ts
rishabh3112: Okay, lets keep it then.

packages/serve/src/parseArgs.ts
```diff
@@ -23,13 +23,8 @@ type ArgsType = {
...
 export default function parseArgs(cli: WebpackCLIType, args: string[]): ArgsType {
-    let devServerFlags: object[];
-    try {
-        // eslint-disable-next-line node/no-extraneous-require
-        devServerFlags = require('webpack-dev-server/bin/cli-flags').devServer;
-    } catch (err) {
-        throw new Error(`You need to install 'webpack-dev-server' for running 'webpack serve'.\n${err}`);
-    }
+    // eslint-disable-next-line node/no-extraneous-require
+    const devServerFlags = require('webpack-dev-server/bin/cli-flags').devServer;
```
rishabh3112, line 27: Same logic goes here too, error can occur here as well if dev server changes files.

evilebottnawi: Yes, you are right

anshumanv: done

snitin315 (changes requested):
Need a rebase

anshumanv:
rebase done

evilebottnawi (approved):

packages/serve/src/index.ts
```diff
@@ -10,6 +10,12 @@ import parseArgs from './parseArgs';
  * @returns {Function} invokes the devServer API
  */
 export default function serve(...args: string[]): void {
+    try {
+        // eslint-disable-next-line node/no-extraneous-require
+        require('webpack-dev-server');
+    } catch (err) {
+        throw new Error(`You need to install 'webpack-dev-server' for running 'webpack serve'.\n${err}`);
+    }
```
snitin315, line 18: exitCode ?

evilebottnawi: /cc @anshumanv

anshumanv: exit code is missing in all tests, we'll have to change utils to export it, future PR

evilebottnawi: I think we should use `process.exit(2)` here to avoid problems

anshumanv: throwing err will exit the process, doing process.exit() after throw would be unreachable code

evilebottnawi: So let's use `logger.error(); process.exit(2)`, because throw will be use `1` exit code, but `1` exit code only for webpack bundle problems

evilebottnawi (approved):
/cc @webpack/cli-team 

Will be great to add more tests, but let's focus on coverage in other PRs

snitin315 (approved):

rishabh3112 (approved):

PR #2027, opened 2020-11-03
feat: add WEBPACK_SERVE environment variable

### LINKED ISSUE #1973: The WEBPACK_DEV_SERVER environment variable is not set when running "webpack serve"
**Describe the bug**

Running webpack via `webpack-dev-server` used to expose an environment variable that allowed you to determine if the application is being served by the dev server. This was quite useful on a React app since you could use this variable to determine when to hydrate the first rendering (SSR) or when to render the full application on first render.

Unfortunately this behaviour is no longer present when running `webpack serve`. I could replicate this behavior by creating my own env variables for this but it would be a lot more ideal if webpack would do this out of the box. If that's not desirable then I would say that this should at least be included in the [migration guide for v5](https://webpack.js.org/migrate/5/#getoptions-method-for-loaders). Mainly because I didn't see a way around using `webpack serve` due to this issue: https://github.com/webpack/webpack-dev-server/issues/2029#issuecomment-707034614.

When running `webpack-dev-server` the value of `process.env.WEBPACK_DEV_SERVER` is set to `true`.

**To Reproduce**

1. Place `console.log(process.env.WEBPACK_DEV_SERVER);` somewhere in your `webpack.config.js`.
2. Run webpack with `webpack serve`
3. Look at the printed value in the console.

An example can be found here: https://github.com/bensampaio/dev-server-issue.

**Expected behavior**

I expected the `console.log` to print `true`.

System:
* OS: macOS 10.15.7
*    CPU: (8) x64 Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz
*    Memory: 192.31 MB / 16.00 GB

Binaries:
*    Node: 12.18.3 - /usr/local/bin/node
*    Yarn: 1.22.5 - /usr/local/bin/yarn
*    npm: 6.14.6 - /usr/local/bin/npm

Browsers:
*    Chrome: 86.0.4240.80
*    Chrome Canary: 88.0.4298.0
*    Edge: 86.0.622.48
*    Firefox: 80.0.1
*    Firefox Developer Edition: 46.0a2
*    Safari: 14.0

**Additional context**

This issue was originally reported in: https://github.com/webpack/webpack-dev-server/issues/2781.

rishabh3112:
Add WEBPACK_SERVE variable in environment and in main process if `webpack serve` is used.

Yes
No
**Summary**
Fixes #1973 

Yes,
1. Exposing `WEBPACK_SERVE` variable which was earlier `WEBPACK_DEV_SERVER` when dev server was not integrated into CLI.
2. New variable `WEBPACK_SERVE` is now accessable only through `env` argument supplied to webpack configuration of type `function`.

**Other information**
/cc @evilebottnawi

### DIFF
--- a/packages/serve/src/index.ts
+++ b/packages/serve/src/index.ts
@@ -12,6 +12,10 @@ import parseArgs from './parseArgs';
 export default function serve(...args: string[]): void {
     const cli = new WebpackCLI();
 
+    // add WEBPACK_SERVE to main process and compilation environment
+    process.env['WEBPACK_SERVE'] = JSON.stringify(true);
+    args = [...args, '--env', 'process.env.WEBPACK_SERVE=true'];
+
     const { webpackArgs, devServerArgs } = parseArgs(cli, args);
 
     cli.getCompiler(webpackArgs).then((compiler): void => {
--- a/packages/serve/src/index.ts
+++ b/packages/serve/src/index.ts
@@ -12,9 +12,8 @@ import parseArgs from './parseArgs';
 export default function serve(...args: string[]): void {
     const cli = new WebpackCLI();
 
-    // add WEBPACK_SERVE to main process and compilation environment
-    process.env['WEBPACK_SERVE'] = JSON.stringify(true);
-    args = [...args, '--env', 'process.env.WEBPACK_SERVE=true'];
+    // add WEBPACK_SERVE to compilation environment
+    args = [...args, '--env', 'WEBPACK_SERVE=true'];
 
     const { webpackArgs, devServerArgs } = parseArgs(cli, args);
 
--- a/packages/serve/src/index.ts
+++ b/packages/serve/src/index.ts
@@ -12,9 +12,6 @@ import parseArgs from './parseArgs';
 export default function serve(...args: string[]): void {
     const cli = new WebpackCLI();
 
-    // add WEBPACK_SERVE to compilation environment
-    args = [...args, '--env', 'WEBPACK_SERVE=true'];
-
     const { webpackArgs, devServerArgs } = parseArgs(cli, args);
 
     cli.getCompiler(webpackArgs).then((compiler): void => {
--- a/packages/serve/src/parseArgs.ts
+++ b/packages/serve/src/parseArgs.ts
@@ -38,6 +38,13 @@ export default function parseArgs(cli: WebpackCLIType, args: string[]): ArgsType
     const parsedWebpackArgs = cli.argParser(core, parsedDevServerArgs.unknownArgs, true, process.title);
     const webpackArgs = parsedWebpackArgs.opts;
 
+    // Add WEBPACK_SERVE environment variable
+    if (webpackArgs.env) {
+        webpackArgs.env.WEBPACK_SERVE = true;
+    } else {
+        webpackArgs.env = { WEBPACK_SERVE: true };
+    }
+
     // pass along the 'hot' argument to the dev server if it exists
     if (webpackArgs && webpackArgs.hot !== undefined) {
         devServerArgs['hot'] = webpackArgs.hot;
--- a/test/serve/serve-variable/webpack.config.js
+++ b/test/serve/serve-variable/webpack.config.js
@@ -1,17 +1,22 @@
 class CustomTestPlugin {
+    constructor(isInEnvironment) {
+        this.isInEnvironment = isInEnvironment;
+    }
     apply(compiler) {
-        compiler.hooks.beforeCompile.tap('testPlugin', () => {
-            if (process.env.WEPACK_SERVE) {
-                console.log('Variable Defined!');
+        compiler.hooks.done.tap('testPlugin', () => {
+            if (process.env.WEBPACK_SERVE && this.isInEnvironment) {
+                console.log('PASS');
             } else {
-                console.log('Not Defined');
+                console.log('FAIL');
             }
         });
     }
 }
 
-module.exports = {
-    mode: 'development',
-    devtool: false,
-    plugins: [new CustomTestPlugin()],
+module.exports = (env) => {
+    return {
+        mode: 'development',
+        devtool: false,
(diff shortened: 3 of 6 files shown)


packages/serve/src/index.ts
```diff
@@ -12,6 +12,10 @@ import parseArgs from './parseArgs';
 export default function serve(...args: string[]): void {
     const cli = new WebpackCLI();
 
+    // add WEBPACK_SERVE to main process and compilation environment
+    process.env['WEBPACK_SERVE'] = JSON.stringify(true);
+    args = [...args, '--env', 'process.env.WEBPACK_SERVE=true'];
```
evilebottnawi, line 17: I think it is unsafe, why do you use `process.env['WEBPACK_SERVE'] = true` in code?

rishabh3112: In the reproducible repo of the issue, it was expected to access this variable through `process.env` directly in webpack config. W/O recieving it from `env` argument. To support this we would need to update it directly in process.

rishabh3112: Should I enforce use of `env` argument in webpack configs?

evilebottnawi: hm, I think we need small discussion here, in webpack-dev-server we setup process.env['WEBPACK_SERVER'], but I found it some weird, so we need to decide how we will do it better:
- use old approach - just set `process.env['WEBPACK_SERVER']`
- passing `serve` in `env` flags by default

evilebottnawi: > Should I enforce use of env argument in webpack configs?

Can you clarify?

rishabh3112: Both have their pros and cons.
1. Old approach: The variable is accessible by webpack config globally. But in this scenario, it would not be available outside webpack config. As it is old behaviour, people would be easily able to migrate for this breaking change.

2. New approach: The variable is not accessible by webpack config globally, use of `env` argument is required by webpack configuration. Plus point is, we would be able to access the variable outside of configuration as well. Con would be that migration of configuration would now be difficult as compared to previous approach.

P.S.
I have done both in this PR for now.

rishabh3112: > > Should I enforce use of env argument in webpack configs?
> 
> Can you clarify?

Yes, I mean access variable like this in webpack configuration.
```
module.exports = (env) => {
  env.WEBPACK_SERVE
  return {...};
}
```

evilebottnawi: Honestly I found `env` is better solution

rishabh3112: Okay, will enforce the `env` argument then 😃.

rishabh3112: Done.

packages/serve/src/index.ts
```diff
@@ -12,6 +12,9 @@ import parseArgs from './parseArgs';
 export default function serve(...args: string[]): void {
     const cli = new WebpackCLI();
 
+    // add WEBPACK_SERVE to compilation environment
+    args = [...args, '--env', 'WEBPACK_SERVE=true'];
```
evilebottnawi, line 16: Maybe we can avoid it? It can be unsafe...

rishabh3112: Updated, see if fine now.

P.S.
>  It can be unsafe

Can you explain this a bit or point me to a place where can I learn more about this?

packages/serve/src/parseArgs.ts
```diff
@@ -38,6 +38,13 @@ export default function parseArgs(cli: WebpackCLIType, args: string[]): ArgsType
     const parsedWebpackArgs = cli.argParser(core, parsedDevServerArgs.unknownArgs, true, process.title);
     const webpackArgs = parsedWebpackArgs.opts;
 
+    // Add WEBPACK_SERVE environment variable
+    if (webpackArgs.env) {
+        webpackArgs.env.WEBPACK_SERVE = true;
```
evilebottnawi, line 43: We should check what it is `serve` command

rishabh3112: I didn't get what you said. Can you explain a bit more?

evilebottnawi: We should add `WEBPACK_SERVE` to env only when we run `webpack serve` command

rishabh3112: `serve` package would only be accessed by users when we run `serve` command. As this code is in serve package, I don't think we need to check again. Or I am missing something ?

anshumanv: nope, this will only run during serve command invocation, we're good here

evilebottnawi: oh, yes, my mistake

evilebottnawi (approved):
/cc @webpack/cli-team

packages/serve/src/index.ts
anshumanv: @evilebottnawi can you give example?

evilebottnawi: we remove this code :smile:

test/serve/serve-variable/webpack.config.js
```diff
@@ -0,0 +1,24 @@
+const isInProcess = process.env.WEBPACK_SERVE;
```
anshumanv, line 1: We don't need to assign the property on the process object?

evilebottnawi: Yes

anshumanv: We should document then that previously the var was on process object now it lives inside the env supplied to config

rishabh3112: Yes, I already have mentioned in PR that this is a breaking change. Will add the bit about enforcement of use of `env` argument as well.

evilebottnawi: Yes, let's update it after merge, I will open an issue

snitin315 (approved):

anshumanv (approved):

PR #2029, opened 2020-11-03
tests: clear old output files before each test

snitin315:
tests improvements

yes
no
**Summary**

clear old output files before each test

No

**Other information**
No

### DIFF
--- a/setupTest.js
+++ b/setupTest.js
@@ -1,1 +1,19 @@
-jest.setTimeout(240000);
+/* eslint-disable */
+const del = require('del');
+const { posix } = require('path');
+
+jest.setTimeout(300000);
+
+jasmine.getEnv().addReporter({
+    specStarted: (result) => (jasmine.currentTest = result),
+    specDone: (result) => (jasmine.currentTest = result),
+});
+
+beforeEach(async () => {
+    await del(
+        [posix.join(jasmine.testPath, `../bin/*`), posix.join(jasmine.testPath, `../binary/*`), posix.join(jasmine.testPath, `../dist/*`)],
+        {
+            force: true,
+        },
+    );
+});
--- a/package.json
+++ b/package.json
@@ -71,6 +71,7 @@
     "commitlint-config-cz": "^0.13.2",
     "concat-stream": "^2.0.0",
     "cz-customizable": "^6.3.0",
+    "del": "^6.0.0",
     "del-cli": "^3.0.1",
     "eslint": "^7.12.1",
     "eslint-config-prettier": "^6.15.0",
--- a/test/entry/defaults-index/entry-multi-args.test.js
+++ b/test/entry/defaults-index/entry-multi-args.test.js
@@ -2,7 +2,6 @@
 
 const { stat } = require('fs');
 const { resolve } = require('path');
-
 const { run } = require('../../utils/test-utils');
 
 describe('single entry flag index present', () => {
--- a/test/mode/mode-with-config/mode-with-config.test.js
+++ b/test/mode/mode-with-config/mode-with-config.test.js
@@ -44,9 +44,8 @@ describe('mode flags with config', () => {
         expect(stdout).toBeTruthy();
 
         // Should generate the appropriate files
-        stat(resolve(__dirname, './bin/main.js.OTHER.LICENSE.txt'), (err, stats) => {
-            expect(err).toBe(null);
-            expect(stats.isFile()).toBe(true);
+        stat(resolve(__dirname, './bin/main.js.OTHER.LICENSE.txt'), (err) => {
+            expect(err).toBeTruthy();
         });
 
         stat(resolve(__dirname, './bin/main.js'), (err, stats) => {
@@ -75,9 +74,8 @@ describe('mode flags with config', () => {
         expect(stdout).toBeTruthy();
 
         // Should generate the appropriate files
-        stat(resolve(__dirname, './bin/main.js.OTHER.LICENSE.txt'), (err, stats) => {
-            expect(err).toBe(null);
-            expect(stats.isFile()).toBe(true);
+        stat(resolve(__dirname, './bin/main.js.OTHER.LICENSE.txt'), (err) => {
+            expect(err).toBeTruthy();
         });
 
         stat(resolve(__dirname, './bin/main.js'), (err, stats) => {
--- a/test/mode/mode-with-config/webpack.config.js
+++ b/test/mode/mode-with-config/webpack.config.js
@@ -6,7 +6,7 @@ module.exports = () => {
     const config = {
         entry: './index.js',
         output: {
-            path: path.join(dirname, 'dist'),
+            path: path.join(dirname, 'bin'),
             filename: '[name].js',
         },
         optimization: {
--- a/yarn.lock
+++ b/yarn.lock
@@ -4482,6 +4482,20 @@ del@^5.1.0:
     rimraf "^3.0.0"
     slash "^3.0.0"
 
+del@^6.0.0:
+  version "6.0.0"
+  resolved "https://registry.yarnpkg.com/del/-/del-6.0.0.tgz#0b40d0332cea743f1614f818be4feb717714c952"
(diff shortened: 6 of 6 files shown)


anshumanv (commented):
they are already removed in global teardown, use case for this?

snitin315:
> they are already removed in global teardown, use case for this?

global teardown removes them after completing all tests. 

Sometimes we check the existence of files in consecutive `it` blocks. So we need to clear them in `beforeEach`.

anshumanv:
> Sometimes we check the existence of files in consecutive it blocks. So we need to clear them in beforeEach.

Won't the new test override them?

test/entry/flag-entry/entry-with-flag.test.js
```diff
@@ -39,8 +49,8 @@ describe('entry flag', () => {
         });
     });
 
-    it('should resolve the path to src/a.js as ./src/a.js', (done) => {
-        const { stderr, stdout } = run(__dirname, ['--entry', 'src/a.js']);
+    it('should resolve the path to /src/a.js as ./src/a.js', (done) => {
+        const { stderr, stdout } = run(__dirname, ['--entry', '/src/a.js']);
```
snitin315, line 53: webpack 5 doesn't resolve `src/a.js` as `./src/a.js` but this test was passing earlier because of the old output file created above.

but it resolves `/src/a.js` as `./src/ajs`.

test/entry/multiple-entries/multi-entries.test.js
```diff
@@ -24,7 +31,7 @@ describe(' multiple entries', () => {
     });
 
     it('should allow multiple entry flags', (done) => {
-        const { stderr, stdout } = run(__dirname, ['--entry', 'src/a.js', '--entry', 'src/b.js']);
+        const { stderr, stdout } = run(__dirname, ['--entry', '/src/a.js', '--entry', '/src/b.js']);
```
snitin315, line 34: same here

test/entry/flag-entry/entry-with-flag.test.js
anshumanv: ah I see, seems valid then, let's do it at top level then? Keep a list of test dirs where the files are to be removed and run it for that list

anshumanv: for each test

snitin315: Can you explain more?

anshumanv: can we use global beforeEach for certain tests

snitin315: I am not sure if that is possible. Also, all beforeEach statements aren't exactly the same.

test/no-hot/no-hot.test.js
```diff
@@ -1,10 +1,17 @@
...
+const { resolve, join } = require('path');
+const rimraf = require('rimraf');
 const { yellow } = require('colorette');
 
 describe('no-hot flag', () => {
+    beforeEach((done) => {
+        rimraf(join(__dirname, './bin/*'), () => {
+            done();
+        });
+    });
```
evilebottnawi, line 13: Yep, we need move it for global options and avoid using in each test, maybe https://jestjs.io/docs/en/configuration#globalteardown-string?

snitin315: `globalTearDown` runs after each `test suite` we need to run it before/after each `test case`.

snitin315: We already clean after each `test suite`. 

https://github.com/webpack/webpack-cli/blob/0038ff282931748903fe746d1ce79b5dbf333a80/jest.config.js#L19

evilebottnawi:
@webpack/cli-team What do you think?

test/no-hot/no-hot.test.js
anshumanv: cab you try with `global.beforeEach`?

snitin315: Can you point me to the corresponding docs of `global.beforeEach`?

anshumanv: Just try defining before each in https://github.com/webpack/webpack-cli/blob/master/setupTest.js

snitin315: Screenshot at 2020-11-04 18-20-24

anshumanv: try `globals.beforeEach`?

anshumanv: or maybe just `beforeEach`

anshumanv: https://github.com/facebook/jest/blob/master/docs/Configuration.md#injectglobals-boolean

snitin315: Working with global 😄 .

```
const { beforeEach } = require('@jest/globals')
const rimraf = require('rimraf');
const { join } = require('path');
jest.setTimeout(240000);

beforeEach((done) => {
    rimraf(join(__dirname, './test/mode/mode-with-config/bin/*'), () => {
        done();
    });
});
```

anshumanv: nice nice, let's maintain a list of dirs which are to be deleted before every test

snitin315:
fixing CI.

setupTest.js
```diff
@@ -1 +1,23 @@
...
+            done();
+        });
+        await rimraf(join(__dirname, `./test/${directory}/binary/*`), () => {
+            done();
+        });
+        await rimraf(join(__dirname, `./test/${directory}/dist/*`), () => {
+            done();
+        });
+    }
+});
```
snitin315, line 23: This will run before each and every test case. Multiple test suites run simultaneously in jest. This can lead to the removal of files when they shouldn't be. 

I think we can overcome this if we can access the directory of the particular test suite when it's running if that is possible.

snitin315: /cc @SimenB, what do you think is this a good idea to inject `beforeEach` globally like this?

evilebottnawi: Use `Promise.all` and avoid using `node`

SimenB: `beforeEach` should be available globally already, no need to import? I might be misunderstanding.

As for the deletion, you should use `del` or something which returns a promise, then `map` over the directories and feed that array into `Promise.all`

(node 14 has https://nodejs.org/api/fs.html#fs_fspromises_rm_path_options, but I assume this needs to work on more node versions than that)

evilebottnawi (approved):
Great job, thanks

/cc @webpack/cli-team

setupTest.js
```diff
@@ -1 +1,22 @@
...
+const { join } = require('path');
+
+jest.setTimeout(300000);
+
+jasmine.getEnv().addReporter({
+    specStarted: (result) => (jasmine.currentTest = result),
+    specDone: (result) => (jasmine.currentTest = result),
+});
+
+beforeEach(async (done) => {
```
SimenB, line 12: Don't mix the `done` callback and promises

setupTest.js
```diff
@@ -1 +1,22 @@
...
+jest.setTimeout(300000);
+
+jasmine.getEnv().addReporter({
+    specStarted: (result) => (jasmine.currentTest = result),
+    specDone: (result) => (jasmine.currentTest = result),
+});
+
+beforeEach((done) => {
+    rimraf(join(jasmine.testPath, `../bin/*`), () => {
+        done();
```
SimenB, line 14: Jest will only wait for the first `done` call. Which is why I recommended [`del`](https://www.npmjs.com/package/del) instead, so you can do

```js
await del([
  join(jasmine.testPath, `../bin/*`),
  join(jasmine.testPath, `../binary/*`),
  join(jasmine.testPath, `../dist/*`),
]);
```

Also note that you don't want `join` as windows slashes are not valid glob file separators. You can use `path.unix.join`, tho (as noted in `del`'s readme)

SimenB: or use `util.promisify(rimraf)` and `Promise.all`

evilebottnawi:
@snitin315 Can you rebase and finish this PR?

snitin315:
Yes, closing this one due to conflicts. I will open a new one.

Issue #2019, opened 2020-11-02
[Question] How to remove logging in watch mode?


BigMeat233:
I find cli always printing message(ex: Compilation finished, watching files for updates) in v4.

But I only need the error message.

So how to close the success message in v4?

snitin315:
I don't think you can.

evilebottnawi:
Yes, no ability, why do you need this?

piecyk:
This will be possible when webpack-cli would use `InfrastructureLogger` for logging. Similar as webpack-dev-server v4 will. 

There are two issues to use it in plugin 
1. webpack InfrastructureLogger is logging to stderr, that would be change from current stdout.
2. InfrastructureLogger is created from compiler, how we should log before?

BigMeat233:
@evilebottnawi Hello, there is my scenarios:

I use webpack-cli watch mode to build project automatically. 

And I use a webpack plugin named ```friendly-errors-webpack-plugin``` to hide log message in building progress.

But the console always print '99% done plugins watchInfo[webpack-cli] Compilation finished [webpack-cli] watching files for updates...' when building success.

BigMeat233:
@piecyk Hello, I want a flag to open/close the logger in **Line 288** and **Line 340** of the file ```/webpack-cli.js```.

rishabh3112:
Seems to be a valid use-case to me. Many users may prefer to have a less verbose output in watch-mode.

BigMeat233:
> Seems to be a valid use-case to me. Many users may prefer to have a less verbose output in watch-mode.

bingo！

piecyk:
@BigMeat233 output from webpack should be controlled by the level in infrastructureLogging

 
So you would configure it in webpack config via 
```js
infrastructureLogging: {
  level: "error"
},
```

https://webpack.js.org/api/logging

BigMeat233:
@piecyk Thanks for your anwser.

But I still can't hide the cli logging message in watch mode:

```
[webpack-cli] Compilation finished

[webpack-cli] watching files for updates...
```

piecyk:
@BigMeat233 because it's not implemented yet.

evilebottnawi:
In theory we can respect `infrastructureLogging` from config, it should be not hard

piecyk:
@evilebottnawi yep i think that also, would go into this direction

https://github.com/webpack/webpack-cli/compare/master...piecyk:webpack-cli/infrastructure-logging

[...]

PR #2036, opened 2020-11-04
feat(webpack-cli): infrastructure logging

piecyk:

Using InfrastructureLogger for logging output from webpack compiler

I plan to do it

No

**Summary**

Related #2019

Yes

**Other information**

### DIFF
--- a/packages/webpack-cli/__tests__/info.test.js
+++ b/packages/webpack-cli/__tests__/info.test.js
@@ -3,18 +3,17 @@ const path = require('path');
 
 describe('Info', () => {
     it('should run with cli', () => {
-        const { stdout, stderr } = spawnSync(path.resolve(__dirname, '../bin/cli.js'), ['info'], {
+        const { stdout } = spawnSync(path.resolve(__dirname, '../bin/cli.js'), ['info'], {
             cwd: path.resolve(__dirname),
             reject: false,
         });
         expect(stdout).toContain('System');
         expect(stdout).toContain('Binaries');
         expect(stdout).toContain('OS');
-        expect(stderr).toBeFalsy();
     });
 
     it('should work with flags', () => {
-        const { stdout, stderr } = spawnSync(path.resolve(__dirname, '../bin/cli.js'), ['info', '--output=json'], {
+        const { stdout } = spawnSync(path.resolve(__dirname, '../bin/cli.js'), ['info', '--output=json'], {
             cwd: path.resolve(__dirname),
             reject: false,
         });
@@ -25,6 +24,5 @@ describe('Info', () => {
             expect(output['System']['OS']).toBeTruthy();
         };
         expect(testJSON).not.toThrow();
-        expect(stderr).toBeFalsy();
     });
 });
--- a/packages/webpack-cli/lib/plugins/WebpackCLIPlugin.js
+++ b/packages/webpack-cli/lib/plugins/WebpackCLIPlugin.js
@@ -1,62 +1,119 @@
 const packageExists = require('../utils/package-exists');
 const webpack = packageExists('webpack') ? require('webpack') : undefined;
-const logger = require('../utils/logger');
-
-const PluginName = 'webpack-cli';
+const { getStatsOptions } = require('../utils/stats-options');
+const { PluginName } = require('../utils/name');
 
 class WebpackCLIPlugin {
     constructor(options) {
         this.options = options;
+
+        this.isWatch = false;
     }
 
     async apply(compiler) {
+        const { progress } = this.options;
+        if (progress) {
+            this.appendProgressPlugin(compiler, progress);
+        }
+
+        const isWebpack5 = Boolean(compiler.webpack);
+
+        const logger = compiler.getInfrastructureLogger(PluginName);
+
+        const resolveName = (obj) => (obj.name ? `${obj.name} ` : '');
+
+        const done = (stats) => {
+            const printStats = (childCompiler, childStats) => {
+                const name = resolveName(childCompiler);
+
+                const status = childStats.toString({
+                    all: false,
+                    version: true,
+                    timings: true,
+                });
+
+                const normalizedStatus = (() => {
+                    if (isWebpack5) {
+                        return status;
+                    } else {
+                        const [version, time] = status.split('\n').map((line) => {
+                            const value = line.split(':')[1] || '';
+                            return value.trim();
+                        });
+
+                        return `${name ? name + `(${version})` : version} compiled in ${time}`;
+                    }
+                })();
+
+                if (childStats.hasErrors()) {
+                    logger.error(normalizedStatus);
+                } else if (childStats.hasWarnings()) {
+                    logger.warn(normalizedStatus);
+                } else {
+                    logger.info(normalizedStatus);
+                }
+
+                const statsString = childStats.toString(getStatsOptions(childCompiler));
+                process.stdout.write(statsString + '\n');
+
+                if (this.isWatch) {
+                    logger.info(name + 'watching files for updates...');
+                }
+            };
+
+            if (compiler.compilers) {
+                compiler.compilers.forEach((compilerFromMultiCompileMode, index) => {
+                    printStats(compilerFromMultiCompileMode, stats.stats[index]);
+                });
+            } else {
+                printStats(compiler, stats);
+            }
+        };
+
+        compiler.hooks.run.tap(PluginName, (compilation) => {
+            logger.info(resolveName(compilation) + 'compilation starting...');
+        });
+        compiler.hooks.watchRun.tap(PluginName, (compilation) => {
+            logger.info(resolveName(compilation) + 'compilation starting...');
+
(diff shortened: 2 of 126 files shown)


piecyk:
As mention before webpack infrastructure logging outputs to `stderr` that affect's lot's of test.

packages/webpack-cli/__tests__/info.test.js
```diff
@@ -3,28 +3,26 @@ const path = require('path');
...
     it('should work with flags', () => {
-        const { stdout, stderr } = spawnSync(path.resolve(__dirname, '../bin/cli.js'), ['info', '--output=json'], {
+        const { stderr } = spawnSync(path.resolve(__dirname, '../bin/cli.js'), ['info', '--output=json'], {
             cwd: path.resolve(__dirname),
             reject: false,
         });
 
         const testJSON = () => {
-            const output = JSON.parse(stdout);
+            const output = JSON.parse(stderr);
```
evilebottnawi, line 22: Breaking change

packages/webpack-cli/lib/plugins/WebpackCLIPlugin.js
```diff
@@ -1,62 +1,143 @@
...
-const logger = require('../utils/logger');
+const { getStatsOptions } = require('../utils/stats-options');
+const { PluginName } = require('../utils/name');
 
-const PluginName = 'webpack-cli';
-
-class WebpackCLIPlugin {
-    constructor(options) {
-        this.options = options;
+class WebpackCLIRuntimePlugin {
```
evilebottnawi, line 6: No need to rename, keep WebpackCLIPLugin

packages/webpack-cli/lib/plugins/WebpackCLIPlugin.js
```diff
@@ -1,62 +1,143 @@
...
+    constructor() {
+        this.isWatch = false;
     }
 
     async apply(compiler) {
-        const compilers = compiler.compilers || [compiler];
+        const isWebpack5 = !!compiler.webpack;
```
evilebottnawi, line 12: Boolean(compiler.webpack)

packages/webpack-cli/lib/utils/logger.js
```diff
@@ -1,11 +1,25 @@
...
+    log: write(),
+
+    /**
+     * @deprecated use info
+     */
+    success: write(undefined, green),
+    /**
+     * @deprecated use info
+     */
+    raw: write(''),
```
evilebottnawi, line 24: Don't touch it here

packages/webpack-cli/lib/utils/logger.js
```diff
@@ -1,11 +1,25 @@
 const util = require('util');
 const { red, cyan, yellow, green } = require('colorette');
 
+const write = (prefix = '[webpack-cli] ', color = (val) => val) => {
+    return (...args) => {
+        const str = prefix + color(util.format(...args));
+        process.stderr.write(str + '\n');
```
evilebottnawi, line 7: No, it is real bad idea

evilebottnawi (changes requested):
Sorry, it is big breaking change, do not touch stdout and stderr, it was big problems for previously versions and we fix it

piecyk:
@evilebottnawi weback InfrastructureLogger is writing to `stderr` https://github.com/webpack/webpack/blob/master/lib/node/nodeConsole.js#L48

Aligned the utils/logger for simpler refactor test, and not go one by one 🙈  Let's revert logger to console and fix the test, just ping me if that is something we want as it's not a small thing.

packages/webpack-cli/lib/plugins/WebpackCLIPlugin.js
piecyk: Why it's better ?

packages/webpack-cli/lib/plugins/WebpackCLIPlugin.js
piecyk: Yep, just was playing around with some ideas. Would be nice if `webpack-dev-middleware` and `webpack-cli` output messages in same format. Basic sharing the plugin https://github.com/webpack/webpack-dev-middleware/blob/master/src/utils/setupHooks.js

evilebottnawi:
hm, I need dicussion with sokra

evilebottnawi:
Anyway we can't change it here, because it is breaking changte

piecyk:
Looks like reverting logger was not that bad. Still needed to disable few test, like with `--bail` is something off...

But yeah, also proposing some changes in what we log, now we show 

`Compilation finished` at end, and simliar information we show in stats, 
`webpack 5.4.0 compiled successfully in 610 ms`. Imho better to use this rather finished

something like
 
v4
```
$ webpack --watch --progress
<i> [webpack-cli] compilation starting...
<i> [webpack-cli] output:
<i> Hash: 01af0357780e7ff104a9
<i> Built at: 11/06/2020 12:41:15 PM
<i>   Asset      Size  Chunks             Chunk Names
<i> main.js  4.01 KiB    main  [emitted]  main
<i> Entrypoint main = main.js
<i> [./src/index.ts] 53 bytes {main} [built]
<i> [webpack-cli] webpack 4.44.2 compiled in 1871ms
<i> [webpack-cli] watching files for updates...
```

v5
```
$ webpack --watch --progress
<i> [webpack-cli] compilation starting...
<i> [webpack-cli] output:
<i> asset main.js 3.79 KiB [emitted] (name: main)
<i> runtime modules 668 bytes 3 modules
<i> ./src/index.ts 53 bytes [built] [code generated]
<i> [webpack-cli] webpack 5.4.0 compiled in 632 ms
<i> [webpack-cli] watching files for updates...
```

evilebottnawi:
Yes, we should remove `Compilation finished`

evilebottnawi:
`<i> [webpack-cli] output:` is unnecessary and looks like spam

evilebottnawi:
And yes, you will right we should - All diagnostic output goes to stderr. only the stats output goes to stdout

packages/webpack-cli/lib/utils/stats-options.js
```diff
@@ -0,0 +1,26 @@
...
+            options = webpack.Stats.presetToOptions(options);
+        }
+    }
+
+    options.colors = typeof options.colors !== 'undefined' ? options.colors : coloretteOptions.enabled;
+
+    return options;
+};
+
+module.exports = { getStatsOptions };
```
evilebottnawi, line 26: No utils relates to complation, use methods classes

evilebottnawi (commented):
I think we should avoid do here a lot of changes, let's separate them, I can't remove and merge this all big changes, sorry, let's focus on migrate infrastructure logger

piecyk:
> I think we should avoid do here a lot of changes, let's separate them, I can't remove and merge this all big changes, sorry, let's focus on migrate infrastructure logger

Yes i know. Basic problem is that migrating to infrastructure logger is bigger one as it touch a lot of testes as we switch from.

> And yes, you will right we should - All diagnostic output goes to stderr. only the stats output goes to stdout

So basic when logging stats we should not use infrastructure logger ?

packages/webpack-cli/lib/utils/stats-options.js
piecyk: This was execrated to utils as, moved stats display to WebpackCLIPlugin, but wanted to keep `json output` in WebpackCLI compiler callback.

evilebottnawi:
Yes, for stats we should use `stdout` and do not use infrastructure logic

evilebottnawi:
> Yes i know. Basic problem is that migrating to infrastructure logger is bigger one as it touch a lot of testes as we switch from.

So let's separate it

piecyk:
> So let's separate it

ok current plan 

* [x] revert utils/logger changes. 
* [x] use infrastructure logger in context of webpack
* [ ] stats should log to stdout 
* [ ] fix test

packages/webpack-cli/lib/plugins/WebpackCLIPlugin.js
```diff
@@ -1,63 +1,120 @@
...
+                        });
 
-                    const isProfile = this.options.progress === 'profile';
+                        return `${name ? name + `(${version})` : version} compiled in ${time}`;
+                    }
+                })();
 
-                    new ProgressPlugin({ profile: isProfile }).apply(compiler);
+                if (childStats.hasErrors()) {
+                    logger.error(normalizedStatus);
```
snitin315, line 49: We should set exitCode = 1 for errors.

piecyk: Exit code is set in callback in webpack-cli.js this wasn't changed https://github.com/webpack/webpack-cli/pull/2036/files#diff-7c711e2a4ef4b16d2cbbe1658495723402f9692da654cb6fa959127f3dc779f2R357-R359

Wondering if we should reset it if next complication is has no errors 🤔

piecyk:
Suppressed by #2144

PR #2041, opened 2020-11-05
fix: parsing of --amd flag

snitin315:
fix

yes
no
**Summary**
`amd` can accept only `false` as boolean. Hence only `--no-amd` is valid.

### Before
Screenshot at 2020-11-05 08-38-11

### After
Screenshot at 2020-11-05 08-36-39

No

**Other information**
Open to suggestions 😄

### DIFF
--- a/packages/webpack-cli/lib/groups/basicResolver.js
+++ b/packages/webpack-cli/lib/groups/basicResolver.js
@@ -30,6 +30,9 @@ function resolveArgs(args) {
         if (arg === 'watch') {
             finalOptions.options.watch = true;
         }
+        if (arg === 'amd') {
+            finalOptions.options.amd = args[arg];
+        }
         if (arg === 'entry') {
             finalOptions.options[arg] = args[arg];
         }
--- a/test/core-flags/amd-flag.test.js
+++ b/test/core-flags/amd-flag.test.js
@@ -3,6 +3,15 @@
 const { run } = require('../utils/test-utils');
 
 describe('--no-amd flag', () => {
+    it('throw error for --amd', () => {
+        const { stderr, stdout, exitCode } = run(__dirname, ['--amd']);
+
+        expect(stderr).toContain('Invalid configuration object');
+        expect(stderr).toContain('configuration.amd should be false');
+        expect(exitCode).toBe(2);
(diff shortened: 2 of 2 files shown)


packages/webpack-cli/lib/groups/basicResolver.js
```diff
@@ -30,6 +30,9 @@ function resolveArgs(args) {
         if (arg === 'watch') {
             finalOptions.options.watch = true;
         }
+        if (arg === 'amd') {
+            finalOptions.options.amd = args[arg];
+        }
```
snitin315, line 35: I think here we can tell the user to use `--no-amd` instead. Any thoughts?

evilebottnawi: Yes, I think we should avoid it in our code, what is webpack.cli return on amd?

evilebottnawi: If only `false` we should support on `--no-amd`

snitin315: yes only `amd: false` is allowed.

evilebottnawi: So let's support only `--no-amd`, do we need fix something on webpack side?

snitin315: Webpack doesn't export any negative flags. They are added by CLI only. I think we can do something like - 

```js
if (arg === 'amd') {
       if (args[arg] === true) {
           logger.error('configuration.amd can accept only false as boolean. Only --no-amd is supported.');
            process.exit(2);
        }
          finalOptions.options.amd = args[arg];
   }
```

evilebottnawi: We should avoid put logic for special values of flags inside webpack-cli, we should fix it on webpack side

snitin315: /cc @sokra

alexander-akait:
Fixed in master, webpack return `problems`, so we output error message when value is invalid, anyway you can help with https://github.com/webpack/webpack/issues/12113 for better DX

PR #2046, opened 2020-11-05
refactor: utils

anshumanv:

refactor

Yes

NA

**Summary**
- default export
- group similar utils
- rm redundant 

no

**Other information**


evilebottnawi (approved):
Good job, I will send a PR with refactor more in future, found a lot of edge cases when refactor it too

packages/webpack-cli/lib/utils/unknown-args.js
```diff
@@ -1,5 +1,25 @@
...
+// Contains an array of strings with core cli flags and their aliases
+const flagNames = core
+    .map(({ alias, name }) => {
+        if (name === 'help') return [];
+        if (alias) {
+            return [`--${name}`, `-${alias}`];
+        }
+        return [`--${name}`];
+    })
+    .reduce((arr, val) => arr.concat(val), []);
```
evilebottnawi, line 22: Ideally we should avoid it, because we do it on `require` and reduce performance in initial start, because we have very many flags, but I will fix it in future too

anshumanv: yep will improve it next

evilebottnawi:
Let's merge and continue refactor

PR #2042, opened 2020-11-05
fix: allow to disable stats color via --no-color

snitin315:
**What kind of change does this PR introduce?**
bugfix

**Did you add tests for your changes?**
Yes
**If relevant, did you update the documentation?**
no
**Summary**

`stats` colors should be disabled if we use `--no-color`.

Fixes #2038  as well

### Before
Screenshot at 2020-11-05 09-11-53

### After
Screenshot at 2020-11-05 09-12-28

**Does this PR introduce a breaking change?**
Nope

**Other information**
No

PR #2049, opened 2020-11-05
fix: merge resolvers

anshumanv:

refactor

Yes

NA

**Summary**

This merges multiple arg resolvers so we don't run them all separately even when they're simple. Not merged with config resolver as it's quite complex.

No

**Other information**


packages/webpack-cli/lib/groups/basicResolver.js
```diff
@@ -12,7 +17,41 @@ const WEBPACK_OPTION_FLAGS = core
...
+    let finalMode;
+    if (mode) {
+        finalMode = mode;
+    } else if (configMode) {
+        finalMode = configMode;
+    } else if (NODE_ENV && (NODE_ENV === PRODUCTION || NODE_ENV === DEVELOPMENT)) {
+        finalMode = NODE_ENV;
+    } else {
+        finalMode = PRODUCTION;
+    }
```
evilebottnawi, line 48: Ideally we should remove it, because we set production by default, but it is out of scope webpack-cli, otherwise webpack will not throw warning on empty mode value

anshumanv: yep lets remove

evilebottnawi: Also let's add test for this to avoid regression in future

anshumanv: done

packages/webpack-cli/lib/groups/basicResolver.js
```diff
@@ -34,7 +71,76 @@ function resolveArgs(args) {
...
+        });
+    } else {
+        const computedMode = assignMode(mode, configOptions);
+        if (computedMode) {
+            finalOptions.options.mode = computedMode;
+        }
+    }
     return finalOptions;
-}
+};
```
evilebottnawi, line 144: Let's move it to WebpackCLI classs as method `resolveArgs`

evilebottnawi: No groups anymore

evilebottnawi (commented):
Anyway good job, one note and we can merge it

packages/webpack-cli/lib/groups/basicResolver.js
anshumanv: I think we should do it separately, a lot of tests are breaking

packages/webpack-cli/lib/groups/basicResolver.js
anshumanv: yes we can move but then it will be a big file, still continue?

evilebottnawi: I don't think it will be very big, keeping logic in one places sometimes is better than separate it, this will help us in future improvements and refactor to see everything

anshumanv: cool lets do it

packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -1,23 +1,65 @@
...
+    .filter((coreFlag) => {
+        return coreFlag.group === groups.BASIC_GROUP;
+    })
+    .reduce((result, flagObject) => {
+        result.push(flagObject.name);
+        if (flagObject.alias) {
+            result.push(flagObject.alias);
+        }
+        return result;
+    }, []);
```
evilebottnawi, line 29: Can we move it inside `resolveArgs`? no need decrease initial start

packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -1,23 +1,65 @@
...
+
+const PRODUCTION = 'production';
+const DEVELOPMENT = 'development';
```
evilebottnawi, line 32: Let's move it inside `resolveArgs` too

packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -1,23 +1,65 @@
...
+        finalMode = mode;
+    } else if (configMode) {
+        finalMode = configMode;
+    } else if (NODE_ENV && (NODE_ENV === PRODUCTION || NODE_ENV === DEVELOPMENT)) {
+        finalMode = NODE_ENV;
+    } else {
+        finalMode = PRODUCTION;
+    }
+    return finalMode;
+};
```
evilebottnawi, line 61: Move this inside `resolveArgs` too

packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -48,6 +90,96 @@ class WebpackCLI {
         this._mergeOptionsToConfiguration(configWithDefaults);
     }
 
+    async resolveArgs(args, configOptions = {}) {
+        const { outputPath, stats, json, mode, target, prefetch, hot, analyze } = args;
```
evilebottnawi, line 94: Can we add `if (args.length > 0) { return; }` before `const`

evilebottnawi (commented):
Couple improvements and we can merge it

packages/webpack-cli/lib/webpack-cli.js
anshumanv: done

packages/webpack-cli/lib/webpack-cli.js
anshumanv: done

packages/webpack-cli/lib/webpack-cli.js
anshumanv: done

packages/webpack-cli/lib/webpack-cli.js
anshumanv: args is object

evilebottnawi: `Object.keys()`, so we will very fast on initial start (no parsing)

evilebottnawi:
And let's fix tests

packages/webpack-cli/lib/webpack-cli.js
anshumanv: yep done

evilebottnawi:
Great, I will merge after CI green, next - let's refactor and resolving config as method `resolveConfig`

anshumanv:
Yep

evilebottnawi:
@anshumanv tests still broken

anshumanv:
Yes wip 😄

packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -48,7 +48,8 @@ class WebpackCLI {
 
     async resolveArgs(args, configOptions = {}) {
         // Since color flag has a default value, when there are no other args then exit
-        if (Object.keys(args).length === 1) return {};
+        // eslint-disable-next-line no-prototype-builtins
+        if (Object.keys(args).length === 1 && args.hasOwnProperty('color')) return {};
```
evilebottnawi: Why we have `color` property here?

anshumanv: because color is default in args, we only need to skip it in that case

anshumanv: in tests we're passing properties which are not color

evilebottnawi: hm, why we have `color` in args?

anshumanv: I think default is assigned there, will take a look and remove next

evilebottnawi: Yep, let's do it in the new PR

evilebottnawi (approved):
Good improve, A couple more of these things and we will normalize the architecture, and it will be very easy to fix errors in future

packages/webpack-cli/lib/webpack-cli.js
snitin315: WIP already on color https://github.com/webpack/webpack-cli/pull/2042#discussion_r518730261

snitin315 (approved):

Issue #2021, opened 2020-11-02
Could not find a yeoman configuration file.


arokia-innominds:
when i run the webpack-cli init ./ i am getting this error

Could not find a yeoman configuration file.

Please make sure to use 'this.config.set('configuration', this.configuration);' at the end of the generator.

Unexpected Error
please file an issue here https://github.com/webpack/webpack-cli/issues/new?template=Bug_report.md

[TypeError: Cannot read property 'usingDefaults' of undefined]

rishabh3112:
@arokia-innominds `webpack init` doesn't accept path for scaffolding, instead you run this in folder you may want to scaffold project.
Here is a example assuming a global installation of `webpack-cli` as per your report:
```
mkdir project
cd project
webpack init
```

P.S.
Please don't remove our issue template while filing issues in future. This may lead to closing of issue without response in future.

evilebottnawi:
I want to mark it as feature, allow to generate loader/plugin code in any directory

evilebottnawi:
@webpack/cli-team should be easy to implement

jamesgeorge007:
I'll take a look at it.

jamesgeorge007:
For `init`, isn't it better to have a flag for the purpose? (say, `--generate-path`)

jamesgeorge007:
Closing.

PR #2050, opened 2020-11-06
feat(init): add --generation-path flag

jamesgeorge007:
feature

Yes

Nope

**Summary**
Added new `--generation-path` flag for `init` subcommand for scaffolding in a specified path. Addresses #2021

**Other information**
https://github.com/webpack/webpack-cli/issues/2021#issuecomment-722395490

### DIFF
--- a/packages/init/src/index.ts
+++ b/packages/init/src/index.ts
@@ -3,7 +3,7 @@ import { modifyHelperUtil, npmPackagesExists } from '@webpack-cli/utils';
 
 const AUTO_PREFIX = '--auto';
 const CONFIG_PREFIX = '--force';
-const PATH_PREFIX = '--generate-path';
+const PATH_PREFIX = '--generation-path';
 
 /**
  *
@@ -22,14 +22,14 @@ export default function initializeInquirer(...args: string[]): Function | void {
     const generateConfig = packages.includes(CONFIG_PREFIX);
     const genPathPrefix = packages.includes(PATH_PREFIX);
 
-    let generatePath: string;
+    let generationPath: string;
     if (genPathPrefix) {
         const idx = packages.indexOf(PATH_PREFIX);
-        // Retrieve the path supplied along with --generate-path
-        generatePath = packages[idx + 1];
+        // Retrieve the path supplied along with --generation-path
+        generationPath = packages[idx + 1];
     }
     if (packages.length === 0 || includesDefaultPrefix || generateConfig || genPathPrefix) {
-        return modifyHelperUtil('init', initGenerator, null, null, includesDefaultPrefix, generateConfig, generatePath);
+        return modifyHelperUtil('init', initGenerator, null, null, includesDefaultPrefix, generateConfig, generationPath);
     }
     return npmPackagesExists(packages);
 }
--- a/packages/init/src/index.ts
+++ b/packages/init/src/index.ts
@@ -3,6 +3,8 @@ import { modifyHelperUtil, npmPackagesExists } from '@webpack-cli/utils';
 
 const AUTO_PREFIX = '--auto';
 const CONFIG_PREFIX = '--force';
+const PATH_PREFIX = '--generate-path';
+
 /**
  *
  * First function to be called after running the init flag. This is a check,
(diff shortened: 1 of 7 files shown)


test/init/generate-path/init-generate-path.test.js
```diff
@@ -0,0 +1,45 @@
...
+
+const genPath = join(__dirname, 'test-assets');
+
+describe('init generate-path flag', () => {
+    beforeAll(() => {
+        rimraf.sync(genPath);
+    });
+
+    it('should scaffold in the specified path with --generate-path', () => {
+        const { stdout } = run(__dirname, ['init', '--generate-path', 'test-assets', '--auto'], false);
```
evilebottnawi, line 18: Can we refactor it to `webpack init ./path/to/something`? And why we need `--auto`?

packages/init/src/index.ts
```diff
@@ -18,8 +20,16 @@ export default function initializeInquirer(...args: string[]): Function | void {
...
+    const genPathPrefix = packages.includes(PATH_PREFIX);
+
+    let generatePath: string;
+    if (genPathPrefix) {
+        const idx = packages.indexOf(PATH_PREFIX);
+        // Retrieve the path supplied along with --generate-path
+        generatePath = packages[idx + 1];
+    }
+    if (packages.length === 0 || includesDefaultPrefix || generateConfig || genPathPrefix) {
+        return modifyHelperUtil('init', initGenerator, null, null, includesDefaultPrefix, generateConfig, generatePath);
```
evilebottnawi, line 32: What is the magic logic?

test/init/generate-path/init-generate-path.test.js
jamesgeorge007: `init` already takes in a package (`webpack-scaffold-*`) as an arg. Being a subcommand having a flag to accept any other arguments would be better.

evilebottnawi: I see

/cc @webpack/cli-team What do you think about `--generate-path` name for flag?

packages/init/src/index.ts
evilebottnawi: Maybe you can clarify?

test/init/generate-path/init-generate-path.test.js
jamesgeorge007: Supplying `--auto` flag would allow us to go with the defaults.

packages/init/src/index.ts
jamesgeorge007: I haven't updated the logic, added an argument that would hold a reference to the specified path.

evilebottnawi: I think here is bug, using `packages.indexOf(PATH_PREFIX)` can be wrong, because you can use `PATH_PREFIX` multiple time in path, we need look at logic and fix it

jamesgeorge007: `indexOf()` returns the first occurence.

evilebottnawi: but it is bug, please refactor this logic, it is buggy

test/init/generate-path/init-generate-path.test.js
rishabh3112: > What do you think about --generate-path name for flag?

what out `--dir`?

packages/init/src/index.ts
jamesgeorge007: We can provide support for `--generate-path=path` syntax once we leverage the `arg-parser`.

test/init/generate-path/init-generate-path.test.js
evilebottnawi: @jamesgeorge007 What do you think about arguments name?

jamesgeorge007: IMO `--generate-path` sounds good.

evilebottnawi: /cc @webpack/cli-team need more options

snitin315: Any issues with `--path` only?

evilebottnawi: I think `--path` can be some misleading

snitin315: hmm, then `--generation-path` or `--generate-path` sounds good.

evilebottnawi: `--generation-path` sounds good for me

jamesgeorge007: I can think of a shortened version: `--gen-path`. How does that sound?

evilebottnawi: I prefer do not shorted, it can be difficult to read

snitin315 (commented):
Please update docs too.

evilebottnawi (approved):

snitin315 (approved):

PR #2053, opened 2020-11-06
fix: remove default value for devtool

anshumanv:
fix

NA

NA

**Summary**
Remove the default value of some flags since they are invalid and not used in most cases.

No

**Other information**

### DIFF
--- a/packages/webpack-cli/lib/utils/cli-flags.js
+++ b/packages/webpack-cli/lib/utils/cli-flags.js
@@ -156,7 +156,6 @@ const core = [
         usage: '--devtool <value>',
         type: String,
         alias: 'd',
-        defaultValue: undefined,
         group: BASIC_GROUP,
         description: 'Determine source maps to use',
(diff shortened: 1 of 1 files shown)


packages/webpack-cli/lib/utils/cli-flags.js
```diff
@@ -92,7 +92,6 @@ const core = [
         usage: '--color',
         type: Boolean,
         negative: true,
-        defaultValue: true,
         description: 'Enable/Disable colors on console',
```
snitin315, line 95: the `color` needs some more fixes, I'm removing it in
https://github.com/webpack/webpack-cli/pull/2042#discussion_r518730261

snitin315 (changes requested):
Let's remove it for `devtool` only here.

snitin315 (approved):

evilebottnawi (approved):

PR #2060, opened 2020-11-07
refactor: get-package-manger logic for perf improvements

snitin315:
refactor.

yes updated.
NA
**Summary**

avoid multiple fs calls.

No

**Other information**
No

### DIFF
--- a/packages/webpack-cli/lib/utils/get-package-manager.js
+++ b/packages/webpack-cli/lib/utils/get-package-manager.js
@@ -1,29 +1,46 @@
 const fs = require('fs');
 const path = require('path');
+const logger = require('./logger');
 const { sync } = require('execa');
 
 /**
  *
  * Returns the name of package manager to use,
- * preferring yarn over npm if available
+ * preference order - npm > yarn > pnpm
  *
  * @returns {String} - The package manager name
  */
 function getPackageManager() {
-    const hasLocalYarn = fs.existsSync(path.resolve(process.cwd(), 'yarn.lock'));
     const hasLocalNpm = fs.existsSync(path.resolve(process.cwd(), 'package-lock.json'));
-    const hasLocalPnpm = fs.existsSync(path.resolve(process.cwd(), 'pnpm-lock.yaml'));
+
+    if (hasLocalNpm) {
+        return 'npm';
+    }
+
+    const hasLocalYarn = fs.existsSync(path.resolve(process.cwd(), 'yarn.lock'));
 
     if (hasLocalYarn) {
         return 'yarn';
-    } else if (hasLocalNpm) {
-        return 'npm';
-    } else if (hasLocalPnpm) {
+    }
+
+    const hasLocalPnpm = fs.existsSync(path.resolve(process.cwd(), 'pnpm-lock.yaml'));
+
+    if (hasLocalPnpm) {
         return 'pnpm';
     }
 
     try {
-        // if the sync function below fails because yarn is not installed,
+        // the sync function below will fail if npm is not installed,
+        // an error will be thrown
+        if (sync('npm', ['--version'])) {
+            return 'npm';
+        }
+    } catch (e) {
+        // Nothing
+    }
+
+    try {
+        // the sync function below will fail if yarn is not installed,
         // an error will be thrown
         if (sync('yarn', ['--version'])) {
             return 'yarn';
@@ -32,7 +49,16 @@ function getPackageManager() {
         // Nothing
     }
 
-    return 'npm';
+    try {
+        // the sync function below will fail if pnpm is not installed,
+        // an error will be thrown
+        if (sync('pnpm', ['--version'])) {
+            return 'pnpm';
+        }
+    } catch (e) {
+        logger.error('No package manager found.');
+        process.exit(2);
+    }
 }
 
 module.exports = getPackageManager;
--- a/packages/webpack-cli/lib/utils/__tests__/get-package-manager.test.js
+++ b/packages/webpack-cli/lib/utils/__tests__/get-package-manager.test.js
@@ -12,6 +12,7 @@ jest.setMock('execa', {
 const getPackageManager = require('../get-package-manager');
 
 jest.mock('cross-spawn');
+jest.mock('../get-package-manager', () => jest.fn());
 const globalModulesNpmValue = 'test-npm';
 jest.setMock('global-modules', globalModulesNpmValue);
 jest.setMock('enquirer', {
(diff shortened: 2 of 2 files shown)


evilebottnawi (commented):
https://github.com/webpack/webpack-cli/blob/1719dcfff6bbc9543b4453f0a5c17f105395ee5e/packages/webpack-cli/lib/utils/get-package-manager.js#L40 let's improve logic for pnpm here too

snitin315:
> https://github.com/webpack/webpack-cli/blob/1719dcfff6bbc9543b4453f0a5c17f105395ee5e/packages/webpack-cli/lib/utils/get-package-manager.js#L40
> 
> let's improve logic for pnpm here too


That will prioritize global `pnpm` over global `npm`.

packages/webpack-cli/lib/utils/get-package-manager.js
```diff
@@ -10,9 +10,19 @@ const { sync } = require('execa');
...
  */
 function getPackageManager() {
-    const hasLocalYarn = fs.existsSync(path.resolve(process.cwd(), 'yarn.lock'));
     const hasLocalNpm = fs.existsSync(path.resolve(process.cwd(), 'package-lock.json'));
-    const hasLocalPnpm = fs.existsSync(path.resolve(process.cwd(), 'pnpm-lock.yaml'));
+
+    let hasLocalYarn = false;
+
+    if (!hasLocalNpm) {
+        hasLocalYarn = fs.existsSync(path.resolve(process.cwd(), 'yarn.lock'));
```
jamesgeorge007, line 18:
```suggestion
        hasLocalYarn = fs.existsSync(path.resolve('yarn.lock'));
```

packages/webpack-cli/lib/utils/get-package-manager.js
```diff
@@ -10,9 +10,19 @@ const { sync } = require('execa');
...
+    }
+
+    let hasLocalPnpm = false;
+
+    if (!hasLocalYarn) {
+        hasLocalPnpm = fs.existsSync(path.resolve(process.cwd(), 'pnpm-lock.yaml'));
```
jamesgeorge007, line 24:
```suggestion
        hasLocalPnpm = fs.existsSync(path.resolve('pnpm-lock.yaml'));
```

packages/webpack-cli/lib/utils/get-package-manager.js
```diff
@@ -10,9 +10,19 @@ const { sync } = require('execa');
  * @returns {String} - The package manager name
  */
 function getPackageManager() {
-    const hasLocalYarn = fs.existsSync(path.resolve(process.cwd(), 'yarn.lock'));
     const hasLocalNpm = fs.existsSync(path.resolve(process.cwd(), 'package-lock.json'));
```
snitin315, line 13:
```suggestion
    const hasLocalNpm = fs.existsSync(path.resolve('package-lock.json'));
```

packages/webpack-cli/lib/utils/get-package-manager.js
snitin315: `process.cwd()` is required. else it will resolve to `/home/snitin315/Projects/webpack-cli/package-lock.json` when it should be `/home/snitin315/Projects/webpack-cli/packages/webpack-cli/lib/utils/__tests__/test-all-lock/package-lock.json`

packages/webpack-cli/lib/utils/get-package-manager.js
```diff
@@ -10,9 +10,19 @@ const { sync } = require('execa');
...
-    const hasLocalPnpm = fs.existsSync(path.resolve(process.cwd(), 'pnpm-lock.yaml'));
+
```
rishabh3112, line 14: Use early returns instead.
```
const hasLocalNpm = ...
if (hasLocalNpm) {
return "npm";
}

const hasLocalYarn = ...
.
.
.
```

rishabh3112 (dismissed):
looks good, left a suggestion.

packages/webpack-cli/lib/utils/get-package-manager.js
jamesgeorge007: Ah yes, it is required; since we are mocking `process.cwd()`.

packages/webpack-cli/lib/utils/get-package-manager.js
```diff
@@ -1,37 +1,33 @@
...
-    } else if (hasLocalPnpm) {
-        return 'pnpm';
     }
 
-    try {
-        // if the sync function below fails because yarn is not installed,
-        // an error will be thrown
-        if (sync('yarn', ['--version'])) {
-            return 'yarn';
-        }
```
jamesgeorge007, line 30: It seems mocking `execa.sync` isn't required anymore.

snitin315: yep removed.

evilebottnawi: No, it was for projects where lock files disabled, it is rare case right now, but better support it

snitin315: just because `yarn` is installed does not mean users want to use `yarn`. I'm forced to have it installed for a project or two, but I have hundreds of other projects that use npm.

The default should be `npm` since that's the node's default.

evilebottnawi: Ideally we have here more powerful logic, if we don't found lock files, we should start to check `npm -v`/`yarn -v`/`pnpm -v`, and if we not found something here too, we should throw an error - `No package managers found please install package manually`, it is possible when node was wtihout package manager, npm can be removed too

snitin315: I see so the order should be `global npm` > `global yarn` > `global pnpm` ?

evilebottnawi: Yes, logic:
- search npm lock file
- search yarn lock file
- search pnpm lock file
- run `npm -v`
- run `yarn -v`
- run `pnpm -v`
- error - No package manager found

packages/webpack-cli/lib/utils/get-package-manager.js
```diff
@@ -32,7 +49,16 @@ function getPackageManager() {
...
-    return 'npm';
+    try {
+        // the sync function below will fail if pnpm is not installed,
+        // an error will be thrown
+        if (sync('pnpm', ['--version'])) {
+            return 'pnpm';
+        }
+    } catch (e) {
+        logger.error('No package manager found.');
+        process.exitCode = 2;
```
evilebottnawi, line 60: I think we need run `process.exit(2)` here, we should not continue work

snitin315: I tried that earlier but the unit test seems to fail too early.
Screenshot at 2020-11-09 18-06-05

evilebottnawi: Yes, please mock `process` in this case

evilebottnawi (approved):

anshumanv (approved):
looks good

PR #2087, opened 2020-11-09
fix(init): warn the user on supplying an invalid value for --generation-path

jamesgeorge007:
bugfix

Yes

N/A

**Summary**
An invalid value supplied for `--generation-path` results in an error.

Nope

**Other information**
N/A

### DIFF
--- a/packages/init/src/index.ts
+++ b/packages/init/src/index.ts
@@ -1,5 +1,8 @@
 import { initGenerator } from '@webpack-cli/generators';
 import { modifyHelperUtil, npmPackagesExists } from '@webpack-cli/utils';
+import { utils } from 'webpack-cli';
+
+const { logger } = utils;
 
 const AUTO_PREFIX = '--auto';
 const CONFIG_PREFIX = '--force';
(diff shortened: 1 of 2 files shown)


packages/init/src/index.ts
```diff
@@ -27,6 +30,11 @@ export default function initializeInquirer(...args: string[]): Function | void {
         const idx = packages.indexOf(PATH_PREFIX);
         // Retrieve the path supplied along with --generation-path
         generationPath = packages[idx + 1];
+        // Throw an error in case an invalid value is supplied
+        if (generationPath.startsWith('-')) {
+            logger.error('Please specify a valid path');
+            process.exit(2);
+        }
```
evilebottnawi, line 37: There are a lot of other edge cases, we really need to switch on arg parser here

jamesgeorge007: Indeed, the most likely being supplying other flags just after the `--generation-path` flag, which is what this PR aims to address.

alexander-akait (commented):
There is not problems here, `--auto` is value for the `generation-path` option

PR #2120, opened 2020-11-19
(fix): reduce size of stats string

nickshevr:

refactoring

No, it's minor

**Summary**

You don't need indents for `fs.write(json)`.

No

**Other information**

### DIFF
--- a/packages/webpack-cli/lib/webpack-cli.js
+++ b/packages/webpack-cli/lib/webpack-cli.js
@@ -656,9 +656,9 @@ class WebpackCLI {
                 : getStatsOptionsFromCompiler(compiler);
 
             if (outputOptions.json === true) {
-                process.stdout.write(JSON.stringify(stats.toJson(foundStats), null, 2) + '\n');
+                process.stdout.write(JSON.stringify(stats.toJson(foundStats)) + '\n');
             } else if (typeof outputOptions.json === 'string') {
-                const JSONStats = JSON.stringify(stats.toJson(foundStats), null, 2);
+                const JSONStats = JSON.stringify(stats.toJson(foundStats));
 
                 try {
                     writeFileSync(outputOptions.json, JSONStats);


evilebottnawi (commented):
It will be unreadable

nickshevr:
> It will be unreadable

You have a lot of tools to:
- prettify json
- analyze result

You don't need to have readable output. It's much better rather than get a `Range error: Invalid string lengh`

nickshevr:
Or it should be `stream`, not `string`. On normal project (> 5k modules) command:
```
webpack --stats > stats.json
```
Is broken.

antonk52:
A potential workaround that can preserve compatibility with current behavior in terms of readability would be as @nickshevr mentioned to write the stringified stats to stdout from a stream instead of the current implementation which indeed fails for large projects.

evilebottnawi:
@nickshevr who is it `webpack --stats > stats.json` broken?

evilebottnawi:
You need to use `webpack --stats --json > stats.json`

evilebottnawi:
> A potential workaround that can preserve compatibility with current behavior in terms of readability would be as @nickshevr mentioned to write the stringified stats to stdout from a stream instead of the current implementation which indeed fails for large projects.

What?

nickshevr:
> You need to use `webpack --stats --json > stats.json`

Ofc I mean that.

> This fix nothing change, jest remove spaces from output

And dramatically reduce size of the string. See the difference:
image

How the *** u didn't recognize that?

nickshevr:
>> jest remove spaces from output
 Jest is a tool for unit testing.

You need to use it to understand: https://github.com/facebook/jest

nickshevr:
Realwold example:
```
cat stats3.json | tr -d " \t\n\r" > stats3_no_whitespace.json
```
Then:
```
ls -la | grep stats3
-rw-r--r--   1 nickshevr  LD\Domain Users  114272374 Nov 10 21:00 stats3.json
-rw-r--r--   1 nickshevr  LD\Domain Users   89148683 Nov 19 18:22 stats3_no_whitespace.json
```
30% reduction.

evilebottnawi:
hm, okay, make sense, please fix tests

evilebottnawi:
We'll just indicate `--json` only for development purpose, not for readable

nickshevr:
> We'll just indicate `--json` only for development purpose, not for readable

I thinks it's how most of people use it

>  please fix tests

kk

avdotion:
In fact, the improvement does not solve an issue with large stats reports. Everything is what you did is decreased the result, but do not prevent process from falling with `Range error` in any case.

Try this:
```javascript
const JsonStreamStringify = require('json-stream-stringify');

const jsonStream = new JsonStreamStringify(stats.toJson(foundStats));
jsonStream.once('error', () => console.log('Error while writing the result to stdout', jsonStream.stack.join('.')));
jsonStream.pipe(process.stdout);
```

[zero-deps json-stream-stringify in npm](https://www.npmjs.com/package/json-stream-stringify).

PR #2126, opened 2020-11-20
fix(serve): do not default port in webpack-dev-server v4

ylemkimon:
Bug fix and refactor

Unit tests. Integration tests will be possible once after webpack-dev-server v4 has been released.

N/A

**Summary**
With https://github.com/webpack/webpack-dev-server/pull/2845, webpack-dev-server will handle setting the default port.

Depends on the PR above.

N/A

**Other information**

### DIFF
--- a/packages/serve/src/startDevServer.ts
+++ b/packages/serve/src/startDevServer.ts
@@ -16,15 +16,21 @@ const { logger } = utils;
  * @returns {Object[]} array of resulting servers
  */
 export default function startDevServer(compiler, devServerArgs): object[] {
-    let Server;
+    let isDevServer4 = false,
+        devServerVersion,
+        Server;
     try {
+        // eslint-disable-next-line node/no-extraneous-require
+        devServerVersion = require('webpack-dev-server/package.json').version;
         // eslint-disable-next-line node/no-extraneous-require
         Server = require('webpack-dev-server/lib/Server');
     } catch (err) {
         logger.error(`You need to install 'webpack-dev-server' for running 'webpack serve'.\n${err}`);
         process.exit(2);
     }
-    const cliOptions = createConfig(devServerArgs);
+    isDevServer4 = devServerVersion.startsWith('4');
+
+    const cliOptions = createConfig(devServerArgs, isDevServer4);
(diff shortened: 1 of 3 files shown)


evilebottnawi (approved):
Ideally webpack-cli should not set default options for any options

snitin315 (approved):

packages/serve/src/startDevServer.ts
```diff
@@ -33,16 +39,21 @@ export default function startDevServer(compiler, devServerArgs): object[] {
     devServerOptions.forEach((devServerOpts): void => {
         const options = mergeOptions(cliOptions, devServerOpts);
         options.host = options.host || 'localhost';
```
evilebottnawi, line 41: @ylemkimon I think we have a problem here too, I think we used `0.0.0.0` by default here, we should use the same logic as used in Node.js `http`/`https` module

evilebottnawi: But let's solve this in other PR

evilebottnawi (approved):

packages/serve/src/startDevServer.ts
ylemkimon: Changing the default hostname would be a breaking change for webpack-cli.

evilebottnawi: Let's do it only for v4

ylemkimon: I think it's better done on the dev-server's side.

ylemkimon: Actually the default host was `localhost` in the webpack-dev-server CLI: https://github.com/webpack/webpack-dev-server/blob/4ab1f21bc85cc1695255c739160ad00dc14375f1/bin/options.js#L175

evilebottnawi: Yes, but I think it is bug and we should fix it, ideally we should be like
```
const http = require('http');

const requestListener = function (req, res) {
    res.writeHead(200);
    res.end('Hello, World!');
}

const server = http.createServer(requestListener);
server.listen(8080);

console.log(server.address())
```

`::` for `IPv6`
`0.0.0.0` for `IPv4`

PR #2141, opened 2020-11-27
fix(serve): do not default host in webpack-dev-server v4

ylemkimon:
Bug fix

Integration tests will be possible once after webpack-dev-server v4 has been released.

N/A

**Summary**
https://github.com/webpack/webpack-cli/pull/2126#discussion_r529471961

With https://github.com/webpack/webpack-dev-server/pull/2869, it'd be possible to pass `undefined` as the host to let `listen()` handle the address to bind to.

Depends on the PR above.

N/A

**Other information**

### DIFF
--- a/packages/serve/src/startDevServer.ts
+++ b/packages/serve/src/startDevServer.ts
@@ -36,9 +36,9 @@ export default function startDevServer(compiler, cliOptions): object[] {
     const usedPorts: number[] = [];
     devServerOptions.forEach((devServerOpts): void => {
         const options = mergeOptions(cliOptions, devServerOpts);
-        options.host = options.host || 'localhost';
-        // devSever v4 handles the default port itself
+        // devSever v4 handles the default host and port itself
         if (!isDevServer4) {
+            options.host = options.host || 'localhost';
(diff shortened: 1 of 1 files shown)


packages/serve/src/startDevServer.ts
```diff
@@ -36,9 +36,9 @@ export default function startDevServer(compiler, cliOptions): object[] {
     const usedPorts: number[] = [];
     devServerOptions.forEach((devServerOpts): void => {
         const options = mergeOptions(cliOptions, devServerOpts);
-        options.host = options.host || 'localhost';
-        // devSever v4 handles the default port itself
+        // devSever v4 handles the default host and port itself
```
alexander-akait, line 39: I think we can break compatibility with webpack-dev-server@3, I think we need do it by default for `else` branch

ylemkimon: `isDevServer4` is negated in the condition.

alexander-akait: I definitely need a vacation

alexander-akait (approved):

PR #2151, opened 2020-12-01
 feat: allow basic & verbose help output

### LINKED ISSUE #2140: Output from `webpack --help` is mangled
# Bug report

Terminal output from `webpack --help` looks mangled:

- Excessive newlines
- Strange column sizing / layout
- Terminal escape codes retained when piping into something that is not a TTY (e.g. `| less`; `| more` works, but presumably due to a happy accident, as terminal escapes are also retained, it just so happens that `more` doesn't break the terminal output when paging such "enriched" text)

A few screenshots showing the problem in various terminal emulators:

gnome-terminal:
gnome-terminal

rxvt-unicode:
urxvt

konsole:
konsole

piping through `less` in gnome-terminal:
gnome-terminal-less

The problems persist across terminal emulators, and regardless of shell used.

Run any of the following commands in a terminal emulator in a context where `webpack-cli` is available.

- `webpack --help`
- `webpack --help | head -n100`
- `webpack --help | more`
- `webpack --help | less`

- Terminal escapes are only inserted when output is a TTY (or, ideally, if forced via a suitable CLI option)
- Output is formatted sensibly (no excessive newlines, reasonable column widths)

**Other relevant information:**
webpack version:  5.7.0
Node.js version: v15.2.1
Operating System: Debian GNU/Linux 10 (buster)
Additional tools: urxvt, gnome-terminal, konsole, less, more, head, bash, zsh (all installed from Debian repositories)

snitin315:
feature

Yes
Yes
**Summary**

Resolve #2140 

Now `webpack --help` will show basic help output.

Screenshot at 2020-11-29 18-00-51

For advance help output - 

- `webpack --help advance`
- `webpack --help=advance`

No

**Other information**
No


rishabh3112 (commented):
Looks good 👍🏻 . One suggestion, I think `--help=verbose` is better suited for this imo.
wdyt @webpack/cli-team ?

snitin315:
`--help=verbose` sounds good to me.

alexander-akait:
@snitin315 Can you rebase?

alexander-akait:
Agree about `verbose`

alexander-akait (commented):
I think we should improve logic here, i.e. keep in flags `help: 'base'` and `help: 'verbose'` and added this to some flags, right now looks good, but I think we should exclude `amd`, `dependencies`, `node`, `externals`

snitin315:
Done

alexander-akait (commented):
Do we need update docs? And can you provide new screenshot?

snitin315:
## base
Screenshot at 2020-12-03 19-26-31

## verbose
Screenshot at 2020-12-03 19-26-57

snitin315:
I found an issue `--help=verbose` is showing only `verbose` flags I think it should display all flags 😄

snitin315:
> Do we need update docs?

Yeah, we need to update here and on webpack.js.org too.

rishabh3112:
> I found an issue --help=verbose is showing only verbose flags I think it should display all flags smile

Yup, should display all.
Also, I think we should keep only some basic flags by default and prompt users to use `--help=verbose` for all.

snitin315:
> prompt users to use --help=verbose for all. 

Prompting with the logger? - `You can use  --help=verbose to see a list of all supported arguments & commands`

alexander-akait:
@snitin315 Let's improve this
- when you write `--help` we should provide more ifnromation like `To see more flags please write "--help=verbose"` (maybe paraphrase)
- I think we should change verbose output like:
```
Options:
--option-name
Description
--option-name
Description
```

otherwise it is not readable

snitin315:
Yep Let's improve it.

snitin315:
## --help

Screenshot at 2020-12-05 19-21-22

## --help=verbose

Screenshot at 2020-12-05 19-42-32

snitin315:
## webpack --help 

Screenshot at 2020-12-05 19-21-22

## webpack --help=verbose

Screenshot at 2020-12-05 19-16-13

alexander-akait:
@snitin315 let's use remame `Available Commands` to `Commands` and `Available Options` to `Options` in both output

alexander-akait:
And change `&` to `and`

packages/webpack-cli/lib/groups/runHelp.js
```diff
@@ -90,46 +98,85 @@ const outputHelp = (args) => {
...
-
-                        return e;
-                    })
-                    .concat(negatedFlags),
-            },
-        ]);
-
-        logger.raw(output);
+
+        if (args.includes('verbose') || args.includes('--help=verbose')) {
```
alexander-akait, line 118: Very bad - `args.includes('verbose')`

packages/webpack-cli/lib/utils/cli-flags.js
```diff
@@ -238,24 +256,34 @@ const builtInFlags = [
...
 
+// single word verbose options
+const verboseFlags = ['amd', 'node', 'externals', 'dependencies'];
+
 // Extract all the flags being exported from core.
 // A list of cli flags generated by core can be found here https://github.com/webpack/webpack/blob/master/test/__snapshots__/Cli.test.js.snap
 const coreFlags = cli
     ? Object.entries(cli.getArguments()).map(([flag, meta]) => {
+          if (flag.includes('-') || verboseFlags.includes(flag)) {
+              meta.help = 'verbose';
```
alexander-akait, line 283: Need tests, up coverage

packages/webpack-cli/lib/utils/cli-flags.js
```diff
@@ -238,24 +256,34 @@ const builtInFlags = [
...
+          } else {
+              meta.help = 'base';
```
alexander-akait, line 285: Need tests, up coverage

packages/webpack-cli/lib/groups/runHelp.js
snitin315: `help` doesn't get parsed like other flags. This check here is to support `webpack --help verbose`

alexander-akait: We really should refactor this hack stuff, `commander` provide great ability to do it https://github.com/tj/commander.js?#custom-help, no need to look at `args`, it is bad, I will merge this PR, but the next is refactor, no PR to help while we don't migrate on `commander` API

snitin315: makes sense 👍🏻

packages/webpack-cli/lib/utils/cli-flags.js
```diff
@@ -238,24 +256,34 @@ const builtInFlags = [
...
 ];
 
+// single word verbose options
+const verboseFlags = ['amd', 'node', 'externals', 'dependencies'];
+
 // Extract all the flags being exported from core.
 // A list of cli flags generated by core can be found here https://github.com/webpack/webpack/blob/master/test/__snapshots__/Cli.test.js.snap
 const coreFlags = cli
     ? Object.entries(cli.getArguments()).map(([flag, meta]) => {
+          if (flag.includes('-') || verboseFlags.includes(flag)) {
```
alexander-akait, line 282: Why we need `flag.includes('-')` and this complex logic? Just create `notVerboseFlags` array and put flags to simple help to this array

snitin315: This is not needed anymore. Old leftover 😞

alexander-akait:
Don't worry about `macos` tests here, I will look how to fix it

alexander-akait (dismissed):
/cc @webpack/cli-team

alexander-akait:
@snitin315 Please rebase too

packages/webpack-cli/lib/groups/runHelp.js
```diff
@@ -90,48 +98,89 @@ const outputHelp = (args) => {
...
+                },
+                { header: 'Options', raw: true },
+            ]);
+
+            // print help-header & commands
+            logger.raw(headerAndCommands);
+            // print all options
+            for (const flag of flagsToDisplay) {
+                let flagType;
+                if (Array.isArray(flag.type)) {
```
alexander-akait, line 139: You need to mapping for type, what is will have `[Number, String]`?

packages/webpack-cli/lib/groups/runHelp.js
```diff
@@ -90,48 +98,89 @@ const outputHelp = (args) => {
...
+                    // only combo of boolean & string is allowed
+                    flagType = 'boolean, string';
+                } else if (flag.multiple) {
```
alexander-akait, line 142: Multiple can be not only string

packages/webpack-cli/lib/groups/runHelp.js
snitin315: right now only the `[Boolean, String]` combo is supported by `arg-parser`. But yes logic should be flexible and future proof.
Fixed 👍🏻

packages/webpack-cli/lib/groups/runHelp.js
alexander-akait: And multiple types can be `multiple` too, it should be not in `else if`

snitin315: fixed

snitin315: Not supported by webpack yet.

alexander-akait: It is logic problem and should be fixed even webpack doesn't provide this flags

packages/webpack-cli/lib/groups/runHelp.js
```diff
@@ -90,48 +98,93 @@ const outputHelp = (args) => {
...
+                { header: 'Options', raw: true },
+            ]);
+
+            // print help-header & commands
+            logger.raw(headerAndCommands);
+            // print all options
+            for (const flag of flagsToDisplay) {
+                let flagType;
+
+                if (flag.multiple && !Array.isArray(flag.type)) {
```
alexander-akait, line 140: Still invalid, again, types can be multiple regardless of their type

snitin315: can you give example? I have covered the following cases - 

- `multiple: true, type: [Boolean, String] (Any combo)`  => `type = boolean, string[]`
- `multiple: flase, type: [Boolean, String] (Any combo)` => `type = boolean, string`
- `multiple: true, type: String` => `type = string[]`
- `multiple: flase, type: String` => `type = string`

snitin315: /cc @alexander-akait

alexander-akait:
```
if (Array.isArray(flag.type)) {
                    const allowedTypes = flag.type.reduce((allTypes, type) => {
                        const currentType = flag.multiple ? `${type.name.toLowerCase()}[]` : type.name.toLowerCase();
                        return [...allTypes, currentType];
                    }, []);
                    flagType = allowedTypes.join(', ');
} else {
   flagType = flag.type.name.toLowerCase();

  if (flag.multiple) {
   flagType += `${flagType}[]`
  }
}
```

Pseudo code

packages/webpack-cli/lib/groups/runHelp.js
```diff
@@ -90,48 +98,95 @@ const outputHelp = (args) => {
...
+                let flagType;
+
+                if (Array.isArray(flag.type)) {
+                    const allowedTypes = flag.type.reduce((allTypes, type) => {
+                        const currentType = flag.multiple ? `${type.name.toLowerCase()}[]` : type.name.toLowerCase();
+                        return [...allTypes, currentType];
+                    }, []);
+                    flagType = allowedTypes.join(', ');
+                } else {
+                    flagType = flag.type.name.toLowerCase();
```
alexander-akait, line 147: Can be simplified: `${flag.type.name.toLowerCase()}${flag.multiple ? '[]' : ''}`

snitin315: done

alexander-akait (approved):
/cc @webpack/cli-team

alexander-akait (approved):
Let's merge and improve it better in future

PR #2157, opened 2020-12-01
refactor: error message

alexander-akait:

refactor, it is regression from big refactor, problems doesn't have helpful information about possible values to fix, it should be implemented on webpack side (I requested it)

Yes

No need

**Summary**

No

No

**Other information**

No

### DIFF
--- a/packages/webpack-cli/lib/webpack-cli.js
+++ b/packages/webpack-cli/lib/webpack-cli.js
@@ -256,8 +256,6 @@ class WebpackCLI {
                             }`,
                         );
                     });
-
-                    process.exit(2);
                 }
 
                 return options;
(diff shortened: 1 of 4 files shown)


packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -256,8 +256,6 @@ class WebpackCLI {
                             }`,
                         );
                     });
-
-                    process.exit(2);
```
alexander-akait, line 260: We will get the exit code 2 from webpack (when we will validate options)

anshumanv (approved):

snitin315 (approved):

rishabh3112 (approved):
Need to fix `--parallelism` test, rest looks good.

alexander-akait:
We need improve problems on webpack side, i.e. webpack should return helpful message

alexander-akait:
Fixed in master, we have `expected` in problems

PR #2152, opened 2020-12-01
fix: warn the user if the JSON stats file already exists

snitin315:
**What kind of change does this PR introduce?**
fix

**Did you add tests for your changes?**
Yes
**If relevant, did you update the documentation?**
No
**Summary**
 warn the user if the JSON stats file already exists.

`file 'stats.json' already exists. CLI will overwrite the existing file.`

**Does this PR introduce a breaking change?**
No

**Other information**
No

PR #2161, opened 2020-12-02
fix: continue to print stats after storing as json

snitin315:
fix

yes
no
**Summary**
continue to print stats after storing as JSON. 

### Before
Screenshot at 2020-12-02 07-47-15

### After
Screenshot at 2020-12-02 07-29-51

No

**Other information**
https://github.com/webpack/webpack-cli/pull/2152#issuecomment-736934318


alexander-akait:
I think we should not continue output stats with json flags, it is some mesleading

alexander-akait:
Fixed in master

PR #2163, opened 2020-12-02
fix: remove unnecessary group filtering in flags

anshumanv:
refactor/perf improvement

Yes

NA

**Summary**
- Removes unnecessary group filtering for core flags, we can use them directly

No

**Other information**

### DIFF
--- a/packages/webpack-cli/lib/webpack-cli.js
+++ b/packages/webpack-cli/lib/webpack-cli.js
@@ -6,7 +6,7 @@ const { writeFileSync, existsSync } = require('fs');
 const { options: coloretteOptions, yellow } = require('colorette');
 
 const logger = require('./utils/logger');
-const { cli, flags } = require('./utils/cli-flags');
+const { cli, flags, coreFlags } = require('./utils/cli-flags');
 const argParser = require('./utils/arg-parser');
 const CLIPlugin = require('./plugins/CLIPlugin');
 const promptInstallation = require('./utils/prompt-installation');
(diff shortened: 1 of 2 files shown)


packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -230,13 +230,11 @@ class WebpackCLI {
 
         if (cli) {
             const processArguments = (options) => {
-                const coreFlagMap = flags
-                    .filter((flag) => flag.group === 'core')
-                    .reduce((accumulator, flag) => {
-                        accumulator[flag.name] = flag;
+                const coreFlagMap = coreFlags.reduce((accumulator, flag) => {
```
alexander-akait, line 233: We should avoid `reduce` too

anshumanv: we can avoid it but creating a map helps so we don't loop over core flags for every arg supplied to the cli

anshumanv: we should ideally do this from the start as exporting array of flags is expensive in computation to check whether one flag is supplied or not

alexander-akait: And we don't need this `we can avoid it but creating a map helps so we don't loop over core flags for every arg supplied to the cli`, please consider merging objects using `...`

anshumanv: then we have problems in `cli.processArguments(coreFlagMap, options, coreConfig);` because then we'll be passing flags which are not supported in core which will return problems of unknown argument

alexander-akait:
Right now I do small refactor, we still need to avoid extra filtering, so I will add todo in source core and we will do this in the next iteration

anshumanv:
Yep will pick it up in next iteration 👍

PR #2170, opened 2020-12-03
fix: defer handling v5 flags to core

anshumanv:

fix

Yes

Not yet

**Summary**
- We're handing devtool flag in our own way, core already exports it and hence should be supported by core and not by cli side.

No

**Other information**


packages/webpack-cli/lib/utils/cli-flags.js
```diff
@@ -210,15 +210,6 @@ const builtInFlags = [
...
     },
-    {
-        name: 'devtool',
-        usage: '--devtool <value>',
-        type: String,
-        negative: true,
-        alias: 'd',
-        description: 'Determine source maps to use',
-        link: 'https://webpack.js.org/configuration/devtool/#devtool',
-    },
```
snitin315, line 221: will `webpack-4` still support `webpack --devtool` ?

anshumanv: seems like no then we should add it conditionally

anshumanv: added not to drop it once webpack@4 is no longer supported

alexander-akait (commented):
We should use union object here, it is reduce quality of code

alexander-akait (changes requested):
Also it should be done for all arguments from core and built-in, not only for devtool

anshumanv:
yep rest wip

anshumanv:
some problems with target - core is not merging it correctly, will take a look at it later

anshumanv:
Added a note to move it later

alexander-akait (approved):
Looks good

alexander-akait:
Need fir tests

anshumanv:
yep something strange, lemme see

packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -328,19 +315,27 @@ class WebpackCLI {
...
                 options.mode = process.env.NODE_ENV;
             }
-
-            if (args.name) {
-                options.name = args.name;
-            }
-
-            if (typeof args.stats !== 'undefined') {
-                options.stats = args.stats;
+            if (args.entry) {
```
alexander-akait, line 318: Why it here?

packages/webpack-cli/lib/utils/cli-flags.js
```diff
@@ -269,12 +275,13 @@ const coreFlags = cli
           }
 
           const inBuiltIn = builtInFlags.find((builtInFlag) => builtInFlag.name === flag);
+          const legacyMetadata = flagsForWebpack4[flag] || {};
 
           if (inBuiltIn) {
-              return { ...meta, name: flag, group: 'core', ...inBuiltIn };
+              return { ...meta, name: flag, group: 'core', ...inBuiltIn, ...legacyMetadata };
```
alexander-akait, line 281: I don't think it is right, why we need double `...`?

packages/webpack-cli/lib/webpack-cli.js
anshumanv: some bugs in cli.processArguments, it's not able to handle multiple flags, will check later

alexander-akait: We need to fix it here

packages/webpack-cli/lib/utils/cli-flags.js
anshumanv: yep redundant, removed

packages/webpack-cli/lib/utils/cli-flags.js
```diff
@@ -105,10 +105,10 @@ const builtInFlags = [
     },
     {
         name: 'merge',
-        usage: '--merge',
+        usage: '--config <first-config> --config <second-config> --merge',
         alias: 'm',
         type: Boolean,
-        description: 'Merge two or more configurations using webpack-merge e.g. -c ./webpack.config.js -c ./webpack.test.config.js --merge',
+        description: 'Merge two or more configurations using webpack-merge',
```
alexander-akait, line 111: Can we do it in separate PR?

snitin315: It's already done. need rebase I think.

anshumanv: yep need rebase, in near future

PR #2180, opened 2020-12-04
chore: better description for built-in negated flags

jamesgeorge007:
chore

Nope

Already up to date.

**Summary**
A better description for built-in negated flags.

Screenshot 2020-12-04 at 6 12 38 PM

Nope

**Other information**
N/A

### DIFF
--- a/packages/webpack-cli/lib/groups/runHelp.js
+++ b/packages/webpack-cli/lib/groups/runHelp.js
@@ -93,7 +93,9 @@ const outputHelp = (args) => {
         const negatedFlags = flags
             .filter((flag) => flag.negative)
             .reduce((allFlags, flag) => {
-                return [...allFlags, { name: `no-${flag.name}`, description: `Negates ${flag.name}`, type: Boolean }];
+                // Use available description for built-in negated flags
+                const description = flag.negatedDescription ? flag.negatedDescription : `Negates ${flag.name}`;
+                return [...allFlags, { name: `no-${flag.name}`, description, type: Boolean }];
             }, []);
         const title = bold('⬡                     ') + underline('webpack') + bold('                     ⬡');
         const desc = 'The build tool for modern web applications';
--- a/packages/webpack-cli/lib/utils/cli-flags.js
+++ b/packages/webpack-cli/lib/utils/cli-flags.js
@@ -129,6 +129,7 @@ const builtInFlags = [
         type: Boolean,
         negative: true,
         description: 'Enables Hot Module Replacement',
+        negatedDescription: 'Disables Hot Module Replacement',
         link: 'https://webpack.js.org/concepts/hot-module-replacement/',
     },
     {
@@ -182,6 +183,7 @@ const builtInFlags = [
         type: Boolean,
         negative: true,
         description: 'Enable colors on console',
+        negatedDescription: 'Disable colors on console',
     },
 
     // For webpack@4
@@ -217,6 +219,7 @@ const builtInFlags = [
         negative: true,
         alias: 'd',
         description: 'Determine source maps to use',
+        negatedDescription: 'Do not generate source maps',
         link: 'https://webpack.js.org/configuration/devtool/#devtool',
     },
     {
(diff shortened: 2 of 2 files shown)


packages/webpack-cli/lib/utils/cli-flags.js
```diff
@@ -129,6 +129,7 @@ const builtInFlags = [
         type: Boolean,
         negative: true,
         description: 'Enables Hot Module Replacement',
```
alexander-akait, line 131: `Enables/Disables Hot Module Replacement`, we are using it for any boolean flags with negative in docs

jamesgeorge007: Currently, this is how the description shows up for built-in negated flags:-

Screenshot 2020-12-04 at 7 13 24 PM

alexander-akait: oh, yep, right

alexander-akait: Anyway we should solve this for all negative flag

jamesgeorge007: Currently, this isn't possible for core flags since we aren't receiving enough information.

alexander-akait: What is not enouth?

jamesgeorge007: We already have the built-in flags as part of [cli-flags](https://github.com/webpack/webpack-cli/blob/master/packages/webpack-cli/lib/utils/cli-flags.js), and hence adding a key corresponding to the description for negated flags would do. However, this isn't the case with core flags; the description for a negated flag is to be made available.

jamesgeorge007: We already have the built-in flags as part of [cli-flags](https://github.com/webpack/webpack-cli/blob/master/packages/webpack-cli/lib/utils/cli-flags.js), and hence adding a key corresponding to the description for negated flags would do. However, this isn't the case with core flags; the description for a negated flag is to be made available.

alexander-akait: What we output for `--no-amd`?

jamesgeorge007: Negates `amd`

alexander-akait: Can you provide example? For help output

jamesgeorge007: The description for built-in flags show up as provided, while core flags go by the format `Negates ${flag.name}`.

Screenshot 2020-12-04 at 6 12 38 PM

alexander-akait (dismissed):

packages/webpack-cli/lib/utils/cli-flags.js
```diff
@@ -182,6 +183,7 @@ const builtInFlags = [
         type: Boolean,
         negative: true,
         description: 'Enable/Disable colors on console',
```
alexander-akait, line 185: I think we should change it to `Enable`

jamesgeorge007: #2188

packages/webpack-cli/lib/groups/runHelp.js
```diff
@@ -93,7 +93,9 @@ const outputHelp = (args) => {
         const negatedFlags = flags
             .filter((flag) => flag.negative)
             .reduce((allFlags, flag) => {
-                return [...allFlags, { name: `no-${flag.name}`, description: `Negates ${flag.name}`, type: Boolean }];
+                // Use available description for built-in negated flags
+                const description = flag.negatedDescription ? flag.negatedDescription : `Negates ${flag.name}`;
+                return [...allFlags, { name: `no-${flag.name}`, description, type: Boolean }];
```
snitin315, line 98: I'm afraid this will conflict with #2151 😞

rishabh3112 (approved):

PR #2189, opened 2020-12-05
fix: flaky cache test on macOS

snitin315:
tests, fix

yes
no
**Summary**
I see some flakiness in the cache test on macOS after the release of `5.10.0`.

References - 
https://github.com/webpack/webpack-cli/runs/1503646276?check_suite_focus=true
https://github.com/webpack/webpack-cli/runs/1503607410?check_suite_focus=true

no

**Other information**
no

### DIFF
--- a/test/cache/cache.test.js
+++ b/test/cache/cache.test.js
@@ -35,7 +35,7 @@ describe('cache', () => {
     });
 
     it('should work in multi compiler mode', () => {
-        rimraf.sync(path.join(__dirname, '../../node_modules/.cache/webpack/cache-test-{first,second}-development'));
+        rimraf.sync(path.join(__dirname, '../../node_modules/.cache/webpack/**'));
 
         let { exitCode, stderr, stdout } = run(__dirname, ['-c', './multi.config.js'], false);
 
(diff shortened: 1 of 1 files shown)


alexander-akait (changes requested):
Bad idea, we should search way why it happens, not break our tests

snitin315:
from CI -
Screenshot at 2020-12-05 20-30-24

I will investigate 👍🏻

alexander-akait:
Do you run test locally?

snitin315:
yeah, passing locally 😕

alexander-akait:
It means other tests generate cache too, so we got duplicate

test/cache/cache.test.js
```diff
@@ -35,7 +35,7 @@ describe('cache', () => {
     });
 
     it('should work in multi compiler mode', () => {
-        rimraf.sync(path.join(__dirname, '../../node_modules/.cache/webpack/cache-test-{first,second}-development'));
+        rimraf.sync(path.join(__dirname, '../../node_modules/.cache/webpack/**'));
```
alexander-akait, line 38: It is unsafe, don't remove full cache, it breaks other tests, in test you need delete only necessary caches

snitin315: How about defining a separate cache-directory for separate tests 😄

alexander-akait: Can you clarify?

snitin315: `node_modules/.cache/webpack` is the default directory for the cache we can overwrite with `cacheDirectory` option in config.

alexander-akait: I am already do it on tests level

alexander-akait:
Thanks for the PR, fixed in master, temporary disable check on `2`, I will investigate with the problem in the future

PR #2190, opened 2020-12-05
Stringify stats using streaming approach

lahmatiy:

This PR changes the approach for stats stringify to use stream instead of JSON.stringify() + sync write. No API or functionality is changed.

No

Not needed

**Summary**

It makes stats available for huge projects. Changes has the follow effects:
- Memory
  - Allows to avoid max string length limit of V8 engine (500MB). In other words, stats becomes available for huge projects, whose stats exceed 500MB.
  - Reduces memory consumption on serialization.
- Adds time penalty. However, this make sense for huge stats only and seems to be an acceptable, given that otherwise stats may be unavailable (due max string limit hit or out of memory).
- Adds new dependency "json-ext", which is dependency free and most effective solution for a huge JSON parse/stringify at the moment. Other solutions have troubles to handle huge JSON (see [benchmark](https://github.com/discoveryjs/json-ext/blob/master/benchmarks/README.md#stream-stringifying)).

No

**Other information**

Tested on AST Explorer (https://github.com/fkling/astexplorer)

Command (cwd website):
> npm run build -- --json stats.json

Time: 4583ms
Memory used: +1_850_027_831 (heap: +1_508_459_432, external: +341_568_399)

Time: 5881ms
Memory used: -90_171_587 (heap: -32_339_984, external: -57_831_603)

So it's a bit slower, but much effective in memory consumption.

### DIFF
--- a/packages/webpack-cli/lib/webpack-cli.js
+++ b/packages/webpack-cli/lib/webpack-cli.js
@@ -2,8 +2,9 @@ const path = require('path');
 const packageExists = require('./utils/package-exists');
 const webpack = packageExists('webpack') ? require('webpack') : undefined;
 const webpackMerge = require('webpack-merge');
-const { writeFileSync, existsSync } = require('fs');
+const { createWriteStream, existsSync } = require('fs');
 const { options: coloretteOptions, yellow } = require('colorette');
+const { stringifyStream: createJsonStringifyStream } = require('@discoveryjs/json-ext');
 
 const logger = require('./utils/logger');
 const { cli, flags } = require('./utils/cli-flags');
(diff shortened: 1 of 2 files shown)


TrySound:
I think node 12+ support is too optimistic until April.

lahmatiy:
@TrySound Didn't know that requirement. Will fix it a bit later.

anshumanv:
I've re-run the CI job, some random install failure

lahmatiy:
Fixed, now json-ext supports Node.js 10. All tests passed

alexander-akait (approved):

alexander-akait:
Big thanks

PR #2205, opened 2020-12-07
fix: correct usage of cli-flags

snitin315:
fix

NA
NA
**Summary**
correct usage of cli-flags

No

**Other information**
No


alexander-akait (approved):
/cc @webpack/cli-team

jamesgeorge007 (approved):

packages/webpack-cli/lib/utils/cli-flags.js
```diff
@@ -114,7 +114,7 @@ const builtInFlags = [
     // Complex configs
     {
         name: 'env',
-        usage: '--env',
+        usage: '--env <variable> | --env <variable> --env <variable=value>',
```
anshumanv, line 117: Ah I think we should also mention that we support nested flags here, var1.var2.var3=value, will PR

/cc @alexander-akait

PR #2206, opened 2020-12-07
fix: reduce spammy logs

alexander-akait:

bugfix

Existing

No

**Summary**

No need extra logs, please disable them by default, so I change `info` to `log`

No

**Other information**

we should output minimum information, stats is enough by default, developers don't love it

### DIFF
--- a/packages/webpack-cli/__tests__/serve/serve.test.js
+++ b/packages/webpack-cli/__tests__/serve/serve.test.js
@@ -14,8 +14,7 @@ describe('Serve', () => {
     it('should run with cli', async () => {
         const { stderr, stdout } = await runServe([], __dirname);
 
-        expect(stderr).toContain('Compilation starting...');
-        expect(stderr).toContain('Compilation finished');
+        expect(stderr).toBeFalsy();
         expect(stdout).toContain('main.js');
         expect(stdout).not.toContain('HotModuleReplacementPlugin');
     });
@@ -23,8 +22,7 @@ describe('Serve', () => {
     it('should work with flags', async () => {
         const { stderr, stdout } = await runServe(['--hot'], __dirname);
 
-        expect(stderr).toContain('Compilation starting...');
-        expect(stderr).toContain('Compilation finished');
+        expect(stderr).toBeFalsy();
         expect(stdout).toContain('main.js');
         expect(stdout).toContain('HotModuleReplacementPlugin');
     });
--- a/packages/webpack-cli/lib/plugins/CLIPlugin.js
+++ b/packages/webpack-cli/lib/plugins/CLIPlugin.js
@@ -42,10 +42,10 @@ class CLIPlugin {
 
     setupHelpfulOutput(compiler) {
         const pluginName = 'webpack-cli';
-        const getCompilationName = (compilation) => (compilation.name ? ` '${compilation.name}'` : '');
+        const getCompilationName = () => (compiler.name ? ` '${compiler.name}'` : '');
 
-        compiler.hooks.run.tap(pluginName, (compiler) => {
-            this.logger.info(`Compilation${getCompilationName(compiler)} starting...`);
+        compiler.hooks.run.tap(pluginName, () => {
+            this.logger.log(`Compilation${getCompilationName()} starting...`);
         });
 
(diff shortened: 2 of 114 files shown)


anshumanv:
lint fixed in master, we can skip rebase and merge on CI green

anshumanv (dismissed):

codymikol:
How can this be turned back on? This actually obfuscated some issues I was experiencing.

eamodio:
Yeah, I would like to know the same -- as this change breaks problem matcher support in VS Code.

@alexander-akait ?

alexander-akait:
You should not rely on logger output, it is not safe, but you can enable them using https://webpack.js.org/configuration/other-options/#infrastructurelogging (set `level` to `log`)

eamodio:
@alexander-akait do you have a suggestion on what to rely on then?

alexander-akait:
stats output

alexander-akait:
What is use case? Why you need this lines?

eamodio:
For VS Code problem matchers -- scraping task (terminal) output to report problems, etc.

See:
https://code.visualstudio.com/Docs/editor/tasks#_processing-task-output-with-problem-matchers
https://code.visualstudio.com/Docs/editor/tasks#_background-watching-tasks

alexander-akait:
In this case please use https://webpack.js.org/configuration/other-options/#infrastructurelogging

eamodio:
Will that still provide start and stop makers in watch mode?

Also the fact that these are now gone by default adds a lot of extra friction for users since they need a custom config for things to "just work"

alexander-akait:
> Will that still provide start and stop makers in watch mode?

Yes, we just reduce log level

> Also the fact that these are now gone by default adds a lot of extra friction for users since they need a custom config for things to "just work"

It is already works fine, I don't know why you use it in this form, most of developers uses `scripts` and `npm run`/`yarn`

Some developers say - please remove extra logs
Some developers say - we need these extra logs

Sorry, I can't solve this problem for everyone

codymikol:
> > Will that still provide start and stop makers in watch mode?
> 
> Yes, we just decrease log level
> 
> > Also the fact that these are now gone by default adds a lot of extra friction for users since they need a custom config for things to "just work"
> 
> It is already works fine, I don't know why you use it in this form, most of developers uses `scripts` and `npm run`/`yarn`
> 
> Some developers say - please remove extra logs
> Some developers say - we need these extra logs
> 
> Sorry, I can't solve this problem for everyone

I'm totally fine with logs off by default, I was just ignorant as to how to enable them.  Your help is much appreciated!

eamodio:
> It is already works fine, I don't know why you use it in this form, most of developers uses scripts and npm run/yarn

I thought I was pretty clear as to what we are/were using it for — allowing VS Code to provide a richer build experience when using built tasks. These tasks can just be (and often are) just `package.json` `scripts` run by npm or yarn, or they can have some extra sugar on top to provide more functionality. For example, many devs configure a default build task, and then when they press `F5` to start debugging, the debugger waits until the build task completes successfully before launching. Also errors output from typescript/eslint/etc can be scraped and aggregated into the VS Code _Problems_ view:

image

Also to be clear, I'm only arguing for a start/end marker when in `watch`, and IMO (even outside the problem matcher issue) not having an end marker at least, doesn't provide a user a nice clear indicator that the watch task has actually completed (especially when you have multiple webpack configs exported)

alexander-akait:
If you need some extra logs - PR welcome, we don't remove previous logs, just decrease level

eamodio:
Also, while I can certainly appreciate the _you can't please everyone_ sentiment (and believe be I get it), I am raising this for a community/ecosystem. I am a core dev on VS Code and the maintainer of the webpack typescript/eslint problem matchers.

Yes, I understand that the logs still exist, but we would still like to have the ability to integrate with webpack without a user having to manually alter their configuration -- which was the case before these changes.

alexander-akait:
@eamodio Can you open a new issue and describe how you run webpack and what you need, I think we can find solution

eamodio:
@alexander-akait Sorry for the delay, opened it here: https://github.com/webpack/webpack-cli/issues/2374

PR #2214, opened 2020-12-08
fix: help header alignment

snitin315:
fix

nope
nope
**Summary**

help header alignment was a bit off. 
## Before
Screenshot at 2020-12-08 16-54-32

## After
Screenshot at 2020-12-08 17-01-45

**Other information**

### DIFF
--- a/packages/webpack-cli/lib/groups/runHelp.js
+++ b/packages/webpack-cli/lib/groups/runHelp.js
@@ -110,12 +110,12 @@ const outputHelp = (args) => {
                 }, []);
         };
 
-        const title = bold('⬡                     ') + underline('webpack') + bold('                     ⬡');
+        const title = bold('⬡                ') + underline('webpack') + bold('                ⬡');
         const desc = 'The build tool for modern web applications';
-        const websitelink = '         ' + underline('https://webpack.js.org');
+        const websitelink = underline('https://webpack.js.org');
         const usage = bold('Usage') + ': ' + '`' + green('webpack [...options] | <command>') + '`';
         const examples = bold('Example') + ': ' + '`' + green('webpack help --flag | <command>') + '`';
-        const hh = `          ${title}\n\n${websitelink}\n\n${desc}\n\n${usage}\n${examples}`;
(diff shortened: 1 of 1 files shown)


alexander-akait:
Now it looks on small console? less 80 characters?

packages/webpack-cli/lib/groups/runHelp.js
```diff
@@ -112,10 +112,10 @@ const outputHelp = (args) => {
 
         const title = bold('⬡                     ') + underline('webpack') + bold('                     ⬡');
```
rishabh3112, line 113: Convert to tabs as well
```suggestion
        const title = bold('⬡\t\t\t\t\t') + underline('webpack') + bold('\t\t\t\t\t⬡');
```

snitin315:
> Now it looks on small console? less 80 characters?

@alexander-akait can you clarify?

alexander-akait:
open terminal do very small width on opened window and run `webpack --help`, it should be readable

snitin315:
> open terminal do very small width on opened window and run `webpack --help`, it should be readable

Hmm, then we should align everything to `left`. including `docs link` & `⬡                       webpack                         ⬡`

alexander-akait:
Yes,or we can use https://github.com/Automattic/cli-table, but need investigate

packages/webpack-cli/lib/groups/runHelp.js
snitin315: some issues with using `tabs` here. Don't know why but indentation is not equal on both sides if using tabs.

snitin315:
Let's keep it simple and align text to left. I'm afraid using  https://github.com/Automattic/cli-table for help ouput will require a lot of refactor.

snitin315:
Screenshot at 2020-12-08 18-09-50

alexander-akait:
Because I say again, we need refactor, no more merging before refactoring. we should migrate on `help` callback (commander API)

alexander-akait:
Help was refactored, so we don't need this anymore, commander do it automatically, but without colors, we will improves colors in future, it is not high priority, also our help was broken in many cases

alexander-akait:
Anyway thanks

PR #2217, opened 2020-12-08
fix: handle stdin end

rishabh3112:
Ends main CLI process when stdin ends.

Yes

No

**Summary**
Attempt at fixing stdin issue

No

**Other information**

### DIFF
--- a/packages/webpack-cli/bin/cli.js
+++ b/packages/webpack-cli/bin/cli.js
@@ -23,11 +23,6 @@ process.stdin.on('end', () => {
     process.exit(0);
 });
 
-// Exit process when stream closes
-process.stdin.on('close', () => {
-    process.exit(0);
-});
-
 process.stdin.resume();
 
 const [, , ...rawArgs] = process.argv;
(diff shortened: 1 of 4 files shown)


packages/webpack-cli/bin/cli.js
```diff
@@ -18,6 +18,18 @@ if (importLocal(__filename)) {
 
 process.title = 'webpack';
 
+// Exit process when no data is to be consumed
+process.stdin.on('end', () => {
+    process.exit(0);
+});
+
+// Exit process when stream closes
+process.stdin.on('close', () => {
```
alexander-akait, line 27: Why on `close`? I think `end` is enough

rishabh3112: Yup, changing.

mveytsman:
Hello!

I think this issue is blocking upgrading webpack in the Phoenix framework (see https://github.com/phoenixframework/phoenix/pull/4054). Is there a timeline on getting this PR in? 

Also, just curious, it looks like this will supersede `--watch-options-stdin` and webpack will always exit when stdin is closed. Is that correct?

alexander-akait:
@mveytsman I think we will fix it today/tomorrow and do release, we should respect `--watch-options-stdin` option, you are right

Issue #1918, opened 2020-10-11
[DEP_WEBPACK_WATCH_WITHOUT_CALLBACK] error when using `--watch` flag


sanex3339:
**Describe the bug**

PR with the configuration
https://github.com/javascript-obfuscator/javascript-obfuscator/pull/776

After upgrade from webpack@4 to webpack@5 and webpack-cli@3 to webpack-cli@4.0.0 I got the following error:
```
[DEP_WEBPACK_WATCH_WITHOUT_CALLBACK] DeprecationWarning: A 'callback' argument need to be provided to the 'webpack(options, callback)' function when the 'watch' option is set. There is no way to handle the 'watch' option without a callback.
```

This error happens only when using `--watch` flag:
```
webpack --config ./webpack/webpack.node.config.js --mode development --watch
```

Full error trace:
```
(node:62890) [DEP_WEBPACK_WATCH_WITHOUT_CALLBACK] DeprecationWarning: A 'callback' argument need to be provided to the 'webpack(options, callback)' function when the 'watch' option is set. There is no way to handle the 'watch' option without a callback.
    at webpack (/Users/sanex/WebstormProjects/javascript-obfuscator/node_modules/webpack/lib/webpack.js:145:5)
    at f (/Users/sanex/WebstormProjects/javascript-obfuscator/node_modules/webpack/lib/index.js:35:15)
    at Compiler.createCompiler (/Users/sanex/WebstormProjects/javascript-obfuscator/node_modules/webpack-cli/lib/utils/Compiler.js:129:35)
    at WebpackCLI.run (/Users/sanex/WebstormProjects/javascript-obfuscator/node_modules/webpack-cli/lib/webpack-cli.js:220:32)
    at async runCLI (/Users/sanex/WebstormProjects/javascript-obfuscator/node_modules/webpack-cli/lib/bootstrap.js:74:24)
```

**Expected behavior**

No error.

evilebottnawi:
@webpack/cli-team high priority

photonstorm:
I get the same error, but I think it's worth mentioning that the watch mode does actually work. Yes, it throws that warning at the start, but it still bundles and still watches for changes correctly.

raduchiriac:
Indeed, webpack v5 does not watch anymore if I remove `--watch` to go with `webpack --config webpack.config.js`
I use this:
```js
{
  mode: "development",
  watch: true,
  devtool: "inline-cheap-source-map",
  watchOptions: {
[...]

PR #2268, opened 2020-12-28
Using bundleCommand instead of createCompiler in webpack-cli/serve

superhit0:

bugfix

Need help as this is my first time contributing to webpack

N/A

**Summary**
- The serve command uses [webpack-cli](https://github.com/webpack/webpack-cli/blob/master/packages/serve/src/index.ts#L83) but it creates a compiler without options which gives a deprecation warning on webpack@5, this one uses changes the use to [bundleCommand](https://github.com/webpack/webpack-cli/blob/master/packages/webpack-cli/lib/webpack-cli.js#L1124), which wraps the callabck inside it

This results in this error - [Issue#1918](https://github.com/webpack/webpack-cli/issues/1918) when we run
- `npx webpack-dev-server --watch`

No

**Other information**
@alexander-akait

### DIFF
--- a/packages/serve/src/index.ts
+++ b/packages/serve/src/index.ts
@@ -80,7 +80,7 @@ class ServeCommand {
 
                 webpackOptions.env = { WEBPACK_SERVE: true, ...options.env };
 
-                const compiler = await cli.createCompiler(webpackOptions);
+                const compiler = await cli.bundleCommand(webpackOptions);
 
                 if (!compiler) {
                     return;


snitin315:
> npx webpack-dev-server --watch

this is not the right way anymore. `webpack-cli serve` should be used to invoke webpack-dev-server.

alexander-akait:
Remove `--watch` from options, expected

alexander-akait:
It is invalid fix and break serve command

PR #2271, opened 2020-12-28
tests: publicPath

### LINKED ISSUE #2242: outputPath is not respected with 4.2.0
**Describe the bug**

We can server our repostiory against our production build with this config. I did cut some (hopefully useless) information, to make it easier to get.

```js
const devSerer = {
      historyApiFallback: true,
      publicPath: '/',
      disableHostCheck: true,
      host: 'localhost.some-domain.com',
      port: 443,
      inline: true,
      hot: true,
      noInfo: false,
      stats: 'minimal',
      proxy: [
        {
          context: ['/api/',],
          target: 'https://www.some-domain.com'',
          changeOrigin: true,
        },
      ],
    }

const output = {
        filename: '[name].[hash].js',
        chunkFilename: '[name].[hash].chunk.js',
        path: path.join(__dirname, 'dist'),
        publicPath: '/app/',
      },
```

Using HtmlWebpackPlugin with some pretty boring index.ejs file

Before 4.2.0 everything work as expected:

Scripts were loaded from the `/app`, index was served from /

Now the html contains

```html
<script src="/app/runtime.4f07461f24ae6827cfbb.js"></script>
```

This returns 404, while accessing just `/runtime.4f07461f24ae6827cfbb.js` works.
So I assume either index is served from the wrong folder, or assets are not put in the correct folder.

Using `publicPath: '/'` does work, but would also effect production builds, which is not desired.

**To Reproduce**

~~

**Expected behavior**

It should work as before.

**Screenshots**

**Additional context**

Webpack Version is 4.44.2

alexander-akait:

tests

Yes

No need

**Summary**

fixes #2242

No

**Other information**

WIP, need set `devServer.publicPath` when publicPath is passed

### DIFF
--- a/packages/serve/__tests__/__snapshots__/startDevServer.test.ts.snap
+++ b/packages/serve/__tests__/__snapshots__/startDevServer.test.ts.snap
@@ -4,6 +4,7 @@ exports[`startDevServer should set default port and host if not provided 1`] = `
 Object {
   "host": "localhost",
   "port": 8080,
+  "publicPath": "/",
 }
 `;
 
@@ -22,6 +23,7 @@ Object {
   "hot": true,
   "port": 9000,
   "progress": true,
+  "publicPath": "/",
 }
 `;
 
@@ -40,6 +42,7 @@ Object {
   "hot": true,
   "port": 9000,
   "progress": true,
+  "publicPath": "/",
 }
 `;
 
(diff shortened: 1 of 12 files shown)


anshumanv (dismissed):

snitin315 (approved):

alexander-akait:
Found bugs with multi compiler mode, WIP

PR #2280, opened 2020-12-31
refactor: code

alexander-akait:

refactoring

Yes

No

**Summary**

Refactor code + resolve TODO + tests

No

**Other information**

More powerful option creator


DelfikPro:
Absolutely loved the [new parameter logic](https://github.com/webpack/webpack-cli/commit/10fb66ecd46f6d1c83122cc039a274e737834610#diff-b808549c190b87885089bbe99c3e8b8a57019aaee02d245032b089608b0fe901) of `runCLI(...)` 

Now webpack-cli wrappers that rely on `runCLI` need to be updated to pass two additional arguments:
```js
runCLI(['--config', webpackConfigPath, ...webpackArgs])
```
`--config ...` part is now ignored, causing not just an immediate crash, but rather weird and glitchy behavior. Took me two hours to investigate and fix.

Solved using:
```js
runCLI(['', '', '--config', webpackConfigPath, ...webpackArgs])
```

alexander-akait:
@DelfikPro If you need we can add third argument `runCLI(['--config', webpackConfigPath, ...webpackArgs], { from: 'user' })` (and it will work), you need two empty `''` because `commader` wait `node ./path/to/script.js --arg value --arg value`

alexander-akait:
Anyway you should not use `bootstrap` file, it is internal, please use `CLI.js` class (look at source code), it is no documented, because it is not public right now, but I think it is safe to use

PR #2325, opened 2021-01-08
tests: add configtest case for non-existent config

anshumanv:

test

yes

NA

**Summary**
- Adds a case for non-existent config for configtest command
- 
No

**Other information**

### DIFF
--- a/test/configtest/configtest.test.js
+++ b/test/configtest/configtest.test.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const { run } = require('../utils/test-utils');
+const path = require('path');
 
 describe('basic info usage', () => {
     it('should validate webpack config successfully', () => {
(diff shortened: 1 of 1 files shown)


jamesgeorge007 (commented):
Already present.

https://github.com/webpack/webpack-cli/blob/50bbe56c0ae9d72301c4ac51fdc2b04df7b66451/test/configtest/configtest.test.js#L40-L46

anshumanv:
> Already present.

Ah yes missed it, nvm improved existing test to assert on filename

test/configtest/configtest.test.js
```diff
@@ -41,7 +42,7 @@ describe('basic info usage', () => {
         const { exitCode, stderr, stdout } = run(__dirname, ['configtest', './a.js'], false);
 
         expect(exitCode).toBe(2);
-        expect(stderr).toContain(`The specified config file doesn't exist`);
+        expect(stderr).toContain(`The specified config file doesn't exist in '${path.join(__dirname, './a.js')}`);
```
rishabh3112, line 45: This change doesn't add any value to test IMO.
The aim for the test is to look for error when configuration doesn't exists, which previous one was also doing. Unnecessary regressions must be avoided as they lead to unstability.

rishabh3112 (changes requested):
Left a comment.

anshumanv:
Yes this seems unnecessary, lets close

PR #2326, opened 2021-01-08
fix: remove legacy tooling

anshumanv:

fix

yes

NA

**Summary**
- Removes babel which is legacy tooling for us since it was kept for node@6 which we have deprecated.

no

**Other information**


alexander-akait (commented):
So we don't use babel for jest?

anshumanv:
Nope not anymore, it was to support async await in Node 6

alexander-akait (approved):

PR #2338, opened 2021-01-09
fix: show exact package name while prompting for installation

jamesgeorge007:
bugfix

Nope

N/A

**Summary**

Currently, the package name that shows up while prompting for installation is the same as the argument supplied as part of `process.argv`. However, this would be misleading if the user supplies the alias or the `loader`/`plugin` command (`generators`).

https://github.com/webpack/webpack-cli/blob/767fe6b4b4d4980b5b5cf8fd5dafb0aa84465625/packages/webpack-cli/lib/webpack-cli.js#L359-L361

```bash
$ webpack loader
```

#### Before
before

#### After
Screenshot 2021-01-10 at 2 29 10 PM

Nope

**Other information**
N/A

### DIFF
--- a/packages/webpack-cli/lib/webpack-cli.js
+++ b/packages/webpack-cli/lib/webpack-cli.js
@@ -357,7 +357,7 @@ class WebpackCLI {
 
                     try {
                         pkg = await promptInstallation(pkg, () => {
-                            logger.error(`For using this command you need to install: '${green(commandName)}' package`);
+                            logger.error(`For using this command you need to install: '${green(pkg)}' package`);
                         });
                     } catch (error) {
                         logger.error(`Action Interrupted, use '${cyan('webpack-cli help')}' to see possible commands`);


packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -357,7 +357,8 @@ class WebpackCLI {
 
                     try {
                         pkg = await promptInstallation(pkg, () => {
-                            logger.error(`For using this command you need to install: '${green(commandName)}' package`);
+                            const pkgName = pkg.slice(pkg.indexOf('/') + 1);
```
rishabh3112, line 360:
```suggestion
                            const pkgName = builtInExternalCommandInfo.name;
```

rishabh3112 (commented):
Left a suggestion.

packages/webpack-cli/lib/webpack-cli.js
jamesgeorge007: The package associated with `loader`/`plugin` commands is `generators`, which won't be suggested accordingly in this case.

rishabh3112: Yup, you are right. Then just specify the full package name `@webpack-cli/generators` instead of `generators` which is current output.
```suggestion
```

packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -357,7 +357,8 @@ class WebpackCLI {
...
+                            logger.error(`For using this command you need to install: '${green(pkgName)}' package`);
```
rishabh3112, line 361:
```suggestion
                            logger.error(`For using this command you need to install: '${green(pkg)}' package`);
```

jamesgeorge007: In fact, the subsequent prompt displays the package name along with the `@webpack-cli` prefix.

Screenshot 2021-01-10 at 12 24 44 PM

rishabh3112 (approved):

snitin315 (approved):

PR #2357, opened 2021-01-15
feat: added the `watch` command

alexander-akait:

feature

Yes

Yes

**Summary**

Many popular tools have the `watch` command, let's do the same and we will have good built-in command:
- webpack configtest
- webpack build
- webpack watch
- webpack serve
- webpack info
- webpack help
- webpack version
- webpack init (we need focus on it)

No

**Other information**

After this I will focus on `webpack index.js`, after refactor we can return this again, but with some limitations, I will describe them in the PR


packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -1373,6 +1396,7 @@ class WebpackCLI {
             }
 
             if (options.json) {
+                const { stringifyStream: createJsonStringifyStream } = require('@discoveryjs/json-ext');
```
alexander-akait, line 1399: Reduce time to initial run

test/watch/basic/basic.test.js
```diff
@@ -0,0 +1,169 @@
...
+        proc.stderr.on('data', (chunk) => {
+            const data = stripAnsi(chunk.toString());
+
+            hasWarning = true;
+
+            expect(data).toContain("No need to use the '--watch, -w' option together with the 'watch' command, it does not make sense");
+        });
+    });
+
+    it('should recompile upon file change using the `command` option and the `--no-wwatch` option and log warning', (done) => {
```
jamesgeorge007, line 128:
```suggestion
    it('should recompile upon file change using the `command` option and the `--no-watch` option and log warning', (done) => {
```

Issue #2370, opened 2021-01-21
[DEP_WEBPACK_WATCH_WITHOUT_CALLBACK] error when using webpack serve with --watch flag


benediktarnold:
**Describe the bug**

When running `webpack serve` with `--watch` the deprecation warning `DEP_WEBPACK_WATCH_WITHOUT_CALLBACK` shows up. Initially I thought it's just the same as #1918 but it's a different code path.

**What is the current behavior?**

When running node with `--trace-deprecation` this is the full deprecation stack:

    10% building(node:54129) [DEP_WEBPACK_WATCH_WITHOUT_CALLBACK] DeprecationWarning: A 'callback' argument need to be provided to the 'webpack(options, callback)' function when the 'watch' option is set. There is no way to handle the 'watch' option without a callback.
        at webpack (/Users/ben/dev/playground/webpack-deprecation-warning/node_modules/webpack/lib/webpack.js:143:5)
        at f (/Users/ben/dev/playground/webpack-deprecation-warning/node_modules/webpack/lib/index.js:37:15)
        at WebpackCLI.createCompiler (/Users/ben/dev/playground/webpack-deprecation-warning/node_modules/webpack-cli/lib/webpack-cli.js:1337:24)
        at async Command.<anonymous> (/Users/ben/dev/playground/webpack-deprecation-warning/node_modules/@webpack-cli/serve/lib/index.js:65:30)
        at async Promise.all (index 1)
        at async Command.<anonymous> (/Users/ben/dev/playground/webpack-deprecation-warning/node_modules/webpack-cli/lib/webpack-cli.js:824:13)

**To Reproduce**

1. checkout https://github.com/benediktarnold/webpack-deprecation-warning
2. run `npm install && npm start`

**Expected behavior**

No deprecation warning

**Please paste the results of `webpack-cli info` here, and mention other relevant information**

    System:
      OS: macOS 11.1
      CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
      Memory: 74.99 MB / 32.00 GB
    Binaries:
      Node: 15.5.0 - /usr/local/bin/node
      Yarn: 1.22.10 - /usr/local/bin/yarn
      npm: 7.3.0 - /usr/local/bin/npm
    Browsers:
      Chrome: 87.0.4280.141
      Firefox: 84.0.1
      Safari: 14.0.2
    Packages:
      webpack: ^5.16.0 => 5.16.0
[...]

PR #2375, opened 2021-01-25
fix: show a warning if serve and `--watch` are used together

snitin315:
fix

yes
no
**Summary**
fixes https://github.com/webpack/webpack-cli/issues/2370
show a warning if `serve` and `--watch` are used together

No

**Other information**
I will work on removing `--watch` & `--no-watch` options from `serve` help output in next PR.

### DIFF
--- a/packages/serve/src/index.ts
+++ b/packages/serve/src/index.ts
@@ -24,7 +24,7 @@ class ServeCommand {
                     process.exit(2);
                 }
 
-                const builtInOptions = cli.getBuiltInOptions();
+                const builtInOptions = cli.getBuiltInOptions().filter((option) => option.name !== 'watch');
 
                 return [...builtInOptions, ...devServerFlags];
             },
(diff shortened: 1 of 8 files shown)


packages/serve/src/index.ts
```diff
@@ -54,6 +54,11 @@ class ServeCommand {
                         kebabedOption !== 'hot' && builtInOptions.find((builtInOption) => builtInOption.name === kebabedOption);
 
                     if (isBuiltInOption) {
+                        // warn user about using '--watch' with serve
+                        if (optionName === 'watch') {
+                            logger.warn("Do not use '--watch' with 'serve'. It has no effect.");
+                        }
```
alexander-akait, line 60: It should be inside `serve` command, because only `serve`command know what it is unnecessary

alexander-akait:
Ideally we should:
- remove `--watch` and `--no-watch` for serve, so developers will not use it
- show warning when configuration has `watch: true` and set it to false to avoid `Deprecation`

I think it is the best DX

PR #2381, opened 2021-01-28
fix: support esm format for configurations

### LINKED ISSUE #1622: [Enhancement]: Error [ERR_REQUIRE_ESM]: Must use import to load ES Module
system: macOS
node version: 14.2
webpack@4.43.0
webpack-cli@3.3.11

So, i have config file "webpack.config.js"
`
export default  {
    entry: './i.js',
    output: {
        filename: 'bundle.js'
    }
}
`

in package.json i have "type": "module"

in i.js i have code
```
import * as path from 'path';
console.log(path);
```

so i use cmd
```webpack --config webpack.config.js```

and got error:
```
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /Users/x8core/Projects/anyx-cli/webpack.config.js
require() of ES modules is not supported.
```

In documentation i see. "When using webpack to bundle your application, you can pick from a variety of module syntax styles including ES6"

Why i am getting such error? Am i need to use babel?

alexander-akait:

bugfix

yes

Yep

**Summary**

fixes #1622

No

**Other information**

No


alexander-akait:
Something has global side effects, it breaks our tests...

alexander-akait:
I hate our tests for `init` and other generators, it is unstable on 100%, we need fully rewrite them

alexander-akait:
Found the problem, somebody install `@babel/core` in out tests so, recoil try to use it, also we need test for this I think

alexander-akait:
I hate `nyc`, it wrap `require` and can't provide option to disable it for specific lines, so it return invalid core and I can't find how we can test it

PR #2392, opened 2021-02-01
test: default loader name and plugin name

rishabh3112:
feat

Not needed, already present

No

**Summary**
Enforced use of process.cwd() as default `output-path`

No

**Other information**
Part of Roadmap

### DIFF
--- a/packages/generators/src/index.ts
+++ b/packages/generators/src/index.ts
@@ -18,9 +18,6 @@ class GeneratorsCommand {
             },
             [],
             async (outputPath) => {
-                if (!outputPath) {
-                    outputPath = process.cwd();
-                }
                 const env = yeoman.createEnv([], { cwd: outputPath });
                 const generatorName = 'webpack-loader-generator';
 
@@ -42,9 +39,6 @@ class GeneratorsCommand {
             },
             [],
             async (outputPath) => {
-                if (!outputPath) {
-                    outputPath = process.cwd();
-                }
                 const env = yeoman.createEnv([], { cwd: outputPath });
                 const generatorName = 'webpack-plugin-generator';
 
--- a/packages/generators/src/index.ts
+++ b/packages/generators/src/index.ts
@@ -18,6 +18,9 @@ class GeneratorsCommand {
             },
             [],
             async (outputPath) => {
+                if (!outputPath) {
+                    outputPath = process.cwd();
+                }
                 const env = yeoman.createEnv([], { cwd: outputPath });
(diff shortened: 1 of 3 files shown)


jamesgeorge007 (commented):
`yeoman-environment` handles it already.
https://github.com/yeoman/environment/blob/main/lib/environment.js#L223

alexander-akait:
@jamesgeorge007 Good catch, maybe we just need add test(s)

rishabh3112:
> Good catch, maybe we just need add test(s)

Tests are already present. Why this was included in roadmap?

alexander-akait:
Can you provide link on tests? Maybe I something missed

rishabh3112:
> Can you provide link on tests? Maybe I something missed

`output-path` is not provided in all the tests below:

https://github.com/webpack/webpack-cli/blob/master/test/loader/loader.test.js#L22
https://github.com/webpack/webpack-cli/blob/master/test/loader/loader.test.js#L29

https://github.com/webpack/webpack-cli/blob/master/test/plugin/plugin.test.js#L22
https://github.com/webpack/webpack-cli/blob/master/test/plugin/plugin.test.js#L29

alexander-akait:
No tests, https://github.com/webpack/webpack-cli/blob/master/test/loader/loader.test.js#L22 is not enough for testing, it is stupid test on questions, it should be fully test with checking on files https://github.com/webpack/webpack-cli/blob/master/test/loader/loader.test.js#L43

rishabh3112:
Line 29 has it.

alexander-akait:
Line 29 pass `loaderName` - `let { stdout } = await runPromptWithAnswers(__dirname, ['loader'], [`${loaderName}${ENTER}`]);`, we should have test without passing loader name (full test)

rishabh3112:
Loader name is not being passed as `output-path`, it is an answer to an inquirer question asked by generator.

Nevertheless will add test for default loader name as well .

packages/generators/src/index.ts
```diff
@@ -18,6 +18,9 @@ class GeneratorsCommand {
             },
             [],
             async (outputPath) => {
```
anshumanv, line 20: can just do default param?

```suggestion
            async (outputPath = process.cwd()) => {
```

rishabh3112: Yup we can but I personally don't like calling functions as default parameter 😄, looks hacky to me.

alexander-akait:
Thanks

PR #2404, opened 2021-02-03
chore: disable colors on CI by default

snitin315:
chore, CI related

Already present, updated
No
**Summary**

 disable colors on CI by default.
 

No

**Other information**
No


alexander-akait:
Good idea, now we need found all tests with `--color` and use `FORCE_COLOR`, for `--no-color` we should delete `NO_COLOR` from `process.env`

test/stats/watch/stats-and-watch.test.js
```diff
@@ -3,22 +3,26 @@
 const { runWatch } = require('../../utils/test-utils');
 
 describe('stats and watch', () => {
+    beforeAll(() => {
+        process.env.FORCE_COLOR = true;
```
alexander-akait, line 7: It will not work, because we inside `execa`, we should pass it to options, i.e. `await runWatch(__dirname, ['-c', './webpack.config.js', '--color'], { env: { FORCE_COLOR: true } });`

snitin315: resolved

alexander-akait (commented):
Ideally we need remove `stripAnsi` in many places, because we don't need it anymore

snitin315:
Will remove 👍🏻

snitin315:
Will fix the remaining tests tommorow.

alexander-akait:
I will finish tests :+1:

PR #2409, opened 2021-02-04
docs: remove init generator questions

jamesgeorge007:
Docs

N/A

Yup

**Summary**
Remove init generator questions as part of the docs.

Nope

**Other information**
N/A


rishabh3112 (changes requested):
We need documentation for questions. Can I get the reason why this is proposed?

alexander-akait:
@rishabh3112 I think we don't need, nobody reads them and the each change in code require change questions, it is unnecessary

alexander-akait:
`init` should be simple, and should not require reading something to start

rishabh3112:
> @rishabh3112 I think we don't need, nobody reads them and the each change in code require change questions, it is unnecessary

In my opinion we need this, Many people don't like to work with tools which hides things in black box. And when main function of the tool is to generate webpack configuration, people should know how it is being created.

> the each change in code require change questions

This is true for any change throughout the project + we shouldn't need to change questions in ideal scenario.

alexander-akait:
> In my opinion we need this, Many people don't like to work with tools which hides things in black box. And when main function of the tool is to generate webpack configuration, people should know how it is being created.

It means we should improve/clarify our questions, I read don't want to read docs when using `init`, it require additional knowledge, it may seem obvious to you, but not to those who start working, this is another reason why I want to simplify the logic here

> This is true for any change throughout the project + we shouldn't need to change questions in ideal scenario.

In ideal scenario we should not have questions, for example, our questions is `Do you use other bundlers?`. What is bundler? Which other? Why we ask this? How does this affect the configuration generation logic?

rishabh3112:
> for example, our question is Do you use other bundlers?. What is bundler? Which other? Why we ask this? How does this affect the configuration generation logic?

That is why we need documentation to be precise. (btw we dont ask this particular question)

> In ideal scenario we should not have questions.

Yup I agree with this. For that we can implement some opinionated templates for common usecases (For eg. one for react similar to ejected CRA),  I have earlier suggested this to you few months back.

alexander-akait:
> That is why we need documentation to be precise. (btw we dont ask this particular question)

Run `npx webpack init`

We ask:

>  Will your application have multiple bundles? (y/N) 

> For that we can implement some opinionated templates for common usecases (For eg. one for react similar to ejected CRA), I have earlier suggested this to you few months back.

I am not against that, but firstly we should simplify our code, to be honestly I myself do not understand what is happening in many places

rishabh3112:
> > That is why we need documentation to be precise. (btw we dont ask this particular question)
> 
> Run `npx webpack init`
> 
> We ask:
> 
> > Will your application have multiple bundles? (y/N)
> 

Yup I know, you originally said the question is `Do you use other bundlers?` both are different a lot as you may see. Nevertheless, the point remains the same we do need documentation for questions as many users would not be able to understand the questions as they are asked as well as how they are effecting the configurartion + we can't ask extremely verbose questions in terminal.

> > For that we can implement some opinionated templates for common usecases (For eg. one for react similar to ejected CRA), I have earlier suggested this to you few months back.
> 
> I am not against that, but firstly we should simplify our code, to be honestly I myself do not understand what is happening in many places

Yes + it is broken for custom scaffolds as well.

alexander-akait:
> Yes + it is broken for custom scaffolds as well.

So I think we need new major release for `init`:
- remove `scaffolds` features, I really think we don't need it, very few people use it but it requires a lot of work from us + it is broken (I try to use other old versions and it is broken too)
- avoid using `jscodeshift` (and other packages) and use only `yeoman-generator` and `yeoman-environment`

/cc @webpack/cli-team

rishabh3112:
On it

> > Yes + it is broken for custom scaffolds as well.
> 
> So I think we need the new major release for `init`:
> 
>     * remove `scaffolds` features, I really think we don't need it, very few people use it but it requires a lot of work from us + it is broken (I try to use other old versions and it is broken too)
> 
>     * avoid using `jscodeshift` (and other packages) and use only `yeoman-generator` and `yeoman-environment`, just generate files install deps and update `package.json`
> 
> 
> /cc @webpack/cli-team

On it.

rishabh3112:
init is underwork at #2429 with many more improvements, I think it is safe to close this one.

PR #2420, opened 2021-02-06
test: add snapshot tests for help

anshumanv:
tests

yes

NA

**Summary**
- Adds snapshot tests for help to make it easier to see the diff when something in the output changes

no

**Other information**
- I've only added it in places where it makes sense, any other place in help we need to add this?


alexander-akait (commented):
We need refactor tests more - i.e. remove `expect().toContain()`, because it is unnecessary + we need move all tests on snapshots where we use `--help` or `help`

anshumanv:
yep lets do it

test/help/help.test.js
```diff
@@ -37,6 +37,7 @@ describe('help', () => {
         expect(pureStdout).toContain("To see list of all supported commands and options run 'webpack --help=verbose'.");
         expect(pureStdout).toContain('CLI documentation: https://webpack.js.org/api/cli/.');
         expect(pureStdout).toContain('Made with ♥ by the webpack team.');
+        expect(stdout).toMatchSnapshot();
```
snitin315, line 40: This is actually a bit tricky because `weback-4` and `webpack-5` have different snapshots. for ex `--target <value>` and `--target <value...>`.

anshumanv: yeah need to come up with different strategy

anshumanv:
One way is to generate different snapshots for both webpack versions but it's very bad, @alexander-akait do you have any ideas?

alexander-akait:
Yep, we need improve snapshot serialiazer

PR #2429, opened 2021-02-10
feat: simplify init

rishabh3112:
Feature

WIP

Will do after implementation

**Summary**
Simplify the init feature to just scaffold configuration, without ast transformations.

Yes

**Other information**
WIP


packages/generators/src/init-generator.ts
```diff
@@ -19,216 +19,45 @@ const { logger, getPackageManager } = utils;
...
-        if (typeof entryOption === 'string') {
-            // single entry apply when default is not used
-            if (entryOption.length > 0 && entryOption !== "'./src/index.js'") {
-                this.configuration.config.webpackOptions.entry = entryOption;
-            }
-        } else if (typeof entryOption === 'object') {
-            // multiple entries
-            this.configuration.config.webpackOptions.entry = entryOption;
+        if (!existsSync(this.resolvedGenerationPath)) {
+            logger.log(`${logSymbols.info}${blue(' INFO ')} supplied generation path doesn't exists, required folders willbe created`);
```
jamesgeorge007, line 40:
```suggestion
            logger.log(`${logSymbols.info}${blue(' INFO ')} supplied generation path doesn't exist, required folders will be created`);
```

packages/generators/src/init-generator.ts
```diff
@@ -19,216 +19,45 @@ const { logger, getPackageManager } = utils;
...
+    public supportedTemplates: string[];
 
-    public constructor(args, opts) {
-        super(args, opts);
+    public constructor(args, options) {
+        super(args, options);
 
-        this.autoGenerateConfig = opts.autoSetDefaults;
-        this.generationPath = opts.generationPath;
+        this.template = options.template;
```
jamesgeorge007, line 30:
```suggestion
        this.template = options.template || 'default';
```

packages/generators/src/index.ts
```diff
@@ -4,41 +4,34 @@ import pluginGenerator from './plugin-generator';
...
-                {
-                    name: 'auto',
-                    type: Boolean,
-                    description: 'To generate default config',
-                },
                 {
                     name: 'generation-path',
                     type: String,
                     description: 'To scaffold in a specified path',
                 },
```
jamesgeorge007, line 24: Specify the current directory as a default value:-
```js
                {
                    name: 'generation-path',
                    type: String,
                    description: 'To scaffold in a specified path',
                    defaultValue: '.',
                },
```

packages/generators/src/init-generator.ts
```diff
@@ -19,216 +19,45 @@ const { logger, getPackageManager } = utils;
...
-            this.autoGenerateConfig,
-        );
-
-        // only apply when output dir is not default
-        if (outputDir !== 'dist') {
-            this.configuration.config.webpackOptions.output = {
-                path: `path.resolve(__dirname, '${outputDir}')`,
-            };
+        if (!this.template) {
+            this.template = 'default';
```
jamesgeorge007, line 44: The fallback takes care of it.

packages/generators/src/index.ts
```diff
@@ -4,41 +4,34 @@ import pluginGenerator from './plugin-generator';
...
-import { npmPackagesExists } from './utils/npm-packages-exists';
-
 class GeneratorsCommand {
     async apply(cli): Promise<void> {
         const { logger } = cli;
 
         await cli.makeCommand(
             {
-                name: 'init [scaffold...]',
+                name: 'init [template] [options]',
```
jamesgeorge007, line 13:
```suggestion
                name: 'init [template]',
```

rishabh3112:
Self Reminder: Add tests.

alexander-akait (commented):
Very good job, but I think we should still leave some questions:
- dev server
- CSS
- babel
- typescript
- html-webpack-plugin

Because it is very popular and will be great to newbie + adding scripts - `serve`, `build`, `dev`, `watch` (maybe more or less). What do you think? Maybe I be some misleading. I mean we should not analyze `webpack.config.js` and just ask questions.

rishabh3112:
> I think we should still leave some questions:

    dev server
    CSS
    babel
    typescript
    html-webpack-plugin

Updating configuration based on questions would require editing the configuration. That itself requires the editing ast of config (therefore needs jscodeshift). But I have a feeling we can achieve same using `webpack-merge`. But I would do that later on (For that I want to take some time to design architecture so that everything remains stable) because currently I am very busy with university exams and projects.

> adding scripts - serve, build, dev, watch (maybe more or less)

I have added dev server by default for now with script for `build` and `serve`.

alexander-akait:
> Updating configuration based on questions would require editing the configuration. That itself requires the editing ast of config (therefore needs jscodeshift). But I have a feeling we can achieve same using webpack-merge. But I would do that later on (For that I want to take some time to design architecture so that everything remains stable) because currently I am very busy with university exams and projects.

Okay, I will look how we can achieve this without `jscodeshift`, I think it is not hard

jamesgeorge007:
> Okay, I will look how we can achieve this without jscodeshift, I think it is not hard

The intended way would be to extend a template based on the provided arguments [extendJson()](https://github.com/sboudrias/mem-fs-editor#extendjsonfilepath-contents-replacer--space).

https://github.com/webpack/webpack-cli/blob/f2e935e8acf6c37e4ff777a80342b5991c1f7485/packages/generators/loader-template/src/_index.js.tpl#L8-L11

rishabh3112:
Can't seem to understand why `isWebpack5` is wrong in v4 tests, works fine locally on v4. :(

rishabh3112:
Everything is working without cache.
Let me try if cache updates if I add it again.

rishabh3112:
> Everything is working without cache.
> Let me try if cache updates if I add it again.

It didn't, is there a way to clear cache @webpack/cli-team? (I found a issue but with not solution https://github.com/actions/cache/issues/2)
If not, changing the key to cache is only solution in my mind.

jamesgeorge007 (commented):
Do we need the `--template` flag as of now? IMO it would be better to have it once we decide to onboard custom templates.

rishabh3112:
> Do we need the --template flag as of now? IMO it would be better to have it once we decide to onboard custom templates.

Yes we need, I am looking to add more templates apart from default in future.

rishabh3112:
TODO: add questioning as well

rishabh3112:
TODO: add flag for enforcing defaults

TODO: add questioning for following in default template
- [x] Javascript
- [ ] CSS
- [ ] Dev-server
- [ ] html-webpack-plugin

TODO: update tests

rishabh3112 (commented):
@alexander-akait All deliverables are done. Having a review would be great 😄 !

packages/generators/src/init-generator.ts
```diff
@@ -19,217 +20,57 @@ const { logger, getPackageManager } = utils;
...
-            this.dependencies.push('workbox-webpack-plugin');
-            (this.configuration.config.webpackOptions.plugins as string[]).push(`new WorkboxWebpackPlugin.GenerateSW({
-				swDest: 'sw.js',
-				clientsClaim: true,
-				skipWaiting: false,
-			})`);
-        }
-
-        this.configuration.config.webpackOptions.mode = this.autoGenerateConfig ? "'production'" : "'development'";
+        await handlers[this.template].questions(this, Question);
```
jamesgeorge007, line 73: Isn't the `Question` argument extraneous? The required prompts can be imported directly within [packages/generators/src/handlers/default.ts](https://github.com/webpack/webpack-cli/pull/2429/files#diff-50d01c8f19701cb38c78b1743da73ab584d818f35d9e0db63c9310b17a522e1e)

```js
import { Confirm, Input, List } from './utils/scaffold-utils';
```

jamesgeorge007 (changes requested):
Left a few comments

packages/generators/src/handlers/default.ts
```diff
@@ -0,0 +1,124 @@
...
+            entryPath += 'js';
+        }
+        self.fs.copyTpl(resolveFile('index.js'), self.destinationPath(entryPath));
+    };
+
+    // Generate entry file/files
+    const entry = './src/index.';
+    if (typeof entry === 'string') {
+        generateEntryFile(entry);
+    }
```
jamesgeorge007, line 99: Are you going to add the prompt for multiple bundles?

packages/generators/src/init-generator.ts
rishabh3112: Not exactly, I am making `handlers` independent of the `webpack-cli` code. So that we can implement support 3rd party templates as well in future without any issues.

packages/generators/init-template/default/webpack.configjs.tpl
```diff
@@ -0,0 +1,47 @@
...
+    plugins: [<% if (htmlWebpackPlugin) { %>
+        new HtmlWebpackPlugin({
+            template: 'index.html',
+        }),
+<% } %>
+        // Add your plugins here
+        // Learn more obout plugins from https://webpack.js.org/configuration/plugins/
+    ],
+    module: {
+        rules: [<% if (lang == "ES6") { %>
```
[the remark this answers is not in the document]
rishabh3112: Thanks, `langType` name looks much better.

test/init/init.test.js
```diff
@@ -0,0 +1,93 @@
...
+        const { stdout, stderr } = await runPromptWithAnswers(assetsPath, ['init', '--use-defaults', '--template=apple'], [`${ENTER}`]);
+        expect(stdout).toContain('Project has been initialised with webpack!');
+        expect(stderr).toContain('webpack.config.js');
+
+        // Test files
+        const files = ['package.json', 'src', 'src/index.js', 'webpack.config.js'];
+        files.forEach((file) => {
+            expect(existsSync(resolve(assetsPath, file))).toBeTruthy();
+        });
+    });
```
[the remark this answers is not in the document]
rishabh3112: Sure, adding it.

packages/generators/src/handlers/default.ts
rishabh3112: Not for now. Will update this logic.

alexander-akait:
@rishabh3112 Can you rebase?

rishabh3112:
@alexander-akait done

INIT.md
```diff
@@ -11,47 +11,16 @@
     -   [Running Locally](#running-locally)
     -   [Running Globally](#running-globally)
     -   [CLI options](#cli-options)
-    -   [Description of questions asked by generator](#description-of-questions-asked-by-generator)
+-   [Description of questions asked by generator](#description-of-questions-asked-by-generator)
```
jamesgeorge007, line 14:
```suggestion
-   [Description of questions asked by the generator](#description-of-questions-asked-by-the-generator)
```

INIT.md
```diff
@@ -87,59 +56,34 @@ webpack-cli init --auto
...
+## Description of questions asked by the generator
 
-3. `In which folder do you want to store your generated bundles? (dist)`
+### Default template
 
-> _Property/key resolved: [output.path](https://webpack.js.org/configuration/output/#output-path)_
+1. `Which of the following JS solutions do you want to use?`
 
-The output directory is where your bundled application will be. Your `index.html` will read the generated files from this folder, that is usually named `dist`.
+> _Property/key resolved: [module.rules](https://webpack.js.org/configuration/module/#module-rules) (for .js, .ts and other related files files)_
```
jamesgeorge007, line 71:
```suggestion
> _Property/key resolved: [module.rules](https://webpack.js.org/configuration/module/#module-rules) (for .js, .ts and other related files)_
```

jamesgeorge007 (commented):
Looking good, left a couple of suggestions.

alexander-akait (changes requested):
- Why we need `--auto` flag here? We should avoid it
- I think we need change `webpack init|c [scaffold...] [options]` to `webpack init|c [path] [options]` (by default `[path]` is `process.cwd()`, in future `scaffold` can be options, so you choose a `scaffold` package and set output path)

Also let's keep questions for `Less`/`Sass`/`Stylus`, because it is very popular

Also we need question for https://webpack.js.org/guides/asset-modules/ (please use `type: 'asset'`), but don't need to ask this if `CSS` or `JS` is true, otherwise `url`, `new URL` will not work, by default please use regexp for `images` and `fonts`

Also we need question for `target`, i.e. `node`/`web`/etc, because it is popular

rishabh3112:
> Why we need --auto flag here? We should avoid/remove it

Yup as we already discussed, it has been removed.

> I think we need change webpack init|c [scaffold...] [options] to webpack init|c [path] [options] (by default [path] is process.cwd(), in future scaffold can be options, so you choose a scaffold package and set output path)
Also let's keep questions for Less/Sass/Stylus, because it is very popular
>
> Also we need question for https://webpack.js.org/guides/asset-modules/ (please use type: 'asset'), but don't need to ask this if CSS or JS is true, otherwise url, new URL will not work, by default please use regexp for images and fonts
>
> Also we need question for target, i.e. node/web/etc, because it is popular

Adding it to TODO now.

rishabh3112:
So here is todo now:
## TODO
- [ ] SASS
- [ ] LESS
- [ ] Stylus
- [ ] asset modules
- [ ] target (need a discussion on that as I was planning to implement seperate template for `node`, default being `web` one)

rishabh3112:
> Also we need question for https://webpack.js.org/guides/asset-modules/ (please use type: 'asset'), but don't need to ask this if CSS or JS is true, otherwise url, new URL will not work, by default please use regexp for images and fonts

What question should we ask regarding this? I am thinking of `Will you be using assets like images and fonts?` currently.

alexander-akait:
@rishabh3112 

> Will you be using assets like images and fonts?

Sounds good and simple

rishabh3112:
>  Also we need question for https://webpack.js.org/guides/asset-modules/ (please use type: 'asset'), but don't need to ask this if CSS or JS is true, otherwise url, new URL will not work, by default please use regexp for images and fonts

I didn't understand what you meant in case we have CSS and JS rules, currently I am not adding asset module support when either of CSS and JS languages solutions are opted.

alexander-akait:
> I didn't understand what you meant in case we have CSS and JS rules, currently I am not adding asset module support when either of CSS and JS languages solutions are opted.

If we don't add it, all `url('./path/to/image.png')` will not work. Solutions:
- ask this always
- ask this only when CSS is `none`

alexander-akait (approved):

snitin315 (approved):
💯

anshumanv (approved):
Well donee 💯

jamesgeorge007 (approved):

PR #2431, opened 2021-02-10
feat: display used config path when logging level=log 

anshumanv:

feat

yes

NA

**Summary**
- Displays the used config when infra logging level is log, helpful to know which config is in use when there are multiple

No

**Other information**

### DIFF
--- a/packages/webpack-cli/lib/plugins/CLIPlugin.js
+++ b/packages/webpack-cli/lib/plugins/CLIPlugin.js
@@ -39,15 +39,18 @@ class CLIPlugin {
 
     setupHelpfulOutput(compiler) {
         const pluginName = 'webpack-cli';
-        const getCompilationName = () => (compiler.name ? ` '${compiler.name}'` : '');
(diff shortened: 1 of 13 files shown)


packages/webpack-cli/lib/plugins/CLIPlugin.js
```diff
@@ -52,7 +60,13 @@ class CLIPlugin {
...
             }
 
-            this.logger.log(`Compilation${getCompilationName()} starting...`);
+            const name = getCompilationName();
+
+            this.logger.log(`Compiler${name ? ` ${name}` : ''} starting...`);
+
+            if (configPath) {
+                this.logger.log(`Compiler${name ? ` ${name}` : ''} is using config: '${configPath}'`);
+            }
```
alexander-akait, line 69: I think we was wrong when we use `compilation` word here, we even use hooks for `compiler`

snitin315 (dismissed):

alexander-akait:
Again problems with cached, we should be careful when setup cache for CI, latest commits for improve it break CI for webpack-cli and webpack-dev-server

PR #2526, opened 2021-03-20
feat: add postcss support to default template

rishabh3112:
Add postcss support as well to default template

WIP

N/A

**Summary**
Add PostCSS support as well to default template because it is popular.

No

**Other information**
/cc @alexander-akait

### DIFF
--- a/packages/generators/init-template/default/postcss.config.js
+++ b/packages/generators/init-template/default/postcss.config.js
@@ -1,5 +1,5 @@
 module.exports = {
     // Add you postcss configuration here
     // Learn more about it at https://github.com/webpack-contrib/postcss-loader#config-files
-    plugins: [['postcss-preset-env']],
+    plugins: [['autoprefixer']],
 };
--- a/packages/generators/init-template/default/postcss.config.js
+++ b/packages/generators/init-template/default/postcss.config.js
@@ -1,0 +1,5 @@
+module.exports = {
+    // Add you postcss configuration here
+    // Learn more about it at https://github.com/webpack-contrib/postcss-loader#config-files
+    plugins: [['postcss-preset-env']],
+};
--- a/packages/generators/src/handlers/default.ts
+++ b/packages/generators/src/handlers/default.ts
@@ -75,7 +75,7 @@ export async function questions(self: CustomGenerator, Question: Record<string, 
             self.dependencies = [...self.dependencies, 'stylus-loader', 'stylus'];
             break;
         case 'PostCSS':
-            self.dependencies = [...self.dependencies, 'postcss-loader', 'postcss'];
+            self.dependencies = [...self.dependencies, 'postcss-loader', 'postcss', 'autoprefixer'];
     }
 
     // store all answers for generation
--- a/packages/generators/src/handlers/default.ts
+++ b/packages/generators/src/handlers/default.ts
@@ -105,7 +105,7 @@ export function generate(self: CustomGenerator): void {
     // Generate README
     self.fs.copyTpl(resolveFile('README.md'), self.destinationPath('README.md'), {});
 
-    // Generate HTML template file
+    // Generate HTML file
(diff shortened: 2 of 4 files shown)


evenstensberg (commented):
Postcss is an addition to sass, css etc and should be added automatically

rishabh3112:
> Postcss is an addition to sass, css etc and should be added automatically

Thanks for review. Here are my opinions:
- Postcss with sass is certainly not a popular option, people either use sass or move all together to using Postcss with css by adding relevent postcss plugins.

- Adding it automatically is not correct as well, many people want to use just css as it works better for many with less dependancy and less build time. (This includes me when I do small uni projects)

Nevertheless, I think I should tag @alexander-akait here as he knows more about how most projects use webpack for css preprocessing being maintainer of many loaders for them.

packages/generators/init-template/default/postcss.config.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+    // Add you postcss configuration here
+    // Learn more about it at https://github.com/webpack-contrib/postcss-loader#config-files
+    plugins: [['postcss-preset-env']],
```
alexander-akait, line 4: I think better put `autoprefixer` here, it is more popular

packages/generators/src/handlers/default.ts
```diff
@@ -56,7 +56,7 @@ export async function questions(self: CustomGenerator, Question: Record<string,
         self,
         'cssType',
         'Which of the following CSS solutions do you want to use?',
-        ['none', 'CSS', 'SASS', 'LESS', 'Stylus'],
+        ['none', 'CSS', 'SASS', 'LESS', 'Stylus', 'PostCSS'],
```
alexander-akait, line 59: I think we need refactor here, ideally:
- Do you want to use CSS? - None/Yes (style-loader and css-loader), will be great to add `mini-css-extract-plugin` when mode is production, but we can do it in separate PR
- Do you want to use `CSS` preprocessor: `SASS`/`Less`/`Stylus`
- Do you want to use `CSS` postprocessolor : `postcss`

What do you think?

rishabh3112: `mini-css-extract-plugin` seems like like a very good idea, will add this in.
Meanwhile, I couldn't understand how adding three separate questions for it seems to bring any value as it is always nice to keep things simple and quick to use. A little more explanation on your intentions would be nice.

alexander-akait: Some developers use `postcss`, some not, maybe we can do it in our questions. 

Do you need CSS?
- None
- CSS
- CSS and PostCSS
- CSS and SASS (for less and Stylus too)
- CSS, PostCSS and SASS (for less and Stylus too)

I was thinking about separate this into different questions... There are a lot of answers...

rishabh3112: Hmm, I understood. I think I would do that in seperate PR. Currently in this PR, I am adding PostCSS  with CSS only.

alexander-akait: Yep, let's do it in separate PR

packages/generators/init-template/default/postcss.config.js
rishabh3112: I found one more thing while exploring, `autoprefixer` is already included in `postcss-preset-env` (https://github.com/csstools/postcss-preset-env#autoprefixer).

alexander-akait: `postcss-preset-env` is very poorly updated and contains rather controversial solutions, if someone wants to add it, he can do it very easily, most developers do not use it due a lot of bugs and problems

alexander-akait (dismissed):
cc @webpack/cli-team

snitin315 (approved):

alexander-akait (approved):

PR #2546, opened 2021-03-23
feat: add env var to skip local import

### LINKED ISSUE #2543: importLocal check may invoke wrong version of WDS in a Yarn workspace
- Operating System: Ubuntu 16
- Node Version: 12.16.1
- NPM Version: 6.13.4
- webpack Version: 4.43.0
- webpack-dev-server Version: 3.11.0
- Browser: N/A

- [x] This is a **bug**
- [x] This is a **modification** request

### Code

Folder structure (including installed packages):

```
- yarn.lock
- node_modules // workspace root
  - import-local
  - webpack-dev-server@3.7.2
- my-project
  - webpack.config.js
  - package.json // includes an npm script to invoke WDS
  - node_modules
    - webpack-dev-server@3.11.0
```

### Expected Behavior

When I invoke WDS from the nested `my-project` directory it should use v3.11.0

### Actual Behavior

v3.11.0 is invoked correctly at first, but the `importLocal` check in the `/bin/webpack-dev-server.js` script causes v3.7.2 to be used. This is because of a bug in the `import-local` package (which I have reported [here](https://github.com/sindresorhus/import-local/issues/7)) where it doesn't take into account that there might be multiple `node_modules` folders, like in a Yarn workspace.

The bug is intermittent because it depends on Yarn's hoisting behaviour; if `import-local` had been placed in the nested `node_modules` directory then v3.11.0 would be used as intended.

The bug can occur for any versions of WDS; 3.11.0 and 3.7.2 are just the versions in my project.

### For Bugs; How can we reproduce the behavior?

It's tricky to reproduce this reliably as it depends on Yarn's hoisting behaviour, but it should be possible to manually recreate a folder structure like the above and observe the behaviour.

### For Features; What is the motivation and/or use-case for the feature?

I would like to propose a flag to skip the `importLocal` check. In my scenario I can be sure there won't be a global installation of WDS, so there's no need to run the check at all, which would avoid any issues like this.

anshumanv:
feat

Yes

**Summary**

Add env var to skip import-local

No

**Other information**
Fix #2543

### DIFF
--- a/test/build/import-local/import-local.test.js
+++ b/test/build/import-local/import-local.test.js
@@ -13,7 +13,6 @@ describe('import local', () => {
         const { exitCode, stderr, stdout } = run(__dirname, [], {
             env: { WEBPACK_CLI_SKIP_IMPORT_LOCAL: true },
         });
-        console.log({ exitCode, stderr, stdout });
         expect(importLocalMock).toHaveBeenCalledTimes(0);
         expect(exitCode).toBe(0);
         expect(stderr).toBeFalsy();
--- a/test/build/import-local/import-local.test.js
+++ b/test/build/import-local/import-local.test.js
@@ -1,0 +1,22 @@
+'use strict';
+
+const { run } = require('../../utils/test-utils');
+
+const importLocalMock = jest.fn();
+jest.setMock('import-local', importLocalMock);
+
+describe('import local', () => {
+    beforeEach(() => {
+        importLocalMock.mockClear();
+    });
+    it('should skip import local when supplied', () => {
+        const { exitCode, stderr, stdout } = run(__dirname, [], {
+            env: { WEBPACK_CLI_SKIP_IMPORT_LOCAL: true },
+        });
+        console.log({ exitCode, stderr, stdout });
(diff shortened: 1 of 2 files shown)


test/build/import-local/import-local.test.js
```diff
@@ -0,0 +1,22 @@
...
+
+describe('import local', () => {
+    beforeEach(() => {
+        importLocalMock.mockClear();
+    });
+    it('should skip import local when supplied', () => {
+        const { exitCode, stderr, stdout } = run(__dirname, [], {
+            env: { WEBPACK_CLI_SKIP_IMPORT_LOCAL: true },
+        });
+        console.log({ exitCode, stderr, stdout });
```
alexander-akait, line 16: Let's remove it

anshumanv: hm done

alexander-akait:
After merge we need improve docs

anshumanv:
will improve

alexander-akait (approved):
/cc @webpack/cli-team

snitin315 (approved):

PR #2553, opened 2021-03-25
build: add bootstrap step in workflow

snitin315:
**What kind of change does this PR introduce?**
build

**Did you add tests for your changes?**
NA
**If relevant, did you update the documentation?**
NA
**Summary**

run `yarn lerna bootstrap` on CI.

**Does this PR introduce a breaking change?**
No

**Other information**

PR #2555, opened 2021-03-25
fix: generation of `negative` flags

snitin315:
bugfix

Yes
No need
**Summary**
fix generation of `negative` flags. currently, we have negative flags for all boolean flags regardless of `option.negative` value.

No

**Other information**
No


alexander-akait:
@snitin315 we should better to solve it, I will finish your PR, dont' worry

alexander-akait:
/cc @webpack/cli-team I want to clarify something here, we have three types:
- `type: 'boolean'` means `--option` and `--no-options`, any boolean for CLI always have positive and negative value
- `type: 'enum'` and `values: [true]` means only `--option` (no negative)
- `type: 'enum'` and `values: [false]` means only `--no-option` (no positive)

Here situation when you can use `type: 'enum'` and `values: [false, true]`, it is possible, but it doesn't make sense, because it means `boolean`

alexander-akait:
Also need review, we should merge it before release

OPTIONS.md
```diff
@@ -858,7 +858,7 @@ Commands:
   configtest|t [config-path]                                                         Validate a webpack configuration.
   help|h [command] [option]                                                          Display help for commands and options.
   info|i [options]                                                                   Outputs information about your system.
-  init|c [generation-path] [options]                                                 Initialize a new webpack project.
+  init|c [generation-path] [options]                                                 Initialize a new webpack configuration.
```
snitin315, line 861: This should not have changed.

alexander-akait: Good catch

snitin315:
> Here situation when you can use `type: 'enum'` and `values: [false, true]`, it is possible, but it doesn't make sense, because it means `boolean`

Yes `'boolean'` should be preferred here.

anshumanv (dismissed):
Much better

alexander-akait:
@snitin315 Let's wait green https://github.com/webpack/webpack-cli/pull/2553 and merge, then rebase

alexander-akait (approved):

anshumanv (approved):

PR #2570, opened 2021-03-27
fix: imports in generator

anshumanv:

fix

Yes

NA

**Summary**
- No not import from webpack-cli, use instance
- 

no

**Other information**

### DIFF
--- a/packages/generators/src/addon-generator.ts
+++ b/packages/generators/src/addon-generator.ts
@@ -33,15 +33,14 @@ const addonGenerator = (
     templateFn: (instance: any) => Record<string, unknown>,
 ): Generator.GeneratorConstructor => {
     return class extends Generator {
-        public logger: Record<string, (string) => void>;
-        public getPackageManager: (val: void) => string;
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        public utils: any;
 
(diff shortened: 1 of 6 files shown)


packages/generators/src/addon-generator.ts
```diff
@@ -37,6 +33,17 @@ const addonGenerator = (
     templateFn: (instance: any) => Record<string, unknown>,
 ): Generator.GeneratorConstructor => {
     return class extends Generator {
+        public logger: Record<string, (string) => void>;
+        public getPackageManager: (val: void) => string;
```
alexander-akait, line 37: let's move it under `utils`

alexander-akait: so we can add more utils in future

anshumanv: yep lets do it

anshumanv: need to union types for all utils, it is nightmare 😞

alexander-akait: You can use `any` right now, after migrate on ts or jsdocs we will fix it

anshumanv: thanks

alexander-akait (approved):

PR #2583, opened 2021-03-31
test: update

cap-Bernardito:
Tests edited on async await


alexander-akait (commented):
Need improve:
- fix failed test on windows
- let's use `run` instead `runAsync` (I think we don't need sync tests) 
- let's implement `normalizeStdout` function in `test-utils` where we strip ansi (for future snapshots tests)

alexander-akait (approved):

alexander-akait:
Thanks

Issue #2430, opened 2021-02-10
feature: new init generator


rishabh3112:
**Is your feature request related to a problem? Please describe.**
Currently, the init generation is a bad DX with support for custom scaffolds completely broken.

**Describe the solution you'd like**
Generate pre-defined configurations instead of creating one using AST transforms. The implementation will be done in steps:

- [x] Simplify the generation flow for init (#2429)
- [x] Add default template for init (#2429)
- [x] Add tests for init
- [x] Update documentation

**Describe alternatives you've considered**
N/A

**Additional context**
#2409
/cc @webpack/cli-team

alexander-akait:
After this we should not have `jscodeshift` in our deps

alexander-akait:
Also we should always generate `webpack.config.js`

alexander-akait:
And we should not exist with `-1` code when somebody press `CRTL + C`  on init (try it using `npm init`)

alexander-akait:
@rishabh3112 What is status? I want to finish it for the next release

rishabh3112:
> @rishabh3112 What is status? I want to finish it for the next release

 It works as we need. I have added tests as well. Will rebase and then we can merge it.

alexander-akait:
@rishabh3112 Do we drop `jscodeshift` from our deps?

rishabh3112:
> @rishabh3112 Do we drop `jscodeshift` from our deps?

Yup its not needed now for init. Will remove it from deps as well.

rishabh3112:
# Status
1. Simplification of code structure done
2. Created default template
2. Testing is done as well

# Nice to dos 
1. Add questioning for webpack configuration. (Trying on #2429 it self)

rishabh3112:
All done ✨ 
Will be adding more templates for `react`, `vue`, `node-packages` etc in further PRs 😎

rishabh3112:
Documentation will be tracked on https://github.com/webpack/webpack.js.org/issues/4741

rishabh3112:
Now available in latest release ✅ 

TODO: Adding templates for common usecases (https://github.com/webpack/webpack-cli/issues/2430#issuecomment-793683956) - WIP

rishabh3112:
### Todo
- [ ] Add support add varied options as per `mode` (eg. `css-minimizer-webpack-plugin` only in production)
- [ ] Update Mini-CSS-extract-plugin configuration after above change. For example:
  ```
[...]

PR #2585, opened 2021-04-01
feat: add support for mode specific config

rishabh3112:
feat

WIP

**Summary**
part of #2430

No

**Other information**
No


alexander-akait (commented):
hm, maybe we can avoid using `webpack-merge` to avoid unnecessary packages for simple configurations?

rishabh3112:
> hm, maybe we can avoid using webpack-merge to avoid unnecessary packages for simple configurations?

Okay, What should differentiate "simple" config from "complex" config?

alexander-akait:
Let's finish, I will look at result and thinking about can be it simplified and is it worth doing

packages/generators/init-template/default/webpack.configjs.tpl
```diff
@@ -64,3 +60,25 @@ module.exports = {
...
+        host: 'localhost',
+    },<% } %>
+}
+
+module.exports = function (env, argv) {
+    if (argv.mode == 'production') {
+        return merge(base, production);
+    }
+    return merge(base, development);
+}
```
alexander-akait, line 84: We need format it better :smile: 
No `;` in some places, no newline at the end and etc problems, maybe we can generate config, run prettier and format this file as prettier

rishabh3112: Yup we can, should I do this in this PR itself or open another PR?

rishabh3112: I think in separate as this will be a big change in scaffolding process itself.

alexander-akait: Yep, let' do separately, I will review this deeply today

evenstensberg (dismissed):

packages/generators/init-template/default/webpack.configjs.tpl
rishabh3112: I am fixing things manually for this particular PR. Will open separate PR for prettier.

rishabh3112:
@webpack/cli-team need review

packages/generators/init-template/default/webpack.configjs.tpl
```diff
@@ -68,3 +64,25 @@ module.exports = {
...
+        host: 'localhost',
+    },<% } %>
+};
+
+module.exports = function (env, argv) {
+    if (argv.mode == 'production') {
+        return merge(base, production);
+    }
+    return merge(base, development);
+};
```
alexander-akait, line 88: I think we can simplify this now, we set `process.env.NODE_ENV` so we can change configuration based on in, `webpack-merge` is good package for multi compiler modes, but I think for fast start we don't need extra packages now, `process.env._NODE_ENV` is enough

rishabh3112: Okay, looking into it.

rishabh3112: Done, Thanks.

rishabh3112:
@webpack/cli-team need review

alexander-akait (commented):
Now we set `process.env.NODE_ENV` by default, so we can use this to enable/disable some features, so I think we can avoid using `webpack-merge`

rishabh3112:
> Now we set process.env.NODE_ENV by default, so we can use this to enable/disable some features, so I think we can avoid using webpack-merge

removed it already. Please take a look at changes once.

packages/generators/init-template/default/webpack.configjs.tpl
```diff
@@ -68,3 +63,16 @@ module.exports = {
...
+        config.mode = 'production';
+    } else {
+        config.mode = 'development';<% if (devServer) { %>
+        config.devServer = {
+            open: true,
+            host: 'localhost',
+        };<% } %>
+    }
+    return config;
+};
```
alexander-akait, line 78: hm, I think some developers use serve for development and production mode (for testing and etc), so I think keeping `devServer` is good as part of basic configuration, because we use `devServer` option only in `serve` mode

rishabh3112: Okay, on it. Thanks.

rishabh3112: Done. Will push snaps (currently running locally) then we can merge.

alexander-akait (approved):
/cc @webpack/cli-team

anshumanv (approved):

snitin315 (approved):
🚀

jamesgeorge007 (approved):

PR #2603, opened 2021-04-06
fix: `configs` property for `--devtool`

snitin315:
bugfix

Yes
No
**Summary**

`devtool` can also accept `false`.

No

**Other information**
None


alexander-akait (commented):
Only for webpack v5, we already support it

snitin315:
I see then we should remove `negative: true`?

snitin315:
It is allowed for v4 too https://github.com/webpack/webpack/blob/webpack-4/schemas/WebpackOptions.json#L2039

alexander-akait:
yep, okay, let's do it, but I don't want to accept any bugfixes/features (expect security bugs for v4), because when we will drop webpack v4 we will remove these flags

snitin315:
Makes sense 👍🏻

alexander-akait (approved):

alexander-akait:
Thanks

PR #2624, opened 2021-04-14
fix(serve): do not set port for modern client

snitin315:
fix

NA

NA
**Summary**

we removed `client.port` in https://github.com/webpack/webpack-dev-server/pull/3181

Refer https://github.com/webpack/webpack-dev-server/pull/3181#issuecomment-818921391

No

**Other information**
No

### DIFF
--- a/packages/serve/src/startDevServer.ts
+++ b/packages/serve/src/startDevServer.ts
@@ -72,16 +72,7 @@ export default async function startDevServer(
         const options = mergeOptions(compilerWithDevServerOption.options.devServer || {}, devServerCliOptions);
 
         if (isDevServer4) {
-            // eslint-disable-next-line node/no-extraneous-require
-            const devServerFlags = require('webpack-dev-server/bin/cli-flags').devServer;
-
-            const isModernClient = Boolean(devServerFlags.find((flag) => flag.name === 'client-web-socket-url'));
-
-            if (!isModernClient) {
-                options.port = await findPort(options.port);
-                options.client = options.client || {};
-                options.client.port = options.client.port || options.port;
-            }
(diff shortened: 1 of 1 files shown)


packages/serve/src/startDevServer.ts
```diff
@@ -72,9 +72,17 @@ export default async function startDevServer(
         const options = mergeOptions(compilerWithDevServerOption.options.devServer || {}, devServerCliOptions);
 
         if (isDevServer4) {
+            // eslint-disable-next-line node/no-extraneous-require
+            const devServerFlags = require('webpack-dev-server/bin/cli-flags').devServer;
+
+            const isModernClient = Boolean(devServerFlags.find((flag) => flag.name === 'client-web-socket-url'));
+
             options.port = await findPort(options.port);
             options.client = options.client || {};
```
alexander-akait, line 81: We move it under `isModerClient` too

snitin315: Oh yes!

alexander-akait:
To be honestly I don't understand why we do it, if client port is not specified or `0` we just get `self.location.port` (basic workflow), so when you run without proxy, we already get port from browser, if we proxy hostname developer uses the same port so `self.location.port` will work, only one situation possible when you proxy dev server on other port, but in this case you need manually set `client.port`, because we don't know original hostname/port

alexander-akait:
It was my mistake here, we should not do it here, we should improve it on dev server, i.e. `options.client.port || options.port`, when we do release and improve `public` here https://github.com/webpack/webpack-dev-server/pull/3181/ and do release again

Difficult...

PR #2626, opened 2021-04-14
feat: improve performance

anshumanv:

performance 

existing

NA

**Summary**
- Improve performance
- inbuiltOptions as object
- unnecessary iterations purged

no

**Other information**


anshumanv:
Looking at CI

anshumanv:
Looks like some spacing diff in help output

snitin315:
Try a rebase.

anshumanv:
hm looks like no help

test/build/core-flags/optimization-flags.test.js
```diff
@@ -4,7 +4,7 @@ const { run, hyphenToUpperCase } = require('../../utils/test-utils');
 const CLI = require('../../../packages/webpack-cli/lib/index');
 
 const cli = new CLI();
-const optimizationFlags = cli.getBuiltInOptions().filter(({ name }) => name.startsWith('optimization-'));
+const optimizationFlags = Object.values(cli.getBuiltInOptions().coreFlags).filter(({ name }) => name.startsWith('optimization-'));
```
evenstensberg, line 7: Util func?

packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -316,7 +316,7 @@ class WebpackCLI {
             return this.builtInOptionsCache;
         }
 
-        const minimumHelpFlags = [
+        const minimumHelpFlags = new Set([
```
evenstensberg, line 319: Perf here would have such little impact that there's no point. If there is 10000 objects I understand, but this is Max 30, so O(30ns) vs O(30^2) wouldnt make much difference

alexander-akait (commented):
Ideally we should reduce number of `Object.entries`/`Object.values`/`Object.keys`/etc

For example we can use this for help:
```
Object.entries(minimumHelpFlags).forEach(([name, meta]) => {
  options[name].helpLevel = 'minimum';
});
```

So we avoid unnecessary looping over 1k options

packages/webpack-cli/lib/webpack-cli.js
anshumanv: we did a lookup for every flag, there could be ~300 flags

alexander-akait:
Found more places to improve perf, I will finish it

anshumanv:
Big thanks! 👍

PR #2640, opened 2021-04-17
feat: prettify generated config

rishabh3112:
Feature

WIP

N/A

**Summary**
Prettify configuration after config is generated

No

**Other information**
No


packages/generators/src/init-generator.ts
```diff
@@ -70,6 +73,24 @@ export default class InitGenerator extends CustomGenerator {
...
+
+        // Handle installation of prettier
+        try {
+            // eslint-disable-next-line node/no-extraneous-require
+            require.resolve('prettier');
+        } catch (err) {
+            const { installPrettier } = await Question.Confirm(
+                this,
+                'installPrettier',
+                'Do you like to install prettier to prettify generated configuration?',
```
evenstensberg, line 85: to format*

packages/generators/src/init-generator.ts
```diff
@@ -86,4 +107,24 @@ export default class InitGenerator extends CustomGenerator {
...
+            const formattedSource = prettier.format(source, {
+                singleQuote: true,
+                trailingComma: 'all',
+                printWidth: 140,
+                tabWidth: 4,
+                parser: 'babel',
+            });
+            writeFileSync(this.configurationPath, formattedSource);
+        } catch (err) {
+            this.utils.logger.log(`${yellow(`Generated configuration may not be properly formatted as prettier was not found.`)}`);
```
evenstensberg, line 126: found in dependencies

packages/generators/src/init-generator.ts
```diff
@@ -86,4 +107,24 @@ export default class InitGenerator extends CustomGenerator {
...
+                printWidth: 140,
+                tabWidth: 4,
+                parser: 'babel',
+            });
+            writeFileSync(this.configurationPath, formattedSource);
+        } catch (err) {
+            this.utils.logger.log(`${yellow(`⚠ Generated configuration may not be properly formatted as prettier is not installed.`)}`);
+            return;
+        }
+    }
```
alexander-akait, line 129: Maybe we avoid extra reading file? But I agree, it is good idea

rishabh3112: I did that because if user opts to install prettier, we can only prettify file after `installing` context of yeoman. Only context that comes after `installing` in yeoman life-cycle is `end`. I considered above solution better as compared to installing prettier earlier using child process because we are not wasting time running npm install twice.

rishabh3112: If you have a suggestion let me know.

alexander-akait: Maybe `yeoman` have hooks before writing files? I don't familiar, so maybe you approach is good, just want check it

rishabh3112: > Maybe yeoman have hooks/callbacks before writing files?

Nothing I am aware of, Files are written first, then packages are installed (https://yeoman.io/authoring/running-context.html). No way other than installing prettier using child process before writting.

packages/generators/src/init-generator.ts
```diff
@@ -86,4 +107,24 @@ export default class InitGenerator extends CustomGenerator {
...
+            // eslint-disable-next-line node/no-extraneous-require, @typescript-eslint/no-var-requires
+            const prettier = require('prettier');
+            const source = readFileSync(this.configurationPath, { encoding: 'utf8' });
+            const formattedSource = prettier.format(source, {
+                singleQuote: true,
+                trailingComma: 'all',
+                printWidth: 140,
+                tabWidth: 4,
+                parser: 'babel',
+            });
```
alexander-akait, line 123: I think better do it without our options, because some developers can prefer other options, prettier should read configuration(s) by default

rishabh3112: Okay, on it

rishabh3112:
@webpack/cli-team need review

packages/generators/src/init-generator.ts
```diff
@@ -86,4 +107,18 @@ export default class InitGenerator extends CustomGenerator {
...
+    public end(): void {
+        // Prettify configuration file if possible
+        try {
+            // eslint-disable-next-line node/no-extraneous-require, @typescript-eslint/no-var-requires
+            const prettier = require('prettier');
+            const source = readFileSync(this.configurationPath, { encoding: 'utf8' });
+            const formattedSource = prettier.format(source, { parser: 'babel' });
+            writeFileSync(this.configurationPath, formattedSource);
+        } catch (err) {
+            this.utils.logger.log(`${yellow(`⚠ Generated configuration may not be properly formatted as prettier is not installed.`)}`);
```
jamesgeorge007, line 120:
```suggestion
            this.utils.logger.warn('⚠ Generated configuration may not be properly formatted as prettier is not installed.');
```

packages/generators/src/init-generator.ts
```diff
@@ -70,6 +73,24 @@ export default class InitGenerator extends CustomGenerator {
...
+            const { installPrettier } = await Question.Confirm(
+                this,
+                'installPrettier',
+                'Do you like to install prettier to format generated configuration?',
+                true,
+                false,
+            );
+
+            if (installPrettier) {
+                this.dependencies = [...this.dependencies, 'prettier'];
```
jamesgeorge007, line 91:
```suggestion
                this.dependencies.push('prettier');
```

jamesgeorge007 (changes requested):
Left a couple of suggestions.

alexander-akait (commented):
Can we add test(s) where developer say `No`?

rishabh3112:
> Can we add test(s) where developer say No?

Will smoketest be fine? (as this would require simulate `prettier` being missing. It would be difficult to implement a normal test as all tests run in parallel)

alexander-akait:
Yep, smoke is enough

packages/generators/src/init-generator.ts
```diff
@@ -70,6 +73,24 @@ export default class InitGenerator extends CustomGenerator {
         }
 
         await handlers[this.template].questions(this, Question);
+
+        // Handle installation of prettier
+        try {
+            // eslint-disable-next-line node/no-extraneous-require
+            require.resolve('prettier');
+        } catch (err) {
+            const { installPrettier } = await Question.Confirm(
```
alexander-akait, line 82: We have some problems with init and no deps :disappointed: try to run `webpack init` without installed deps locally and look at result, I think we should solve it before merge it

alexander-akait: I known it is not directly problem in this PR, but if we want integration something with post install, we need solve it, otherwise it is very bad DX

alexander-akait: /cc @webpack/cli-team make sense to priority this problem and search fixes/workarounds/solutions

rishabh3112: Yup that something I feel to address as well. `webpack` and `webpack-cli` are now generally used as devDependencies, This adds noise of installing generators package as well as overriding package.json file during scaffolding. If we can make `webpack-cli` to be recommended for global installation (Like create-react-app, Vue CLI, etc), then we can enhance the DX here.

rishabh3112:
@webpack/cli-team I am unable to see https://github.com/webpack/webpack-cli/commit/1e1e93cf3a39a815f31e2fd6d331a92aa62b38c9 commit in pull request here but have push it to the branch of this PR (as link suggests). Does anyone knows reason for this?

alexander-akait:
@rishabh3112 Let's rebease, I improve our tests in some places and fix timeouts, macos is real slow

rishabh3112:
@alexander-akait can we merge this? CI is green now. I have added smoketest for it so shouldn't be a problem.

alexander-akait (approved):
Let's merge

packages/generators/src/init-generator.ts
```diff
@@ -70,6 +73,24 @@ export default class InitGenerator extends CustomGenerator {
...
+                this,
+                'installPrettier',
+                'Do you like to install prettier to format generated configuration?',
+                true,
+                false,
+            );
+
+            if (installPrettier) {
+                this.dependencies.push('prettier');
```
anshumanv, line 91: should be devDeps

rishabh3112: Every dep in this.dependencies is installed as devDep.

snitin315 (approved):

anshumanv (approved):

jamesgeorge007 (approved):

PR #2632, opened 2021-04-15
Refactor test

alexander-akait:
**What kind of change does this PR introduce?**

refactor/tests

**Did you add tests for your changes?**

it is tests

**If relevant, did you update the documentation?**

No need

**Summary**

Part of https://github.com/webpack/webpack-cli/pull/2573 to avoid big changes

**Does this PR introduce a breaking change?**

No

**Other information**

In future we should migrate on stdout and stderr snapshot  tests everywhere, so feel free to send PRs with improve it, better do it by command, for `build` command by group of options, `serve` no need touch right now WIP on this

PR #2643, opened 2021-04-18
fix: improve parsing of `--env` flag

### LINKED ISSUE #2642: --env foo="" is true instead of being "".
**Describe the bug**

--env foo="" is true instead of being "".

**To Reproduce**

1. Call webpack with the option `--env foo=""`
2. In webpack.config.js foo is equal to true instead of "".

**Expected behavior**

Foo should be equal to "" instead of true.

snitin315:
bugfix

Yes

will add an example in webpack.js.org after merge.

**Summary**

Fixes #2642 

Maybe

**Other information**
No

### DIFF
--- a/packages/webpack-cli/lib/webpack-cli.js
+++ b/packages/webpack-cli/lib/webpack-cli.js
@@ -374,10 +374,12 @@ class WebpackCLI {
                 name: 'env',
                 type: (value, previous = {}) => {
                     // for https://github.com/webpack/webpack-cli/issues/2642
-                    const regExpForSplitting = (value.match(/=/g) || []).length === 1 ? /=/ : /=(.+)/;
+                    if (value.endsWith('=')) {
+                        value.concat(`''`);
+                    }
 
                     // This ensures we're only splitting by the first `=`
-                    const [allKeys, val] = value.split(regExpForSplitting, 2);
+                    const [allKeys, val] = value.split(/=(.+)/, 2);
                     const splitKeys = allKeys.split(/\.(?!$)/);
 
                     let prevRef = previous;
--- a/packages/webpack-cli/lib/webpack-cli.js
+++ b/packages/webpack-cli/lib/webpack-cli.js
@@ -373,8 +373,11 @@ class WebpackCLI {
             {
                 name: 'env',
                 type: (value, previous = {}) => {
+                    // for https://github.com/webpack/webpack-cli/issues/2642
+                    const regExpForSplitting = (value.match(/=/g) || []).length === 1 ? /=/ : /=(.+)/;
+
                     // This ensures we're only splitting by the first `=`
-                    const [allKeys, val] = value.split(/=(.+)/, 2);
+                    const [allKeys, val] = value.split(regExpForSplitting, 2);
                     const splitKeys = allKeys.split(/\.(?!$)/);
 
                     let prevRef = previous;
@@ -389,7 +392,11 @@ class WebpackCLI {
                         }
 
                         if (index === splitKeys.length - 1) {
-                            prevRef[someKey] = val || true;
+                            if (typeof val === 'string') {
+                                prevRef[someKey] = val;
+                            } else {
+                                prevRef[someKey] = true;
+                            }
                         }
 
                         prevRef = prevRef[someKey];
--- a/packages/webpack-cli/lib/webpack-cli.js
+++ b/packages/webpack-cli/lib/webpack-cli.js
@@ -375,7 +375,7 @@ class WebpackCLI {
                 type: (value, previous = {}) => {
                     // for https://github.com/webpack/webpack-cli/issues/2642
                     if (value.endsWith('=')) {
-                        value.concat(`''`);
+                        value.concat('""');
                     }
 
                     // This ensures we're only splitting by the first `=`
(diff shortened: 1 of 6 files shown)


packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -373,8 +373,11 @@ class WebpackCLI {
             {
                 name: 'env',
                 type: (value, previous = {}) => {
+                    // for https://github.com/webpack/webpack-cli/issues/2642
+                    const regExpForSplitting = (value.match(/=/g) || []).length === 1 ? /=/ : /=(.+)/;
```
anshumanv, line 377: why?

snitin315: When passing `--env foo=""` via terminal the value is actually `foo=` and not `foo=""` hence setting `env.foo=` = `true`,

I don't think it is problem with our CLI, the terminal just ignores `""`.

To get the desired behavior we need to pass `--env foo=\"\"` with this the value is `foo=""` and hence the desired behavior.

snitin315: I am not sure if we should accept this change it is more of a workaround for the situation. Maybe we can mention this somewhere in docs and leave it as is.

packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -389,7 +392,11 @@ class WebpackCLI {
                         }
 
                         if (index === splitKeys.length - 1) {
-                            prevRef[someKey] = val || true;
```
anshumanv, line 392: @snitin315 bug is here, I don't think you need diff regex, can you verify? 

what's the result with original regex and this removed?

snitin315: Screenshot 2021-04-18 at 3 28 48 PM

anshumanv: hmm yes it won't help, lemme thing of something else

rishabh3112 (approved):
lgtm

packages/webpack-cli/lib/webpack-cli.js
anshumanv: There is logical error, for any value containing = this will match, better to look for last char in the string and if it's `=` then set value as `""`

anshumanv (changes requested):

packages/webpack-cli/lib/webpack-cli.js
snitin315: Can you give example?

rishabh3112:
> There is logical error, for any value containing = this will match, better to look for last char in the string and if it's = then set value as ""

Yup, good catch. One more thing we can do is to introduce some special character for empty string like we do in parser grammars. ``` ` ``` could be tried for this.

packages/webpack-cli/lib/webpack-cli.js
anshumanv: yep, look how it uses first regex in all cases

image

snitin315: For any value containing only one `=` it will match and the corresponding regexp will parse it correctly. I believe test cases are already in place. Am I missing something here?

anshumanv: maybe we are good with this, I'm not sure why the other was used, but there was some suggestion on the PR, try with this regex in all cases, we might be good with this (`/=/`)

anshumanv: i.e remove the regexExpForSplitting logic

snitin315: I will improve and add more tests 👍

anshumanv: 💯 💯 💯

snitin315: Other is for the case of multiple `=`. 

`--env foo=bar=app` => `foo: 'bar=app'`

anshumanv: hmm yes, what about cases when `foo=bar=""` quite absurd but possibility

snitin315: Updated 👍🏻

packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -373,6 +373,11 @@ class WebpackCLI {
             {
                 name: 'env',
                 type: (value, previous = {}) => {
+                    // for https://github.com/webpack/webpack-cli/issues/2642
+                    if (value.endsWith('=')) {
+                        value.concat('""');
```
alexander-akait, line 378: Something weird with coverage here

alexander-akait:
Added more tests + fix edge cases with `--verison=` (empty string)

alexander-akait (approved):

alexander-akait:
After CI will be green I want to merge it, because we need finish https://github.com/webpack/webpack-cli/pull/2632 and when finish testing on webpack-dev-server@4, after this we need do release (and webpack-dev-server release too)

anshumanv (approved):

PR #2646, opened 2021-04-20
test: refactor core-flags

cap-Bernardito:

Gets flags from `webpack.cli.getArguments` function

No

**Other information**

No

### DIFF
--- a/test/utils/test-utils.js
+++ b/test/utils/test-utils.js
@@ -256,14 +256,21 @@ const uniqueDirectoryForTest = async (assetsPath) => {
     return result;
 };
 
-const getWebpackCliArguments = () =>
-    Object.entries(cli.getArguments()).map((flag) => {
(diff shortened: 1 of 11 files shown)


test/utils/test-utils.js
```diff
@@ -256,6 +256,15 @@ const uniqueDirectoryForTest = async (assetsPath) => {
     return result;
 };
 
+const getWebpackCliArguments = () =>
+    Object.entries(cli.getArguments()).map((flag) => {
```
anshumanv, line 260: not available in v4

cap-Bernardito: this is only for tests `core-flags` (only fo v5)

anshumanv: can we take a param and send filtered flags? doing in every test seems redundant

alexander-akait (approved):

PR #2658, opened 2021-04-23
test: snapshots more

alexander-akait:

tests

it is tests

**Summary**

cc @webpack/cli-team We need to rewrite most of our tests on snapshot testing, here example how we should do it, just improve normalized and add this lines everywhere 

No

**Other information**

No

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "ts-jest": "^26.5.5",
     "ts-node": "^9.1.1",
     "typescript": "^4.1.3",
-    "webpack": "^4.46.0",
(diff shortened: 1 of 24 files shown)


package.json
```diff
@@ -81,7 +81,7 @@
     "ts-jest": "^26.5.5",
     "ts-node": "^9.1.1",
     "typescript": "^4.1.3",
-    "webpack": "^5.35.0",
+    "webpack": "^4.46.0",
```
anshumanv, line 84: need revert

PR #2674, opened 2021-04-28
refactor(generators): remove extraneous args to addon-generator

jamesgeorge007:
refactoring

Already present

N/A

**Summary**
https://github.com/webpack/webpack-cli/blob/65ae77d98e918ea5907de89c9c87b797f57f1b2f/packages/generators/src/addon-generator.ts#L30

Directory content is inferred from the `templateDir` argument. Hence, it is not necessary to explicitly specify the template files.

No

**Other information**
N/A

### DIFF
--- a/packages/generators/src/addon-generator.ts
+++ b/packages/generators/src/addon-generator.ts
@@ -100,8 +100,14 @@ const addonGenerator = (
             // eslint-disable-next-line @typescript-eslint/no-var-requires
             this.fs.extendJSON(this.destinationPath('package.json'), require(packageJsonTemplatePath)(this.props.name));
 
-            // An array of file paths (relative to `./templates`) of files to be copied to the generated project
-            const files = getFiles(this.resolvedTemplatePath);
+            let files = [];
+            try {
+                // An array of file paths (relative to `./templates`) of files to be copied to the generated project
+                files = getFiles(this.resolvedTemplatePath);
+            } catch (error) {
+                this.utils.logger.error(`Failed to generate starter template.\n ${error}`);
+                process.exit(2);
+            }
 
             // Template file paths should be of the form `path/to/_file.js.tpl`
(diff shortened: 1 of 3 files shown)


alexander-akait (dismissed):
/cc @webpack/cli-team

packages/generators/src/addon-generator.ts
```diff
@@ -4,6 +4,16 @@ import Generator from 'yeoman-generator';
 
 import { List } from './utils/scaffold-utils';
 
+// Helper to get the template-directory content
+
+const getFiles = (dir) => {
+    return fs.readdirSync(dir).reduce((list, file) => {
+        const filePath = path.join(dir, file);
+        const isDir = fs.statSync(filePath).isDirectory();
```
alexander-akait, line 12: Maybe we need handle this case better, in some edge cases here can be exception, for example permissions

jamesgeorge007: How do you see it? In fact, this is used by projects like [vite](https://github.com/vitejs/vite)
https://github.com/vitejs/vite/blob/978d991293f5b8e1a4197ac4e3aee4fa2e838a88/packages/create-app/index.js#L318

alexander-akait: Yep, used, but not handle exceptions, just add `try/catch` around `getFiles`, i.e. `try { getFiles() } catch (error) { logger.error(error); process.exit(2) }`

packages/generators/src/addon-generator.ts
```diff
@@ -1,5 +1,6 @@
 import fs from 'fs';
 import path from 'path';
+import createLogger from 'yeoman-environment/lib/util/log';
```
alexander-akait, line 3: Let's avoid it,w e ahve built-in logger^^

jamesgeorge007: Oops

alexander-akait (approved):

snitin315 (approved):

PR #2684, opened 2021-05-01
chore: remove `scaffolding` from keywords

snitin315:
chore

NA
NA
**Summary**

Remove the `scaffolding` keyword.

No

**Other information**
No

### DIFF
--- a/packages/webpack-cli/package.json
+++ b/packages/webpack-cli/package.json
@@ -18,7 +18,6 @@
   "keywords": [
     "webpack",
     "cli",
-    "scaffolding",
     "module",
     "bundler",
     "web"


packages/webpack-cli/package.json
```diff
@@ -18,7 +18,6 @@
   "keywords": [
     "webpack",
     "cli",
-    "scaffolding",
```
anshumanv, line 21: we actually do scaffold things using init/loader/plugins, just not with webpack-scaffold

evenstensberg: I agree here

noharashinchan (approved):

packages/webpack-cli/package.json
snitin315: Hmm, If we think closely it's actually another package `@webpack-cli/generators` for the scaffolding stuff. But I think we are good here. I don't have a strong opinion on this one.

rishabh3112: scaffolding is meant to be done through CLI only, so we should keep this keyword and should not be removed.

PR #2687, opened 2021-05-03
chore(generators): send warning regarding invalid template to stderr

jamesgeorge007:
chore

Update test suite.

N/A

**Summary**
https://github.com/webpack/webpack-cli/blob/af6c02b130a35b62b94542680aa7ea41561e77d7/packages/generators/src/addon-generator.ts#L60

Nope

**Other information**
N/A

### DIFF
--- a/packages/generators/src/addon-generator.ts
+++ b/packages/generators/src/addon-generator.ts
@@ -1,4 +1,3 @@
-import { yellow } from 'colorette';
 import fs from 'fs';
 import path from 'path';
 import Generator from 'yeoman-generator';
@@ -58,7 +57,7 @@ const addonGenerator = (
 
         public async prompting(): Promise<void> {
             if (!this.supportedTemplates.includes(this.template)) {
-                this.utils.logger.log(`${yellow(`⚠ ${this.template} is not a valid template, please select one from below`)}`);
+                this.utils.logger.warn(`⚠ ${this.template} is not a valid template, please select one from below`);
 
                 const { selectedTemplate } = await List(
                     this,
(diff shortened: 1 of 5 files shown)


packages/generators/src/addon-generator.ts
```diff
@@ -57,7 +58,7 @@ const addonGenerator = (
 
         public async prompting(): Promise<void> {
             if (!this.supportedTemplates.includes(this.template)) {
-                this.utils.logger.warn(`⚠ ${this.template} is not a valid template, please select one from below`);
+                this.utils.logger.log(`${yellow(`⚠ ${this.template} is not a valid template, please select one from below`)}`);
```
alexander-akait, line 61: hm, why we need change warn on log?

jamesgeorge007: We're writing to `stdout` for the case of `init`.
https://github.com/webpack/webpack-cli/blob/af6c02b130a35b62b94542680aa7ea41561e77d7/packages/generators/src/init-generator.ts#L61

We follow a similar implementation for the `loader` and `plugin` subcommands. However, using `logger.warn()` writes to `stdin`.
https://github.com/webpack/webpack-cli/blob/af6c02b130a35b62b94542680aa7ea41561e77d7/packages/generators/src/addon-generator.ts#L60

evenstensberg: writing to stderr is better

alexander-akait (approved):

ddsmatt (commented):
Lots of changes without explicit approval. Thanks.

PR #2667, opened 2021-04-26
fix: double require problem

rishabh3112:
**What kind of change does this PR introduce?**
Fix

**Did you add tests for your changes?**
N/A

**If relevant, did you update the documentation?**
N/A

**Summary**
Fix double require problem. Described as in https://github.com/webpack/webpack-dev-server/pull/3219

**Does this PR introduce a breaking change?**
No

**Other information**
No

PR #2716, opened 2021-05-13
fix: module not found problem after ask installation

alexander-akait:

bgufix

hard to test, maybe we can use smoke tests for this, I will look

No need

**Summary**

https://github.com/webpack/webpack-cli/pull/2667

No

**Other information**

/cc @rishabh3112 I think it is more clear

### DIFF
--- a/packages/webpack-cli/lib/utils/package-exists.js
+++ b/packages/webpack-cli/lib/utils/package-exists.js
@@ -1,9 +1,14 @@
+const childProcess = require('child_process');
+
+// Node.js has cache on `require.resolve` and second `require` throws the error with `MODULE_NOT_FOUND` code
 function packageExists(packageName) {
-    try {
-        return require.resolve(packageName);
-    } catch (error) {
-        return false;
-    }
+    const nodePath = process.execPath;
+    const nodeOptions = process.execArgv.filter((arg) => !arg.startsWith('--inspect'));
+    const command = `require.resolve(${JSON.stringify(packageName)})`;
+
+    const result = childProcess.spawnSync(nodePath, [...nodeOptions, '-e', command], { encoding: 'utf8', maxBuffer: 1000 * 1000 * 100 });
+
+    return result.status === 0;
 }
 
 module.exports = packageExists;
--- a/packages/webpack-cli/lib/utils/prompt-installation.js
+++ b/packages/webpack-cli/lib/utils/prompt-installation.js
@@ -19,12 +19,7 @@ async function promptInstallation(packageName, preMessage) {
     }
 
     // yarn uses 'add' command, rest npm and pnpm both use 'install'
-    const commandToBeRun = `${packageManager} ${[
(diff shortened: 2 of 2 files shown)


packages/webpack-cli/lib/utils/prompt-installation.js
```diff
@@ -45,7 +50,7 @@ async function promptInstallation(packageName, preMessage) {
             process.exit(2);
         }
 
-        return utils.packageExists(packageName);
+        return packageName;
```
alexander-akait, line 48: I think we don't need check here, because `require` fails itself if we will have problem. 

Before we have `require.resolve(package)` and then `require()`, it doesn't make sense, because `require` call `require.resolve` internally

packages/webpack-cli/lib/utils/package-exists.js
```diff
@@ -1,9 +1,14 @@
...
-    } catch (error) {
-        return false;
-    }
+    const nodePath = process.execPath;
+    const nodeOptions = process.execArgv.filter((arg) => !arg.startsWith('--inspect'));
+    const command = `require.resolve(${JSON.stringify(packageName)})`;
+
+    const result = childProcess.spawnSync(nodePath, [...nodeOptions, '-e', command], { encoding: 'utf8', maxBuffer: 1000 * 1000 * 100 });
+
+    return result.status === 0;
```
alexander-akait, line 11: I think we don't need check error message/path/something else, if it fail we should always try to install, even if something goes wrong we will catch it in `require` anyway

rishabh3112: Yeah you are right. This seems much better check 😄

packages/webpack-cli/lib/utils/prompt-installation.js
rishabh3112: Yes , we can skip it as we reach here after successful installation only.

P.S.
I guess this was added here to make sure `webpack` is avaliable before running CLI (talking about code at `bin/cli.js`) and as it is not being checked in `then` block.

alexander-akait: I see

evenstensberg (approved):

packages/webpack-cli/lib/utils/package-exists.js
alexander-akait: Here one bottleneck - perf, before it was fast, no it is not fast due external process, maybe we can cache result, need time to think

sokra: Yeah, I also think this isn't the best solution performance-wise.

You can create a very simple lookup solution:

``` js
let dir = __dirname;
do {
  try {
    if(fs.statSync(path.join(dir, "node_modules", packageName)).isDirectory()) return true;
  } catch(e) { }
} while(dir !== (dir = path.dirname(dir)));
return false;
```

PR #2718, opened 2021-05-14
refactor: loading packages

alexander-akait:

refactor

Existing

No

**Summary**

Unify code, we should avoid using `require` for external packages due Node.js is out and we some modules can migrate on ES. Also reduce duplicate of code and error handling

No

**Other information**

No

### DIFF
--- a/packages/serve/src/startDevServer.ts
+++ b/packages/serve/src/startDevServer.ts
@@ -27,7 +27,7 @@ export default async function startDevServer(
         // eslint-disable-next-line node/no-extraneous-require
         devServerVersion = require("webpack-dev-server/package.json").version;
         // eslint-disable-next-line node/no-extraneous-require
-        Server = require("webpack-dev-server/lib/Server");
+        Server = require("webpack-dev-server");
     } catch (err) {
         logger.error(
             `You need to install 'webpack-dev-server' for running 'webpack serve'.\n${err}`,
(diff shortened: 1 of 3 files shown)


snitin315 (approved):

anshumanv (dismissed):
lgtm

alexander-akait:
Also fix the random failed test, we should not use `__dirname` :smile:

packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -780,17 +843,13 @@ class WebpackCLI {
                 let loadedCommand;
 
                 try {
-                    loadedCommand = require(pkg);
+                    loadedCommand = await this.tryRequireThenImport(pkg, false);
```
evenstensberg, line 846: why false here and default true in function?

alexander-akait: Because we ignore errors in some case, we have tests on this

packages/serve/src/startDevServer.ts
```diff
@@ -27,7 +27,7 @@ export default async function startDevServer(
         // eslint-disable-next-line node/no-extraneous-require
         devServerVersion = require("webpack-dev-server/package.json").version;
         // eslint-disable-next-line node/no-extraneous-require
-        Server = require("webpack-dev-server/lib/Server");
+        Server = require("webpack-dev-server");
```
alexander-akait, line 30: We should use public API

PR #2743, opened 2021-06-01
chore: update jest to v27

snitin315:
chore

NA
NA
**Summary**

Update jest to v27.

No

**Other information**
No


jest.config.js
```diff
@@ -9,6 +9,7 @@ const ignorePattern =
 module.exports = {
     testPathIgnorePatterns: ignorePattern,
     testEnvironment: "node",
+    testRunner: "jest-jasmine2",
```
snitin315, line 12: As per https://jestjs.io/blog/2021/05/25/jest-27#flipping-defaults

anshumanv: why? We should use `jest-circus`

snitin315: Actually we rely on jasmine, maybe we can update I will look into it.

https://github.com/webpack/webpack-cli/blob/7dad5817ca3a39a5782411d3629512ac7dc161a5/setupTest.js#L5

anshumanv: oh yep, we need to migrate as jasmine will soon be deprecated

alexander-akait: @snitin315 Let's merge it as is and do `jest-circus` migration in the separate PR

snitin315: Will do 👍

anshumanv (approved):

PR #2747, opened 2021-06-02
chore: migrate to `jest-circus`

snitin315:
chore

No
No
**Summary**

Migrate to `jest-circus` which is the default test runner as of `jest` v27

No

**Other information**
No

### DIFF
--- a/test/api/CLI.test.js
+++ b/test/api/CLI.test.js
@@ -8,19 +8,17 @@ describe("CLI API", () => {
     });
 
     describe("makeCommand", () => {
-        it("should make command", async (done) => {
+        it("should make command", async () => {
             cli.program.commands = [];
 
             const command = await cli.makeCommand({ name: "command" }, [], (options) => {
                 expect(options).toEqual({});
-
-                done();
             });
 
             command.parseAsync([], { from: "user" });
         });
 
-        it("should make command with Boolean option by default", async (done) => {
+        it("should make command with Boolean option by default", async () => {
             cli.program.commands = [];
 
             const command = await cli.makeCommand(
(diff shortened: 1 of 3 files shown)


anshumanv (approved):
nice nice

test/api/CLI.test.js
```diff
@@ -8,19 +8,17 @@ describe("CLI API", () => {
     });
 
     describe("makeCommand", () => {
-        it("should make command", async (done) => {
+        it("should make command", async () => {
```
alexander-akait, line 11: I think we should return `Promise` here based on https://jestjs.io/blog/2021/05/25/jest-27#features-coming-with-breaking-changes

snitin315: I think we already return a void promise here as per the error - 

Screenshot 2021-06-02 at 6 08 02 PM

alexander-akait: We should return https://github.com/webpack/webpack-cli/blob/03c5030537dd47e90abbca1af133f7d8c8176a8d/test/api/CLI.test.js#L18, because now we nothing return, other solution is using https://jestjs.io/ru/docs/expect#expectassertionsnumber

snitin315: Updated with `expect.assertions()`

test/api/CLI.test.js
```diff
@@ -59,6 +63,8 @@ describe("CLI API", () => {
             );
 
             command.parseAsync(["--boolean"], { from: "user" });
+
+            expect.assertions(1);
```
alexander-akait, line 67: It should be on top

alexander-akait (approved):

alexander-akait:
Feel free to merge after CI will be green, thanks!

JAREDARVIND:
can you gave me a hotspot

PR #2761, opened 2021-06-04
fix: not found module after ask installation

alexander-akait:

bugfix

Hard to add, manually tested

No need

**Summary**

fixes #2716

No

**Other information**

@webpack/cli-team 

- we need backport this solution, in `webpack-dev-server` and `webpack` itself, just replace `require.resolve` on this function
- also we need small refactor, `promptInstallation` should not return something, only install package

### DIFF
--- a/packages/webpack-cli/lib/utils/prompt-installation.js
+++ b/packages/webpack-cli/lib/utils/prompt-installation.js
@@ -22,6 +22,7 @@ async function promptInstallation(packageName, preMessage) {
     const commandToBeRun = `${packageManager} ${[
         packageManager === "yarn" ? "add" : "install",
         "-D",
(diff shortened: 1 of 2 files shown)


packages/webpack-cli/lib/utils/prompt-installation.js
```diff
@@ -51,7 +51,7 @@ async function promptInstallation(packageName, preMessage) {
             process.exit(2);
         }
 
-        return utils.packageExists(packageName);
+        return packageName;
```
alexander-akait, line 54: No need to check package on exists, because if it was failed we will exit with `2` code above

anshumanv:
> we need backport this solution, in webpack-dev-server and webpack itself, just replace require.resolve on this function

👍

Issue #15, opened 2016-12-24
Feature: Watch for webpack config changes in watch mode


bebraw:
**I'm submitting a feature request**

**Webpack version:**
2.x

**Please tell us about your environment:**
OSX 10.x

**Current behavior:**
I'm not sure if this was discussed before, but it's one of the pain point for me in webpack workflow.

When config / config's dep is updated — hot reloadable server (or watch mode) must be restarted manually. Usually this hot reloadable server is not stand-alone process, but part of the bigger task or even lives inside docker container. And in case of config related change — the only way to pick this up is to restart everything. 

**Expected/desired behavior:**

It'd be cool if webpack could track such changes and rebuild everything from scratch w/o killing current process.

**What is the motivation / use case for changing the behavior?**

It'll make easier:
- work on webpack config setup
- work w/ loader's configs, eg postcss plugin's options (js vars etc)

**Moved from https://github.com/webpack/webpack/issues/3153**

DanielaValero:
@ev1stensberg, @pksjce what you guys think, should this be part of our roadmap to v1 ?

bebraw:
Minimum - document how to achieve it outside of the cli. I guess it depends on the overall goals and priority. I expect v1 would be more or less what it is right now and you can build more features in v1.1 and so on.

DanielaValero:
I'd expect the same. So far, did not have time to look deeper into the already available options, but if this is in there, for sure will be part of v1. Otherwise, I'd agree with you by adding this to a v1.1 milestone

ev1stensberg:
Fixed #23

nmschulte:
@ev1stensberg, does this apply to hot-module-reloading / webpack-dev-server?  I'm looking for a solution to reset Webpack so it'll pick up the new configuration, and have HMR communicate this change appropriately (likely, a "you must manually refresh to pick up latest changes" type message).

How do we enable this in that scenario; is it as simple as watching the files that modify Webpack configuration?  How does one use that without causing all watched files to incite a Webpack reset?

ev1stensberg:
The resave flag applies to HMR, indeed. Let me get back to you about your issues later this week :)

yacinehmito:
@ev1stensberg How does #24 solves this issue? It seems like it's not related.

yacinehmito:
IMHO we should either reopen this issue or https://github.com/webpack/webpack/issues/3153

ev1stensberg:
Pinging @bebraw for advice

[...]

PR #2797, opened 2021-06-24
feat: watch mode config change

evenstensberg:
Fixes https://github.com/webpack/webpack-cli/issues/15

No
N/A
**Summary**
Need to check if CLIplugin is added to the config, otherwise there will be a memory leak

No
**Other information**
N/A

### DIFF
--- a/packages/webpack-cli/lib/webpack-cli.js
+++ b/packages/webpack-cli/lib/webpack-cli.js
@@ -1993,7 +1993,12 @@ class WebpackCLI {
             if (!configOptions.plugins) {
                 configOptions.plugins = [];
             }
-
+            const alreadyHasPLugin = configOptions.plugins.find(
+                (plugin) => plugin instanceof CLIPlugin,
+            );
+            if (alreadyHasPLugin) {
+                return configOptions;
+            }
             configOptions.plugins.unshift(
                 new CLIPlugin({
                     configPath: config.path.get(configOptions),
@@ -2039,7 +2044,7 @@ class WebpackCLI {
         let config = await this.resolveConfig(options);
         config = await this.applyOptions(config, options);
         config = await this.applyCLIPlugin(config, options);
-
+        console.log(config);
         let compiler;
 
         try {
@@ -2075,7 +2080,6 @@ class WebpackCLI {
     }
 
     async buildCommand(options, isWatchCommand) {
-        // eslint-disable-next-line prefer-const
         let compiler;
         let createJsonStringifyStream;
 
--- a/packages/webpack-cli/lib/webpack-cli.js
+++ b/packages/webpack-cli/lib/webpack-cli.js
@@ -24,13 +24,15 @@ class WebpackCLI {
                 ),
         });
     }
-
     async tryRequireThenImport(module, handleError = true) {
         let result;
 
         try {
+            delete require.cache[module];
             result = require(module);
+            console.log(result);
         } catch (error) {
+            console.log(error, "HEY");
             let previousModuleCompile;
 
             // TODO Workaround https://github.com/zertosh/v8-compile-cache/issues/30
@@ -67,7 +69,6 @@ class WebpackCLI {
                 throw error;
             }
         }
-
         // For babel/typescript
         if (result.default) {
             result = result.default;
@@ -1545,10 +1546,8 @@ class WebpackCLI {
 
                 process.exit(2);
             }
-
             return { options, path: configPath };
         };
-
         const evaluateConfig = async (loadedConfig, argv) => {
             const isMultiCompiler = Array.isArray(loadedConfig.options);
             const config = isMultiCompiler ? loadedConfig.options : [loadedConfig.options];
@@ -1581,7 +1580,7 @@ class WebpackCLI {
             return loadedConfig;
         };
 
-        const config = { options: {}, path: new WeakMap() };
+        const config = { options: {}, path: new Map() };
 
         if (options.config && options.config.length > 0) {
             const evaluatedConfigs = await Promise.all(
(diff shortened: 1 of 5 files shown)


alexander-akait (changes requested):
Wrong solution

packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -1581,7 +1580,7 @@ class WebpackCLI {
             return loadedConfig;
         };
 
-        const config = { options: {}, path: new WeakMap() };
+        const config = { options: {}, path: new Map() };
```
alexander-akait, line 1583: Leaking

packages/webpack-cli/package.json
```diff
@@ -33,13 +33,15 @@
...
     "@webpack-cli/serve": "^1.5.1",
+    "async-require": "^1.2.2",
     "colorette": "^1.2.1",
     "commander": "^7.0.0",
     "execa": "^5.0.0",
     "fastest-levenshtein": "^1.0.12",
     "import-local": "^3.0.2",
     "interpret": "^2.2.0",
     "rechoir": "^0.7.0",
+    "require.async": "^0.1.1",
```
alexander-akait, line 44: No need this

packages/webpack-cli/package.json
```diff
@@ -33,13 +33,15 @@
     "@webpack-cli/configtest": "^1.0.4",
     "@webpack-cli/info": "^1.3.0",
     "@webpack-cli/serve": "^1.5.1",
+    "async-require": "^1.2.2",
```
alexander-akait, line 36: Bad idea, we have `import().then()`

evenstensberg:
PTAL

packages/webpack-cli/package.json
evenstensberg: yeah this is outdated

packages/webpack-cli/package.json
evenstensberg: resolved

packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -24,11 +24,11 @@ class WebpackCLI {
                 ),
         });
     }
-
     async tryRequireThenImport(module, handleError = true) {
         let result;
 
         try {
+            delete require.cache[module];
```
alexander-akait, line 31: Unsafe and very bad for perf

packages/webpack-cli/lib/webpack-cli.js
alexander-akait: Look at CLI plugin

packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -1998,7 +1993,12 @@ class WebpackCLI {
             if (!configOptions.plugins) {
                 configOptions.plugins = [];
             }
-
+            const alreadyHasPLugin = configOptions.plugins.find(
+                (plugin) => plugin instanceof CLIPlugin,
+            );
+            if (alreadyHasPLugin) {
+                return configOptions;
+            }
```
alexander-akait, line 2001: Rerun `applyOptions` is very bad idea, it has very bad performance and it is not safe too

packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -2171,7 +2168,15 @@ class WebpackCLI {
         }
 
         compiler = await this.createCompiler(options, callback);
-
+        if (options.watch) {
+            // eslint-disable-next-line
+            for (const [_, filePath] of compiler.config.path.entries()) {
```
alexander-akait, line 2173: You can do the same using `WeakMap`

packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -2171,7 +2168,15 @@ class WebpackCLI {
...
+                // eslint-disable-next-line
+                fs.watchFile(filePath, async () => {
+                    compiler = await this.createCompiler(options, callback);
```
alexander-akait, line 2176: It should rerun `commander`, not create compiler

alexander-akait (commented):
Ideally we should not rerun command whole, webpack can revalidate own options and run new compilation(s), everything webpack needs to know is path to configuration(s) for each compiler, so this task should be done on webpack side, not here

packages/webpack-cli/lib/webpack-cli.js
evenstensberg: its cache for one file, perf wont change that much because of this

packages/webpack-cli/lib/webpack-cli.js
evenstensberg: cant iterate over a weakmap

evenstensberg:
@alexander-akait you have the source, maybe you can have a look at it?

packages/webpack-cli/lib/webpack-cli.js
alexander-akait: You have configuration keys

evenstensberg:
Some points:

1. We only have to worry about the compiler re-rednering after a config change is noticed
2. We only have access to the configuration at this level


Due this, I think this is the right level of abstraction. Since the CLI is the one that reads a config, I think this is where we naturally add this feature. I don't understand why webpack itself should re-validate a config, core should only be worried about doing a compilation.

WDYT?

alexander-akait:
@evenstensberg Because only webpack knows about dependencies

evenstensberg:
We just want config changes, not dependency changes. Watchmode + watch a config file

alexander-akait:
In this case we don't look at `import`/`require` inside `webpack.config.js`/package.json and etc, so reload will not work for many cases

evenstensberg:
The watch API will sniff those changes if the text inside  the config changes. this is what we want

alexander-akait:
We should watch dependencies too like `package.json`/`import`/`require` and etc

evenstensberg:
No need, if we refer to the original issue

alexander-akait:
It is require to right implementation

evenstensberg:
PTAL. Uses https://github.com/evenstensberg/watch-traverse

alexander-akait:
I wanted to close it, because it is invalid, we need to improve it on webpack side, we need:
- move `buildDependecies` from `options.cache.buildDependencies` to top options (it will be useful for any configuration, not only for webpack itself)
- implement `watchRun` hook and get files from `compilation.modifiedFiles`, if we have confiugration file there restart our CLI

PR #2809, opened 2021-06-29
chore(generators): do not prompt if multiple installers are not available

jamesgeorge007:
chore

N/A

**Summary**

### Before
before

### After
after

No

**Other information**
N/A


alexander-akait (approved):
Good idea :+1:

snitin315 (approved):

anshumanv (approved):

alexander-akait:
@jamesgeorge007 Also I think we should not ask question about installation when you run `npx webpack init`, because I already run init and want to init

PR #2815, opened 2021-07-04
feat: install required packages for loading command automatically

snitin315:
feature

WIP in tests
NA
**Summary**

https://github.com/webpack/webpack-cli/pull/2809#issuecomment-870755830

No

**Other information**
No

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -55,7 +55,7 @@ jobs:
       matrix:
         # Debug
         os: [windows-latest, macos-latest]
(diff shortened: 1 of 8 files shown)


snitin315:
smoke tests passing locally 😞.

alexander-akait (commented):
hm, do you think it is better do for all commands? I think it will be good for init only, but I am fine with install all by default

snitin315:
Yes, we should do it for all.

alexander-akait:
@snitin315 Do we need this, I just checked and we don't ask about installation when I use `webpack init`

PR #2842, opened 2021-07-20
refactor(generators): leverage helpers

jamesgeorge007:
refactoring

N/A

N/A

**Summary**
Move common logic to helper functions.

No

**Other information**
N/A


packages/generators/src/utils/helpers.ts
```diff
@@ -20,3 +22,44 @@ export function toUpperCamelCase(str: string): string {
         .map((x) => x.slice(0, 1).toUpperCase() + x.slice(1).toLowerCase())
         .join("");
 }
+
+export async function getInstaller(): Promise<string> {
+    const installers = this.utils.getAvailableInstallers();
+
+    if (installers.length === 1) {
+        return installers[0];
```
alexander-akait, line 30: Let's add test on this line

test/api/generators/helpers.test.js
```diff
@@ -0,0 +1,73 @@
...
+        List.mockReturnValue({ selectedTemplate: "default" });
+
+        const loggerMock = jest.spyOn(logger, "warn").mockImplementation(() => {});
+
+        // Invoke the helper function`
+        const template = await getTemplate.call(context);
+        expect(template).toBe("default");
+        expect(loggerMock).toHaveBeenCalled();
+    });
+});
```
alexander-akait, line 73: We should avoid using unit tests, these tests are not tested real cases, please convert them to integration tests with real scenarios

jamesgeorge007: The only scenario for which we're missing out on an integration test in this context is for the case in which the prompt doesn't show up when multiple installers are not available. An attempt to write one won't go well, as both `npm` and `yarn` are available in the CI environment. I think unit tests better serve the purpose here.

alexander-akait: hm, I think there is a method to disable npm/yarn in CI, but yep, it is not easy, okay, let's merge it as is, maybe in future we will improve this

Issue #2832, opened 2021-07-14
feature: add init templates for popular webpack use cases


rishabh3112:
**Is your feature request related to a problem? Please describe.**
We recently updated the init architecture. This makes it easier to add template for initializing new projects.

**Describe the solution you'd like**
Here are few high priority ones:
- [ ] React/Preact
- [ ] Vue
- [ ] Svelte
- [ ] Node

Feel free to add more suggestions below 😄 

**Describe alternatives you've considered**
N/A

**Additional context**
#2430

rishabh3112:
Will be doing the React/Preact first!

jamesgeorge007:
Hi @rishabh3112, I'd love to join forces. Can I take up Svelte?

alexander-akait:
Yes, feel free to send any PR

rishabh3112:
@jamesgeorge007 sure, thank you for helping out! Sorry for late reply, was busy with other stuffs and I missed this issue some how.

rishabh3112:
bump

PR #2859, opened 2021-07-27
feat(init-generator): add svelte template

jamesgeorge007:
feature

Yes

N/A

**Summary**
This PR aims at adding a `svelte` template for `init-generator` #2832 

No

**Other information**
N/A


packages/generators/init-template/svelte/webpack.config.js
```diff
@@ -0,0 +1,62 @@
...
+                        compilerOptions: {
+                            dev: !isProduction,
+                        },
+                        emitCss: isProduction,
+                        hotReload: !isProduction,
+                    },
+                },
+            },
+            {
+                test: /\.css$/,
```
rishabh3112, line 39: we should also add support for postcss, sass, etc here.

alexander-akait: I agree, ideally we should extend default template and add custom loaders/plugins

evenstensberg (commented):
We don’t need this. CWA (create webpack app) is going to be a standalone tool soon. Gsoc project.

PR #2862, opened 2021-07-29
feat: react template

rishabh3112:
Refer #2832 

WIP

Will do

**Summary**
Adds a react template in init command.

No

**Other information**
No


rishabh3112:
# Status 
## TODO
- [x] Decide good defaults (dev-server, asset-modules and  html-webpack-plugin) + directory structure
- [ ] Allow for ES6 as well as Typescript
- [ ] Allow for all css options (except `none`) as in default template
- [ ] Design a good looking first page

## Opinions
- [ ] style-components or css-modules or both or ask question for this?

alexander-akait (commented):
Sam problem for `css` here... Any ideas how we can improve our generator?

rishabh3112:
> Sam problem for css here... Any ideas how we can improve our generator?

Yeah, I have a solution in mind. Will do some more research on it and create a pull request.

rishabh3112:
ejs partials is is not feasible, as one may need to provide different options for a loader/plugin which are already supplied by the default one.

Only one solution in my mind now is writing babel AST transforms ( like for adding plugins/loader ) which can be used by other templates to update the configuration generated by default one. Let me know if you have some better suggestions 😅

alexander-akait:
@rishabh3112 ideally will be great to avoid using babel AST transforms, maybe we can create `hook` in our generator and use it?

rishabh3112:
@alexander-akait the problem is not with extending the default generator (this can be done without implementing a hook like interface for it as well easily), but is with templating the webpack configuration file for it.

alexander-akait:
@rishabh3112 maybe we can add to default template `addional.loader`/`addional.plugin`/etc and set to empty on default template, but fill out on custom templates?

alexander-akait:
Other solution, just add more questions to our default generator and do not complex logic

rishabh3112:
> maybe we can add to default template addional.loader/addional.plugin/etc and set to empty on default template, but fill out on custom templates?

This is what I wanted to try using the ejs partials actually. This would create duplication of options if any two options clashes.

> Other solution, just add more questions to our default generator and do not complex logic

Yeah, but it would not scale well with frameworks/libraries I am afraid, the configuration template is already to big, it would become a lot bigger and complex with addition of support for each new framework/library. Reason behind this is one need to write code for all possible scenarios in ejs template itself.

alexander-akait:
hm, I will look at this in near future and will think about better solution

rishabh3112:
Finally I decided to do a tradeoff and duplicate the default's webpack config template as base for react one.
It turned out be great as it kept the integration pretty simple and also will make it easier to create web based templates ahead!

rishabh3112:
@webpack/cli-team should we add support for css modules by default?

rishabh3112:
Starting rebasing this.

rishabh3112:
@alexander-akait this template is functional and works. we should land this for now and improve later on.

rishabh3112:
We can merge this now I think @webpack/cli-team

packages/generators/src/handlers/react.ts
```diff
@@ -0,0 +1,80 @@
...
+      break;
+    case "Stylus":
+      files.push("./src/styles/global.styl");
+      break;
+  }
+
+  for (const file of files) {
+    self.fs.copyTpl(resolveFile(file + ".tpl"), self.destinationPath(file as string), self.answers);
+  }
+}
```
alexander-akait, line 80: We really should avoid duplication of code

rishabh3112: Well, these are two separate templates for very different use-cases and should be independent from each other. Coupling them for the sake of it shouldn't be a way to go, would lead to way more bugs in future.

snitin315: I am okay with separate templates, we can always make improvements based on user needs and feedback.

rishabh3112: /cc @alexander-akait

alexander-akait: I am okay, if other developers will support it :+1: we need more reviews

rishabh3112: /cc @anshumanv

anshumanv: I'm okay with different handler because different templates will have different options, in fact I like how the default template is being reused here to minimize duplication.

anshumanv (commented):
Can we have a simple test to ensure the generation works end to end?

rishabh3112:
> Can we have a simple test to ensure the generation works end to end?

Will do, once approach will be finalized by us all.

packages/generators/init-template/react/webpack.config.js.tpl
```diff
@@ -0,0 +1,105 @@
...
+
+            // Add your rules for custom modules here
+            // Learn more about loaders from https://webpack.js.org/loaders/
+        ],
+    },
+    resolve: {
+        alias: {
+            "@": path.resolve(__dirname, "./src/"),
+        },<% if (langType == "Typescript") {%>
+        extensions: ['.tsx', '.ts', '.js'],<% } %>
```
alexander-akait, line 88: Let's use `['.tsx', '.ts', '.jsx', '.js', '...']`, we should use `...` everythere

rishabh3112: Okay 👍🏽

packages/generators/src/handlers/react.ts
rishabh3112: Closing this discussion now. Thanks everyone.

packages/generators/init-template/react/webpack.config.js.tpl
```diff
@@ -85,7 +85,7 @@ const config = {
         alias: {
             "@": path.resolve(__dirname, "./src/"),
         },<% if (langType == "Typescript") {%>
-        extensions: ['.tsx', '.ts', '.js'],<% } %>
+        extensions: ['.tsx', '.ts', '.js', '...'],<% } %>
```
alexander-akait, line 88: Let's add `.jsx` here too

packages/generators/init-template/default/webpack.configjs.tpl
```diff
@@ -76,7 +76,7 @@ const config = {
         ],
     },<% if (langType == "Typescript") {%>
     resolve: {
-        extensions: ['.tsx', '.ts', '.js'],
+        extensions: ['.tsx', '.ts', '.js', '...'],
```
alexander-akait, line 79: > Let's add .jsx here too

packages/generators/init-template/react/webpack.config.js.tpl
rishabh3112: Okay 👍🏽

packages/generators/init-template/default/webpack.configjs.tpl
rishabh3112: Done

packages/generators/init-template/react/webpack.config.js.tpl
```diff
@@ -85,7 +85,7 @@ const config = {
         alias: {
             "@": path.resolve(__dirname, "./src/"),
         },<% if (langType == "Typescript") {%>
-        extensions: ['.tsx', '.ts', '.js', '...'],<% } %>
+        extensions: ['.tsx', '.ts', '.js', '.jsx. '...'],<% } %>
```
alexander-akait, line 88: Please put it before `js`, jsx should be prefered

packages/generators/src/handlers/default.ts
```diff
@@ -16,15 +16,17 @@ export async function questions(
   self: CustomGenerator,
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   Question: Record<string, any>,
+  forceSelection = false,
+  isSkip: Record<string, boolean> = {},
```
alexander-akait, line 20: Can you describe why we change it?

rishabh3112: So that we can reuse `default` template handler in other we based templates as base.

alexander-akait: oh, I see, maybe we can union two variables in one?

packages/generators/src/handlers/default.ts
```diff
@@ -10,21 +10,27 @@ const resolveFile = (file: string): string => {
...
  */
 
 export async function questions(
   self: CustomGenerator,
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   Question: Record<string, any>,
+  config: { forceSelection: boolean; skip: Record<string, boolean> } = {
+    forceSelection: false,
+    skip: {},
+  },
```
alexander-akait, line 23: I mean we don't need `forceSelection`, you can pass answers directly and avoid ask question if you pass answers, example:
```
answers: { 
  langType: "ES6",
  devServer: true,
} 
```

and etc

rishabh3112: Variable just removes 'none' as possible answer, user an still select typescript for langType for example.

rishabh3112: Yup, thats the point. But in default template it does makes sense. So, therefore need of this variable, so that we can reuse default handler's questions.

alexander-akait: If something in `answers`, it is already forced, no need to do complex and confusing logic based on different variables

rishabh3112: Again saying,  answer for a question is not being forced, we are only enforcing that answer can't be `none` (It can be anything other than `none`) and introducing the variable is the simplest possible solution to it, not a complicating one.

alexander-akait: Why we should ask questions, if we know what we want?

rishabh3112: We don't know!
That is why we are asking question. We just know it can't be `none`.

alexander-akait: Why we need skip in this case? I really don't understand your logic

rishabh3112: Hmm it is simple, please answer this question.

I want to ask user whether they want to use ES6 or Typescript. So why should I skip this question?

alexander-akait: Let's use better names and structures - we can add `override: { langType: {  variants: [...], default: 'default', force: true } }`, it will allow to override whole questions

rishabh3112: Okay 👍🏽

rishabh3112: @alexander-akait Updated it, I guess new change struck nice balance between simplicity and verbosity for configuring it. wdyt?

alexander-akait: Yes, now we have simple and flexible logic

alexander-akait (approved):

alexander-akait:
Let's fix lint

rishabh3112:
> Let's fix lint

Okay, on it.

packages/generators/src/handlers/default.ts
rishabh3112: Closing this.

alexander-akait (dismissed):

alexander-akait:
@rishabh3112 I will merge it after types, I will finish it  - today/tomorrow

rishabh3112:
> @rishabh3112 I will merge it after types, I will finish it in near future - today/tomorrow

Thanks, looking forward to it.

rishabh3112:
friendly ping @alexander-akait

alexander-akait (dismissed):
/cc @webpack/cli-team

alexander-akait:
Let'sm fix lint

rishabh3112:
> Let's fix lint

Yeah sure, will do as soon as I get time.

rishabh3112:
> Let's fix lint

done

snitin315 (approved):
💯

alexander-akait (approved):

alexander-akait:
Thanks

PR #2868, opened 2021-08-02
feat: sharabale configs

snitin315:
FEATURE

yes, more tests WIP
WIP
**Summary**
Fixes  https://github.com/webpack/webpack-cli/issues/2748

No

**Other information**

WIP in covering more edge cases with multiple nested configs and multi compiler mode.

### DIFF
--- a/packages/webpack-cli/lib/webpack-cli.js
+++ b/packages/webpack-cli/lib/webpack-cli.js
@@ -401,6 +401,7 @@ class WebpackCLI {
         const minimumHelpFlags = [
             "config",
             "config-name",
+            "extends",
             "merge",
             "env",
             "mode",
@@ -451,6 +452,17 @@ class WebpackCLI {
                     },
                 ],
                 description: "Merge two or more configurations using 'webpack-merge'.",
+            },
+            {
+                name: "extends",
+                alias: "e",
+                configs: [
+                    {
+                        type: "string",
+                    },
+                ],
+                multiple: true,
+                description: "Extends a webpack configuration.",
             },
             // Complex configs
             {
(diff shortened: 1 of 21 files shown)


alexander-akait (commented):
Let's add tests with `--merge` too

snitin315:
I am also implementing `--extends` flag with this.

alexander-akait:
Yep, it is great idea

packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -1691,6 +1703,35 @@ class WebpackCLI {
...
+
+                configOptions.extends.map(async (value) => {
+                    newOptions.config.push(path.resolve(value));
+                });
+
+                delete configOptions.extends;
+
+                await this.resolveConfig(newOptions);
+            }
+        });
```
alexander-akait, line 1733: We do not use `webpack-merge` logic here? it is working good

snitin315: We will also need to load the extended configurations, here we have only `paths` for extended configurations and not their `options`. 

Or am I missing something here?

alexander-akait: Ideally `--extends` means `--config=webpack.config.js --config=webpack.other.config.js --merge`, so I think we already have logic for this, for the next major release we will change logic for multiple `--config` https://github.com/webpack/webpack-cli/blob/master/packages/webpack-cli/lib/webpack-cli.js#L1619, I think we will throw an error on two and more config options without `--merge`, but we can reuse this logic for `--extends`, `--extends` like syntax sugar, also do not forget `--extends` in CLI should override `extends` on `webpack.config.js` (I think eslint works the same, but need check)

alexander-akait:
Fixed https://github.com/webpack/webpack-cli/pull/3738

PR #2908, opened 2021-08-21
chore: setup prettier as standalone

alexander-akait:

chore

No need

No need

**Summary**

Run prettier as standalone tool. Why we don't need `eslint-plugin-prettier`:

1. We already run prettier as standalone for markdown files
2. We don't need extra dependency(ies) to run prettier
3. We already run prettier before sending a PR using `husky` and `lint-staged`
4. Most of IDE supports prettier, so you can run prettier in your IDE without any problems
5. Extra errors in IDE from prettier (i.e. `eslint-plugins-prettier`) annoying during development, you should not worry about code style because we will fix it using `husky` and `lint-staged`
6. 

No

**Other information**

I change our indentation on `2`, everywhere in webpack and webpack-contrib we use `2` spaces, let's do the same, also when you have classes using `2` value is the best solution, otherwise you will have a lot of empty spaces


alexander-akait:
After merge this we need rebase our next branch, I will care about it

snitin315 (approved):
LGTM

anshumanv (approved):

PR #2911, opened 2021-08-22
feat: add alias `a` for `analyze` option

snitin315:
feature

Yes

**Summary**

add alias `a` for `analyze` option

No

**Other information**
No

### DIFF
--- a/OPTIONS.md
+++ b/OPTIONS.md
@@ -12,7 +12,7 @@ Options:
   --node-env <value>                                                                 Sets process.env.NODE_ENV to the specified value.
   -h, --hot [value]                                                                  Enables Hot Module Replacement
   --no-hot                                                                           Disables Hot Module Replacement.
-  --analyze                                                                          It invokes webpack-bundle-analyzer plugin to get bundle information.
+  -a, --analyze                                                                      It invokes webpack-bundle-analyzer plugin to get bundle information.
   --progress [value]                                                                 Print compilation progress during build.
   --prefetch <value>                                                                 Prefetch this request.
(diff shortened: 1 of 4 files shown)


evenstensberg (approved):

alexander-akait:
@snitin315 I think we should implement short alias for all flags, let's do it, it should be not hard, just put this logic in `makeOption`

snitin315:
Yep, lets do it.

rishabh3112:
> I think we should implement short alias for all flags, let's do it, it should be not hard, just put this logic in makeOption

Definitely should automate this! Although a heads up, algorithm for this should make sure flags which are already assigned an alias don't change if we add or remove flags. This is because, unstable alias will result in bad DX (a potential breaking change in every new flag addition).
Example:
`--serve` is assigned `-s`
`--save` is now added
`--save` should not be assigned `-s`, `--serve` should continue to have `-s` alias.

This should also hold true even if we rename a flag imo (eg. if `--serve` is renamed to `--dev-server`, alias should remain `-s` only).

alexander-akait:
Yes, we should respect existing short alias, but it is not hard

evenstensberg (commented):
Could you rebase?

alexander-akait:
@snitin315 Do we really need this? What do you think?

snitin315:
@alexander-akait I think alias for all options are not needed, only for popular options.

alexander-akait:
Yeah, let's close maybe I will find time to autogenerate short keys for all options, we just need to create a script and json file and generate them to avoid overlapping, but no one request it, so let's close

PR #2919, opened 2021-08-25
fix: close compiler on SIGINT

### LINKED ISSUE #2918: filesystem cache corruption on SIGINT
If webpack-cli is terminated by SIGINT while writing to the filesystem cache (e.g. in watch mode during an incremental rebuild of a project with large external source maps), the filesystem cache will be corrupted and must be erased before webpack will build again.

webpack-cli should [hook `SIGINT`](https://nodejs.org/api/process.html#process_signal_events) and await a clean exit from [compiler.close](https://webpack.js.org/api/node/#compiler-instance).

Note that the child processes spawned by e.g. `fork-ts-checker-webpack-plugin` or `thread-loader` will hear SIGINT from the parent process, so they'll need to be isolated by spawning the compiler into a detached child process.

snitin315:
fix

No
No
**Summary**

Fix #2918 

None

**Other information**
No

### DIFF
--- a/packages/serve/src/index.ts
+++ b/packages/serve/src/index.ts
@@ -132,6 +132,40 @@ class ServeCommand {
 
         const servers = [];
 
+        const stopAllServers = () => {
+          Promise.all(
+            servers.map((server) => {
+              if (typeof server.stop === "function") {
+                return server.stop();
+              }
+
+              // TODO remove in the next major release
+              return new Promise<void>((resolve) => {
+                server.close(() => {
+                  resolve();
+                });
+              });
+            }),
+          ).then(() => {
+            servers.map((server) => {
+              if (typeof server.stopCallback === "function") {
+                return server.stopCallback(() => {
+                  process.exit(0);
+                });
+              }
+
+              // TODO remove in the next major release
+              return server.close(() => {
+                process.exit(0);
+              });
+            });
+          });
+        };
+
+        process.on("SIGINT", () => {
+          stopAllServers();
+        });
+
         if (cli.needWatchStdin(compiler) || devServerCLIOptions.stdin) {
           // TODO remove in the next major release
           // Compatibility with old `stdin` option for `webpack-dev-server`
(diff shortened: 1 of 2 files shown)


packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -2125,13 +2125,18 @@ class WebpackCLI {
...
+    });
+
     const isWatch = (compiler) =>
       compiler.compilers
         ? compiler.compilers.some((compiler) => compiler.options.watch)
         : compiler.options.watch;
 
     if (isWatch(compiler) && this.needWatchStdin(compiler)) {
       process.stdin.on("end", () => {
+        compiler.close();
```
alexander-akait, line 2139: ` compiler.close(() => { process.exit(0) });`

packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -2125,13 +2125,18 @@ class WebpackCLI {
       return;
     }
 
+    process.on("SIGINT", () => {
+      compiler.close();
```
alexander-akait, line 2129: We need test on `130` code here

alexander-akait (commented):
Also let's add test on 130 code and close for serve (it is already implemented in dev server)

alexander-akait (commented):
Looks good, let's add tests for `build`/`watch`/`serve`, just enable logging and check on https://github.com/webpack/webpack/blob/main/lib/cache/PackFileCacheStrategy.js#L1263 or you can implement custom plugin and use `compiler.cache.hooks.shutdown.tapPromise()` with `{ stage: 9999 }` to ensure we will run after cache was stored

evenstensberg (approved):

alexander-akait:
I will add tests

alexander-akait:
@snitin315 We need improve logic here: we need allow to graceful close and allow immutability disable, my idea - when developer run `CTRL + C` we try to `close()`, if `CTRL + C` pressed again we close immediately, also we need handle ` const signals = ["SIGINT", "SIGTERM"];` signals, for `serve` we need disable `setupExitSignals` and move this logic here https://github.com/webpack/webpack-dev-server/blob/master/lib/Server.js#L1088 (in future major release we will remove this option, because it is wrong place for this, developer should handle closing manually)

snitin315:
Makes sense, I will update it in near future.

alexander-akait:
@snitin315 hello, let's finish, just copy logic from dev server

snitin315:
@alexander-akait sure 👍🏻

evenstensberg (changes requested):
Could you rebase?

packages/serve/src/index.ts
```diff
@@ -132,6 +132,40 @@ class ServeCommand {
 
         const servers = [];
 
+        const stopAllServers = () => {
+          Promise.all(
```
evenstensberg, line 136: I'd love to make this a variable and passing the variable to the promise expression

PR #2947, opened 2021-09-13
fix: no serve when dev-server is false

rishabh3112:
Fix

Yup

Yes

**Summary**
No dev server when `devServer: false` is supplied.

Nope

**Other information**
/cc @alexander-akait


rishabh3112:
@alexander-akait devServer is `object` in schema, should change/publish that first to support `devServer: false`.

alexander-akait:
We should return before using schema

rishabh3112:
> We should return before using schema

We should not read configuration without validating against schema ideally.

evenstensberg (approved):
All tests green. LGTM 🧑🏽‍💻

alexander-akait:
We need to revert this, because we should not exit, just run a regular compilation with a warning, I will fix it

alexander-akait:
Also `webpack build` with `devServer: false` should work too without any problems

PR #2956, opened 2021-09-18
Migrate to colorette 2.0.x

### LINKED ISSUE #2955: Migrate to new API in Colorette 2.0.x
Currently webpack-cli relies on modifying global state, which potentially results in affecting other parts of the code that might rely on Colorette as well.

https://github.com/jorgebucaran/colorette/issues/70

**Additional context**

You can see how we migrated pino-pretty to a new API here: https://github.com/pinojs/pino-pretty/pull/230

kibertoad:
fixes #2955

Update colorette to a new semver major version.

Not relevant.

Not relevant.

**Summary**

This ensures webpack-cli is staying on up-to-date dependency version, and also removes reliance on global state, which is never a good idea.

It shouldn't, although there was a change in colorette behavior: it no longer generates colouring for empty strings, which is unlikely to be an issue for majority of users.

### DIFF
--- a/packages/webpack-cli/lib/utils/index.js
+++ b/packages/webpack-cli/lib/utils/index.js
@@ -1,6 +1,22 @@
+let colorette = null;
+
 module.exports = {
+
   get colors() {
-    return require("colorette");
+    if (!colorette || colorette.options.changed) {
+      const col = require('colorette');
+      const options = {
+        changed: false,
+        enabled: colorette? colorette.options.enabled : true,
+      }
+
+      colorette = {
+        ...col.createColors({ useColor: options.enabled }),
+        options,
+      }
+    }
+
+    return colorette;
   },
 
   get levenshtein() {
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "@typescript-eslint/parser": "^4.14.1",
     "@webpack-cli/migrate": "^1.1.2",
     "coffeescript": "^2.5.1",
-    "colorette": "^1.2.1",
+    "colorette": "^2.0.2",
     "concat-stream": "^2.0.0",
     "cz-customizable": "^6.3.0",
     "del-cli": "^3.0.1",
--- a/packages/generators/package.json
+++ b/packages/generators/package.json
@@ -22,7 +22,7 @@
     "plugin-template"
   ],
   "dependencies": {
-    "colorette": "^1.2.1",
+    "colorette": "^2.0.2",
     "yeoman-environment": "^2.10.3",
     "yeoman-generator": "^4.12.0"
   },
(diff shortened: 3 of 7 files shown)


packages/webpack-cli/lib/utils/index.js
```diff
@@ -1,6 +1,22 @@
+let colorette = null;
```
alexander-akait, line 1: Why we need this?

kibertoad: To avoid recreating colorette instance on every getter call. We could simply import default colouring operations, but these would default to always on and ignore configuration overrides.

packages/webpack-cli/lib/utils/index.js
```diff
@@ -1,6 +1,22 @@
...
+
 module.exports = {
+
   get colors() {
-    return require("colorette");
+    if (!colorette || colorette.options.changed) {
+      const col = require('colorette');
+      const options = {
+        changed: false,
+        enabled: colorette? colorette.options.enabled : true,
```
alexander-akait, line 10: It seems some weird, colorette already has logic for `enabled`, why we need check `colorette`?

kibertoad: See existing code at https://github.com/webpack/webpack-cli/blob/3813e7ae22b22be63c84d530072073254bc2104a/packages/webpack-cli/lib/webpack-cli.js#L995
From my understanding this is an existing solution to trigger changes in the behaviour of instance, returned by getter. Alternatively, we can expose a method that would instantly recreate instance, stored here, to avoid this.

alexander-akait: We should rewrite logic, because we already cache module, we can keep `changed` on object level

kibertoad: caching module is not enough, we need to store initialized coloretto instance as well

packages/webpack-cli/lib/utils/index.js
jorgebucaran: We decided to classify the always-color behavior as a bug after all (https://github.com/jorgebucaran/colorette/pull/72). If you don't really care about the new `createColors` API, you may continue using Colorette as you were before, just upgrade to `2.0.5` or later.

jorgebucaran (changes requested):

jorgebucaran:
`2.x` is largely compatible with `1.x` as is. `createColors` is probably not needed here. There was an issue, but we fixed it a few days ago (https://github.com/jorgebucaran/colorette/pull/72).

cc @alexander-akait @kibertoad 👋

kibertoad:
superseded by #2966

PR #2962, opened 2021-09-21
fix: allow falsy values for "port" option

henryqdineen:

Fix

Open to suggestions for testing this.

N/A

**Summary**

This fixes a regression that we noticed when updating to `webpack-cli@4.7.2`. We are still on `webpack-dev-server@3.11.2` which allows using `null` or `0` to automatically use a free port. In our case we are setting `port: 0` and it was causing this error:

```
Error: listen EADDRINUSE: address already in use 127.0.0.1:8080
```

If `webpack-cli` still supports `webpack-dev-server@3.11.2` it seems like it shouldn't prevent passing `null` or `0` as an option to `port`. Our current workaround is using the string `'0'`, which is not falsy.

No

**Other information**

### DIFF
--- a/packages/serve/src/index.ts
+++ b/packages/serve/src/index.ts
@@ -296,7 +296,8 @@ class ServeCommand {
             };
 
             devServerOptions.host = devServerOptions.host || "localhost";
-            devServerOptions.port = devServerOptions.port || 8080;
+            devServerOptions.port =
+              typeof devServerOptions.port !== "undefined" ? devServerOptions.port : 8080;
             devServerOptions.stats = getStatsOption();
             devServerOptions.publicPath = getPublicPathOption();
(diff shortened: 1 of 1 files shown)


alexander-akait:
Please accept CLA, also I want to note - please prefer to use `port: 'auto'` in v4, using `0` will enable Node.js default behavior (i.e. not try to release port again)

henryqdineen:
Thank you @alexander-akait. I will make a note of that for when we upgrade to v4.

alexander-akait (approved):

anshumanv (approved):

evenstensberg (approved):
LGTM

snitin315 (approved):
Thanks!

PR #2965, opened 2021-09-24
chore: migrate to `colorette` v2

snitin315:
chore

Already present
No
**Summary**
Fix https://github.com/webpack/webpack-cli/pull/2956
Fix https://github.com/webpack/webpack-cli/issues/2955
migrate to `nanocolors`

No

**Other information**
No


alexander-akait:
our tests failed, need to check

jorgebucaran:
Colorette, which is already being used, works exactly the same, what's the reason behind this change?

alexander-akait:
@jorgebucaran better API

jorgebucaran:
It's the exact same API.

alexander-akait:
@jorgebucaran no extra `createColors` call, also we can't get `isColorSupported`

jorgebucaran:
@alexander-akait No, that's incorrect. It's exactly the same API.

alexander-akait:
hm, as I can see nanocolors is faster:
```
$ ./test/loading.js
chalk          3.465 ms
cli-color     21.849 ms
ansi-colors    1.101 ms
kleur          1.628 ms
kleur/colors   0.508 ms
colorette      2.610 ms
nanocolors     0.486 ms
```

https://github.com/ai/nanocolors

jorgebucaran:
Results should be the same as nanocolors is essentially a copy of Colorette. They're also outdated. 

- [nanocolors](https://github.com/ai/nanocolors/blob/44cff8a0271b37f34550d942baeb92dea0f9175e/index.js#L22-L34)
- [Colorette](https://github.com/jorgebucaran/colorette/blob/77ff01b077babe8f8877fa6317353c8d3ff44107/index.js#L20-L29)

jorgebucaran:
Those are loading benchmarks, measuring loading performance is tricky. Except for `cli-colors` perhaps (but still), every package in that list loads in around 1 or 2 milliseconds (thousandth of a second).

alexander-akait:
Are you sure, these benchmarks is very new and nanocolors wins

jorgebucaran:
Yes. Because nanocolors is essentially a copy of Colorette, the numbers tend to fluctuate.

alexander-akait:
But time of startup is faster as you can see

jorgebucaran:
That was a result of importing `process` directly as opposed to using the global object. We really don't need to do that. I've updated https://github.com/webpack/webpack-cli/pull/2966 to `2.0.9`. 

```console
chalk         14.422 ms
cli-color     111.140 ms
ansi-colors    2.554 ms
kleur          2.903 ms
kleur/colors   0.791 ms
colorette      0.754 ms
nanocolors     1.146 ms
```

alexander-akait:
Now we have two same projects :smile:

jorgebucaran:
#2966 is now using `2.0.10`. It loads and runs even faster now. Also smaller: https://packagephobia.com/result?p=colorette.

jorgebucaran:
#2966 is using `2.0.10` now. It loads and runs even faster. Also smaller: https://packagephobia.com/result?p=colorette.

jorgebucaran:
[Running the benchmarks in a CI](https://github.com/jorgebucaran/color-bench-test/runs/3707054689?check_suite_focus=true) you'll see that measuring loading time is not worthwhile. Import order, top-level _vs_ `node_modules`, `require` _vs_ static `import` _vs_ dynamic `import()`, all matter and make results fluctuate. Shift the call order around, run the benchmarks again and you'll see varying results. 

```console
nanocolors     5.27ms
colorette      1.68ms
nanocolors     0.01ms
colorette      0.01ms
nanocolors     0.01ms
colorette      0.00ms
```
```console
nanocolors     10,468,299 ops/sec
colorette      10,637,481 ops/sec
nanocolors     10,497,637 ops/sec
colorette      10,494,131 ops/sec
nanocolors     10,463,827 ops/sec
colorette      10,266,131 ops/sec
```

The idea that we are considering switching to nanocolors, essentially a dupe of Colorette that was published just a few days ago, is problematic and discouraging. Not in the true spirit of open-source software.

snitin315:
I think we should stick with `colorette` here as the original package. migrating to colorette 2.x and nanocolors is exactly the same for us as nanoncolors is almost a duplicate package.

anshumanv:
Yep, I vote for sticking with colorette too, been in active development since last few years and nanocolors feels like a copy

packages/webpack-cli/lib/utils/logger.js
```diff
@@ -2,10 +2,26 @@ const utils = require("./index");
...
+  info: (val) => {
+    const { cyan } = utils.colors.createColors({ useColor: utils.colors.isColorSupported });
+
+    return console.info(`[webpack-cli] ${cyan(val)}`);
+  },
+  success: (val) => {
+    const { green } = utils.colors.createColors({ useColor: utils.colors.isColorSupported });
+
+    return console.log(`[webpack-cli] ${green(val)}`);
+  },
```
alexander-akait, line 24: Very bad idea

alexander-akait:
@jorgebucaran the new version has a very inconvenient api:
1. we can't change `isColorSupported` https://github.com/jorgebucaran/colorette/blob/2.0.12/index.js#L16
2. Due `1.` if developer provide `--no-color` flag we should create colors again, otherwise it will be colored

Maybe I am missing something?

jorgebucaran:
Does https://github.com/jorgebucaran/colorette/issues/77 fix this? I can release a new patch right away.

packages/webpack-cli/lib/utils/logger.js
snitin315: This was now required for `--no-color` support.

snitin315:
> Does [jorgebucaran/colorette#77](https://github.com/jorgebucaran/colorette/issues/77) fix this? I can release a new patch right away.

Yes, that would be helpful. Otherwise we will need to write some extra code.

alexander-akait:
@webpack/cli-team I do big refactor our utils:
1. simple API, no need utils anymore and avoid dirty circular logic (when utils require util and utils require utils )
2. Better performance 

Also I will refactor loadConfig logic and processing options for better performance in the next PRs

Issue #2944, opened 2021-09-10
This happens when I use `npx webpack init`. I worked around it by doing `./node_modules/bin/webpack init`.


alexander-akait:
This happens when I use `npx webpack init`. I worked around it by doing `./node_modules/bin/webpack init`.

_Originally posted by @volkanunsal in https://github.com/webpack/webpack-cli/issues/191#issuecomment-916912387_

alexander-akait:
Just run `npx webpack init` on new package

alexander-akait:
@webpack/cli-team we need to rename `@webpack-cli/init` to `create-webpack-app` package, also it should work without `webpack-cli` itself, i.e. we should move `commander` and other API inside `create-webpack-app` package also fix docs everywhere, because `npx webpack init` and `npx webpack-cli` doesn't work

anshumanv:
On it

alexander-akait:
Still valid

rishabh3112:
It's been two months now, so I am taking this up.

rishabh3112:
WIP: https://github.com/webpack/webpack-cli/pull/3036

adriancuadrado:
Any progress on this?

abrehan2:
Hello Sir @snitin315,

I bring with me extensive expertise in full-stack development and open-source contributions, including the publication of my own npm packages. I am eager to tackle some starter bugs; could you kindly point me in the right direction? ;)

Best regards,
Rehan

snitin315:
@MysteryHawk17 @abrehan2 Let's use Discord for introductions & questions and avoid commenting on issues.

abrehan2:
> @MysteryHawk17 @abrehan2 Let's use Discord for introductions & questions and avoid commenting on issues.

Sure! Can you please share the link?

MysteryHawk17:
@snitin315 please share the discord link to connect

samrudh3125:
@snitin315 please share the dicord connect link

urizennnn:
Hello guys I would love to work on this topic as it seems quite intuitive. Here are my questions on this , what are the targeted languages for this  like the main languages one would like to create a boiler code for and also if there should be an other option if the developer just wants to create something else but still use the webpack bundler

c6z3h:
@samrudh3125 I tried clicking on the discord link but it seems expired or have no permission, is there an updated link maybe?

urizennnn:
[...]

PR #2980, opened 2021-10-01
fix: npx init

alexander-akait:

bugfix

manually tests

yes

/cc @webpack/cli-team we need update our docs, for `init` we should always use `npx webpack-cli init`

**Summary**

It is fast fix, in future we should create rename `init` to `create-webpack-app` and use `npx create-webpack-app`

No

**Other information**

Partial fix https://github.com/webpack/webpack-cli/issues/2944

### DIFF
--- a/packages/generators/package.json
+++ b/packages/generators/package.json
@@ -23,11 +23,11 @@
   ],
   "dependencies": {
     "yeoman-environment": "^2.10.3",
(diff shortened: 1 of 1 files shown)


alexander-akait:
No `npx` install `webpack-cli` inself, because we need this package and allow to run `init`

PR #3019, opened 2021-10-30
chore(info): rename `--additional-package` flag to `additional-packages`

jamesgeorge007:
chore

Update test suite.

Yes

**Summary**
Rename `--additional-package` to `--additional-packages` as it supports variadic arguments.

Yes, `--additional-package` is not recognized anymore.

**Other information**
N/A

### DIFF
--- a/packages/info/README.md
+++ b/packages/info/README.md
@@ -32,10 +32,10 @@ webpack info [options]
 
 #### Output format
 
-| Flag                                  | Description                             | Type   |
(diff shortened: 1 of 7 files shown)


alexander-akait:
Sorry for delay, not sure we need it, because we have `--entry=foo.js --entry=bar.js` and there are a lot of other options with the same logic in names, so let's keep it as it and avoid breaking changes, I would say that this is our design, sorry for losing your time and feel free to feedback

PR #3045, opened 2021-12-05
chore: upgrade to ESLint 8

nschonni:

Bump parser and ESLint to v8 and cleanup now unneeded suppressions

**Summary**

**Other information**


packages/serve/src/types.ts
```diff
@@ -43,6 +43,7 @@ export type devServerOptionsType = {
     | false
     | string
     | transportModeEnum
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
nschonni, line 46: This was just an oustanding warning in CI

snitin315 (changes requested):
We cannot update it right now, we still support Node v10.

nschonni:
> We cannot update it right now, we still support Node v10.

No problem, notice the same issue with bumping Husky. Is there a plan to drop v10 support? It hit EOL at the end of April

snitin315:
Yes, we will drop node 10 and webpack 4 support in next major release.

packages/serve/src/types.ts
snitin315: We can accept this change, if you can revert the others.

nschonni: Split this out into https://github.com/webpack/webpack-cli/pull/3048, if you want to leave this open till you drop 10, or just close this one since it will get outdated till you're read to drop 10

PR #3047, opened 2021-12-06
chore: remove lint matrix

nschonni:

**Summary**

Was only setup for a single execution, so most of the values were unused

**Other information**

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -16,23 +16,17 @@ on:
 
 jobs:
   lint:
-    name: Lint - ${{ matrix.os }} - Node v${{ matrix.node-version }}
+    name: Lint Code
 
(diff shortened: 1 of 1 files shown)


alexander-akait (commented):
Why? Solution before works the same, we prefer to use matrix

nschonni:
> Why? Solution before works the same, we prefer to use matrix

Like the Commit lint job, there is only a single execution of the task, so a matrix doesn't make sense for this job. The Node version in the task name was also confusing your branch policy before (looks like you've unstuck the v12 version)

evenstensberg (commented):
Why?

nschonni:
> Why?

https://github.com/webpack/webpack-cli/pull/3047#issuecomment-986940958

alexander-akait:
Sorry, we don't need it, current solution look good for me, anyway thanks

PR #3075, opened 2022-01-07
feat: graceful exit terminal

rishabh3112:
feature

WIP

**Summary**
Added code for graceful exit.
For reference: https://github.com/webpack/webpack-dev-server/blob/75999bb9bb8803de633fcb037405f45a5bf7d029/lib/Server.js#L1808-L1838

Yup can be for tools that consume webpack-cli's output.

**Other information**
/cc @alexander-akait


rishabh3112:
@alexander-akait should we add flag to enable this behavior as dev server does?
If we keep it by default, it can be a breaking change for tools that rely on webpack-cli's output.

packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -2324,6 +2324,27 @@ class WebpackCLI {
...
+
+        needForceShutdown = true;
+
+        compiler.close(() => {
+          process.exit();
+        });
+      };
+
+      process.on(signal, listener);
+    });
```
alexander-akait, line 2346: We should run it only when `cache` is not `false`/`undefined`

rishabh3112: Done. Will do for watch as well then we are good to go.

packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -2324,26 +2324,30 @@ class WebpackCLI {
...
 
-    const signals = ["SIGINT", "SIGTERM"];
-    let needForceShutdown = false;
-
-    signals.forEach((signal) => {
-      const listener = () => {
-        if (needForceShutdown) {
-          process.exit();
-        }
+    if (compiler.cache) {
```
alexander-akait: It can be multi compiler

rishabh3112: Please check in latest commit, I updated it to use options.

packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -2324,14 +2324,42 @@ class WebpackCLI {
       return;
     }
 
+    let needForceShutdown = false;
+    if (compiler.options.cache) {
```
alexander-akait, line 2328: Please change it on:
```
const needGracefullShutdown = (compiler) =>
      compiler.compilers
        ? compiler.compilers.some((compiler) => compiler.options.cache)
        : compiler.options.cache;
```

rishabh3112: `needGracefullShutdown` is semaphore for signals, but I got what change you want :)

alexander-akait (commented):
Looks good, let's add couple tests - just create custom plugin with infinity waiting and send `SIGINT` signals to close it

packages/webpack-cli/lib/webpack-cli.js
rishabh3112: marking it as resolved

packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -2326,14 +2326,47 @@ class WebpackCLI {
...
+
     const isWatch = (compiler) =>
       compiler.compilers
         ? compiler.compilers.some((compiler) => compiler.options.watch)
         : compiler.options.watch;
 
     if (isWatch(compiler) && this.needWatchStdin(compiler)) {
       process.stdin.on("end", () => {
-        process.exit(0);
+        compiler.close(() => {
```
alexander-akait, line 2366: It can be problem for webpack v4, because no `close` method, we should check it `if (typeof compiler.close === "function")`

packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -2326,14 +2326,47 @@ class WebpackCLI {
...
+            process.exit();
+          }
+
+          this.logger.info(
+            "Gracefully shutting down. To force exit, press ^C again. Please wait...",
+          );
+
+          needForceShutdown = true;
+
+          compiler.close(() => {
```
alexander-akait, line 2350: Same here

packages/webpack-cli/lib/webpack-cli.js
rishabh3112: Okay on it 👍🏽

rishabh3112: Done

packages/webpack-cli/lib/webpack-cli.js
rishabh3112: Done

packages/webpack-cli/lib/webpack-cli.js
```diff
@@ -2326,14 +2326,52 @@ class WebpackCLI {
...
```
alexander-akait, line 2350: Let's move check here, otherwise we will not print log for webpack v4

rishabh3112: Is there a way to gracefully exit in webpack v4?

alexander-akait: Just output logger and then run `process.exit()`, `if (typeof compiler.close === "function") { compiler.close(() => { process.exit(); }) } else { process.exit(); }`

rishabh3112: Okay, will do.

rishabh3112: Done

evenstensberg (commented):
could you rebase?

evenstensberg (commented):
No activity, closing.

alexander-akait:
@evenstensberg let's keep, it is a bug on our side

alexander-akait:
Let's keep this open, because there is a bug on our side, we need fix it here - https://github.com/webpack/webpack-cli/blob/master/packages/webpack-cli/src/webpack-cli.ts#L2488

```
isWatch(compiler) || isFileSystemCacheEnabled(compiler) 
```

Easy fix

evenstensberg:
I'll pick this up tomorrow.

PR #3076, opened 2022-01-07
Feat added types

alexander-akait:
**What kind of change does this PR introduce?**

**Did you add tests for your changes?**

**If relevant, did you update the documentation?**

**Summary**

**Does this PR introduce a breaking change?**

**Other information**

PR #3078, opened 2022-01-07
Feat strict types

alexander-akait:

types

No need

No need

**Summary**

Use `strict`

No

**Other information**

- refactor our code
- remove unused things
- avoid extra any

there are still a lot of any, so still WIP, also we need to refactor our generators more, to use one CustomGenerator class


alexander-akait:
@webpack/cli-team Can you review, I rewrite our logic on using only one custom generator, so we don't need copy/paste logic, also improve types, now we use `strict`, any `cli` itself is `any`, but we will finish it in https://github.com/webpack/webpack-cli/pull/3076

packages/generators/src/handlers/default.ts
```diff
@@ -198,7 +192,6 @@ export function generate(self: CustomGenerator): void {
     ...self.answers,
     entry,
   });
-  self.configurationPath = self.destinationPath("webpack.config.js");
```
rishabh3112, line 201: Not a big deal for now, but IMO, This should be handled here only, as templates might generate multiple configuration files as well in future.

rishabh3112 (approved):
Looks good, Just one doubt.

packages/generators/src/handlers/default.ts
alexander-akait: Yep, let's do refactor when we will implement it, I think now it good, focus on types for webpack-cli

PR #3128, opened 2022-02-12
Feature/added cli types

taranek:
Adds Typescript support for `webpack-cli` package. 

No

No

**Summary**
Inspired by [PR #3076](https://github.com/webpack/webpack-cli/pull/3076) I managed to propose my solution with reduced usage of `any`. It it still missing some types from `webpack`(v5.68.0) package (which are declared there but not exported) - for now I duplicated them in `webpack-internal-types.ts` - I can open a seperate PR in `webpack` repo in order to export them (please tell me if it's needed). It's my first PR here, so I had to reverse-engineer some types and names for them - which means I probably didn't pick the best names for types.

No

**Other information**
N/A


packages/webpack-cli/src/webpack-cli.ts
```diff
@@ -2214,49 +2377,50 @@ class WebpackCLI {
...
     }
 
     return compiler;
   }
 
-  needWatchStdin(compiler) {
-    if (compiler.compilers) {
-      return compiler.compilers.some(
-        (compiler) => compiler.options.watchOptions && compiler.options.watchOptions.stdin,
+  needWatchStdin(compiler: Compiler | MultiCompiler): undefined | boolean {
```
alexander-akait, line 2387: It should never return `undefined`, we should refactor logic

packages/webpack-cli/src/webpack-internal-types.ts
```diff
@@ -0,0 +1,176 @@
...
+interface ArgumentConfig {
+  description: string;
+  negatedDescription?: string;
+  path: string;
+  multiple: boolean;
+  type: "string" | "number" | "boolean" | "path" | "enum" | "RegExp" | "reset";
+  values?: any[];
+}
+
+export type EntryOptions = { name?: string } & Omit<EntryDescriptionNormalized, "import">;
```
alexander-akait, line 176: We should fix it in webpack and export them, no need to duplicate, it is just mistake on webpack side

packages/webpack-cli/src/webpack-cli.ts
```diff
@@ -2320,14 +2487,14 @@ class WebpackCLI {
...
-    compiler = await this.createCompiler(options, callback);
+    compiler = await this.createCompiler(options as WebpackDevServerOptions, callback);
 
     if (!compiler) {
       return;
     }
 
-    const isWatch = (compiler) =>
-      compiler.compilers
+    const isWatch = (compiler: WebpackCompiler): undefined | boolean =>
```
alexander-akait, line 2496: Same here, only `boolean` should be returned, technically `undefined` is `false`, but it is not good practice

alexander-akait (commented):
Great work! Let's fix some stuff and we can merge

packages/webpack-cli/src/webpack-internal-types.ts
taranek: Thank you, I've submitted a [PR](https://github.com/webpack/webpack/pull/15377)

packages/webpack-cli/src/webpack-cli.ts
taranek: Thanks, I'll just cast both method using `Boolean()`

alexander-akait (commented):
Good job, currently we use `any` https://github.com/webpack/webpack-cli/blob/master/packages/info/src/index.ts#L14 on CLI packages, can you fix it to check we really don't missing something in our API (some things can be complex, put TODO in this case)

packages/webpack-cli/src/webpack-internal-types.ts
snitin315: I believe we need to update webpack here.

taranek:
@alexander-akait Thank you for mentioning that part. I managed to import the types from `webpack-cli` by using project references. I guess we can run CI checks once again.

.eslintignore
```diff
@@ -2,10 +2,7 @@ coverage
 .nyc_output
 node_modules
 dist
-packages/configtest/lib
-packages/generators/lib
-packages/info/lib
-packages/serve/lib
+packages/*/lib
```
snitin315, line 5: This will now ignore `packages/webpack-cli/lib` too which we don't want.

snitin315:
@taranek  Let's fix lint

packages/webpack-cli/package.json
```diff
@@ -59,5 +59,8 @@
       "optional": true
     }
   },
-  "gitHead": "fb50f766851f500ca12867a2aa9de81fa6e368f9"
+  "gitHead": "fb50f766851f500ca12867a2aa9de81fa6e368f9",
+  "devDependencies": {
+    "webpack-dev-server": "^4.7.4"
+  }
```
taranek, line 65: I've added `webpack-dev-server` as dev dep in `webpack-cli` package, but the smoketests are still failing when resolving `webpack-dev-server` in `serve` package. [(check the details here)](https://github.com/webpack/webpack-cli/runs/5214548107?check_suite_focus=true) 

Should I add `webpack-dev-server` as a peerDependency in `serve` package? 🤔

alexander-akait: In theory yes, but there is the problem, `serve` is part of `webpack-cli`, so if we add `webpack-dev-server` to `peerDependency` new npm verison will install it automatically, it is not good, because many projects don't use dev server, I think better refactor our tests

taranek: Thank you for the explanation. I managed to reproduce the CI error locally and fix it by removing the dev dep from `webpack-cli` and surpressing extraneous import in TS and eslint.

alexander-akait (approved):
Looks good, let's wait CI

alexander-akait (dismissed):
/cc @webpack/cli-team Can anyone review this as well?

snitin315 (approved):
This is great ⭐ , thank you ❤️

rishabh3112 (approved):

alexander-akait:
@taranek Sorry for this, can you rebase?

taranek:
@alexander-akait  Sure, but are we ready to merge now? If we still need a review from the core team, can we rebase after the approval?

snitin315:
Yes, we can merge after rebase.

rishabh3112 (approved):

alexander-akait:
Big thanks

taranek:
@alexander-akait You're welcome - do you think converting tests code to TS would be worth doing as well?

PR #3146, opened 2022-02-25
refactor(webpack-cli): don't spawn needless shells and replace execa with cross-spawn

ludofischer:
Refactor: use a simple child process instead of a shell and replace `execa` with `cross-spawn`.

No, I only replaced the `execa` mocks with `cross-spawn` mocks, as the functionality should be identical as before.
I've tested `webpack init` on my local machine.

**Summary**
Executing a shell is less performant, less secure and less cross-platform, according to both the [execa](https://github.com/sindresorhus/execa#shell) and [cross-spawn](https://github.com/moxystudio/node-cross-spawn#using-optionsshell-as-an-alternative-to-cross-spawn) documentation.

Replacing `execa` with `cross-spawn` removes 9 dependencies on a production install:
- execa
- get-stream
- human-signals
- is-stream
- mimic-fn
- npm-run-path
- onetime
- signal-exit
- strip-final-newline
 
I can't  find any options unique to execa in the codebase and cross-spawn should take care of the [Windows compatibility issues](https://github.com/moxystudio/node-cross-spawn#why).

No.

**Other information**
 execa uses `cross-spawn` under the hood to parse the command arguments, so nothing should change in that regard: https://github.com/sindresorhus/execa/blob/main/index.js

### DIFF
--- a/packages/webpack-cli/src/webpack-cli.ts
+++ b/packages/webpack-cli/src/webpack-cli.ts
@@ -154,7 +154,7 @@ class WebpackCLI implements IWebpackCLI {
   }
 
   getAvailablePackageManagers(): PackageManager[] {
-    const { sync } = require("execa");
+    const { sync } = require("cross-spawn");
     const installers: PackageManager[] = ["npm", "yarn", "pnpm"];
     const hasPackageManagerInstalled = (packageManager: PackageManager) => {
       try {
@@ -179,7 +179,7 @@ class WebpackCLI implements IWebpackCLI {
   }
 
   getDefaultPackageManager(): PackageManager | undefined {
-    const { sync } = require("execa");
+    const { sync } = require("cross-spawn");
     const hasLocalNpm = fs.existsSync(path.resolve(process.cwd(), "package-lock.json"));
 
     if (hasLocalNpm) {
@@ -294,10 +294,10 @@ class WebpackCLI implements IWebpackCLI {
     }
 
     if (needInstall) {
-      const execa = require("execa");
+      const spawn = require("cross-spawn");
 
       try {
-        await execa(commandToBeRun, [], { stdio: "inherit", shell: true });
+        spawn(commandToBeRun, [], { stdio: "inherit", shell: true });
       } catch (error) {
         this.logger.error(error);
 
--- a/packages/webpack-cli/src/webpack-cli.ts
+++ b/packages/webpack-cli/src/webpack-cli.ts
@@ -244,13 +244,6 @@ class WebpackCLI implements IWebpackCLI {
       options.preMessage();
     }
 
-    // yarn uses 'add' command, rest npm and pnpm both use 'install'
-    const commandToBeRun = `${packageManager} ${[
-      packageManager === "yarn" ? "add" : "install",
-      "-D",
-      packageName,
-    ].join(" ")}`;
-
     const prompt = ({ message, defaultResponse, stream }: PromptOptions) => {
       const readline = require("readline");
       const rl = readline.createInterface({
@@ -275,6 +268,10 @@ class WebpackCLI implements IWebpackCLI {
       });
     };
 
+    // yarn uses 'add' command, rest npm and pnpm both use 'install'
+    const commandArguments = [packageManager === "yarn" ? "add" : "install", "-D", packageName];
+    const commandToBeRun = `${packageManager} ${commandArguments.join(" ")}`;
+
     let needInstall;
 
     try {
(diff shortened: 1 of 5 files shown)


packages/webpack-cli/src/webpack-cli.ts
```diff
@@ -294,10 +291,10 @@ class WebpackCLI implements IWebpackCLI {
     }
 
     if (needInstall) {
-      const execa = require("execa");
+      const { sync } = require("cross-spawn");
 
       try {
-        await execa(commandToBeRun, [], { stdio: "inherit", shell: true });
+        sync(packageManager, commandArguments, { stdio: "inherit" });
```
ludofischer, line 297: Can't understand why this is the only spot where `sync` was not used and `shell` was set to `true`.  Could be that it was a way to make passing all the arguments in a single string work?

packages/webpack-cli/src/webpack-cli.ts
```diff
@@ -275,6 +268,10 @@ class WebpackCLI implements IWebpackCLI {
       });
     };
 
+    // yarn uses 'add' command, rest npm and pnpm both use 'install'
+    const commandArguments = [packageManager === "yarn" ? "add" : "install", "-D", packageName];
```
ludofischer, line 272: We can re-use the argument array if we keep it separate instead of concatenating immediately into a single string.

packages/webpack-cli/src/webpack-cli.ts
alexander-akait: Good questions, maybe it was just mistake

alexander-akait:
Can you try to update lock file?

ludofischer:
> Can you try to update lock file?

I've updated all versions in `yarn.lock`.

packages/webpack-cli/src/webpack-cli.ts
ludofischer: Is there also a reason the shell must run async? Intuitively it does not seem the case, as there is no way the program can continue if the install does not finish.

It's true that [npm is a bash script on Unix and a cmd file on Windows](https://github.com/npm/cli/tree/latest/bin).

But later on the [comment you've linked](https://github.com/sindresorhus/execa/issues/433#issuecomment-674277826) adds 
> If you specify a shell file (as opposed to a binary), on Windows, and do not use the shell option, Execa emulates that option for you (through [the dependency node-cross-spawn](https://github.com/moxystudio/node-cross-spawn/blob/master/lib/parse.js#L36)) by basically running cmd /d /s /c /q "npm run tsc" instead of npm run tsc

I think that must be true, because f the shell option is required, the previous `execa` calls, all without `shell: true` would fail and I think someone would have reported the bug. I think the `shell: true` is really needed if you want the shell to expand or replace something in the command string.

alexander-akait:
Oh, looks like we should update types when we updating lock file

ludofischer:
> Oh, looks like we should update types when we updating lock file

I think it's the opposite, updating `@types/yeoman-environment` from `2.1.0.5 ` to `2.10.6` breaks the type check. I've removed the commit that updated all dependencies and rebased, since some dependencies have been updated on master in the meantime.

alexander-akait:
hm, we have a problems with:
```
RUN  Missing webpack-dev-server
  swapping webpack-dev-server with .webpack-dev-server
  timeout: killing process
  swapping .webpack-dev-server with webpack-dev-server
FAIL  Missing webpack-dev-server

RUN  Missing webpack
  swapping webpack with .webpack
  timeout: killing process
  swapping .webpack with webpack
FAIL  Missing webpack

RUN  Missing webpack-bundle-analyzer
  swapping webpack-bundle-analyzer with .webpack-bundle-analyzer
  timeout: killing process
  swapping .webpack-bundle-analyzer with webpack-bundle-analyzer
FAIL  Missing webpack-bundle-analyzer
```

That is weird...

ludofischer:
> That is weird...
I think the problem is `timeout: killing process`, not the `missing ...` messages, because the `missing` messages also appear when the tests pass.
The smoketests all pass on my machine, even when I follow the exact same steps as in CI. I think the timeout comes from `smoketests/helpers.js`https://github.com/webpack/webpack-cli/blob/80ae4a2e2fc9ff1274221f04cc6dc23c80f93d05/smoketests/helpers.js#L45
but it's set to 30 seconds, which should be plenty (on my machine each tests runs in a second or so).

alexander-akait:
Can we try to increase timeout? Sorry for this, don't have time on deep investigation for this

ludofischer:
> Can we try to increase timeout? Sorry for this, don't have time on deep investigation for this

I've increased the timeout to 60 seconds.

alexander-akait:
Maybe we need increase some more

ludofischer:
> Maybe we need increase some more

I'm not sure, that timeout only affects the smoketests but now it's the other tests hanging. Anyway, this time the smoketests passed in less than 1 min on Ubuntu, I think either the containers are too slow or there is some deadlock issue.

alexander-akait (approved):
/cc @webpack/cli-team Let's review

rishabh3112 (approved):

snitin315 (approved):

anshumanv (approved):
Thanks

PR #3161, opened 2022-03-13
fix: attempts custom package require without path exists check

solsson:

Bugfix

No

Yes

**Summary**

Firstly this change fixes an issue that looks like a copy-paste error: Custom WEBPACK_DEV_SERVER_PACKAGE would be conditioned on WEBPACK_PACKAGE to exist.

The next change, using require.resolve, widens the usefulness of the override. With plain `existsSync` you basically need an absolute path, unless you know what the current working directory will be when using the override.

I don't think so. Existing WEBPACK_PACKAGE or WEBPACK_DEV_SERVER_PACKAGE would have satisfied `fs.existsSync(require.resolve(path)))` if they worked.

**Other information**

### DIFF
--- a/packages/webpack-cli/src/webpack-cli.ts
+++ b/packages/webpack-cli/src/webpack-cli.ts
@@ -62,8 +62,10 @@ const util = require("util");
 
 const { program, Option } = require("commander");
 
-const WEBPACK_PACKAGE = process.env.WEBPACK_PACKAGE || "webpack";
-const WEBPACK_DEV_SERVER_PACKAGE = process.env.WEBPACK_DEV_SERVER_PACKAGE || "webpack-dev-server";
+const WEBPACK_PACKAGE_IS_CUSTOM = !!process.env.WEBPACK_PACKAGE;
+const WEBPACK_PACKAGE = WEBPACK_PACKAGE_IS_CUSTOM ? process.env.WEBPACK_PACKAGE as string : "webpack";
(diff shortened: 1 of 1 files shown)


alexander-akait:
Thanks, can you accept CLA?

solsson:
CLA signed.

Actually if the `existsSync` check causes any issues I'd be in favor of dropping it. These envs have distinct names that are unlikely to collide with other use of evns. Silently dropping them might cause more confusion than attempting the require regardless of value.

solsson:
I will force push to only the first commit, then get one commit at a time through CI.

snitin315:
@solsson can you fix lint?

rishabh3112 (changes requested):
Looks good, we need to fix lint and types as well.

packages/webpack-cli/src/webpack-cli.ts
```diff
@@ -421,12 +423,12 @@ class WebpackCLI implements IWebpackCLI {
         let skipInstallation = false;
 
         // Allow to use `./path/to/webpack.js` outside `node_modules`
-        if (dependency === WEBPACK_PACKAGE && fs.existsSync(WEBPACK_PACKAGE)) {
+        if (dependency === WEBPACK_PACKAGE && WEBPACK_PACKAGE_IS_CUSTOM) {
```
alexander-akait, line 426: I think we should use `require.resolve()` here and improve error reporting here (same for below) if package was not found

alexander-akait:
@solsson Hello, do you have time to finish this PR? Because it is easy to fix lint, thank you

Issue #3197, opened 2022-04-08
Timestamp in verbose logs is incorrect by a factor of 1000x


ryanwilsonperkin:
**Describe the bug**

When running `webpack build --watch` with an infrastructure log level of `debug`, any time a file is changed a log line will be emitted that specifies which file changed and what the time was of the change. This time is incorrect because of a calculation on the initial timestamp, causing the real datetime `Changed time is Fri Apr 8 2022 09:56:00 EST-0400 (Eastern Daylight Time)` to be reported as `Changed time is Mon Mar 26 54238 16:04:29 GMT-0400 (Eastern Daylight Time)`

**What is the current behavior?**

Reports a datetime that is tens of thousands of years in the future 😄 

**To Reproduce**

Here's a minimal reproduction: https://gist.github.com/ryanwilsonperkin/9b14fcc354e6c24df99c9985e1fa5ccd

1. Set infrastructure logging level to debug
2. Run a build in watch mode
3. Modify one of the entry files to trigger a recompile
4. Note that the modified datetime that is specified in the resulting log will be incorrect

**Expected behavior**

Expected the datetime to be accurate

**Screenshots**

<img width="1840" alt="image" src="https://user-images.githubusercontent.com/3004111/162452062-f7527659-154f-4522-afbb-d987b10845b5.png">

<details>

<summary>Image alt text from terminal</summary>

```
~/src/github.com/ryanwilsonperkin/webpack-watch-timestamp-bug main*
❯ yarn webpack build --watch
yarn run v1.22.17
$ /Users/ryan/src/github.com/ryanwilsonperkin/webpack-watch-timestamp-bug/node_modules/.bin/webpack build --watch
    [webpack-cli] Compiler starting...
    [webpack-cli] Compiler is using config: '/Users/ryan/src/github.com/ryanwilsonperkin/webpack-watch-timestamp-bug/webpack.config.js'
asset main.js 1.2 KiB [compared for emit] (name: main)
./index.js 28 bytes [built] [code generated]
webpack 5.72.0 compiled successfully in 47 ms
    [webpack-cli] Compiler finished
    [webpack-cli] Compiler is watching files for updates...
    [webpack-cli] File '/Users/ryan/src/github.com/ryanwilsonperkin/webpack-watch-timestamp-bug/index.js' was modified
    [webpack-cli] Changed time is Mon Apr 02 54238 08:02:55 GMT-0400 (Eastern Daylight Time) (timestamp is 1649426673775)
[...]

PR #3198, opened 2022-04-08
Bugfix in calculation of modified time of files in watch mode

ryanwilsonperkin:

Bugfix

No, seemed too niche for a test case but happy to add one with a bit of instruction (namely whether `test/plugin/plugin.test.js` is the correct spot)

Not relevant, updates a verbose log

**Summary**

Fixes https://github.com/webpack/webpack-cli/issues/3197

We need to pass the timestamp in milliseconds to the date constructor, this code was assuming that changeTime was in seconds and needed to be multiplied by 1000 to get milliseconds but it is already in milliseconds.

This led to the constructed date being incorrect.

|Before|After|
|---|---|
||image|

No

**Other information**

It appears that the `changeTime` value comes through the `WatchMethod` value of `WatchFileSystem` type, ultimately tracing back to the `watchpack` package. Watchpack uses `fs.lstat` to retrieve a `fs.Stat` object of the file and retuns the mtime by casting it to a number using the `+` operand. This might cause some platform-dependant behaviour, such as if `+new Date()` returns different levels of precision between Mac/Windows/Linux/etc. There's a `mtimeMs` on the Stat object that might be more appropriate in this case to ensure a certain level of precision is returned.

Ref: https://nodejs.org/api/fs.html#stat-time-values

### DIFF
--- a/packages/webpack-cli/src/plugins/CLIPlugin.ts
+++ b/packages/webpack-cli/src/plugins/CLIPlugin.ts
@@ -93,7 +93,7 @@ export class CLIPlugin {
     });
 
     compiler.hooks.invalid.tap(pluginName, (filename, changeTime) => {
-      const date = new Date(changeTime * 1000);
+      const date = new Date(changeTime);
 
       this.logger.log(`File '${filename}' was modified`);
       this.logger.log(`Changed time is ${date} (timestamp is ${changeTime})`);


rishabh3112 (approved):
Looks good, lets wait for CI to be green.

snitin315 (approved):
LGTM, thanks!

rishabh3112:
Thanks

PR #3240, opened 2022-05-12
fix: consider using createroot instead of render

harish-sethuraman:

I see that there is a react template support being added so used `createRoot` instead of `render` as per React 18.

### DIFF
--- a/packages/generators/src/handlers/react.ts
+++ b/packages/generators/src/handlers/react.ts
@@ -25,7 +25,7 @@ export async function questions(
   });
 
   // Add react dependencies
-  self.dependencies.push("react@18.1.0", "react-dom@18.1.0");
+  self.dependencies.push("react@18", "react-dom@18");
 
   // Add webpack-dev-server always
   self.dependencies.push("webpack-dev-server");
--- a/packages/generators/src/handlers/react.ts
+++ b/packages/generators/src/handlers/react.ts
@@ -25,7 +25,7 @@ export async function questions(
   });
 
   // Add react dependencies
-  self.dependencies.push("react", "react-dom");
+  self.dependencies.push("react@18.1.0", "react-dom@18.1.0");
 
   // Add webpack-dev-server always
   self.dependencies.push("webpack-dev-server");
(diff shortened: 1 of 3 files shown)


snitin315 (changes requested):
The scaffolded app can be using react 17?

harish-sethuraman:
I don't see any version specification. Just the package name and this `doInstall` util installs all packages. So it pulls in the latest version?

- https://github.com/webpack/webpack-cli/blob/888bed93f2cbe348c63afc32d6dab6ecd02b7c47/packages/webpack-cli/src/webpack-cli.ts#L272-L273

snitin315:
Yeah, it will use the latest version. Ideally, we should specify the version because it is possible that the current template is not compatible with the latest version.

harish-sethuraman:
https://github.com/webpack/webpack-cli/blob/master/packages/generators/init-template/react/package.json.tpl

Although we add react version to the scaffold when we `doInstall` using the util the latest version only will be fetched right?

alexander-akait:
We have react here https://github.com/webpack/webpack-cli/blob/master/packages/generators/src/handlers/react.ts#L28, just need to look how to provide version

harish-sethuraman:
Done I see that this works when we push deps with the version as @snitin315 mentioned

packages/generators/src/handlers/react.ts
```diff
@@ -25,7 +25,7 @@ export async function questions(
   });
 
   // Add react dependencies
-  self.dependencies.push("react", "react-dom");
+  self.dependencies.push("react@18.1.0", "react-dom@18.1.0");
```
evenstensberg, line 28: I don't think we should pin specific version numbers

alexander-akait: Yep, just `react@18`

harish-sethuraman: Done 👍🏻

rishabh3112 (approved):
Looks fine now, Lets see CI then we can merge.

alexander-akait (approved):

harish-sethuraman:
Bumping for merge :)

snitin315 (approved):
Thanks!

PR #3244, opened 2022-05-14
chore: update dockerfile

evenstensberg:

Updating deps

N/A

N/A
**Summary**
Updates to node v18

No

**Other information**
N/A

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:16
+FROM node:18
 
 # Add global instances of prettier and eslint for vscode
 RUN npm install -g eslint prettier


.devcontainer/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:16
+FROM node:18
```
snitin315, line 1: 16 is still LTS

evenstensberg: Docs on this?

snitin315: https://nodejs.org/en/

rishabh3112 (changes requested):
I think we need to have overhead to updating this with every node release.
I have a simple solution in mind to fix it. Will be creating a PR.

rishabh3112:
Superseded by #3311

rishabh3112:
Thanks

Issue #3227, opened 2022-04-29
Feature Request: Swap from Lerna to another monorepo tool


evenstensberg:
**Is your feature request related to a problem? Please describe.**

Lerna is depreciated, so maybe we should swap to another monorepo tool.

- https://github.com/lerna/lerna

**Describe the solution you'd like**

Something like [Nx](https://nx.dev/), pnpm workspaces

**Describe alternatives you've considered**

N/A

**Additional context**

N/A

alexander-akait:
We can use built-in workspace feature in npm/yarn and avoid extra packages

alexander-akait:
I think we can close it, because lerna is not deprecated anymore

PR #3252, opened 2022-05-17
(Draft): Feature/remove lerna

taranek:

The intention is to remove `lerna` from dependencies list and use yarn workspaces instead. I started with removing `lerna bootsrap` command to see if `yarn install` would suffice for pipelines to succeed. 

After replacing `lerna bootstrap`, I'll have a look on other lerna usages (versioning and publishing) to see how we can replace those.

No

Not yet, I'd like to run a few CI pipelines on my branch to see how it goes.

**Summary**

[Related to feature request #3227](https://github.com/webpack/webpack-cli/issues/3227)

No

**Other information**

This is a draft PR, as migration from `lerna` is not ready yet.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -72,7 +72,6 @@ In case you are suggesting a new feature, we will match your idea with our curre
 - Bootstrap all the submodules before building for the first time
 
   ```bash
-  yarn lerna bootstrap
   yarn build
   ```
 
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "./packages/*"
   ],
   "scripts": {
+    "prune:deps": "find . -name 'node_modules' -type d -prune -exec rm -rf '{}' +",
     "clean": "del-cli \"*.tsbuildinfo\" \"packages/**/*.tsbuildinfo\" \"packages/*/lib/!(*.tpl)\" \"**/.yo-rc.json\"",
     "prebuild": "yarn clean",
     "prebuild:ci": "yarn clean && node ./scripts/setupBuild.js",
--- a/.cspell.json
+++ b/.cspell.json
@@ -93,7 +93,8 @@
     "wagoid",
     "Nitin",
     "dbaeumer",
-    "esbenp"
+    "esbenp",
+    "quickstart"
   ],
   "dictionaries": ["npm", "software-terms"],
   "ignorePaths": [
--- a/.cspell.json
+++ b/.cspell.json
@@ -91,7 +91,9 @@
     "Zenitsu",
     "eslintcache",
     "wagoid",
-    "Nitin"
+    "Nitin",
+    "dbaeumer",
+    "esbenp"
   ],
   "dictionaries": ["npm", "software-terms"],
   "ignorePaths": [
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -5,6 +5,6 @@
   "settings": {
     "terminal.integrated.shell.linux": "/bin/bash"
   },
-  "postCreateCommand": "yarn install && yarn lerna bootstrap && yarn build",
+  "postCreateCommand": "yarn install && yarn build",
   "extensions": ["dbaeumer.vscode-eslint", "esbenp.prettier-vscode"]
 }
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -38,9 +38,6 @@ jobs:
       - name: Install dependencies
         run: yarn --frozen-lockfile
 
-      - name: Bootstrap
-        run: yarn install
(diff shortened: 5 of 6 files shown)


.github/CONTRIBUTING.md
```diff
@@ -72,7 +72,6 @@ In case you are suggesting a new feature, we will match your idea with our curre
 - Bootstrap all the submodules before building for the first time
```
taranek, line 72: "Bootstrap all the submodules" needs to be removed, will do it later

package.json
```diff
@@ -22,6 +22,7 @@
     "./packages/*"
   ],
   "scripts": {
+    "prune:deps": "find . -name 'node_modules' -type d -prune -exec rm -rf '{}' +",
```
taranek, line 25: That's just a helper command to remove all `node_modules` directories nested inside workspaces, we can either remove or keep it if that could somehow be useful later

evenstensberg: Good

snitin315: Does this command work on windows as well?

alexander-akait:
Just intresting why not use npm/yarn/pnpm workspace feature?

taranek:
@alexander-akait I am using `yarn workspaces` feature here - most of the setup was already done, as `lerna` uses yarn workspaces under the hood.

alexander-akait:
Looks good for me

alexander-akait:
The main problem will be here https://github.com/webpack/webpack-cli/blob/adfdc51bd24b24d3e851904ee3751939438aa4c0/package.json#L46

taranek:
@alexander-akait Yes, it seems like we'll need an external dependency if we want to preserve the current flow, let me describe it in detail by breaking down `lerna.json` configuration file:
```typescript
{
 // ... rest of the config
  "command": {
    "version": {
      "message": "chore(release): publish new version", // 1. commit message template to use when publishing
      "conventionalCommits": true, // 2. generate `CHANGELOG.md` file based on commits included in the current version
      "allowBranch": ["master", "next"] // 3. allow versioning only on particular branch name
    }
  }
}
```

Whereas features 1. and 3. can be achieved via `preversion` or `postverion` npm package hooks or simple bash scripts, we can't generate changelog without adding a dependency.
We can either install a small library to generate changelog and handle the release process by ourselves or introduce a  bigger dependency, like e.g. [semantic release](https://www.npmjs.com/package/semantic-release) which can handle both prerelease, postrelease hooks along with generating changelog and determining webpack version.

@alexander-akait What do you think? Which option seems to be better?

alexander-akait:
@taranek Good question, I haven't studied in detail but isn't there built-in mechanism in yarn workspaces? I thin it is possible:
- use small library to generate changelog
- run built-in yarn workspace feature to do release

taranek:
@alexander-akait Yeah, I was mostly concerned about generating changelog files, which I guess it's not possible in yarn workspaces. I'll have a second look on the current publishing flow and will propose a suitable solution with a small dependency.

alexander-akait:
:+1: Also we can create small script to generate changelog (even more we can copy code from lerna to achive it)

taranek:
@alexander-akait One more thing that we need to take care of is resolving the next release's version number - the `lerna version` command does it by using the conventional commits strategy. TBH I'm not sure if it's worth solving these problems (generating changelog, versioning, publishing) by ourselves/ 

I'd rather go with an external dependency instead of reinventing the wheel - it's just an additional code we'll need to maintain. Also, our release setup is one of the simplest possible - I would understand creating customised solution in order to improve performance of the releasing process, but in this case, I think there's really no need to do so 🤔

snitin315:
I suppose we can use changesets for managing reeases

alexander-akait:
@taranek 

> I'd rather go with an external dependency instead of reinventing the wheel - it's just an additional code we'll need to maintain. 

Sounds good

rishabh3112 (changes requested):
@webpack/cli-team I think we don't need to remove `lerna` now, as it is now being maintained by Nrwl.

Discussion for more info: https://github.com/lerna/lerna/discussions/3140

rishabh3112:
ping @alexander-akait

alexander-akait:
@rishabh3112 I am still thinking we should replace lerna on built-in workspace feature of npm/yarn/pnpm, it is faster, more stable and ligth

rishabh3112:
>  I am still thinking we should replace lerna on built-in workspace feature of npm/yarn/pnpm, it is faster, more stable and 

Okay, Makes sense. @webpack/cli-team  lets keep this open to explore more solutions into this.

alexander-akait:
Let's close and maybe revisit it late, I don't think we have a lot of time to migrate, anyway I think we will do migration when we decide to use update yarn to the latest version or migrate on other package manager, because workspaces are built-in now

Issue #3284, opened 2022-06-08
Variables passed to `--env` with an empty value are incorrectly parsed


benjaminblack:
**Describe the bug**

When passing env vars to webpack cli using `--env`, if the value of the variable passed is empty, then the resulting variable name and value is parsed incorrectly. As an example, `webpack --env TEST=""` results in the following `env`: 

```
{
  WEBPACK_BUNDLE: true,
  WEBPACK_BUILD: true,
  'TEST=': true
}
```

**What is the current behavior?**

Empty variables passed to `--env` like `--env VAR=""` or `--env VAR=$NOT_DEFINED` result in the `env` object containing property names like `VAR=` with  the value `true`.

**To Reproduce**

Minimal webpack.config.js demonstrating behavior:

```js
import { fileURLToPath } from "node:url";

export default (env) => {
  console.log(env);

  return {
    mode: "production",
    entry: "./main.js",
    output: {
      path: fileURLToPath(new URL(".", import.meta.url)),
      filename: "main.bundle.js",
    },
  };
};
```

1.
```
$ npx webpack --env TEST1="" --env TEST2=$NON_EXISTENT_ENVVAR --env TEST3=
```

**Expected behavior**

Empty variables could be represented as one of `null`, `undefined`, or `""`, but certainly should not be `true`, and the property name should not include the `=`.

**Screenshots**

N/A

**Please paste the results of `npx webpack-cli info` here, and mention other relevant information**

```bash
npx webpack --env TEST1="" --env TEST2=$NON_EXISTENT_ENVVAR --env TEST3=
{
  WEBPACK_BUNDLE: true,
  WEBPACK_BUILD: true,
  'TEST1=': true,
  'TEST2=': true,
  'TEST3=': true
}
asset main.bundle.js 50 bytes [compared for emit] [minimized] (name: main)
./main.js 28 bytes [built] [code generated]
webpack 5.73.0 compiled successfully in 82 ms
```

**Additional context**

None

alexander-akait:
PR welcome

snitin315:
But yes I feel if it ends with `=` we should set it to undefined instead of setting `foo=: true`. I'll raise a patch for this.

benjaminblack:
I'm not invoking Webpack through the terminal, I'm doing it through package.json.

With this script:
```
 "webpack": "webpack --env TEST=\"\""
```

The result is a property of `TEST=` with a value of `true`.

With this:

```
[...]

PR #3286, opened 2022-06-09
fix: improve parsing of `--env` flag

snitin315:

bugfix.

Yes

I will add a note on webpack.js.org

**Summary**

Fix https://github.com/webpack/webpack-cli/issues/3284

Set the variable to `undefined` instead of setting it to `true`.

```
$ npx webpack --env TEST1="" --env TEST2=$NON_EXISTENT_ENVVAR --env TEST3=
{
  WEBPACK_BUNDLE: true,
  WEBPACK_BUILD: true,
  'TEST1': '',
  'TEST2': undefined,
  'TEST3': undefined
}
asset main.bundle.js 50 bytes [compared for emit] [minimized] (name: main)
./main.js 28 bytes [built] [code generated]
webpack 5.73.0 compiled successfully in 82 ms
```

No

**Other information**
No

### DIFF
--- a/packages/webpack-cli/src/webpack-cli.ts
+++ b/packages/webpack-cli/src/webpack-cli.ts
@@ -756,13 +756,13 @@ class WebpackCLI implements IWebpackCLI {
           value: string,
           previous: Record<string, BasicPrimitive | object> = {},
         ): Record<string, BasicPrimitive | object> => {
-          // for https://github.com/webpack/webpack-cli/issues/2642
-          if (value.endsWith("=")) {
-            value.concat('""');
-          }
-
           // This ensures we're only splitting by the first `=`
-          const [allKeys, val] = value.split(/=(.+)/, 2);
+          let [allKeys, val] = value.split(/=(.+)/, 2);
+          if (typeof val === "undefined" && allKeys.endsWith("=")) {
+            // remove '=' from key
+            allKeys = allKeys.slice(0, -1);
+            val = "undefined";
+          }
           const splitKeys = allKeys.split(/\.(?!$)/);
 
           let prevRef = previous;
@@ -777,8 +777,11 @@ class WebpackCLI implements IWebpackCLI {
             }
 
             if (index === splitKeys.length - 1) {
-              if (typeof val === "string") {
+              if (typeof val === "string" && val !== "undefined") {
                 prevRef[someKey] = val;
+              } else if (val === "undefined") {
+                // @ts-expect-error we explicitly want to set it to undefined
+                prevRef[someKey] = undefined;
               } else {
                 prevRef[someKey] = true;
               }
--- a/packages/webpack-cli/src/webpack-cli.ts
+++ b/packages/webpack-cli/src/webpack-cli.ts
@@ -757,17 +757,21 @@ class WebpackCLI implements IWebpackCLI {
           previous: Record<string, BasicPrimitive | object> = {},
         ): Record<string, BasicPrimitive | object> => {
           // This ensures we're only splitting by the first `=`
-          let [allKeys, val] = value.split(/=(.+)/, 2);
-          if (typeof val === "undefined" && allKeys.endsWith("=")) {
-            // remove '=' from key
-            allKeys = allKeys.slice(0, -1);
-            val = "undefined";
-          }
+          const [allKeys, val] = value.split(/=(.+)/, 2);
           const splitKeys = allKeys.split(/\.(?!$)/);
 
           let prevRef = previous;
 
           splitKeys.forEach((someKey, index) => {
+            // https://github.com/webpack/webpack-cli/issues/3284
+            if (someKey.endsWith("=")) {
+              // remove '=' from key
+              someKey = someKey.slice(0, -1);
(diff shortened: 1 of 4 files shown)


alexander-akait (commented):
Let's add test case ``--foo=undefined`, in this case we should have `undefined` string value

test/build/config/type/function-with-env/function-with-env.test.js
```diff
@@ -142,11 +142,35 @@ describe("function configuration", () => {
...
+      const { exitCode, stderr, stdout } = await run(__dirname, ["--env", `foo=undefined`]);
+
+      expect(exitCode).toBe(0);
+      expect(stderr).toBeFalsy();
+      // should log foo: 'undefined'
+      expect(stdout).toContain("foo: 'undefined'");
+      // Should generate the appropriate files
+      expect(existsSync(resolve(__dirname, "./dist/undefined-foo.js"))).toBeTruthy();
+    });
+  }
```
alexander-akait, line 172: And lets add
```
webpack --env TEST=""
```
```
webpack --env TEST=\\\"\\\"
```

To avoid future regressions

snitin315: @alexander-akait Done ✅

rishabh3112 (approved):
Looks good, left a suggestion.

test/build/config/type/function-with-env/function-with-env.test.js
```diff
@@ -62,6 +63,7 @@ describe("function configuration", () => {
     expect(exitCode).toBe(0);
     expect(stderr).toBeFalsy();
     expect(stdout).toBeTruthy();
```
rishabh3112, line 65:
```suggestion

```
Redundant check

snitin315: Good catch, thanks!

packages/webpack-cli/src/webpack-cli.ts
```diff
@@ -756,18 +756,22 @@ class WebpackCLI implements IWebpackCLI {
...
 
           let prevRef = previous;
 
           splitKeys.forEach((someKey, index) => {
+            // https://github.com/webpack/webpack-cli/issues/3284
+            if (someKey.endsWith("=")) {
+              // remove '=' from key
+              someKey = someKey.slice(0, -1);
+              // @ts-expect-error we explicitly want to set it to undefined
+              prevRef[someKey] = undefined;
```
harish-sethuraman, line 771: I don't think we might want to set `TEST1=""` to be `TEST1: undefined`? The value exists as empty string? It is not left undefined like other cases? thoughts @alexander-akait?

snitin315: `TEST1=""` is not undefined. `TEST1=` see test cases.

snitin315: https://github.com/webpack/webpack-cli/pull/3286/files#diff-e4aa9d12f691761ac121306d5d8b35208d9438d28ab8c1e093ac14703390e673R129

rishabh3112 (approved):
lgtm

packages/webpack-cli/src/webpack-cli.ts
harish-sethuraman: Screenshot 2022-06-10 at 9 36 12 AM

Any Idea on what is happening? I ran it in this branch only?  `TEST1=""` and the env logs TEST1 to be undefined although it was set to be empty string?

snitin315: @harish-sethuraman In the shell, you need to escape `"` character.

`npx webpack --env TEST=\"\"`

[See this test case](https://github.com/webpack/webpack-cli/pull/3286/files#diff-e4aa9d12f691761ac121306d5d8b35208d9438d28ab8c1e093ac14703390e673R166-R176)

harish-sethuraman: Yea If we escape it I get an empty string within a string `TEST1: '""'` which is expected since we escape. But for cases where we don't (TEST="") Consider if Im doing checks with length like `env.TEST1.length === 0` it is going to fail since there it is undefined. Im not sure if this is the expected but I feel we could improve here.

This is what 
When I set the `VITE_SOME_KEY=""` this is what I get in the env object but in webpack it is undefined
Screenshot 2022-06-10 at 9 51 28 AM

Even when I ran commands like TEST="" in the cli and logged the process.env I see TEST: '' and not TEST: undefined


Feel free to resolve if that is the expected behaviour

test/build/config/type/function-with-env/webpack.env.config.js
```diff
@@ -1,4 +1,5 @@
 module.exports = (env) => {
+  console.log(env);
```
harish-sethuraman, line 2: PS: Is this expected 👀

snitin315: Yes, to check env values in test cases on stdout.

evenstensberg (approved):

alexander-akait (approved):
Looks good, we need a patch release after merge

PR #3288, opened 2022-06-10
feat: add summary to smoketests

rishabh3112:
feature

Yes

No Need

**Summary**
It is difficult to detect which test passed and which one failed on CI. Added a summary at the end of run to make it easier.
image

Nope

**Other information**
Nope

### DIFF
--- a/smoketests/index.js
+++ b/smoketests/index.js
@@ -11,6 +11,7 @@ const tests = [
 
 (async () => {
   let isAllPassed = true;
+  const allResults = [];
 
   for await (const test of tests) {
     console.log(`\nRUN  ${test.name}`);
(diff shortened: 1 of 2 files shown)


smoketests/index.js
```diff
@@ -11,6 +11,8 @@ const tests = [
 
 (async () => {
   let isAllPassed = true;
+  const passResults = [];
+  const failResults = [];
 
   for await (const test of tests) {
     console.log(`\nRUN  ${test.name}`);
```
snitin315, line 18: Can we make this bold and colorful using chalk, to make it more prominent?

rishabh3112: Jest output in CI is uncolored but is readable as it gives summary in the end (see screenshot below). So, I think no need.

image

snitin315: Actually, so many output logs confused me at first glance

Screenshot 2022-06-10 at 11 07 51 AM

rishabh3112 (commented):
Updated output, Looks like one below:
image

smoketests/index.js
rishabh3112: > I think we can update this line:

No we should not. We disable colors by default because it breaks snapshots because of uneven color unicode that color text in different OS. This would potentially break many tests.

rishabh3112: > Actually, so many output logs confused me at first glance

That is why I am adding summary now 😄

snitin315: Oh! got it. I was looking at the wrong logs.

>This would potentially break many tests.

I meant update something like this only for smoketest - `if (!expect.getState().testPath.includes("colors.test.js") || !expect.getState().testPath.includes("smoketest")) { `


But yes, not needed. I was looking at the wrong logs.

snitin315 (approved):
LGTM

PR #3317, opened 2022-06-29
feat: failOnWarnings option

### LINKED ISSUE #3315: new option to terminate with non-zero code if builded with warnings

related to https://github.com/webpack/webpack/issues/15960#issuecomment-1168468320

One possible solution could be - add an option to terminate process with non-zero code if any warning exist. e.g. `--strict` option ( `false` by default )

**Additional context**

rishabh3112:
feature

Yes

Done

**Summary**
adds `--fail-on-warnings` option
closes #3315 

No

**Other information**
#3315

### DIFF
--- a/packages/webpack-cli/src/webpack-cli.ts
+++ b/packages/webpack-cli/src/webpack-cli.ts
@@ -976,6 +976,16 @@ class WebpackCLI implements IWebpackCLI {
         description: "Stop watching when stdin stream has ended.",
         negatedDescription: "Do not stop watching when stdin stream has ended.",
       },
+      {
+        name: "fail-on-warnings",
+        configs: [
+          {
+            type: "boolean",
+          },
+        ],
+        negative: false,
+        description: "Stop webpack-cli process on non-zero exit code",
+      },
     ];
 
     // Extract all the flags being exported from core.
@@ -2476,6 +2486,10 @@ class WebpackCLI implements IWebpackCLI {
         // Avoid extra empty line when `stats: 'none'`
         if (printedStats) {
           this.logger.raw(printedStats);
+        }
+
+        if (options.failOnWarnings && stats.hasErrors()) {
+          process.exit(1);
         }
       }
     };
--- a/packages/webpack-cli/src/webpack-cli.ts
+++ b/packages/webpack-cli/src/webpack-cli.ts
@@ -981,7 +981,7 @@ class WebpackCLI implements IWebpackCLI {
         configs: [
           {
             type: "enum",
-            values: [false],
+            values: [true],
           },
         ],
         description: "Stop webpack-cli process with non-zero exit code on warnings from webpack",
--- a/packages/webpack-cli/src/webpack-cli.ts
+++ b/packages/webpack-cli/src/webpack-cli.ts
@@ -984,7 +984,6 @@ class WebpackCLI implements IWebpackCLI {
             values: [false],
           },
         ],
-        negative: true,
         description: "Stop webpack-cli process with non-zero exit code on warnings from webpack",
       },
     ];
@@ -2427,7 +2426,7 @@ class WebpackCLI implements IWebpackCLI {
         process.exit(2);
       }
 
-      if (stats && stats.hasErrors()) {
+      if (stats && (stats.hasErrors() || (options.failOnWarnings && stats.hasWarnings()))) {
         process.exitCode = 1;
       }
 
@@ -2488,10 +2487,6 @@ class WebpackCLI implements IWebpackCLI {
         if (printedStats) {
           this.logger.raw(printedStats);
         }
-      }
-
-      if (options.failOnWarnings && (stats.hasErrors() || stats.hasWarnings())) {
-        process.exit(1);
       }
     };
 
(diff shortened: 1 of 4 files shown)


packages/webpack-cli/src/webpack-cli.ts
```diff
@@ -976,6 +976,16 @@ class WebpackCLI implements IWebpackCLI {
         description: "Stop watching when stdin stream has ended.",
         negatedDescription: "Do not stop watching when stdin stream has ended.",
       },
+      {
+        name: "fail-on-warnings",
+        configs: [
+          {
+            type: "boolean",
+          },
```
snitin315, line 984: Should we allow it to have a `number` type as well so that users can use their own exit code? Or we want to force the exit code as `1`.

`webpack --fail-on-warnings 2`

rishabh3112: Thanks for review, but I don't think we should. We have only 2 non-zero exit codes.
For exit code 2, we exit immediately by default. This option enforces immediate exit for `exitCode == 1`

snitin315: That sound reasonable to me 👍🏻. We can always support that in the future if required.

packages/webpack-cli/src/webpack-cli.ts
```diff
@@ -2478,6 +2488,10 @@ class WebpackCLI implements IWebpackCLI {
           this.logger.raw(printedStats);
         }
       }
+
+      if (options.failOnWarnings && (stats.hasErrors() || stats.hasWarnings())) {
+        process.exit(1);
```
vankop, line 2493:
```suggestion
        process.exitCode = 1;
```
I think we should terminate gracefully

vankop (changes requested):

packages/webpack-cli/src/webpack-cli.ts
```diff
@@ -976,6 +976,16 @@ class WebpackCLI implements IWebpackCLI {
         description: "Stop watching when stdin stream has ended.",
         negatedDescription: "Do not stop watching when stdin stream has ended.",
       },
+      {
+        name: "fail-on-warnings",
+        configs: [
+          {
+            type: "boolean",
```
snitin315, line 983: Because we don't want `--no-fail-on-warnings`
```suggestion
            type: "enum",
            values: [false],
```

packages/webpack-cli/src/webpack-cli.ts
```diff
@@ -2478,6 +2488,10 @@ class WebpackCLI implements IWebpackCLI {
           this.logger.raw(printedStats);
         }
       }
+
+      if (options.failOnWarnings && (stats.hasErrors() || stats.hasWarnings())) {
```
snitin315, line 2492: We should merge this with the condition above - https://github.com/webpack/webpack-cli/pull/3317/files#diff-fbc9bd0ae2f54d7f45607dc7979f8c1721de2e9ccf4a5934504bab29e7204805R2429

packages/webpack-cli/src/webpack-cli.ts
rishabh3112: @vankop I have a doubt, If we not exit here, process will not terminate on warning/errors from webpack in watch mode (with option named as `--fail-on-warnings`). Will this be accurate behavior?

packages/webpack-cli/src/webpack-cli.ts
rishabh3112: Yeah, I was doing that only at first but this - https://github.com/webpack/webpack-cli/pull/3317#discussion_r909393891

packages/webpack-cli/src/webpack-cli.ts
rishabh3112: Okay great, will update this then 👍🏽

packages/webpack-cli/src/webpack-cli.ts
```diff
@@ -976,6 +976,17 @@ class WebpackCLI implements IWebpackCLI {
...
       },
+      {
+        name: "fail-on-warnings",
+        configs: [
+          {
+            type: "enum",
+            values: [false],
+          },
+        ],
+        negative: true,
```
snitin315, line 987: This is misleading, it should be `false` if we don't want `--no-fail-on-warnings`. We actually prefer the `configs` property over this `negative` property so it will be ignored (like it was previously). So you can just remove it.

```suggestion
```

packages/webpack-cli/src/webpack-cli.ts
```diff
@@ -976,6 +976,17 @@ class WebpackCLI implements IWebpackCLI {
         description: "Stop watching when stdin stream has ended.",
         negatedDescription: "Do not stop watching when stdin stream has ended.",
       },
+      {
+        name: "fail-on-warnings",
+        configs: [
+          {
+            type: "enum",
+            values: [false],
```
snitin315, line 984: Sorry, typo in my earlier suggestion. it should be true.
```suggestion
            values: [true],
```

vankop (approved):

alexander-akait (approved):

PR #3342, opened 2022-07-19
refactor!: the minimum supported webpack version is v5.0.0

snitin315:
Breaking change.

Already present.

No

**Summary**

The minimum supported webpack version is `v5.0.0`.

BREAKING CHANGE: 
- webpack-cli no longer supports webpack v4, the minimum supported version is webpack v5.0.0
- Reset flags like `--target-reset`, `--entry-reset` flags are required to be used before using original flags like `--target`, `--entry`.

**Other information**
No


alexander-akait (commented):
Let's separete this PR on parts to avoid regressions:
1. Remove old options
2. Remove workaround for webpack v4

Or verca vice

snitin315:
@alexander-akait I've created https://github.com/webpack/webpack-cli/pull/3346 to remove workarounds, I'll update this PR to remove old options.

snitin315:
@alexander-akait Done ✅

test/build/core-flags/core-flags.test.js
```diff
@@ -239,15 +239,17 @@ describe("core flags", () => {
       expect(stdout).toContain(`devtool: 'source-map'`);
     });
 
-    it("should allow string value devtool option using alias", async () => {
+    // TODO: Enable alias with webpack 5
+    it.skip("should allow string value devtool option using alias", async () => {
```
alexander-akait, line 243: Why skip?

test/help/__snapshots__/help.test.js.snap.devServer3.webpack5
```diff
@@ -102,20 +102,20 @@ Options:
...
-  -o, --output-path <value>                        Output location of the file generated by webpack e.g. ./dist/.
-  --stats [value]                                  It instructs webpack on how to treat the stats e.g. verbose.
-  --no-stats                                       Disable stats output.
-  -t, --target <value...>                          Sets the build target e.g. node.
+  --output-path <value>                            The output directory as **absolute path** (required).
+  --stats [value]                                  Stats options object or preset name.
+  --no-stats                                       Negative 'stats' option.
+  --target <value...>                              Environment to build for. Environment to build for. An array of environments to build for all of them when possible.
   --no-target                                      Negative 'target' option.
-  -w, --watch                                      Watch for files changes.
```
alexander-akait, line 115: I think keeping short `-w`/`-d` and ethc make sense, they are very popular, let's keep them, what do you think?

test/build/core-flags/core-flags.test.js
snitin315: We don't have alias property coming from webpack.

https://github.com/webpack/webpack/blob/8f87b50dc7ac24eb5c91fd0d55a22e34e252863c/test/__snapshots__/Cli.basictest.js.snap#L416

We had defined alias for only few config options which we had for webpack 4 which we removed in this PR.

test/help/__snapshots__/help.test.js.snap.devServer3.webpack5
snitin315: Yes, we can add `alias` support for some popular flags `devtool (-d)`, `target (-t)`, `--output-path (-o)`, `--watch (-w)`

snitin315: Ideally, we should add this in the webpack schema itself. 
Or we can write logic to generate an alias property for each option in webpack-cli. 
Or add some workaround logic to support alias for some popular options only (coming from webpack core).

alexander-akait: > Or we can write logic to generate an alias property for each option in webpack-cli.

I think so, it was my original idea, so we will support all options using short names, we can do it separatly

snitin315: Yes. Let's add them separately. In the meantime should I update the PR to keep the popular ones?

alexander-akait: Yes, let's keep popular options here, some of them we can kept for better compatibility

snitin315: @alexander-akait Added in https://github.com/webpack/webpack-cli/pull/3342/commits/00fae2fca6baa314b46e6207fd9bc54b141cd5f9

alexander-akait (approved):
Looks good

Issue #3436, opened 2022-10-05
Refactor(cli): Migrate from CJS to ESM 


evenstensberg:
**Is your feature request related to a problem? Please describe.**
As node evolves, we need to change from CJS to ESM for webpack-cli.

**Describe the solution you'd like**

N/A

**Describe alternatives you've considered**

N/A
**Additional context**

This issue originates from @alexander-akait 's feedback.

nyngwang:
@evenstensberg @alexander-akait:

Does this refactor mean that we can write `webpack.config.js` in ESM without relying on the package.json `"type": "module"`? There is currently some serious bug on TypeScript's new option `"moduleResolution": "bundler"` when `"type": "module"` is set, and it's clearly stated on their documentation regarding working with bundler: https://www.typescriptlang.org/docs/handbook/modules/guides/choosing-compiler-options.html#im-using-a-bundler.

> In addition to adopting the following settings, it’s also recommended not to set { "type": "module" } or use .mts files in bundler projects for now. [Some bundlers](https://andrewbranch.github.io/interop-test/#synthesizing-default-exports-for-cjs-modules) adopt different ESM/CJS interop behavior under these circumstances, which TypeScript cannot currently analyze with "moduleResolution": "bundler". See [issue #54102](https://github.com/microsoft/TypeScript/issues/54102) for more information.

I'm here to check whether this issue can remove the need for `"type": "module"`. The related issue of TypeScript#54102 is https://github.com/webpack/webpack/issues/17288.

nyngwang:
Does this refactor mean that we can write `webpack.config.js` in ESM without relying on the package.json `"type": "module"` because an ESM can easily import another ESM?

## Background

There is currently some serious bug on TypeScript's new option `"moduleResolution": "bundler"` when `"type": "module"` is set, and it's clearly stated on their documentation regarding working with bundler: https://www.typescriptlang.org/docs/handbook/modules/guides/choosing-compiler-options.html#im-using-a-bundler.

> In addition to adopting the following settings, it’s also recommended not to set { "type": "module" } or use .mts files in bundler projects for now. [Some bundlers](https://andrewbranch.github.io/interop-test/#synthesizing-default-exports-for-cjs-modules) adopt different ESM/CJS interop behavior under these circumstances, which TypeScript cannot currently analyze with "moduleResolution": "bundler". See [issue #54102](https://github.com/microsoft/TypeScript/issues/54102) for more information.

[...]

PR #3439, opened 2022-10-08
refactor: migrate `configtest` package to ESM

snitin315:
refactor.

No
No
**Summary**
Start ESM migration.

No

**Other information**
No


evenstensberg (commented):
Maybe draft these ESM converted packages and group them up, until every package is done migrated?

snitin315:
Yes, I'll convert this to a draft.

alexander-akait:
@snitin315 I think in this PR we need to convert all to ESM (or alternative we can use `import(...)`) to load extensions for webpack-cli, so they will work

packages/configtest/package.json
```diff
@@ -3,7 +3,12 @@
   "version": "1.2.0",
   "description": "Validate a webpack configuration.",
   "type": "module",
-  "exports": "./lib/index.js",
+  "exports": {
+    "./package.json": "./package.json",
+    ".": {
+      "import": "./dist/index.js"
```
alexander-akait, line 9: Always add `default` as fallback :+1:

snitin315: Yes, webpack v5 is using [require](https://github.com/webpack/webpack/blob/9fcaa243573005d6fdece9a3f8d89a0e8b399613/bin/webpack.js#L69) as well.

snitin315:
@alexander-akait I suppose if we move to ESM it will break compatibility with webpack v5 too. Because [webpack 5 is using cli with `require`](https://github.com/webpack/webpack/blob/9fcaa243573005d6fdece9a3f8d89a0e8b399613/bin/webpack.js#L69)

Screenshot 2022-11-06 at 4 05 28 PM

alexander-akait:
What about if we will use `import(...)`? It should work

snitin315:
@alexander-akait yes but it means from webpack v5.0.0 to the current latest version of webpack will not work with this major version. It will be compatible only with future version of webpack which is using import().

snitin315:
@alexander-akait Yes `import(...)` will work. But it means users will have to update webpack to the latest version which includes this PR https://github.com/webpack/webpack/pull/16448

alexander-akait:
@snitin315 Oh, I got it, yeah, let's postpone it then, and mark it for webpack v6, we will need to do it for webpack-dev-server too

snitin315:
@alexander-akait yes, let's postpone.

alexander-akait:
Let's close in favor https://github.com/webpack/webpack-cli/issues/3436, I think we will start migration everywhere on ESM when Node.js@18 will be outdated, because there are some bugs in Node.js@18 with ESM and especial with `import(...)`, we can make a pain for many developers if we do it right now, anyway thank you

PR #3514, opened 2022-12-02
fix: make define-process-env-node-env alias node-env

### LINKED ISSUE #3503: `--define-process-env-node-env` doesn't do as advertised
**Describe the bug**

This is a follow up from [this comment thread](https://github.com/webpack/webpack-cli/pull/3456#issuecomment-1328176668). `--node-env` was renamed in `webpack-cli` v5 with [this comment](https://github.com/webpack/webpack-cli/pull/3279#issuecomment-1280763714) linked as the reasoning.

> I wanted to rename `node-env`, because it is confising, many developers using `node-env` think that it will change `process.env.NODE_ENV` inside `webpack.config.js`, but it only change `process.env.NODE_ENV` in code, that is why I think we need another name for this.

This is not true, not in the previous version and not in the new version. See "To Reproduce" section below for a very minimal demo proving this isn't true.

`process.env.NODE_ENV` is set to the value of `--node-env`/`--define-process-env-node-env` within both `webpack.config.js` and the rest of the application.

**To Reproduce**

These are the only two files needed to reproduce. Install the dependencies and run either command and see the value logged.

Gist: https://gist.github.com/CreativeTechGuy/ae6ec05ee9ec5cb71807f2ff74a99c41

**Expected behavior**

There's a conflict. `--define-process-env-node-env` is named as such to (attempt to) make it clear that it doesn't set the environment variable within `webpack.config.js`. That precondition doesn't seem to be true. So the previous name `--node-env` is much simpler and easier to understand and does do what you'd expect it to do.

CreativeTechGuy:

Restores v4's `--node-env` CLI flag while preserving the new `--define-process-env-node-env` flag.

**Did you add tests for your changes?** Yup

**If relevant, did you update the documentation?** Yup (although depending on what the desired future is for these flags, possibly the [web documentation](https://webpack.js.org/api/cli/#node-env) needs to be updated too).

**Summary**

Fix #3503. Since the full fix discussed in that thread would be breaking, the interim solution is to restore the `--node-env` flag and have both `--define-process-env-node-env` and `--node-env` do the same thing. In the future these may diverge as discussed here: https://github.com/webpack/webpack-cli/issues/3503#issuecomment-1329037767

No. This was very intentionally done in this way to not create a breaking change. It seems unclear as to what the desired long-term behavior is so for now there are two alias commands.

**Other information**

N/A


packages/webpack-cli/src/webpack-cli.ts
```diff
@@ -2177,6 +2188,8 @@ class WebpackCLI implements IWebpackCLI {
   ): Promise<WebpackCompiler> {
     if (typeof options.defineProcessEnvNodeEnv === "string") {
       process.env.NODE_ENV = options.defineProcessEnvNodeEnv;
```
alexander-akait, line 2190: Ideally we need to remove this line, because `defineProcessEnvNodeEnv` should set `NODE_ENV` only for runtime, but I am afraid some people can migrate on `define-process-env-node-env` and we can do breaking change for them, so can you add `TODO` here and we will solve it in the next breaking change, thank you

CreativeTechGuy: Yeah definitely. I didn't want to make a breaking change on the thing that was literally just added! I'll add a TODO :)

alexander-akait (approved):
/cc @webpack/cli-team Please review too

snitin315 (approved):

evenstensberg (approved):

PR #3659, opened 2023-03-11
feat: add `WEBPACK_HOT` environment variable

snitin315:
feature

No.

**Summary**

Resolve https://github.com/webpack/webpack-dev-server/pull/4734

Add `WEBPACK_HOT: 'true'`, when HMR is enabled.

No

**Other information**
No

### DIFF
--- a/packages/serve/src/index.ts
+++ b/packages/serve/src/index.ts
@@ -225,6 +225,16 @@ class ServeCommand {
             usedPorts.push(portNumber);
           }
 
+          // Set WEBPACK_HOT env variable to "true" if hot is specified in devServer options or via CLI
+          // If not specified, default to "true" as it is the default value for the hot option in devServer
+          if (typeof devServerOptions.hot !== "undefined") {
+            process.env.WEBPACK_HOT = String(
+              devServerOptions.hot === true || devServerOptions.hot === "only",
+            );
+          } else {
+            process.env.WEBPACK_HOT = "true";
+          }
+
           try {
             const server = new DevServer(devServerOptions, compiler);
 
--- a/test/serve/basic/serve-basic.test.js
+++ b/test/serve/basic/serve-basic.test.js
@@ -19,6 +19,7 @@ describe("basic serve usage", () => {
 
     expect(normalizeStderr(stderr)).toMatchSnapshot("stderr");
     expect(stdout).toContain("HotModuleReplacementPlugin");
+    expect(stdout).toContain("WEBPACK_HOT: true");
     expect(stdout).toContain("main.js");
   });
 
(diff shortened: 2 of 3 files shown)


snitin315:
cc @alexander-akait

alexander-akait:
@snitin315 On my radar, will look at this today :+1:

packages/serve/src/index.ts
```diff
@@ -225,6 +225,16 @@ class ServeCommand {
...
           }
 
+          // Set WEBPACK_HOT env variable to "true" if hot is specified in devServer options or via CLI
+          // If not specified, default to "true" as it is the default value for the hot option in devServer
+          if (typeof devServerOptions.hot !== "undefined") {
+            process.env.WEBPACK_HOT = String(
+              devServerOptions.hot === true || devServerOptions.hot === "only",
+            );
+          } else {
+            process.env.WEBPACK_HOT = "true";
```
alexander-akait, line 235: It should be early, so if you write `console.log(process.env["WEBPACK_HOT"]);` in `webpack.config.js`, it should be `true`. You need something this:
```js
process.env.WEBPACK_HOT = "true";

const compiler = await cli.createCompiler(webpackCLIOptions);

if (!compiler) {
  return;
}
```

But in other way you can disable `hot` inside `devServer` options, that's why I didn't implement this before, not sure how we can solve it.

We have `env` variables https://webpack.js.org/api/cli/#environment-options, and developer can create commands in npm `scripts` with hot and without hot, and get them in configuration or developer can just add them before `WEBPACK_HOT=true webpack serve`, I think we can't solve it the right way.

What do you think?

snitin315: @alexander-akait I think we can set it at two places. If we have passed `--hot`/`--no-hot` set it to `true`/`false` before loading the config. Then if users have overridden the `devServer.hot` in the config we can again override it after loading the config.

But yes, this behavior can be confusing to developers.

alexander-akait: > @alexander-akait I think we can set it at two places. If we have passed --hot/--no-hot set it to true/false before loading the config. Then if users have overridden the devServer.hot in the config we can again override it after loading the config.

We can, but it doesn't make sense, because you need `WEBPACK_HOT` in `webpack.config.js`, and it will be true in all cases except (`--no-hot`), so I don't see any possibles to solve it valid way, shorty - a developer want to get the value before we can read and evaluate it

And we really have `--env` variables for such purposes

snitin315: Yeah, `--env` makes more sense in this case.

alexander-akait: So let's close and provide examples, don't think we really need it

PR #3738, opened 2023-04-18
feat: shareable webpack configs using `extends`

burhanuday:

feature

Adds `--extends` option to the CLI as well as support for `extends` in webpack config

This PR builds on top of https://github.com/webpack/webpack-cli/pull/2868
Instead of having to update webpack, this takes the approach of modifying the config before webpack is run

yes

no

**Summary**

Closes https://github.com/webpack/webpack-cli/issues/2748

**Other information**


burhanuday:
Before continuing further work. A couple of questions
1. What is the expected behaviour in case of multiple configs. If i use `merge` to then it will merge for `extends`, then everything gets merged into a single config which is not the expected behaviour. should we then limit `extends` to a single config and throw an error when multiple configs are passed?
2. An extended config can also have its own `extends` property. In this case, do we want to recursively keep looking for extends or should we limit to some particular depth?

@alexander-akait @snitin315 Need your insights on this. Thank you

alexander-akait:
> What is the expected behaviour in case of multiple configs. If i use merge for extends, then everything gets merged into a single config which is not the expected behaviour. should we then limit extends to a single config and throw an error when multiple configs are passed?


I think merge it in each other, you can have `webpack.config.base.js` and have multiple configuration which extend base, for example server and client

> An extended config can also have its own extends property. In this case, do we want to recursively keep looking for extends or should we limit to some particular depth?

We can prevent it to keepind array of merged and if it developer has a loop throw an error

snitin315:
> An extended config can also have its own extends property. In this case, do we want to recursively keep looking for extends or should we limit to some particular depth?

>> We can prevent it to keepind array of merged and if it developer has a loop throw an error

I added [a few tests cases](https://github.com/webpack/webpack-cli/pull/2868/files#diff-ed3f02008b20388a38debb35875674ef5e410be3c4d0dce2771f8cf19a129f47) in #2868. I feel those should be covered.

burhanuday:
> > An extended config can also have its own extends property. In this case, do we want to recursively keep looking for extends or should we limit to some particular depth?
> 
> > > We can prevent it to keepind array of merged and if it developer has a loop throw an error
> 
> I added [a few tests cases](https://github.com/webpack/webpack-cli/pull/2868/files#diff-ed3f02008b20388a38debb35875674ef5e410be3c4d0dce2771f8cf19a129f47) in #2868. I feel those should be covered.

Got it. Most of the implementation is done, just need to handle the part where an extended config has its own `extends` property

packages/webpack-cli/src/webpack-cli.ts
```diff
@@ -1950,17 +1950,94 @@ class WebpackCLI implements IWebpackCLI {
...
+            }, {});
+          } else {
+            // load the config from the extends option
+            // @ts-expect-error TODO remove when extends property type is added to webpack
+            extendedConfig = loadConfigByPath(path.resolve(configOptions.extends), options.argv);
+          }
+
+          config.options[index] = merge(extendedConfig, configOptions);
+        }
+      }
```
alexander-akait, line 2006: Please move it to own function and just use `&& { extend: string[] }` to avoid `ts-except-error`

burhanuday:
All checks have passed

alexander-akait:
While refactoring found bugs...

It should work now as expected, also allos to load configurations on any depth + catch recusive calls + setup valid cache invalidation.

But yeah we need more tests here:
- deep extends
- multiple extends in `webpack.config.js`
- multiple `--extends`
- recusive calls
- mix extends in `webpack.config.js` and `--extends` (also check `buildDependencies` to undestand we have all loaded configuration paths)
- test with `--extends` and `--merge` (also check `buildDependencies` to undestand we have all loaded configuration paths)

In theory we can add even more tests, but I think it is enough for basic testing

alexander-akait:
So if somebody have time feel free to add such tests cases

snitin315:
Do we add them in the same PR or can we add them separately?

alexander-akait:
We need to ad them here

alexander-akait (approved):

snitin315:
@burhanuday Would you be willing to send a PR to add this feature to [the CLI docs here](https://webpack.js.org/api/cli/)?

burhanuday:
@snitin315 yup. Will do

burhanuday:
@snitin315 @alexander-akait can we do a release for webpack-cli before webpack release with the type changes get published tomorrow

alexander-akait:
@burhanuday better to do it after webpack release, otherwise we will have types problems, release will be tomorrow, so I will do webpack-cli release tomorrow too

Issue #3747, opened 2023-04-21
implement built-in `.env` support


alexander-akait:
I think we need to implement built-in `.env` support, there is a good plugin https://github.com/mrsteele/dotenv-webpack

**Describe the solution you'd like**

Allow to red env variables from `.env` file and control it using CLI option

**Describe alternatives you've considered**

No need

**Additional context**

No need

silverwind:
How about supporting [vite's files](https://vitejs.dev/guide/env-and-mode.html#env-files)? I think multiple files are almost required if you have default vars checked in git and then have another file to override those vars at build time. If the same files as vite could be supported, it'd be some nice compatibilty.

alexander-akait:
Yeah :+1:

mertcanaltin:
@alexander-akait I want to do this 🚀

silverwind:
@webpack-bot keep open.

MasterOwnerETcorps:
¶`<--'"<!m>={" create&merge:"} ["new pull request;"] 'On', ("#/.mpret").-->`

urizennnn:
Can i work on this too?

Sakib25800:
@alexander-akait, it looks like this feature was close to completion in #3759 but was closed due to inactivity. Could I perhaps submit a new PR picking up from where it was left off previously?

alexander-akait:
@Sakib25800 Yes, you can

alexander-akait:
To be honestly not sure we really need this right now, because we have https://nodejs.org/en/learn/command-line/how-to-read-environment-variables-from-nodejs, I think we need update our docs and that's enough, we provice an example how to use them in `webpack.config.js` and as a build time variables using `DefinePlugin({})`

And both things already work good, but only for `Node.js 20`

So let's update documentation and that's all

alexander-akait:
@snitin315 Do we have example to how to using `.env` files here or on our website?

silverwind:
> To be honestly not sure we really need this right now, because we have https://nodejs.org/en/learn/command-line/how-to-read-environment-variables-from-nodejs

Node's dotenv support is limited to a single file though. Often it's desirable to support two files, one checked into git, and one not. Vite does support `.env` and `.env.local` for  this, see https://vite.dev/guide/env-and-mode#env-files. Also https://github.com/motdotla/dotenv-expand is a very useful addon that is not supported by node's loader.

alexander-akait:
@silverwind 

[...]

PR #3751, opened 2023-04-24
feat: add built-in .env support from dotenv-webpack plugin

mertcanaltin:
@alexander-akait I tried to make an experiment 
I would be very happy if you have a comment

issue:#3747

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -22,6 +22,7 @@ module.exports = {
     "no-loop-func": "error",
     "no-undef": "error",
     "prefer-const": "error",
+    "node/no-unpublished-require": "off",
   },
   overrides: [
     {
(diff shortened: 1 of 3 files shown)


alexander-akait:
Sorry, we don't want reuse `dotenv-plugin`, we want to implement own deep integrated (with CLI) plugin with:
- cache support to avoid bad perf
- ESM support out of box, i.e. `import.meta.env`
- flexible loading logic
- good logging

and other features, so you don't need to install plguins to have `.env` support.

We started it https://github.com/webpack/webpack-cli/pull/3759, so let's close, anyway thank you, if you want to be a part of development feel free to write your ideas in the issue and here https://github.com/webpack/webpack-cli/pull/3759

PR #3761, opened 2023-04-28
`json` parameter description clarified

SetTrend:
## What kind of change does this PR introduce?

Help texts improved, so parameters get clear to the reader without being required to looking up additional information.

## Did you add tests for your changes?

No, not necessary. Only text changes provided.

## Summary

When reading the help, I was often wondering when to use parameters and how to use them. Help texts didn't get clear on this.

I wanted to change this, so I won't need to look up the WebPack CLI documentation over and over again in the future.

## Does this PR introduce a breaking change?

No.


alexander-akait (dismissed):

alexander-akait:
Can you accept CLA?

SetTrend:
Sure, just did. Actually, this is still a WIP. I started off with the `--json` parameter but would like to continue with the `--configuration-name` etc.

alexander-akait:
Yeah, feel free to ping me when it will be ready

OPTIONS.md
```diff
@@ -5,16 +5,16 @@ Alternative usage to run commands: webpack [command] [options]
 The build tool for modern web applications.
 
 Options:
-  -c, --config <value...>                                                            Provide path to a webpack configuration file e.g. ./webpack.config.js.
-  --config-name <value...>                                                           Name of the configuration to use.
+  -c, --config <pathToConfigFile...>                                                 Provide path to one or more webpack configuration files to process, e.g. "./webpack.config.js".
+  --config-name <name...>                                                            Name(s) of particular configuration(s) to use if configuration file exports an array of multiple configurations.
```
alexander-akait, line 9: It was generated by our script, so I think we need to improve this inside webpack-cli logic, otherwise it will be replaced in future

SetTrend:
@alexander-akait:

I now updated documentation for `--config`, `--config-name`, `--env`, `--json`, and `--serve` CLI parameters to be more specific and explanatory.

I also took the opportunity to sort the `cspell` dictionary, so terms in the user dictionary may be found more easily. I added two terms (`myvar` and `myval`) to this array.

Well, though I did the best I could, I don't seem to be able to pass all tests. Somehow, the tests are always comparing to the previous installation.

OPTIONS.md
SetTrend: I see, so there's nothing I can do from my side?

alexander-akait: I think you can.

Here we use `[value]`/`value` - https://github.com/webpack/webpack-cli/blob/master/packages/webpack-cli/src/webpack-cli.ts#L678

But you can add a property (I think the best name is `valueDescription`) for https://github.com/webpack/webpack-cli/blob/master/packages/webpack-cli/src/webpack-cli.ts#L805 (and for other) and use `const valueDescription = option.valueDescription ? valueDescription : "value"` (https://github.com/webpack/webpack-cli/blob/master/packages/webpack-cli/src/webpack-cli.ts#L678
)

SetTrend: I'm not sure I understand correctly.

From the code I have the impression that there is a [`description` property already](https://github.com/webpack/webpack-cli/blob/5134eebd9814036c0fac7c8a11f7a6cc211f64bc/packages/webpack-cli/src/webpack-cli.ts#L814).

So, I would need to add an additional property, (e.g. `parameterName`)? And you would amend your script to take this new property into account?

alexander-akait: No, when we register our options https://github.com/webpack/webpack-cli/blob/5134eebd9814036c0fac7c8a11f7a6cc211f64bc/packages/webpack-cli/src/webpack-cli.ts#L679 and https://github.com/webpack/webpack-cli/blob/5134eebd9814036c0fac7c8a11f7a6cc211f64bc/packages/webpack-cli/src/webpack-cli.ts#L681, that is why you see `value` instead `pathToConfigFile`/`name`/etc, we just need to add a new property to our options, for example https://github.com/webpack/webpack-cli/blob/5134eebd9814036c0fac7c8a11f7a6cc211f64bc/packages/webpack-cli/src/webpack-cli.ts#L952, i.e.
```
{
        name: "json",
        configs: [
          {
            type: "string",
          },
          {
            type: "enum",
            values: [true],
          },
        ],
        alias: "j",
        valueName: "pathToFile", // <- HERE
        description: "Prints result as JSON or store it in a file.",
        helpLev
``` 

And use `option.valueName` instead `value` here https://github.com/webpack/webpack-cli/blob/5134eebd9814036c0fac7c8a11f7a6cc211f64bc/packages/webpack-cli/src/webpack-cli.ts#L679 and https://github.com/webpack/webpack-cli/blob/5134eebd9814036c0fac7c8a11f7a6cc211f64bc/packages/webpack-cli/src/webpack-cli.ts#L681

SetTrend:
I updated my pull request accordingly and created a new branch following contribution guidelines.

Unfortunately, that required me to create new pull request #3789, so I'm closing this pull request here in favour of the new one.

PR #3772, opened 2023-05-06
fix: improve check for custom webpack-dev-server package existance

snitin315:
bugfix

No
No
**Summary**

Fix https://github.com/webpack/webpack-cli/pull/3161

No

**Other information**
No

### DIFF
--- a/packages/webpack-cli/src/webpack-cli.ts
+++ b/packages/webpack-cli/src/webpack-cli.ts
@@ -57,8 +57,14 @@ const { pathToFileURL } = require("url");
 const util = require("util");
 const { program, Option } = require("commander");
 
-const WEBPACK_PACKAGE = process.env.WEBPACK_PACKAGE || "webpack";
-const WEBPACK_DEV_SERVER_PACKAGE = process.env.WEBPACK_DEV_SERVER_PACKAGE || "webpack-dev-server";
+const WEBPACK_PACKAGE_IS_CUSTOM = !!process.env.WEBPACK_PACKAGE;
+const WEBPACK_PACKAGE = WEBPACK_PACKAGE_IS_CUSTOM
+  ? (process.env.WEBPACK_PACKAGE as string)
+  : "webpack";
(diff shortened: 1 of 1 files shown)


packages/webpack-cli/src/webpack-cli.ts
```diff
@@ -526,7 +526,10 @@ class WebpackCLI implements IWebpackCLI {
         }
 
         // Allow to use `./path/to/webpack-dev-server.js` outside `node_modules`
-        if (dependency === WEBPACK_DEV_SERVER_PACKAGE && fs.existsSync(WEBPACK_PACKAGE)) {
+        if (
+          dependency === WEBPACK_DEV_SERVER_PACKAGE &&
+          fs.existsSync(WEBPACK_DEV_SERVER_PACKAGE)
+        ) {
           skipInstallation = true;
         }
```
snitin315, line 534: The current behavior is to check for the package and if it doesn't exist we ask users to install it.

Screenshot 2023-05-06 at 8 17 12 AM

If we use `require.resolve()` it will throw an error and exit the process. Should we use it?

alexander-akait: Good question, and another good question why we used `fs.existsSync` here, can you look at git history?

snitin315: @alexander-akait added on this PR https://github.com/webpack/webpack-cli/pull/2907

alexander-akait: @snitin315 I think we can remove this check, if developer provided a bad path, node.js itself generate a good error, so I think it is extra unnecessary check

alexander-akait: Not sure why I was added it :smile:

alexander-akait (commented):
Looks good, let's add a small test case where we have invalid/unknown path to webpack and/or webpack-dev-server, just ensure we throw an error and it is readable

PR #3831, opened 2023-06-03
fix: support `.cts` and `.mts`

alexander-akait:
Some magic:
- handle `.cts` and `.mts`
- `.mts` is still have hacky support - `NODE_OPTIONS="--loader ts-node/esm" webpack-cli --config ./webpack.config.mts`
- `.cts` works fine
- `.ts` works out of box for `CommonJS`, but need a hack from above for `.mts`
- improve perf for `.mts` to avoid using extra `require`

@hardfist there is basic support for `.cts`and `.mts` (`.ts` already supported), no problems with `commonjs`, but `esm` using is still hacky due `rechoir` doesn't support `ESM` hooks for Node.js (because Node.js changed it multiple times recently and it is still not stable).

I want to say it is a `bugfix`

Soon

No need

**Summary**

Better support typescript configurations

No

**Other information**

No


alexander-akait:
And I really hope we will not have `.mjsx`, `.cjsx` (and same for typescript), it will be a disaster :exploding_head:

packages/webpack-cli/src/webpack-cli.ts
```diff
@@ -322,37 +323,83 @@ class WebpackCLI implements IWebpackCLI {
...
+          if (handleError) {
+            this.logger.error(error);
+            process.exit(2);
+          } else {
+            throw error;
+          }
+        }
+        break;
+      }
+      case "esm": {
```
alexander-akait, line 376: Avoid extra `require` for `esm` (i.e. `mts` and `mjs`)

test/build/config-format/coffee/webpack.config.coffee
```diff
@@ -1,7 +1,7 @@
 path = require 'path'
 
 config =
-    mode: 'production'
+    mode: 'development'
```
alexander-akait, line 4: For faster tests

packages/webpack-cli/src/webpack-cli.ts
```diff
@@ -1897,14 +1964,15 @@ class WebpackCLI implements IWebpackCLI {
...
+        ".mts",
         ...Object.keys(interpret.extensions),
       ];
       // Order defines the priority, in decreasing order
-      const defaultConfigFiles = [
-        "webpack.config",
-        ".webpack/webpack.config",
-        ".webpack/webpackfile",
-      ].flatMap((filename) => extensions.map((ext) => path.resolve(filename + ext)));
+      const defaultConfigFiles = new Set(
```
alexander-akait, line 1971: To avoid duplicates

hardfist:
the current method seems good enough for normal use case and we're going to support this as well

PR #4009, opened 2023-11-16
fix(env): set mode after config loaded

### LINKED ISSUE #3992: `--node-env` doesn't set the `mode` as expected
According to the [docs](https://webpack.js.org/api/cli/#node-env) and [this comment](https://github.com/webpack/webpack-cli/issues/2362#issuecomment-771776945) I should be able to use the `--node-env` option to set _both_ `process.env.NODE_ENV` and `mode` but I cannot get this to work as expected. 

Using the following `webpack.config.js` configuration:

```
const config = {};

module.exports = (env, argv) => {
  console.log("mode is " + argv.mode);
  console.log("process.env.NODE_ENV is " + process.env.NODE_ENV);

  return config;
};

```
...and the following `package.json` file:

```
{
  "name": "WebpackIssue",
  "version": "0.1.0",
  "private": true,
  "devDependencies": {
    "webpack": "^5.89.0",
    "webpack-cli": "^5.1.4"
  },
  "scripts": {
    "build": "webpack build --node-env=production"
  }
}
```

Running the `npm run build` command outputs the following:

```
mode is undefined
process.env.NODE_ENV is production
```

Why is `argv.mode` undefined here? Do I still need to set the `mode` explicitly after all (like below) or what am I doing wrong?

    webpack build --mode=production --node-env=production

evenstensberg:
Fixes #3992

No
N/A
**Summary**
Make sure mode is set based on node_env
No
**Other information**
N/A

### DIFF
--- a/packages/webpack-cli/src/types.ts
+++ b/packages/webpack-cli/src/types.ts
@@ -216,6 +216,7 @@ interface WebpackRunOptions extends WebpackOptionsNormalized {
   env: Env;
   failOnWarnings?: boolean;
   isWatchingLikeCommand?: boolean;
+  nodeEnv: string;
 }
 
 /**
--- a/packages/webpack-cli/src/webpack-cli.ts
+++ b/packages/webpack-cli/src/webpack-cli.ts
@@ -2435,6 +2435,15 @@ class WebpackCLI implements IWebpackCLI {
       createJsonStringifyStream = jsonExt.stringifyStream;
     }
 
+    // Make sure mode gets set after node_env is loaded
+    if (
+      options &&
+      options.nodeEnv &&
+      (options.nodeEnv === "development" || options.nodeEnv === "production")
+    ) {
+      options.mode = options.nodeEnv;
+    }
+
     const callback = (error: Error | undefined, stats: WebpackCLIStats | undefined): void => {
       if (error) {
         this.logger.error(error);


snitin315 (commented):
Can we add a corresponding test case?

Ivan198407:
Д̶а̶ к̶о̶н̶е̶ч̶н̶о̶, п̶о̶ч̶е̶м̶у̶ н̶е̶т̶...

ср, 22 нояб. 2023 г., 13:53 Nitin Kumar ***@***.***>:

alexander-akait (changes requested):
It is a bad idea, node-env and mode are different things and we already support another solution

aoberoi:
@alexander-akait can you clarify what the other solution is? maybe we just need to update the docs to reflect it.

aoberoi:
After reading the code, it seems like the `mode` is automatically __after__ the exported function (which returns a config) is run and returns.

Issues like the one @mgnsm has reported (#3992) are occurring because the documentation isn't describing exactly __how__ `NODE_ENV`, the arguments to the configuration function, and `mode` are coordinated.

Based on the current docs it seems reasonable to expect the automatic application of `NODE_ENV` to `mode` to show up somewhere in the config function. I think its usually the case that developers who implement a config function want to use the `mode` value to implement other logic as well (enabling / disabling plugins, modifying other config values, etc). The fact that this automatic application happens after the config function runs seems -- less than helpful.

In order to implement logic based on `NODE_ENV`/`mode` in the config function, the developer is forced to essentially duplicate the logic that's already in the CLI solution. Which ironically renders the CLI solution moot because the developer could just assign that to `mode` in the returned config themselves. So who does the CLI solution serve? It seems like its really for developers who are not using a config function.

Beyond updating the docs to make the intended behavior and use cases clear, I don't have a suggestion on updating the implementation. I think removing the current solution would be a breaking change, so I'm not in favor of that. I think adding this change would be duplicative, so I'm not in favor of that. I think we just need better guidance.

alexander-akait:
Works as expected due https://github.com/webpack/webpack-cli/issues/3992#issuecomment-1887315505, yes, we need to update our docs here, I will help

alexander-akait:
@aoberoi Yeah, you are right, `argv` is just an object of passed arguments through CLI, I don't know why developers expected to see the `mode` there

No need to duplicate logic because you can use `argv.nodeEnv` or  `process.env.NODE_ENV`, the last solution is more common and using a lot of developers

Anyway feel free to feedback

PR #4113, opened 2024-03-14
feat: dot-env plugin

info-arnav:
This PR adds a plugin directory with a basic dot-env support plugin.

I am not sure on how to go about it further. But provides a basic program in the plugin directory.

#3747


alexander-akait:
We need something like https://github.com/mrsteele/dotenv-webpack, just rewrite it on using cache (i.e. inputFileSystem) and more refactor

info-arnav:
actually i am a bit new here and really wanna contribute so would be great if you could please give me a hint on how to proceed with these changes, like maybe the files where i can proceed from.

info-arnav:
I have made some changes, add the caching to internal file system and refactored everything a bit

alexander-akait (changes requested):
It should be a plugin, no need to modify webpack-cli code

info-arnav:
#3759

packages/webpack-cli/src/loadconfig.ts
```diff
@@ -0,0 +1,353 @@
...
+          mergedConfigPaths.push(...configPath);
+        }
+
+        return mergedOptions;
+      }, {});
+      config.path.set(config.options, mergedConfigPaths);
+    }
+
+    return config;
+  }
```
alexander-akait, line 353: We don't need it, we have an webpack CLI API to do it

packages/webpack-cli/src/webpack-cli.ts
```diff
@@ -1,2544 +1,1931 @@
...
-
-    if (isWatch(compiler) && this.needWatchStdin(compiler)) {
-      process.stdin.on("end", () => {
-        process.exit(0);
-      });
-      process.stdin.resume();
     }
-  }
 }
-
```
alexander-akait, line 2543: Don't do these changes, we need `--dont-env` flag and apply your plugin when it is try

info-arnav:
I am adding the tests as well

.gitignore
```diff
@@ -67,3 +67,6 @@ test/**/dist
 test/**/**/dist
 test/**/**/**/dist
 test/**/stats.json
+
+# private readme
+private_readme.md
```
alexander-akait, line 72: Why we need it?

alexander-akait (commented):
Avoid changes in `.husky`, just rever them

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.js
```diff
@@ -0,0 +1,172 @@
...
+    const variables = this.gatherVariables(compiler.inputFileSystem);
+    const target = compiler.options.target ?? "web";
+    const version = (compiler.webpack && compiler.webpack.version) || "4";
+    const data = this.formatData({
+      variables,
+      target,
+      version,
+    });
+    const DefinePlugin =
+      (compiler.webpack && compiler.webpack.DefinePlugin) || require("webpack").DefinePlugin;
```
alexander-akait, line 36: You don't need `require("webpack").DefinePlugin`, remove it

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.js
```diff
@@ -0,0 +1,172 @@
...
+      },
+      config,
+    );
+    this.cache = {};
+  }
+
+  apply(compiler) {
+    const variables = this.gatherVariables(compiler.inputFileSystem);
+    const target = compiler.options.target ?? "web";
+    const version = (compiler.webpack && compiler.webpack.version) || "4";
```
alexander-akait, line 29: Also the same, remove `4`

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.js
```diff
@@ -0,0 +1,172 @@
...
+    } catch (err) {
+      return {};
+    }
+  }
+
+  /**
+   * @param {String} msg - The message.
+   */
+  warn(msg) {
+    console.warn(msg);
```
alexander-akait, line 168: Avoid using `console` in favor `const logger = compilation.getLogger("copy-webpack-plugin");`, let's use webpack logger

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.js
```diff
@@ -0,0 +1,172 @@
...
+    });
+
+    return vars;
+  }
+
+  initializeVars() {
+    return this.config.systemvars ? Object.assign({}, process.env) : {};
+  }
+
+  getEnvs(inputFileSystem) {
```
alexander-akait, line 66: Store `inputFileSystem` in `this.inputFileSystem` in `apply` method

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.js
```diff
@@ -0,0 +1,172 @@
...
+};
+
+const isMainThreadElectron = (target) => target.startsWith("electron") && target.endsWith("main");
+
+class Dotenv {
+  constructor(config = {}) {
+    this.config = Object.assign(
+      {},
+      {
+        path: "./.env",
```
alexander-akait, line 18: Default value should be other, please read my issues and old PR for this, also we should allow to specify array here

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.js
```diff
@@ -0,0 +1,172 @@
...
+        prefixes: ["process.env.", "import.meta.env."],
+      },
+      config,
+    );
+    this.cache = {};
```
alexander-akait, line 23: That is a potential memory leaking cache, please use webpack cache, you can find an example in https://github.com/webpack-contrib/copy-webpack-plugin/blob/master/src/index.js or WeakMap

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.js
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  apply(compiler) {
+    const variables = this.gatherVariables(compiler.inputFileSystem);
+    const target = compiler.options.target ?? "web";
```
alexander-akait, line 28: Don't fallback on `web` it is not fully true

alexander-akait (commented):
Let's start the first round improvements, code looks good, also don't forget to add tests

alexander-akait:
After then we finish it, we will move it in own `workspace`, like other packages, but it is not a big problem, so let's focus on improvements and tests

info-arnav:
Now changing the cache to webpack cache

info-arnav:
used Map instead of wekmap, hope that works ?

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,179 @@
...
+    this.logger = compiler.getInfrastructureLogger("dotenv-webpack-plugin");
+    const variables = this.gatherVariables();
+    const target = compiler.options.target;
+    const data = this.formatData({
+      variables,
+      target: typeof target === "boolean" ? undefined : target,
+    });
+    new DefinePlugin(data).apply(compiler);
+  }
+  private gatherVariables(): EnvVariables {
```
alexander-akait, line 61: Please use `#` for private

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,179 @@
...
+          (!target.includes("node") && !isMainThreadElectron(target))),
+    );
+  }
+
+  private loadFile(filePath: string): string {
+    if (this.cache[filePath]) {
+      return this.cache[filePath];
+    }
+    try {
+      const content = this.inputFileSystem.readFileSync(filePath, "utf8");
```
alexander-akait, line 166: Avoid using sync method here

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,179 @@
...
+      this.cache[filePath] = content;
```
alexander-akait: We don't need cache here, because `this.inputFileSystem` is cachable by default, this cache doesn't make sense

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,179 @@
...
+
+export class Dotenv {
+  private logger!: Logger;
+  private config: DotenvConfig;
+  private cache: Record<string, string>;
+  private inputFileSystem: any;
+
+  constructor(config: DotenvConfig = {}) {
+    this.config = {
+      paths: ["./.env"],
```
alexander-akait, line 43: Default values should be other

info-arnav: can you please elaborate on this, like do u mean the ./.env should be ./.env.local or something

alexander-akait: it should be:
```
.env.$(NODE_ENV).local
.env.local (Not checked when NODE_ENV is test.)
.env.$(NODE_ENV)
.env
```

NODE_ENV is `process.env.NODE_ENV`, if the environment variable NODE_ENV is unassigned it is `compilation.option.mode`, so:

JUST pseudo code:
```js
const nodeEnv = process.env.NODE_ENV || compilation.options.mode;
const possiblePaths = [".env.[node_env].local"]
  .concat(nodeEnv === "test" ? [] : [".env.local "])
  .concat(".env.[node_env]", ".env")
].map(item => item.replace("[node_env]", nodeEnv))
```

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,179 @@
...
+      ...config,
+    };
+    this.cache = {};
+  }
+
+  public apply(compiler: Compiler): void {
+    this.inputFileSystem = compiler.inputFileSystem;
+    this.logger = compiler.getInfrastructureLogger("dotenv-webpack-plugin");
+    const variables = this.gatherVariables();
+    const target = compiler.options.target;
```
alexander-akait, line 54: `target` can be only string

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
info-arnav: so, shall i remove the cache overall; and just run 

`const content = this.inputFileSystem.readFile(filePath, "utf8");`

wverytime

alexander-akait: Yeah, we will implement the cache later in right way, but let's finish eveything before it

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
info-arnav: This is actually causing some issue

package.json
```diff
@@ -58,9 +58,11 @@
     "@babel/register": "^7.15.8",
     "@commitlint/cli": "^17.1.2",
     "@commitlint/config-conventional": "^17.1.0",
+    "@types/dotenv-defaults": "^2.0.4",
     "@types/jest": "^29.4.0",
     "@types/node": "^20.11.5",
     "@types/rechoir": "^0.6.1",
+    "@types/webpack": "^5.28.5",
```
alexander-akait, line 65: Please remove it, webpack has own types

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,200 @@
...
+    compiler.hooks.thisCompilation.tap("dotenv-webpack-plugin", async (compilation) => {
+      const cache = compilation.getCache("dotenv-webpack-plugin-compiler");
+      const nodeEnv = process.env.NODE_ENV || compilation.options.mode;
+      if (this.#config.paths) {
+        this.#config.paths = [
+          ...this.#config.paths,
+          ...(nodeEnv === "test" ? [] : [".env.local"]),
+          `.env.${nodeEnv}`,
+          `.env.${nodeEnv}.local`,
+        ];
```
alexander-akait, line 64: It is invalid, it should be:
```
this.#config.paths = [
         ".env",
          ...(nodeEnv === "test" ? [] : [".env.local"]),
          `.env.${nodeEnv}`,
          `.env.${nodeEnv}.local`,
        ];
```

and with another order and put this in `constructor`

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,200 @@
...
+      }
+
+      if (!this.#cached) {
```
alexander-akait, line 67: We should check cache using webpack API, there is a simple example usage cache API https://github.com/webpack-contrib/json-minimizer-webpack-plugin/blob/master/src/index.js#L122

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,200 @@
...
+  #loadFile(filePath: string): string {
+    try {
+      const content = this.#inputFileSystem.readFileSync(filePath, "utf8");
+      return content;
+    } catch (err) {
+      return "{}";
+    }
+  }
+
+  public warn(msg: string): void {
```
alexander-akait, line 195: Just remove this mehod, and use `this.#logger.warn(msg);` directly

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,200 @@
...
+          (!target.includes("node") && !isMainThreadElectron(target))),
+    );
+  }
+
+  #loadFile(filePath: string): string {
+    try {
+      const content = this.#inputFileSystem.readFileSync(filePath, "utf8");
+      return content;
+    } catch (err) {
+      return "{}";
```
alexander-akait, line 191: Please don't do it, we need to log this - `this.#logger.log("Unable to upload ${filepath} file due:\n ${err.toString()}");`

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
alexander-akait: i.e. `await cache.getPromise("dotenv-webpack-plugin-data", null)` should return value, so if values is defined - we have cache and we don't need `this.#cached` at all

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,200 @@
...
+        prefix === "process.env." &&
+        this.#config.ignoreStub !== true &&
+        (this.#config.ignoreStub === false ||
+          (!target.includes("node") && !isMainThreadElectron(target))),
+    );
+  }
+
+  #loadFile(filePath: string): string {
+    try {
+      const content = this.#inputFileSystem.readFileSync(filePath, "utf8");
```
alexander-akait, line 188: Each loaded file should be add to compilation to rerun compilation if file was changed, i.e. `compilation.fileDependencies.push(file)` which was loaded and `compilation.buildDependencies.push(...)` and `compilation.missingDependencies.push(...)` on paths which were not loaded

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
info-arnav: somehow if i am adding await before the first execution and going in an asynchronous manner in the first run, i am facing some error. Like the code isnt executing correctly in the test created, that is why the first execution is kept synchronous.
can you just guide on this once

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,196 @@
...
+  return env;
+};
+
+const isMainThreadElectron = (target: string | undefined): boolean =>
+  !!target && target.startsWith("electron") && target.endsWith("main");
+
+export class Dotenv {
+  #logger!: Logger;
+  #config: DotenvConfig;
+  #inputFileSystem: any;
```
alexander-akait, line 38: Wny any?

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,196 @@
...
+
+  #loadFile(filePath: string, compilation: any): string {
+    compilation.fileDependencies.add(filePath);
+    try {
+      const content = this.#inputFileSystem.readFileSync(filePath, "utf8");
+      compilation.buildDependencies.add(filePath);
+      return content;
+    } catch (err: any) {
+      compilation.missingDependencies.add(filePath);
+      this.#logger.warn(`Unable to upload ${filePath} file due:\n ${err.toString()}`);
```
alexander-akait, line 190: Don't generate warn, because it will be too spammy, just use `this.#logger.log`

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,196 @@
...
+      : {};
+  }
+
+  #getEnvs(compilation: any): { env: EnvVariables; blueprint: EnvVariables } {
+    const { paths } = this.#config;
+
+    const env: EnvVariables = {};
+
+    (paths || ["./.env"]).forEach((path) =>
+      Object.assign(env, dotenv.parse(this.#loadFile(path || "./.env", compilation))),
```
alexander-akait, line 113: Avoid fallback to `"./.env"`, just throw an error if there are no paths in `this.#config`

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,196 @@
...
+export class Dotenv {
+  #logger!: Logger;
+  #config: DotenvConfig;
+  #inputFileSystem: any;
+  #cached: boolean;
+
+  constructor(config: DotenvConfig = {}) {
+    const nodeEnv = process.env.NODE_ENV || "development";
+
+    this.#config = {
```
alexander-akait, line 44: Rename it to `#options`, we use this pattern in all plugins

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,196 @@
...
+        this.#config.ignoreStub !== true &&
+        (this.#config.ignoreStub === false ||
+          (!target.includes("node") && !isMainThreadElectron(target))),
+    );
+  }
+
+  #loadFile(filePath: string, compilation: any): string {
+    compilation.fileDependencies.add(filePath);
+    try {
+      const content = this.#inputFileSystem.readFileSync(filePath, "utf8");
```
alexander-akait, line 185: I think we can use `readFile` instead `readFileSync` to avoid blocking loop

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,196 @@
...
+  info(message: string): void;
+  warn(message: string): void;
+  error(message: string): void;
+}
+
+interface DotenvConfig {
+  paths?: string[];
+  prefixes?: string[];
+  systemvars?: boolean;
+  allowEmptyValues?: boolean;
```
alexander-akait, line 18: let's do it true by default, because it is popular `PUBLIC_PATH=""`

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,196 @@
...
+  expand?: boolean;
```
alexander-akait, line 19: Let's do it true by default too, next.js does the same, it is useful

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,196 @@
...
+  ignoreStub?: boolean;
+  defaults?: boolean | string;
```
alexander-akait, line 21: I don't see how we use this option in your code

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,196 @@
...
+  [key: string]: string;
+}
+
+interface Logger {
+  info(message: string): void;
+  warn(message: string): void;
+  error(message: string): void;
+}
+
+interface DotenvConfig {
```
alexander-akait, line 14: Let's implement `safe`, it is useful in some cases https://github.com/mrsteele/dotenv-webpack?tab=readme-ov-file#properties

alexander-akait (commented):
Looks good, some nitpicks, let's finish it on this week, thank you for your work

info-arnav:
Ill conduct the tests and update tom

info-arnav:
Solving the remaining two issues as well

packages/webpack-cli/package.json
```diff
@@ -41,6 +41,8 @@
     "colorette": "^2.0.14",
     "commander": "^10.0.1",
     "cross-spawn": "^7.0.3",
+    "dotenv": "^16.4.5",
+    "dotenv-defaults": "^5.0.2",
```
alexander-akait, line 45: We don't need `dotenv-defaults`

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,253 @@
...
+    this.#inputFileSystem = compiler.inputFileSystem;
+    this.#logger = compiler.getInfrastructureLogger("dotenv-webpack-plugin");
+    if (!this.#options.paths) {
+      this.#options.paths = [
+        ".env",
+        ".env.local",
+        `.env.${compiler.options.mode}`,
+        `.env.${compiler.options.mode}.local`,
+      ];
+    }
```
alexander-akait, line 75: This is invalid, please use in constructor:
```
[
  ".env",
  ...(process.env.NODE_ENV === "test" ? [] : [".env.local"]),
  `.env.[mode]`,
  `.env.[mode].local`,
]
```

then in `for (const path of paths) {` just replace `[mode]` on `process.env.NODE_ENV || compiler.options.mode`

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,253 @@
...
+    compiler.hooks.thisCompilation.tap("dotenv-webpack-plugin", (compilation) => {
```
alexander-akait, line 76: That is wrong hook, it is too later, and will be bad for perf with child compilation

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,253 @@
...
+    return vars;
+  }
+
+  #initializeVars(): EnvVariables {
+    return this.#options.systemvars
+      ? Object.fromEntries(Object.entries(process.env).map(([key, value]) => [key, value ?? ""]))
+      : {};
+  }
+
+  merge(apply = {}, defaults = {}) {
```
alexander-akait, line 135: Avoid it and use `...`

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,253 @@
...
+      (target) =>
+        prefix === "process.env." &&
+        this.#options.ignoreStub !== true &&
+        (this.#options.ignoreStub === false ||
+          (!target.includes("node") && !isMainThreadElectron(target))),
+    );
+  }
+
+  async #loadFile(filePath: string, compilation: any): Promise<string> {
+    compilation.fileDependencies.add(filePath);
```
alexander-akait, line 232: It is not a `fileDependencies`

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,253 @@
...
+          }
+        }
+        obj[key] = JSON.stringify(value);
+        return obj;
+      },
+      {},
+    );
+
+    const formatted: Record<string, string> = {};
+    prefixes?.forEach((prefix) => {
```
alexander-akait, line 198: Never do it, it is unreadable, please use `if (...) { /* code */ }`, same for all places

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,253 @@
...
+      Object.entries(preprocessedVariables).forEach(([key, value]) => {
+        formatted[`${prefix}${key}`] = value;
+      });
+    });
+
+    const shouldStubEnv =
+      prefixes?.includes("process.env.") && this.#shouldStub({ target, prefix: "process.env." });
```
alexander-akait, line 205: Using `?.` here is normal

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,253 @@
...
+    Object.assign(parsed, parsedDefault, parsedSrc);
+    return parsed;
+  }
+
+  async #getEnvs(compilation: any): Promise<{ env: EnvVariables; blueprint: EnvVariables }> {
+    const { paths, safe } = this.#options;
+
+    const env: EnvVariables = {};
+    let blueprint: EnvVariables = {};
+    if (paths) {
```
alexander-akait, line 152: Avoid checks in your code, make all validation in constructor, it decreases readability of your code

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,253 @@
...
+    Object.assign({}, defaults, apply);
+  }
+
+  parse(src: string, defaultSrc = "") {
+    const parsedSrc = dotenv.parse(src);
+    const parsedDefault = dotenv.parse(defaultSrc);
+    const parsed = {};
+    Object.assign(parsed, parsedDefault, parsedSrc);
+    return parsed;
+  }
```
alexander-akait, line 145: What is this? Why we need it?

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,253 @@
...
+
+  async #getEnvs(compilation: any): Promise<{ env: EnvVariables; blueprint: EnvVariables }> {
+    const { paths, safe } = this.#options;
+
+    const env: EnvVariables = {};
+    let blueprint: EnvVariables = {};
+    if (paths) {
+      for (const path of paths) {
+        const fileContent = await this.#loadFile(path, compilation);
```
alexander-akait, line 154: Refactor this function - `this.#loadFile` should return `{ content: Buffer, path: "/path/to/file/.env" } | Error`, one function = one action

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,253 @@
...
+        Object.assign(env, this.parse(fileContent));
+      }
+      blueprint = env;
+      if (safe) {
+        for (const path of paths) {
+          const exampleContent = await this.#loadFile(`${path}.example`, compilation);
+          Object.assign(env, this.parse(exampleContent));
```
alexander-akait, line 161: Why we need `Object.assign` here?

alexander-akait (changes requested):
please ask more if you do not understand, the previous version was more correct

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,253 @@
...
+interface EnvVariables {
+  [key: string]: string;
+}
+
+interface Logger {
+  info(message: string): void;
+  warn(message: string): void;
+  error(message: string): void;
+  log(message: string): void;
+}
```
alexander-akait, line 13: Why we need this type? Just get it from webpack types

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,253 @@
...
+  });
+  return env;
+};
+
+declare interface InputFileSystem {
+  readFile: (
+    arg0: string,
+    arg1: (arg0?: null | NodeJS.ErrnoException, arg1?: string | Buffer) => void,
+  ) => void;
+}
```
alexander-akait, line 38: Same here, just get it from webpack types

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
alexander-akait: Maybe we can't use `async` here, because we should load them before `initialize`, so please return code without async

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,253 @@
...
+          const target: string =
+            typeof compiler.options.target == "string" ? compiler.options.target : "";
+          const data = this.#formatData({
+            variables,
+            target: target,
+          });
+          new DefinePlugin(data).apply(compiler);
+          await cache.storePromise("dotenv-webpack-plugin-data", null, data);
+        }
+      };
```
alexander-akait, line 94: Refactor it for readable:
```
if (!cache) {
   new DefinePlugin()
} else {
 // Other logic
}
```

info-arnav:
> please ask more if you do not understand, the previous version was more correct

so, shall i rever it back to synchronous running and then make the changes suggested ?

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
info-arnav: this was the function to convert "ABCD=1" to {"ABCD":"1"}, however i dont think its required in the program, ill remove it and use dotenv directly

info-arnav:
@alexander-akait I tried to make changes, but like if i am using async then the tests arent executing and even the environment hooks are failing. Shall i make the other chaneges u said and revert stuff to a synchronous manner ?

info-arnav:
@alexander-akait I have a asynchronous version as well, please let me know if i should upload that as well, if that is a major requirment.

package.json
```diff
@@ -61,6 +61,7 @@
...
```
alexander-akait, line 64: Please remove it, webpack has own types, don't use this package

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,210 @@
...
+    config: DotenvConfig = {
+      paths: process.env.NODE_ENV
+        ? [
+            ".env",
+            ...(process.env.NODE_ENV === "test" ? [] : [".env.local"]),
+            `.env.[mode]`,
+            `.env.[mode].local`,
+          ]
+        : [],
+    },
```
alexander-akait, line 44: We should not use this here, if I pass any option in our config, all paths will be overrided, it is wrong

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,210 @@
...
+      ...config,
+    };
+  }
+
+  public apply(compiler: Compiler): void {
+    this.#inputFileSystem = compiler.inputFileSystem;
+    this.#logger = compiler.getInfrastructureLogger("dotenv-webpack-plugin");
+    this.#compiler = compiler;
+    this.#cache = false;
+    compiler.hooks.thisCompilation.tap("dotenv-webpack-plugin", async (compilation) => {
```
alexander-akait, line 59: Wrong hook

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,210 @@
...
+    let blueprint: EnvVariables = {};
+
+    for (const path of paths) {
+      const fileContent = this.#loadFile(path, compilation);
+      Object.assign(env, dotenv.parse(fileContent));
+    }
+    blueprint = env;
+    if (safe) {
+      for (const path of paths) {
+        path.replace("[mode]", process.env.NODE_ENV || this.#compiler.options.mode || "");
```
alexander-akait, line 126: if `process.env.NODE_ENV || this.#compiler.options.mode` is falsy, we should throw an error that we can't replace `[mode]`, and don't fallback to `""`

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
```diff
@@ -0,0 +1,210 @@
...
+        const exampleContent = this.#loadFile(`${path}.example`, compilation);
```
alexander-akait, line 127: Why? `safe` is false by default, why we do extra fs call?

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
alexander-akait: Please use `initialize` because here we always have `this.#inputFileSystem`

info-arnav: So with this, Ill have to remove cache; and also log the errors instead of adding it to the compilation. Right ? If I understood it right, ill proceed to make the changes.

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
info-arnav: Sure doing that

packages/webpack-cli/src/plugins/dotenv-webpack-plugin.ts
info-arnav: > So with this, Ill have to remove cache; and also log the errors instead of adding it to the compilation. Right ? If I understood it right, ill proceed to make the changes.
@alexander-akait

info-arnav:
@alexander-akait can you please guide me on where to get the types for the logger. For input file system the type can not be used as we are using readFileSync

info-arnav:
> @alexander-akait can you please guide me on where to get the types for the logger. For input file system the type can not be used as we are using readFileSync

please let me know about the hook also, as using initialize will remove the compilation object. so `compilation.errors.push` and `compilation.cache` wont work

info-arnav:
@alexander-akait at line 58 I tried adding a 

`compiler.hooks.compilation.tap`

inside the

`compiler.hooks.initialize.tap`

does that make sense ?

info-arnav:
Also do you think adding a readFile in the compilation hook and  SyncReadFile in the initilize hook makes sense ? As in we check if the compilation object was passed as an argument, if it was we proceed in an asynchronous way, else synchronous

> @alexander-akait at line 58 I tried adding a
> 
> `compiler.hooks.compilation.tap`
> 
> inside the
> 
> `compiler.hooks.initialize.tap`
> 
> does that make sense ?

evenstensberg:
open a fresh pr please

PR #4116, opened 2024-03-18
feat: Create webpack app

info-arnav:
This package introduces a command to create webppack app will templates of various frameworks such as react and vue via the command create-webpack-app

#2944

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,3 +67,6 @@ test/**/dist
 test/**/**/dist
 test/**/**/**/dist
 test/**/stats.json
+
+# private readme
+private_readme.md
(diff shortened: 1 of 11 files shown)


evenstensberg (changes requested):
Thanks for your PR! 

As per now, we need to support other libraries/frameworks than just react. We also need vue, angular, svelte support for starters. This idea/feature is also intended for gsoc, so we might close this in favor for a more comprehensive PR on create-webpack-app as theres a lot more work needed to land this.

Anyhow, thank you so much! If you want, I can find another issue for you to fix

info-arnav:
> for gsoc, so we might close this in favor for a more comprehensive PR on create-webpack-app as theres a lot more work needed to land this.

sure, got it.

Would be great if you could tell any more issues to work on. I am working on #4113 as of now

Issue #3537, opened 2022-02-08
Add CLI command to output current in-memory configuration in full


SetTrend:
## Feature request

I propose to add a new command to Webpack CLI that outputs a project's current configuration in full detail – including default values.

The output should be a full-fledged JS file, with actual values used for each single configuration property. If no configuration value is provided, the actual and the computed default value should be added in a comment next to the value (which usually is `undefined`, as it is undefined).

For example, given a configuration like this:

```js
module.exports =
{
  mode: "development",
};
```

… the proposed Webpack CLI `config` command:

```powershell
npx webpack config [config-path [config-name]]
```

… should output the following to `stdout`:

<details>
<br/>

**Excerpt only:**

```js
const path = require('path');

module.exports = {
  target: undefined /* "web" */,
  mode: "development",
  entry: undefined /* "./src" */,
  output: {
    path: undefined /* path.resolve(__dirname, "dist") */ /* "C:\repos\myproject\dist" */,
    filename: undefined /* "[name].js" */,
    publicPath: undefined /* "/assets/" */,

  // ...
  // ...
  // ...

  optimization: {
    chunkIds: undefined /* "size" */,
    moduleIds: undefined /* "size" */,
    mangleExports: undefined /* "size" */,
    minimize: undefined /* true */,
    minimizer: undefined /* null */
  )
}
```

</details>

Please note the following in the above excerpt:

1. `mode` is the only property not being output as `undefined`, because it's being defined in the configuration file.
1. `mode` is the only property without a default value comment, because no default is applied to it.
1. Suppose, no `browserlist` file exists in this example. Then the actual default value for the `target` property is `"web"`, as is depicted in the excerpt above.
1. The actual default value for the `output.path` property is a computation (`path.resolve(__dirname, "dist")`). This will result in a second comment being added to the property, containing the computed default value, which is `"C:\repos\myproject\dist"`.
<br/>

#### What is motivation or use case for adding/changing the behavior?

Although there are many useful default values being applied to Webpack when a small configuration in a Webpack configuration file is used, it's still obscure which actual values are being used for all untouched Webpack configuration properties. So, it's hard to figure out unwanted or less than perfect build behaviour.

Moreover, with the proposed output, GUI tools can be written to provide form based helpers or linters for editing/optimizing webpack configurations.
<br/>

#### What is the expected behavior?

[...]

PR #4126, opened 2024-03-23
feat: added config command

samrudh3125:

It introduces a new feature which is the config command which has three options i.e file, name and include.

I have added tests for name and file option but for include didn't add tests as at this stage it just prints the present configuration in the console.

**Summary**

The issue #3537 is solved in this PR. This was a very good issue to work on as i could understand the complete codebase  of webpack as I am aiming at working in one of the projects of webpack in GSOC.

Yes there are a lot of changes as I have completely deleted the built in commands --config and --config-name and introduced a whole new command in their place.
**Other information**


snitin315 (changes requested):
>Yes there are a lot of changes as I have completely deleted the built in commands --config and --config-name and introduced a whole new command in their place.

This is not the intend of the issue, what we want to achieve is a new `--print-config`/`--show-config` cli flag similar to [other tools like jest](https://jestjs.io/docs/cli#--showconfig) which will output the whole webpack config on console.

samrudh3125:
Yes but in the issue itself you had asked to create a webpack config command which has 3 options:
1 file- which replaces --config
2 name- which replaces --config-name
3 include- which is the additional options which prints the whole configuration in the console.

snitin315:
> Yes but in the issue itself you had asked to create a webpack config command which has 3 options: 1 file- which replaces --config 2 name- which replaces --config-name 3 include- which is the additional options which prints the whole configuration in the console.

No this is not the intent of the issue. Those are options for this command to print selective configurations and can be added later.

samrudh3125:
Then can you please elaborate the issue once again so that I change it accordingly

package.json
```diff
@@ -96,5 +96,8 @@
     "webpack": "^5.90.3",
     "webpack-bundle-analyzer": "^4.5.0",
     "webpack-dev-server": "^5.0.2"
+  },
+  "dependencies": {
+    "yarn": "^1.22.21"
```
alexander-akait, line 101: Why we need this?

packages/webpack-cli/src/webpack-cli.ts
```diff
@@ -517,6 +560,49 @@ class WebpackCLI implements IWebpackCLI {
...
+                }
+                console.log(`\n${key}:`, finalOption);
+              }
+            });
+          })
+        : [];
+    }
+
+    const compiler = webpack(config.options as webpack.Configuration[]);
+    compiler.run((err, stats) => {
```
alexander-akait, line 597: We don't need to run this

packages/webpack-cli/src/webpack-cli.ts
```diff
@@ -517,6 +560,49 @@ class WebpackCLI implements IWebpackCLI {
...
+      configName: options.name ? (options.name as unknown as string[]) : undefined,
+    });
+    if (options.name) {
+      config.options = Array.isArray(config.options)
+        ? config.options.filter((option: any) => options.name?.includes(option.name))
+        : [];
+    }
+
+    if (options.include) {
+      const includeOptions = webpack.config.getNormalizedWebpackOptions({});
```
alexander-akait, line 576: Also I think we need to apply defaults, so you will see how webpack sees your configuration

alexander-akait (changes requested):
Also please don't remove tests

package.json
samrudh3125: I think that was a mistake I will resolve it

packages/webpack-cli/src/webpack-cli.ts
samrudh3125: No it's running only in case of --file and --name for --inckude I will remove it

packages/webpack-cli/src/webpack-cli.ts
samrudh3125: Actually we need to comment the default values beside them which is getting very complex so shall I apply the defaults?

samrudh3125:
My bad actually since I removed the command I deleted the tests

packages/webpack-cli/src/webpack-cli.ts
alexander-akait: You can get default value `applyWebpackOptionsBaseDefaults` and `applyWebpackOptionsDefaults` from:
```
const {
	applyWebpackOptionsDefaults,
	applyWebpackOptionsBaseDefaults
} = require("webpack/config/defaults");
```

How I see this logic, you load config file/files, normalize and them apply default.

Also we have two possible commands:
- `webpack config` just print full resolved configuration, i.e. after apply normalization and default
- `webpack config --diff` show only options which are different after normalize and default, so you can undestand which options you don't need because they are default

Issue #3768, opened 2023-05-04
implement webpack template generator for webpack repos


alexander-akait:
**Is your feature request related to a problem? Please describe.**

For internal usage - generate:
- configuration file for prettier/eslint/etc
- setup basic structure

**Describe the solution you'd like**

```
webpack init --template webpack
```

**Describe alternatives you've considered**

No need

**Additional context**

Replacer for webpack-defaults

samrudh3125:
@alexander-akait  please can you elaborate the issue I didn't get it

maverox:
and what could be the reference for eslint and prettier configurations?

ModalityZ:
https://webpack.js.org/configuration/#set-up-a-new-webpack-project still promotes "npx webpack init" although the init command is no longer listed in "webpack --help".

alexander-akait:
@snitin315 can you help to fix it, thanks

ModalityZ:
> @snitin315 can you help to fix it, thanks

I could not find a replacement for it among the available commands.

snitin315:
I'll raise a fix.

r3yc0n1c:
hi, I'd like to work on this if it's still available

itsAyush5:
Can I work on this ?

alexander-akait:
@snitin315 still make sense, we don't have common webpack repo generator

snitin315:
Sorry, I accidentally closed this

PR #4152, opened 2024-04-22
Feat: webpack template generator for webpack repos

info-arnav:
With reference to #3768 

The following --template flag will be introduced

```
webpack init --template webpack
```

And the template webpack will act as a replacer for webpack-defaults

I have only added the basic idea of execution, please lemme know if it is interprated correctly; then ill continue.


packages/generators/src/init-generator.ts
```diff
@@ -16,13 +16,27 @@ export default class InitGenerator<
...
 
+  #parseArgs(args: string[]): void {
+    const templateIndex = args.findIndex((arg) => arg === "--template");
+    if (
+      templateIndex !== -1 &&
+      templateIndex + 1 < args.length &&
+      args[templateIndex + 1] == "webpack"
+    ) {
+      this.template = args[templateIndex + 1];
+      // now if the template is webpack, then we follow the execution of webpack-defaults ?
```
alexander-akait, line 34: We need to move all logic of webpack-defaults here

alexander-akait: i.e. the same files, configuration and etc

info-arnav: got it, working on it

evenstensberg (changes requested):
we have a gsoc project working on CWA (Create Webpack App) and we will not accept his pr at this time. Thank you for the contribution though! Maybe we can find another issue for you to solve?

PR #4225, opened 2024-07-18
Feat: loader and plugin command [2]

maverox:
feature

Yes

**Summary**
- add loader command
- add plugin command
- add tests for each

**Other information**


packages/create-webpack-app/src/generators/init/default.ts
```diff
@@ -0,0 +1,272 @@
...
+      }
+      actions.push({
+        type: "pkgInstall",
+        path: plop.renderString("{{projectPath}}/", answers),
+        // Custom function don't automatically render hbs template as path hence manual rendering
+        packages: dependencies,
+      });
+      return actions;
+
+      // answers.entryPoint = answers.langType === "Typescript" ? "./src/index.ts" : "./src/index.js";
```
evenstensberg, line 197: whys this commented out?

maverox: This was a previous way of doing the file creation and template logic, which was more restrictive so I commented it out and didnt delete for reference.
I think Now there is no need for this will remove this in next commit.

packages/create-webpack-app/src/utils/logger.ts
```diff
@@ -1,16 +1,16 @@
 import { WebpackCLILogger } from "webpack-cli";
 import { green, yellow, Color, red, cyan, cyanBright, blue, blueBright } from "colorette";
 import { PlopActionHooksChanges, PlopActionHooksFailures } from "../types";
-import { basename } from "path";
+import { relative } from "path";
 
 const prefix: string = blueBright("create-webpack");
 const getLogger = (): WebpackCLILogger => {
```
snitin315, line 7: Can't we use the logger like below?

```js
import webpackCLI from "webpack-cli"

const cli = new webpackCLI();

cli.logger()
```

maverox: I think that I am using logger only and defining it is easier than importing the whole webpackCLI().
also earlier I tried doing it but was getting into a lot of dependency complexities.

So this method was easier.

maverox: Also I told you last time that the issue was that webpack-cli doesnt exports cli, only definitions.
```console
error| '"webpack-cli"' has no exported member named 'webpackCLI'. Did you mean 'IWebpackCLI'?
```

snitin315: We should avoid code duplication wherever possible

snitin315 (approved):

PR #4249, opened 2024-08-19
Feat: File generation and collision handler [6]

maverox:
Introduce a new File Generation and Collision Handler module 
- add ejs support and remove hbs conflict
- add fileCollision Logic in same
  - add abort prompt
  - add overwrite and overwrite all
  - add diff view in vs code
  - conditional rendering based on file type (bin or text)
- Uses the same Config Object as plop libraries add function

**Summary**

Yes

**Other information**


packages/create-webpack-app/src/utils/fileActions.ts
```diff
@@ -0,0 +1,221 @@
...
+      throw new Error("Template or templateFile is required");
+    }
+
+    await fs.mkdir(path.dirname(renderedPath), { recursive: true });
+
+    if (isBinaryFile(templatePath)) {
+      await fs.writeFile(renderedPath, content as Buffer);
+    } else {
+      await fs.writeFile(renderedPath, content as string, "utf8");
+    }
```
alexander-akait, line 217: You can just use `await fs.writeFile(renderedPath, content);`, node will do converting under the hood

maverox: Earlier I was using that way, but then I was getting a problem where image wasnt loading and it turns out that, renderer was changing the content of image file and corrupting in the process. That's why I amhad to introduce this workaround

alexander-akait: I think you have  "utf8" before, just omit it when you write a file

maverox: Yes, oh so if i dont specify it decides based on content?
Okay I will make this change,

Does readFile also work in similar ways?

alexander-akait: > Does readFile also work in similar ways?

If you have `"utf8"` in `readFile` you will get `string`, otherwise `Buffer`

maverox: I tried doing it, the problem is, ejs.render first should be a string,

alexander-akait: Just convert it to string when you pass it to ejs (from `Buffer`), avoid extra check like `isBinaryFile`, it is wrong https://github.com/webpack/webpack-cli/pull/4249/files#diff-1554215a066d0b79435196cfc8e6329e9d77267b8e3780afa6176c35e4c39787R30 and

packages/create-webpack-app/src/utils/fileActions.ts
```diff
@@ -0,0 +1,221 @@
...
+    const existingContent = await fs.readFile(filePath, "utf8");
+    return existingContent === content;
+  } catch {
+    return false;
+  }
+}
+
+function getDiff(filePath: string, tempFilePath: string): Promise<void> {
+  return new Promise((resolve, reject) => {
+    const editor = "code";
```
alexander-akait, line 59: It can be broken on some `OS`, also developer just can't have `code` package...

maverox: initially I had vim as fallback, but then I asked on slack and changed this to vscode.
Would checking for code and then setting work?

packages/create-webpack-app/src/utils/fileActions.ts
maverox: Tried this but it is still causing the same corruption of .png file
```js
async function renderTemplate(
  template: string | undefined,
  templateFile: string | undefined,
  data: Answers | undefined
): Promise<Content | undefined> {
  if (template) {
    return ejs.render(template, data || {}, { async: true });
  }

  if (templateFile) {
    const fileContent: Content = await fs.readFile(templateFile);
    const templateContent = fileContent.toString("utf8");
    const renderedContent = await ejs.render(templateContent, data || {}, { async: true });
    return renderedContent;
  }

  return undefined;
}
```

packages/create-webpack-app/src/utils/fileActions.ts
maverox:
```typescript 
function checkIfCodeInstalled(): boolean {
  try {
    const result = spawn.sync('code', ['--version'], { stdio: 'ignore' });
    return result.status === 0;
  } catch (error) {
    return false;
  }
}

function getDiff(filePath: string, tempFilePath: string): Promise<void> {
  return new Promise((resolve, reject) => {
    const platform = process.platform;
    let editor = '';

    if (platform === 'win32') {
      if (checkIfCodeInstalled()) {
        editor = 'code';
      } else {
        return reject(new Error('Visual Studio Code is not installed. Please install VS Code to continue.'));
      }
    } else if (platform === 'darwin' || platform === 'linux') {
      editor = checkIfCodeInstalled() ? 'code' : 'vim';
    } else {
      return reject(new Error(`Unsupported platform: ${platform}`));
    }

    let diffCommand = '';

    if (editor === 'code') {
      diffCommand = `${editor} --diff ${filePath} ${tempFilePath}`;
    } else if (editor === 'vim') {
      diffCommand = `${editor} -d ${filePath ${tempFilePath}`;
    }

    const diffProcess = spawn(diffCommand, { shell: true, stdio: 'inherit' });

    diffProcess.on('exit', (code) => {
      if (code !== 0) {
        reject(new Error('Error opening diff in editor'));
      } else {
        resolve();
      }
    });
  });
}
```

maverox: @alexander-akait Would this work?

packages/create-webpack-app/src/generators/init/default.ts
```diff
@@ -0,0 +1,198 @@
...
+        type: "list",
+        name: "langType",
+        message: "Which of the following JS solutions do you want to use?",
+        choices: ["none", "ES6", "Typescript"],
+        default: "none",
+      },
+      {
+        type: "confirm",
+        name: "devServer",
+        message: "Would you like to use Webpack Dev server?",
```
evenstensberg, line 42: small w

evenstensberg:
@maverox you need to rebase

evenstensberg:
cc @snitin315

PR #4318, opened 2024-11-05
feat!: rename `--define-process-env-node-env` to `--config-node-env`

snitin315:
Fix https://github.com/webpack/webpack-cli/issues/3503#issuecomment-1328363347

https://github.com/webpack/webpack-cli/blob/c97779b5e7447701f19047df10f661d11118d45f/packages/webpack-cli/src/webpack-cli.ts#L2388-L2390

Yes
Will update in docs website, I have updated the option descriptions.
**Summary**

https://github.com/webpack/webpack-cli/pull/4318#discussion_r1829484978

Yes

`--define-process-env-node-env` was removed in favor of `--config-node-env` option.

**Other information**
No


packages/webpack-cli/src/webpack-cli.ts
```diff
@@ -2363,16 +2364,30 @@ class WebpackCLI implements IWebpackCLI {
     options: Partial<WebpackDevServerOptions>,
     callback?: Callback<[Error | undefined, WebpackCLIStats | undefined]>,
   ): Promise<WebpackCompiler> {
-    if (typeof options.defineProcessEnvNodeEnv === "string") {
-      // TODO: This should only set NODE_ENV for the runtime not for the config too. Change this during next breaking change.
-      process.env.NODE_ENV = options.defineProcessEnvNodeEnv;
-    } else if (typeof options.nodeEnv === "string") {
+    const isNodeEnvOptionDefined = typeof options.nodeEnv === "string";
+    const isDefineProcessEnvNodeEnvOptionDefined =
+      typeof options.defineProcessEnvNodeEnv === "string";
```
alexander-akait, line 2369: @snitin315 I don't think we need this option anymore anyway. We have - https://github.com/webpack/webpack/blob/main/lib/WebpackOptionsApply.js#L599.

The main problem is that we have two options:
- `--node-env` (set `process.env.NODE_ENV` for `webpack.config.js`)
- `--optimization-node-env` (set `process.env.NODE_ENV` for runtime)

And that's why they confuse some developers. The names are almost the same, but they do completely different things.

I think the best solution would be to do the following:
- remove `--define-process-env-node-env` at all
- we leave `--node-env`  because it is widely used by many  
- we come up with a new name for the `--node-env` variable and updates the documentation so that new developers can use it

Shortly `--node-env` and `---new-name-node-env` work and do the same thing, but the documentation will recommend to use a new name.

To be honest I have no ideas for a new name, so I'd be happy to hear any suggestions :smile:

snitin315: @alexander-akait how about `--config-node-env`?

alexander-akait: `--config-node-env` sounds good

snitin315: Updated 👍🏻

alexander-akait (approved):

anshumanv (approved):

c100k:
Hello,

I've just realized one of our production scripts was broken because of this change, after upgrading to 6.x a couple of days ago.

What do you think about adding this as a BREAKING CHANGE in the [CHANGELOG of 6.0.0](https://github.com/webpack/webpack-cli/releases/tag/webpack-cli%406.0.0) ?

It will make it easier for everyone to migrate while upgrading. Note that the previous change regarding this flag was mentioned in the [CHANGELOG of 5.0.0](https://github.com/webpack/webpack-cli/releases/tag/webpack-cli%405.0.0).

junhaoliao:
> Hello,
> 
> I've just realized one of our production scripts was broken because of this change, after upgrading to 6.x a couple of days ago.
> 
> What do you think about adding this as a BREAKING CHANGE in the [CHANGELOG of 6.0.0](https://github.com/webpack/webpack-cli/releases/tag/webpack-cli%406.0.0) ?
> 
> It will make it easier for everyone to migrate while upgrading. Note that the previous change regarding this flag was mentioned in the [CHANGELOG of 5.0.0](https://github.com/webpack/webpack-cli/releases/tag/webpack-cli%405.0.0).

I second this. The --define-process-env-node-env option is still mentioned in the [README.md](https://github.com/webpack/webpack-cli/blame/ad6e380b1210798183835ccfbb036eb789aa1476/packages/webpack-cli/README.md#L88), which could lead to confusion.

PR #4546, opened 2025-09-16
fix: gracefull shutdown on watch/cache

evenstensberg:

#3075

TBD

TBD
**Summary**

No

**Other information**

Need to add tests.

### DIFF
--- a/packages/webpack-cli/src/webpack-cli.ts
+++ b/packages/webpack-cli/src/webpack-cli.ts
@@ -2484,9 +2484,7 @@ class WebpackCLI implements IWebpackCLI {
 
     compiler = await this.createCompiler(options as WebpackDevServerOptions, callback);
 
-    if (!compiler) {
-      return;
-    }
+    if (!compiler) return;
 
     const isWatch = (compiler: WebpackCompiler): boolean =>
       Boolean(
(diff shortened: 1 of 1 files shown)


alexander-akait (commented):
We should include this in the next major release, not now, and please add a test case

evenstensberg:
Still wip, need to send a sigterm signal whenever.

test/build/cache/cache.test.js
```diff
@@ -218,4 +218,30 @@ describe("cache", () => {
...
+
+    const proc = runAndGetProcess(__dirname, [
+      "--config",
+      "./graceful-exit.webpack.config.js",
+      "--name",
+      "graceful-exit-test",
+    ]);
+    proc.stdout.on("data", (chunk) => {
+      const data = chunk.toString();
+      if (data.includes("app.bundle.js")) {
```
evenstensberg, line 236: after compile is done and we got an async plugin the sigterm is invoked while we're in the child process without an infinite loop fyi @alexander-akait

test/build/cache/graceful-exit.webpack.config.js
```diff
@@ -0,0 +1,40 @@
...
+      });
+    });
+  }
+}
+
+module.exports = {
+  mode: "development",
+  name: "graceful-exit-test",
+  cache: {
+    type: "filesystem",
```
evenstensberg, line 22: Might need to specify `cache.type` in the check at `isFileSystemCacheEnabled`

PR #4642, opened 2026-01-31
chore: use commander built-ins over custom implementations

avivkeller:
**Summary**

This is still quite a big PR, but I do believe it's an improvement. This PR changes the `webpack-cli.ts` file to use Commander's `help` and `unknownCommand` handlers over the previous custom-implementation.

This PR introduces a **CLI** change in the Commander program.

Nothing substantial changed that required additional tests, but all previous tests pass. 

Note that tests that tested features that are no handled internally were removed, as those are tested by Commander, and don't need to be tested by us.

Also note that the commander unknown command output is slightly different from the previous implementation, so a few snapshots have been changed, but seeing as it's just one stderr line, I don't think it's a breaking change.

As mentioned above, this PR does not include any breaking changes to CLI. The only difference is a slightly different output when an invalid command is passed (Commander gives recommendations in `()`), but that's not breaking.

### DIFF
--- a/packages/webpack-cli/src/bootstrap.ts
+++ b/packages/webpack-cli/src/bootstrap.ts
@@ -17,6 +17,3 @@ export default runCLI;
 
 // TODO remove me in the next major release and use `default` export
 module.exports = runCLI;
-
-// @ts-expect-error ...
-if (process.env.npm_lifecycle_script === "tsx") runCLI();


alexander-akait (changes requested):
I don't understand why we need to refactor this and I don't understand what you want to achieve, please split PRs into small steps, we have stable cli over a lot of year and I sorry I can't read this code because a lot of different changes mixed in the one PR

alexander-akait:
At least please make logical commits and we can read what and where and why was changed

avivkeller:
A lot of these changes are interconnected, but perhaps if it aint broken, there's no need to fix it.

PR #4727, opened 2026-04-07
Refactor(cli): Migrate from CJS to ESM 

info-arnav:
refactor(cli): migrate webpack-cli from CJS to ESM ( Solves the issue #3436 )

- Add "type": "module" and "exports" to webpack-cli package.json
- Rewrite bin/cli.js as ESM, drop import-local dependency
- Replace eval('import()') with native import() for config loading
- Use createRequire for sync require() calls (webpack colors, CJS config fallback)
- Replace __dirname with path.dirname(fileURLToPath(import.meta.url))
- Fix double .default unwrap on cli-plugin dynamic import
- Update ESLint config to recognize bin/ as ESM
- Update API tests to use async import() instead of require()

### DIFF
--- a/.changeset/migrate-cjs-to-esm.md
+++ b/.changeset/migrate-cjs-to-esm.md
@@ -1,0 +1,5 @@
+---
+"webpack-cli": major
+---
+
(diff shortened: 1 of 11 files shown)


alexander-akait:
It is planned only for the next major release, it is too early to make it right now

PR #4728, opened 2026-04-07
fix(serve): allow multi-compiler configs to share a single dev server when using the same port

### LINKED ISSUE #2408: serve does not work with multi-server
**Describe the bug**

We're having issues updating to webpack 5 and `webpack serve`. Our project is divided into multiple webpack configurations. Say for example we have one configuration for a SPA `/login/` , `/portal/`, our application for end users and `/admin/` for our administrators. We do this, so that every SPA when hosted belongs into its own folder and we can easily make sure that critical code like in our `/admin/` SPA is never served outside of intranet. Now to test locally, we use webpack-dev-server's multi project configuration. Every SPA is identifiable by an unique `publicPath` (e. g. `/admin`). This worked beautifully with webpack-dev-server, but with webpack 5 and webpack-cli this is no longer the case.   We use multiple webpack configurations to test our 

 We used to use `webpack-dev-server` to start our development server directly, however as I understand that is no longer working with webpack 5.

Webpack-cli outputs a an error message:

````
 Error: Unique ports must be specified for each devServer option inyour webpack configuration. Alternatively, run only 1 devServer config using the--config-name flag to specify your desired config.
    at Object.startDevServer [as default] 
````

**To Reproduce**

https://github.com/webpack/webpack-cli/blob/master/test/serve/basic/multi-dev-server.config.js without the explicit ports does show my use case perfectly. 

**Expected behavior**

If `publicPath`s for configurations are unique, `webpack serve` should be able to serve them from the same port. 

**Screenshots**

**Additional context**

Maybe this was first introduced with https://github.com/webpack/webpack-cli/pull/1649 A possible solution might be to change https://github.com/webpack/webpack-cli/blob/ec3baa84479984921e40cd7113e4ab6a14723a7a/packages/serve/src/startDevServer.ts#L14 to add an additional check for unique `publicPath`s if ports are not unique. 

Thanks a lot,
Christian

info-arnav:
Previously, webpack serve would throw "Unique ports must be specified" when multiple
configs had devServer options, even if they used the same port or no port at all.
This broke the common multi-compiler use case where configs with unique publicPaths
should be served from one server.

Now configs sharing the same port (or with no port) use a single DevServer instance
with the full multi-compiler, matching how webpack-dev-server v4 was designed to work.
Different explicit ports still error, since multiple DevServer instances cannot share
a multi-compiler without causing ConcurrentCompilationError.

Also adds test coverage for:
- same-port multi-compiler configs (previously errored, now works)
- shared devServer config across configs
- --no-dev-server disabling all dev servers

Note: --no-dev-server [config-name] for per-config disable from CLI was requested
in #2408 but cannot use that exact flag name since webpack's schema already owns
the dev-server option. Maintainer input needed on naming.

Closes #2408

### DIFF
--- a/packages/webpack-cli/src/webpack-cli.ts
+++ b/packages/webpack-cli/src/webpack-cli.ts
@@ -1677,15 +1677,16 @@ class WebpackCLI {
         const possibleCompilers = compilers.filter((compiler) => compiler.options.devServer);
         const compilersForDevServer =
           possibleCompilers.length > 0 ? possibleCompilers : [compilers[0]];
+        const usedPorts = new Set<string>();
 
-        const portGroups = new Map<string, DevServerConfiguration>();
+        let devServerConfiguration: DevServerConfiguration | undefined;
 
         for (const compilerForDevServer of compilersForDevServer) {
           if (compilerForDevServer.options.devServer === false) {
             continue;
           }
 
-          const devServerConfiguration: DevServerConfiguration =
+          const currentConfig: DevServerConfiguration =
             compilerForDevServer.options.devServer || {};
 
           const args: Record<string, WebpackArgument> = {};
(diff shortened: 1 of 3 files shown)


info-arnav:
@alexander-akait

alexander-akait:
https://github.com/webpack/webpack-cli/issues/2408#issuecomment-2221000927, do you read the issue before asking AI to work on it? Looks like no, your solution will not work as expected, the plugin support was already implemented on the dev server side, we just want to fix some small issues and make a new release, so this issue will be closed, no action needed here except adding support the new version of dev server

PR #4736, opened 2026-04-14
feat(cli): decouple UI renderer from CLI core

ThierryRakotomanana:
**Summary**

As discussed in my previous PR #4723 , this update overhauls the design of the webpack-cli output. The primary goal is to standardize the interface by replacing ad-hoc logging with a structured informations.

| BEFORE | AFTER |
| :--- | :--- |
| help alias | help alias |
| version | version |
| build help | build help |
| info | info |

it's a feature, introuce a new UI for the cli 
- A demo : 

https://github.com/user-attachments/assets/aaefbf82-0c99-4f21-a18c-dda7ceb918aa

- Full Coverage: Added comprehensive unit tests for all ui-renderer components.
- Legacy Support: I implemented a stripRendererChrome utility in the test suite. This sanitizes the new visual elements (icons, borders, indents) from stdout and stderr. This allows our existing legacy tests to pass without any modifications, ensuring full backward compatibility for our testing pipeline.
- Snapshot Updates: Updated the CLI integration snapshots to reflect the new standardized design.

Technically, yes, because the output format has changed. However, since the UI is now fully decoupled from the core CLI logic, it is much more stable for future development. Changes to the "guts" of the CLI won't accidentally break the UI, and vice-versa.

I’ve added comments throughout the code to explain the new logic. I’m not sure if we need to update the official external documentation for these internal UI changes yet, but I am open to suggestions if you think it's necessary!

**Use of AI**
for commit description, some JSDoc comments


alexander-akait (commented):
Some general notes:
- avoid any pagination, it is a job of TTY, custom implementation is a breaking change and should be clear described in changelog, we can't bring it in this version
- no changes in stats output, i.e. `webpack build`/`webpack watch`/`webpack serve`, stats output is critical (and often used many services to parse and get some information about success or failed) and should not output any additional information expect webpack output itself

packages/webpack-cli/src/webpack-cli.ts
```diff
@@ -655,6 +673,36 @@ class WebpackCLI {
...
+      const canPaginate =
+        Boolean(process.stdout.isTTY) &&
+        Boolean(process.stdin.isTTY) &&
+        typeof (process.stdin as NodeJS.ReadStream & { setRawMode?: unknown }).setRawMode ===
+          "function";
+
+      await renderCommandHelp(commandHelpData, this.#renderOptions(), canPaginate);
+      renderFooter(this.#renderOptions(), { verbose: isVerbose });
+      process.exit(0);
+    });
```
alexander-akait: It should be in the separate PR, not here, because it is a new feature, not redesign, feel free to create separate PR and implement this basically for our current output

packages/webpack-cli/src/webpack-cli.ts
```diff
@@ -990,6 +1038,45 @@ class WebpackCLI {
     }
   }
 
+  #renderOptions(): RenderOptions {
+    return {
+      colors: this.colors,
+      log: (line) => this.logger.raw(line),
+      // process.stdout.columns can be undefined in non-TTY environments
+      // (CI, test runners, pipes). Fall back to 80 which fits most terminals.
+      columns: process.stdout.columns ?? 80,
```
alexander-akait, line 1049: we should not do it in here, commander respects for `process.stdout.columns`, we should look how to get this information from them

packages/webpack-cli/src/webpack-cli.ts
```diff
@@ -1167,7 +1257,16 @@ class WebpackCLI {
           process.exit(2);
         }
 
-        this.logger.raw(command.helpInformation());
+        const commandHelpData = this.#buildCommandHelpData(command, program, isVerbose);
+        const canPaginate =
```
alexander-akait: No pagination, it is a job of TTY, otherwise it is a breaking change

packages/webpack-cli/src/webpack-cli.ts
```diff
@@ -1213,62 +1312,58 @@ class WebpackCLI {
...
             .map((value) => (typeof value === "string" ? `'${value}'` : value))
             .join(" | ");
-
-          this.logger.raw(`${bold("Possible values:")} ${possibleValuesUnionTypeString}`);
         }
       }
 
-      this.logger.raw("");
+      const canonicalName = option.long ?? optionName;
+      const docUrl = `https://webpack.js.org/option/${canonicalName.replace(/^--/, "")}/`;
```
alexander-akait, line 1332: I like this idea

packages/webpack-cli/src/webpack-cli.ts
```diff
@@ -1773,9 +1913,28 @@ class WebpackCLI {
...
+        );
+
+        const rawInfo = await this.#getInfoOutput({
+          information: {
+            npmPackages: `{${DEFAULT_WEBPACK_PACKAGES.map((item) => `*${item}*`).join(",")}}`,
+          },
+        });
+
+        renderVersionOutput(rawInfo, renderOpts);
+        renderFooter(renderOpts);
```
alexander-akait: I think we can avoid it to repeat multiple times in each command and looking for hooks for command's output on commander level, I don't like how many code we added just for design in many places, we should search API in commander to hook output and make it more elegant

packages/webpack-cli/src/webpack-cli.ts
```diff
@@ -2771,28 +2971,72 @@ class WebpackCLI {
...
+            } else {
+              summary = { success: true, message: `Compiled successfully${time}` };
+            }
+          } catch {
+            // proceed without summary
+          }
         }
+
+        renderStatsOutput(printedStats, summary, renderOpts);
+        onFirstBuildComplete?.();
```
alexander-akait: Don't touch this at all, your make a lot of breaking changes and perf regressions here, let's avoid here any `runWebpack` modification, we can discuss about improvements build/watch/serve command's output a little bit later, let's focus on the new design and structure of code for help/configtest/etc commands firstly

alexander-akait:
Output looks good, let's focus firstly on refactor and making code more elegant to avoid repeating it multiple times

packages/webpack-cli/src/webpack-cli.ts
ThierryRakotomanana: That's great news. I'll get that improved and open another PR once I've finished a few things. Some of the links are currently broken and don't point to the right spots in the documentation.

packages/webpack-cli/src/webpack-cli.ts
ThierryRakotomanana: It's possible to do this using the pre and post-run hooks from Commander. But, we'd need to set up a specific property for each individual action info, configtest, and version. Let met implement this

ThierryRakotomanana:
Some of my comments got marked as "pending" because I accidentally made the review outdated, just wanted to make sure you saw them

packages/webpack-cli/src/webpack-cli.ts
ThierryRakotomanana: I missed updating this one,I'll get to it tomorrow morning.

ThierryRakotomanana: After investigating a way to retrieve this information from Commander, I've found this:
```js
this.program.configureHelp().helpWidth
```
It extracts the exact dimensions from Commander, but the problem is it causes a crash by initiating an infinite loop inside **renderOption** (which is worse than an `undefined` value). I think it' because the Commander program isn't fully initialized, or is in a lifecycle phase where the internal help state is unavailable when renderOptions is executed during a command.

And after taking a look at how Commander accesses this information, it's the same as what I've proposed, as you can see here: https://github.com/tj/commander.js/blob/master/lib/help.js#L30.
I think we should either use `process.stdout.columns ?? 80` or just default to 80 here.

ThierryRakotomanana:
After re-refactoring, this is likely the most minimal set of changes I can propose right now

packages/webpack-cli/src/webpack-cli.ts
ThierryRakotomanana: I found a way to move it outside of `webpack-cli` without adding new properties or touching with the Commander hooks. But still need minimal change to call it from `ui-renderer`

evenstensberg:
this is a big change, we need to make sure we dont break the way people extract cli output (i.e `webpack | cat -l 2)` and also we need to test --json thoroughly.

alexander-akait:
Let’s close, thanks for this PR, good ideas and good output ⭐️

PR #4765, opened 2026-05-27
perf(webpack-cli): allocate Levenshtein buffer lazily

alexander-akait:
The 256 KB peq Uint32Array was allocated at module load, so every CLI
invocation paid that cost even though distance() only runs on error
paths ("did you mean" suggestions). Defer the allocation to the first
distance() call.

https://claude.ai/code/session_01TKJqMqs6zkot18iBUae52n


psnet:
There was no need to move code from levenshtein.ts into cli file, lazy init could be done in levenshtein.ts directly without moving everyting in one big file.

PR #4791, opened 2026-07-03
chore: use plugin support

bjohansebas:
**Summary**

**Use of AI**


bjohansebas:
I'm not sure if you want to support both dev-server 5 and 6 at the same time, with 6 being the one that has plugin support, or if you'd rather introduce this in a new major version for dev-server 6.

bjohansebas (commented):
I think it's preferable to support dev-server 6 so it's easy to switch between versions, rather than releasing a new major version and making it more difficult to upgrade.

.github/workflows/nodejs.yml
```diff
@@ -85,6 +85,10 @@ jobs:
       - name: Install dependencies
         run: npm ci --ignore-scripts
 
+      - name: Install webpack-dev-server ${{ matrix.dev-server-version }}
+        if: matrix.dev-server-version == '5'
+        run: npm install webpack-dev-server@${{ matrix.dev-server-version }} --no-save --ignore-scripts
```
evenstensberg, line 90: can we use `npm ci` instead?

packages/webpack-cli/src/webpack-cli.ts
```diff
@@ -2280,10 +2305,74 @@ class WebpackCLI {
...
+        }
+
+        // In plugin mode the dev server leaves signal handling to the host and
+        // tears itself down through the compiler's `shutdown` hook, so
+        // `compiler.close()` is the single shutdown path here.
+        this.#setupGracefulShutdown(compiler);
+
+        if (this.#needWatchStdin(compiler)) {
+          process.stdin.on("end", () => {
+            compiler.close(() => {
```
evenstensberg, line 2332: do we need to clean exit fscache here too? 🤔

test/build/error/invalid-schema/invalid-schema.test.js
```diff
@@ -148,7 +149,11 @@ describe("invalid schema", () => {
 
     expect(exitCode).toBe(2);
     expect(stderr).toContain("Invalid value 'Yukihira' for the '--server-type' option");
-    expect(stderr).toContain("Expected: 'http | https | spdy | http2'");
+    expect(stderr).toContain(
+      devServerVersion === "5"
```
evenstensberg, line 153: id split these sinto two seperate statements (assertions) if(cond) expect
else expect

packages/webpack-cli/src/webpack-cli.ts
bjohansebas: No, shouldn't `compiler.close()` already do that natively?

.github/workflows/nodejs.yml
bjohansebas: Nope, there's no other way

avivkeller (approved):

evenstensberg:
@bjohansebas think this PR is obsolete now that we added the latest dev server in a newer pr?

bjohansebas:
Yep, I was planning to close this and create the branch directly in the original repo (I didn’t have access before), I just hadn’t had the time.

evenstensberg:
@bjohansebas could you rebase?

bjohansebas:
on it!

bjohansebas:
closes in favor of #4847

PR #4811, opened 2026-07-21
chore: add allowscripts to work with npm 12

ludofischer:
**Summary**

When using npm 12, post install scripts are not allowed to run by default. Without this change, the user has to manually approve the scripts after install, which modifies the `package.json` file or allow the scripts in their own `.npmrc` which makes the install less reproducible.

See https://github.com/orgs/community/discussions/198547

npm configuration change in package.json

The changes do not touch the code that reaches the end user, only the install procedure for development.

No.

**Use of AI**

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,5 +94,11 @@
   },
   "engines": {
     "node": ">=20.9.0"
+  },
+  "allowScripts": {
+    "@parcel/watcher@2.5.6": true,
+    "@swc/core@1.15.43": true,
+    "esbuild@0.28.1": true,
+    "unrs-resolver@1.11.1": true
   }
 }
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -94,5 +94,11 @@
   },
   "engines": {
     "node": ">=20.9.0"
+  },
+  "allowScripts": {
+    "@parcel/watcher@2.5.6": true,
+    "@swc/core@1.15.43": true,
+    "esbuild@0.28.1": true,
+    "unrs-resolver@1.11.1": true
```
alexander-akait: Can we avoid versions here?

ludofischer: Yes, it is possible: https://docs.npmjs.com/cli/v12/commands/npm-approve-scripts I've unpinned the packages in the next commit.

alexander-akait (approved):

PR #4834, opened 2026-08-22
chore: update webpack-dev-server to v6 and test against v5 and v6

bjohansebas:
**Summary**

Updates the `webpack-dev-server` dev dependency to v6 and makes the test suite cover both v5 and v6, which are the versions the `webpack-cli` peer range allows (`^5.0.0 || ^6.0.0`).

- Snapshots for dev-server 6 are added next to the existing v5 ones. The snapshot resolver already supports this through the `.snap.devServer<major>.webpack<major>` extension, so no tooling change was needed. The only difference in the output is the help text for `--server-type`, since v6 drops `spdy` from the possible values.
- CI runs the matrix against `dev-server-version: [5, latest]`, installing v5 with `npm install ... --no-save --ignore-scripts` for that job, following the same approach as #4791.
- `webpack-dev-server` v6 is ESM and exposes the server as its `default` export, while v5 exports it directly, so `tsc` failed with "This expression is not constructable". The CLI now resolves the constructor type from the default export when there is one, which type-checks against both versions. Runtime behavior is unchanged: the package was already loaded through `(await import(pkg)).default`.

chore, plus a types fix in `webpack-cli`.

No new tests, the existing `test/help` and `test/serve` suites now run against both dev-server versions with their own snapshots.

No. The peer range already allowed v6, and the type change only affects how the constructor type is resolved.

`SERVE-OPTIONS-v6.md` will be generated by the scheduled "Update documentation" workflow, which derives the filename from the installed dev-server major.

**Use of AI**

Yes. I used Claude Code (Opus) to carry out the upgrade under my direction: the dependency bump, the type fix, the CI matrix and the snapshots. I verified the build and ran `test/help` against the new snapshots locally, and reviewed the final diff.

### DIFF
--- a/.changeset/dev-server-6-types.md
+++ b/.changeset/dev-server-6-types.md
@@ -1,0 +1,5 @@
+---
+"webpack-cli": patch
+---
+
+fix: resolve the `webpack-dev-server` type from its default export, so the types work with both v5 and v6
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -63,7 +63,7 @@ jobs:
         os: [ubuntu-latest, windows-latest, macos-latest]
         node-version: [20.x, 22.x, 24.x, 25.x]
         webpack-version: [latest]
-        dev-server-version: [latest]
+        dev-server-version: [5, latest]
 
     steps:
       - name: Checkout code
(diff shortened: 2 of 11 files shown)


avivkeller (approved):

evenstensberg:
@bjohansebas could you rebase against main, so we can see the node v26 workflows in this pr?

evenstensberg (approved):

alexander-akait (approved):

